Issue #9, opened 2015-04-19
Object Model for Use Cases 1.i - 1.iii


hsorby:
This issue is for discussion around the object model that we should implement to at least meet the needs of Use Cases 1.i - 1.iii

jonc125:
When it comes to supporting Maths, is there any chance of extracting relevant parts from libSBML, or at least collaborating? Most of the COMBINE standards need to handle mathematics defined in XML, and it'd be nice if we didn't have to either implement it ourselves (like the CellML API) or rely on libSBML (like libSEDML).

nickerso:
I agree that a common MathML implementation across the COMBINE projects would be useful. And I know that people have expressed an interest in making use of the libSBML MathML support outside of libSBML previously. I would imagine that we would need to do the work to extract the MathML bits from libSBML. There would also be a fair amount of work required to extend the libSBML implementation to support the use of MathML in the way that we use it in CellML (for example, there is no concept in [lib]SBML for expressing an ODE in pure MathML). At HARMONY there was some discussion on an upcoming proposal for extending the range of MathML allowed in SBML (and hence to be supported by libSBML), but I'm not too sure how that will go. But this approach is certainly something that will be considered and investigated.

nickerso:
fixed in #10

PR #10, opened 2015-04-19
Add basic object model documentation

hsorby:
Start of work on the object model documentation.  Possibly a better place to discuss issue #9 since we can make use of Github restructured text rendering.


nickerso:
- typo: AbstradtLlbCellMLError
- maybe typo: libcellml:format (should be libcellml::format?)
- I prefer the camel case version of libCellML to be LibCellml :)

agarny:
Agreed with the camel case version: I also prefer `LibCellml`.

nickerso:
I think this is sufficient to get on with those initial use-cases. And can now be merged in.

jonc125:
+1 from me.  It might be nice to have the source that was used to generate the diagrams as well though?

agarny:
I am mainly concerned with the diagram for the first use cases at this stage, so in that context I am mainly fine with the current diagram. Indeed, I am not sure how `deserialize()` works. Why does it return a string? Also, shouldn't it take a string as an input? Either a URI or the contents of a CellML file? Maybe I missed the point of that method?...

Otherwise, I agree with @jonc125 that it would be nice to have the source for the diagrams, if possible.

Issue #20, opened 2015-05-15
Use Case 1.i


hsorby:
Implement code to satisfy use case 1.i

PR #22, opened 2015-05-15
Usecase 1i

hsorby:
Adding test, skeleton implementation and documentation to address #20.


hsorby:
This work is ready for review.

CMakeLists.txt
```diff
@@ -33,7 +33,12 @@ endif()
 if(COVERAGE AND GCOV_FOUND)
   set( LIBCELLML_COVERAGE "${COVERAGE}" CACHE BOOL ${_PARAM_ANNOTATION} FORCE )
 elseif(COVERAGE)
-  message(WARNING "Coverage testing requested but gcov or it's requirements not found!")
+  if (CMAKE_COMPILER_IS_GCC)
+    message(WARNING "Coverage testing requested but gcov or it's requirements not found!")
```
agarny: "... its requirements..."

CMakeLists.txt
```diff
@@ -44,7 +49,12 @@ endif()
 if( MEMCHECK AND VALGRIND_FOUND )
   set( LIBCELLML_MEMCHECK "${MEMCHECK}" CACHE BOOL ${_PARAM_ANNOTATION} FORCE )
 elseif(MEMCHECK)
-  message(WARNING "Memcheck testing requested but valgrind or it's requirements not found!")
+  if (CMAKE_COMPILER_IS_GCC)
+    message(WARNING "Memcheck testing requested but valgrind or it's requirements not found!")
```
agarny: "... its requirements..."

src/api/libcellml/entity.h
```diff
@@ -0,0 +1,53 @@
...
+class Entity
+{
+protected:
+    /**
+     * @brief Pure virtual serialise method.
+     * Pure virtual serialise method for serialising a CellML object to a std::string.
+     * @param format The format to serialise the object to.
+     * @return std::string representation of the object.
+     */
+    virtual std::string serialise(libcellml::CELLML_FORMATS format) const = 0;
```
agarny, line 37: Don't we want to use a `const` reference for the `format` parameter, i.e.

```
virtual std::string serialise(const libcellml::CELLML_FORMATS &format) const = 0;
```

Also, personally, I like to make it clear that a parameter is indeed a parameter. So, I would have:

```
virtual std::string serialise(const libcellml::CELLML_FORMATS &pFormat) const = 0;
```

As for class members, I would have something like `mFormat`.

Now, this may not be part of the Google coding style, I don't know (well, I haven't checked), but if it isn't then maybe it's something we could have nonetheless?

src/api/libcellml/entity.h
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * @brief Pure virtual deserialise method.
+     * Pure virtual deserialise method for deserialising a CellML object from a std::string
+     * in the given format.
+     * @param string The std::string representation of the object.
+     * @param format The format of the string parameter.
+     */
+    virtual void deserialise(std::string string, libcellml::CELLML_FORMATS format) = 0;
```
agarny: Same suggestion as above, of course.

src/api/libcellml/model.h
```diff
@@ -0,0 +1,82 @@
...
+     */
+    void deserialise(std::string string, libcellml::CELLML_FORMATS format);
+
+private:
+
+    Model(const Model&) = delete; /**< non-copyable */
+    Model& operator=(const Model&) = delete; /**< non-copyable */
+
+    struct ModelImpl; /**< Forward declaration for pImpl idiom. */
+    std::unique_ptr<ModelImpl> _pimpl; /**< Private member to implementation pointer */
```
agarny: Typical case where I would personally use `mPimpl` rather than `_pimpl`.

agarny:
For those looking for the use cases: https://github.com/cellml/libcellml/blob/develop/docs/usecases.rst.

agarny:
Apart from the very minor points above, this looks good to me.

src/api/libcellml/entity.h
hsorby, line 37: In the case of the `format` parameter it is in this case an enumeration and therefore it is really weird to pass it as a `const` reference in my opinion.  But otherwise yes in parameters should be passed by const reference if appropriate.

As for prefixing the parameter the Google style guide doesn't explicitly mention parameter naming, but in all the examples offered none have a parameter prefixed with anything.  I am against adding an unnecessary prefix and would not like to see as part of the libcellml coding standard.

For class members I am ambivalent about the prefix (the Google style guide here says class members should be post-fixed with an underscore) either pre-fixing with `_` or `m` is fine for me.  But we should decide and add it to the libcellml standard.

jonc125, line 37: I'd prefer an `m` prefix, since that's what we use in Chaste :)

We also have a convention that function parameters are written in lowerCamelCase (no prefix) and local variables are written_with_underscores.

agarny, line 37: Hugh, I can see your point about passing an enumeration as a `const`. Still, I do it so so that I don't have to 'worry' about 'special' cases.

Regarding naming conventions for local variables, function parameters and class members, it's really a matter of making the code easier to read. I know that since I have started using a prefix of `m` for class members and `p` for function parameters, my code has been much easier to read (for me, at least). Just my 2c/2p worth of experience.

However, I wouldn't personally go with the convention used in Chaste (except for class members). So, to sum up, I would use:
- Local variable: `myCellmlVariable`
- Function parameter: `pMyCellmlParameter`
- Class member: `mMyCellmlMember`

nickerso, line 37: I've never seen a need to distinguish between parameter and local variables and don't have a problem reading my code ;-) 

I'm not fussed regarding the prefix for member variables, but personally use `m` in my own code. The consensus here seems to be for `m`, so maybe we can decide on that one at least and add it to the code style?

nickerso:
This looks good to me. Although perhaps the model deserialise test either isn't complete yet or shouldn't have been added in 96dcd3a?

hsorby:
The model deserialise test isn't complete for sure, it is possibly that the method Model::deserialise shouldn't have been added yet.  My only excuse is that I intend to use that method at some point and also complete the test,  if you don't want to let that sort of naughtiness slide then I will remove the method and test.

It seems obvious from this that my test for 100% coverage is not working that well.

agarny:
Regarding `Model::deserialise()`, I would remove it at this stage (since it's not complete).

Issue #25, opened 2015-06-09
Use cases 1.ii, 1.iii


hsorby:
Implement use case 1.ii and 1.iii

agarny:
It would be nice to be reminded what the description of a particular use case is. FTR:
- 1.ii: create **a model with a valid name** from scratch and serialise it to XML.
- 1.iii: create **a model with an invalid name** from scratch and serialise it to XML.

PR #26, opened 2015-06-09
Usecases 1ii 1iii

hsorby:
addresses issue #25

### DIFF
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -60,11 +60,9 @@ Model& Model::operator=(Model&& rhs)
 std::string Model::serialise(libcellml::CELLML_FORMATS format) const
 {
     std::string repr = "";
-    if (format == CELLML_FORMAT_XML)
-    {
+    if (format == CELLML_FORMAT_XML) {
         repr += "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\"";
-        if (mPimpl->mName.length())
-        {
+        if (mPimpl->mName.length()) {
             repr += " name=\"" + mPimpl->mName + "\"";
         }
         repr += "></model>";
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -66,5 +66,13 @@ std::string Model::serialise(libcellml::CELLML_FORMATS format) const
     return repr;
 }
 
+void Model::setName(const std::string & /*name*/)
+{
 }
 
+const std::string Model::getName() const
+{
+    return "";
+}
+
+}
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -19,6 +19,8 @@ limitations under the License.Some license of other
 
 namespace libcellml {
 
+static const std::string initial_name = "OhOXiRmNswYoVBA2InM7NRPvQ0HMXFR2wNhoFExd";
+
 /**
  * @brief The private implementation for the Model class.
  * This struct is the private implementation struct for the Model class.  Separating
@@ -31,6 +33,8 @@ struct Model::ModelImpl
     ~ModelImpl(){}
     ModelImpl(const ModelImpl&) = delete;
     ModelImpl& operator=(const ModelImpl&) = delete;
+
+    std::string mName = initial_name;
 };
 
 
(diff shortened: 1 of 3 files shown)


nickerso:
The tests look good to me.

nickerso:
@agarny @jonc125 @mirams I have added a test for "unsetting" the name showing how I expect things to behave (which is not how @hsorby has initially implemented it). What do you think? I imagine similar behaviour for unsetting cmeta:id's once they are implemented...

src/model.cpp
```diff
@@ -19,6 +19,8 @@ limitations under the License.Some license of other
 
 namespace libcellml {
 
+static const std::string initial_name = "OhOXiRmNswYoVBA2InM7NRPvQ0HMXFR2wNhoFExd";
```
agarny: Why this? Why if someone was to use exactly that name for a model? Why not simply have an empty string?

src/model.cpp
```diff
@@ -60,11 +64,25 @@ std::string Model::serialise(libcellml::CELLML_FORMATS format) const
     std::string repr = "";
     if (format == CELLML_FORMAT_XML)
     {
-        repr += "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\"></model>";
+        repr += "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\"";
+        if (mPimpl->mName != initial_name)
+        {
```
agarny: We want `{` at the end of the previous line (see https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Conditionals).

tests/model/model.cpp
```diff
@@ -2,12 +2,50 @@
...
+    const std::string e = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\" name=\"name\"></model>";
+
+    libcellml::Model m = libcellml::Model();
+    std::string n = "name";
+    m.setName(n);
+
+    EXPECT_EQ("name", m.getName());
+
+    EXPECT_EQ(e, m.serialise(libcellml::CELLML_FORMAT_XML));
+}
```
agarny, line 24: I would rather have:

```
TEST(Model, name) {
    const std::string n = "name";
    const std::string e = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\" name=\""+n+"\"></model>";

    libcellml::Model m = libcellml::Model();
    m.setName(n);

    EXPECT_EQ(n, m.getName());
    EXPECT_EQ(e, m.serialise(libcellml::CELLML_FORMAT_XML));
}
```

tests/model/model.cpp
```diff
@@ -2,12 +2,50 @@
...
+    libcellml::Model m = libcellml::Model();
+    std::string n = "name";
+    m.setName(n);
+    EXPECT_EQ("name", m.getName());
+    EXPECT_EQ(eName, m.serialise(libcellml::CELLML_FORMAT_XML));
+
+    m.setName("");
+    EXPECT_EQ("", m.getName());
+    EXPECT_EQ(e, m.serialise(libcellml::CELLML_FORMAT_XML));
+}
```
agarny, line 39: Similarly, I would have something like:

```
TEST(Model, unset_name) {
    const std::string n = "name";
    const std::string eName = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\" name=\""+n+"\"></model>";
    const std::string e = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\"></model>";

    libcellml::Model m = libcellml::Model();
    m.setName(n);
    EXPECT_EQ(n, m.getName());
    EXPECT_EQ(eName, m.serialise(libcellml::CELLML_FORMAT_XML));

    m.setName("");
    EXPECT_EQ("", m.getName());
    EXPECT_EQ(e, m.serialise(libcellml::CELLML_FORMAT_XML));
}
```

tests/model/model.cpp
```diff
@@ -2,12 +2,50 @@
...
+    const std::string e = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\" name=\"invalid name\"></model>";
+
+    libcellml::Model m = libcellml::Model();
+    std::string in = "invalid name";
+    m.setName(in);
+
+    EXPECT_EQ("invalid name", m.getName());
+
+    EXPECT_EQ(e, m.serialise(libcellml::CELLML_FORMAT_XML));
+}
```
agarny, line 51: And finally:

```
TEST(Model, invalid_name) {
    const std::string in = "invalid name";
    const std::string e = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\" name=\""+in+"\"></model>";

    libcellml::Model m = libcellml::Model();
    m.setName(in);

    EXPECT_EQ(in, m.getName());
    EXPECT_EQ(e, m.serialise(libcellml::CELLML_FORMAT_XML));
}
```

agarny:
@nickerso, your test for "unsetting" the name is also how I would do it (as opposed to using `initial_name`, with `initial_name` being equal to `OhOXiRmNswYoVBA2InM7NRPvQ0HMXFR2wNhoFExd`, if that's what you meant).

nickerso:
I'm happy with these changes now. I have no real preference for testing the name value and would be happy to leave it as it is or switch to Alan's preferred style.

Issue #29, opened 2015-06-17
Use cases upto 1.vi


hsorby:
The use cases we wish to satisfy are:
- a model with a single component
  - a component with a valid name
  - a component with an invalid name
- a model with two or more components
- a model with three components and an encapsulation hierarchy
  - one component encapsulating two children
  - one component encapsulating a single child which in turn encapsulates a single child
  - an invalid cyclical encapsulation hierarchy

PR #30, opened 2015-06-17
Implementation for use cases upto 1.vi

hsorby:
Add initial component test and skeleton component class.

addresses #29

### DIFF
--- a/src/api/libcellml/component.h
+++ b/src/api/libcellml/component.h
@@ -27,6 +27,11 @@ public:
      */
     ~Component();
 
+    Component(Component&& rhs); /**< move constructor */
+    Component& operator=(Component&& rhs); /**< move assignment */
+    Component(const Component& rhs); /**< copy constructor */
+    Component& operator=(const Component& rhs); /**< assignment */
+
     /**
      * @brief serialise the Model.
      * This method serialises the Model to a std:string representation in the
@@ -37,23 +42,16 @@ public:
      */
     std::string serialise(CELLML_FORMATS format) const;
 
-    Component(Component&& rhs); /**< move constructor */
-    Component& operator=(Component&& rhs); /**< move assignment */
-
     /**
      * @brief add a child component to this component.
-     * Add a child component to this component.
+     * Add a copy of the given component as a child component of this component.
      * @param c the component to add.
      */
-    void addComponent(Component &c);
+    void addComponent(const Component &c);
 
 private:
-
-    Component(const Component&) = delete; /**< non-copyable */
-    Component& operator=(const Component&) = delete; /**< non-copyable */
-
     struct ComponentImpl; /**< Forward declaration for pImpl idiom. */
-    std::unique_ptr<ComponentImpl> mPimpl; /**< Private member to implementation pointer */
+    ComponentImpl* mPimpl; /**< Private member to implementation pointer */
 
 };
 
--- a/src/api/libcellml/component.h
+++ b/src/api/libcellml/component.h
@@ -10,8 +10,8 @@
 namespace libcellml {
 
 /**
- * @brief The Model class.
- * The Model class is for representing a CellML Model.
+ * @brief The Component class.
+ * The Component class is for representing a CellML Component.
  */
 class LIBCELLML_EXPORT Component: public Nameable
 {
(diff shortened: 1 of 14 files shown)


src/api/libcellml/component.h
```diff
@@ -0,0 +1,63 @@
...
+#include <string>
+#include <memory>
+
+#include "libcellml/libcellml_export.h"
+#include "libcellml/nameable.h"
+
+namespace libcellml {
+
+/**
+ * @brief The Model class.
```
rchristie: Cut & paste typo - should describe component.

src/api/libcellml/component.h
```diff
@@ -0,0 +1,63 @@
...
+     */
+    void addComponent(Component &c);
+
+private:
+
+    Component(const Component&) = delete; /**< non-copyable */
+    Component& operator=(const Component&) = delete; /**< non-copyable */
+
+    struct ComponentImpl; /**< Forward declaration for pImpl idiom. */
+    std::unique_ptr<ComponentImpl> mPimpl; /**< Private member to implementation pointer */
```
rchristie: I was thinking this whole class should have been a handle to a component in which case I expected a shared_ptr here. On thinking about it, directly exposing the actual Component (or any other) class in the API is acceptable as long as the client only ever accesses it via a std::shared_ptr. In that case I would expect the constructors to be private and instead only publicly expose a static create() method which returns a std::shared_ptr, meaning whether you are creating a new component or getting a component from a model or other component, you are always dealing with a shared_ptr to it. Similarly all addComponent methods would take the shared_ptr reference too.
The alternative to a static create() method on the class is to require a model to create a component via a Model::createComponent() method which supports a system where the Model knows all Components under it at all times and permits global defaults to be passed to the constructor, if that's useful.

tests/component/encapsulation.cpp
```diff
@@ -0,0 +1,31 @@
+#include "gtest/gtest.h"
+
+#include <libcellml>
+
+
+TEST(Encapsulation, serialise) {
+    libcellml::Component parent = libcellml::Component();
```
rchristie: For C++ 11 prefer:
auto parent = libcellml::Component();
Incidentally, what you had was equivalent to just:
libcellml::Component parent;
which is even less clear that a new component is being made.

If you follow my other recommendations to return std::shared_ptr and have a static create method this would become:
auto parent = libcellml::Component::create();
which is pretty clear about what it does; the auto type will then be a shared_ptr.
If you make the change all other methods will need to called with pointer dereferencing i.e. parent->addComponent(child)

tests/component/encapsulation.cpp
```diff
@@ -0,0 +1,31 @@
...
+    libcellml::Component child = libcellml::Component();
+    parent.addComponent(child);
```
rchristie, line 9: I don't think the above is tenable unless you change to shared_ptr as I expect the internal representation of a list of children to be an STL container of shared_ptr which can't be generated except from another shared_ptr.

src/api/libcellml/component.h
hsorby: Why would I want to put a restriction on that a client can only access a class via a std::shared_ptr?  If the client is worried about memory handling then perhaps they would prefer to use a language like Python or Java and use the library through the appropriate bindings.

tests/component/encapsulation.cpp
hsorby: I hope <code>libcellml::Component parent;</code> is pretty clear to all C++ programmers about what is happening.

src/api/libcellml/component.h
rchristie: If you're going to store a list of child Component objects I'd expect it to be an STL container of shared_ptr, leading to the conclusion that all objects are on the heap (never on the stack) with lifetimes managed by shared_ptr references. [Unlike the intrusive reference counting we use in Zinc, my understanding is that shared_ptr is implemented as a linked list, requiring disciplined use of shared_ptr for all methods that affect object lifetime: create(), addComponent() etc. For using other methods it is perfectly fine or even preferred to use a reference to the actual object (i.e. obtain it from the shared_ptr).]

nickerso:
I believe this completely addresses #29. The discussion on the use of std::shared_ptr<> can be shelved for now and revisited once we begin addressing use cases which require objects to be retrieved from the object model (getting components from a model, etc.).

Issue #33, opened 2015-07-01
Add use cases demonstrating the modification of an existing model and component encapsulation hierarchy


nickerso:
Before heading too far down the implementation path of the libCellML object model, we should implement some tests demonstrating the use of libCellML to modify an existing model and component encapsulation hierarchy. See the related discussion thread here: https://groups.google.com/forum/#!msg/libcellml-team/ELvmfQiY7EE/uXwpNayqw_MJ

(current use-case list available at: http://libcellml.readthedocs.org/en/latest/usecases.html)

rchristie:
Looks good to me.

PR #34, opened 2015-07-01
New use cases for modification of model and component hierarchy

nickerso:
add initial draft of new use cases for the modification of an existing model and component hierarchy. Addressing issue #33

### DIFF
--- a/docs/usecases.rst
+++ b/docs/usecases.rst
@@ -30,7 +30,6 @@ Use-cases for libCellML
       #. take a component (remove the component and return it to the user)
       #. determine if a component with a given name exists in a model or component
       #. determine the number of components encapsulated by a model or component
-      #. determine the number of components with a given name that are encapsulated by a model or component
 
    #. a model with imported components
 
--- a/docs/usecases.rst
+++ b/docs/usecases.rst
@@ -20,6 +20,17 @@ Use-cases for libCellML
       a. one component encapsulating two children
       #. one component encapsulating a single child which in turn encapsulates a single child
       #. an invalid cyclical encapsulation hierarchy
+      
+   #. manipulation of a model with multi-level component encapsulation hierarchy
+   
+      a. remove a component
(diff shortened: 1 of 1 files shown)


docs/usecases.rst
```diff
@@ -20,6 +20,17 @@ Use-cases for libCellML
...
       #. one component encapsulating a single child which in turn encapsulates a single child
       #. an invalid cyclical encapsulation hierarchy
+      
+   #. manipulation of a model with multi-level component encapsulation hierarchy
+   
+      a. remove a component
+      #. change the name of a top-level component
+      #. change the name of an encapsulated child component
+      #. replace one component with a new component
+      #. take a component
```
rchristie: 'take' is ambiguous to me, could mean either get a component from a model (model is unchanged) or remove a component from a model.

rchristie:
Looks good to me, with the last minor change.

mirams:
I think these sound sensible tasks for use cases.

docs/usecases.rst
```diff
@@ -20,6 +20,16 @@ Use-cases for libCellML
       a. one component encapsulating two children
       #. one component encapsulating a single child which in turn encapsulates a single child
       #. an invalid cyclical encapsulation hierarchy
+      
+   #. manipulation of a model with multi-level component encapsulation hierarchy
+   
+      a. remove a component
```
agarny: As for changing the name of a component, shouldn't we have both "remove a top-level component" and "remove an encapsulated child component"?

agarny: A part from my comment above, this looks good to me.

agarny: In fact, looking at what has been implemented so far (see #36), it looks like we should make it clear that we also want to test the removal of a component from a model (and not just a component of sorts).

nickerso: good point, I've added those now.

PR #36, opened 2015-07-04
Implementation for usecases upto 1vii

hsorby:
Add initial component count test and skeleton code to successfully compile the codebase.


src/api/libcellml/model.h
```diff
@@ -67,6 +67,78 @@ class LIBCELLML_EXPORT Model: public Nameable
...
+     * @brief remove the component at the given index.
+     * Remove the component at the given index.  If the index is
+     * not in the range 0, componentCount then an exception will
+     * be raised.
+     * @param index the index of the component to remove.
+     */
+    void removeComponent(size_t index);
+
+    /**
+     * @brief remove the component at the given index.
```
nickerso: copy-paste error :)

tests/component/component.cpp
```diff
@@ -45,6 +45,33 @@ TEST(Component, unset_name) {
...
+
+    parent.addComponent(child1);
+    parent.addComponent(child2);
+    parent.addComponent(child3);
+    parent.addComponent(child4);
+    EXPECT_EQ(4, parent.componentCount());
+
+    child3.addComponent(child4);
+    parent.addComponent(child3);
+    EXPECT_EQ(4, parent.componentCount());
```
nickerso: should be 5 not 4

tests/model/model.cpp
```diff
@@ -112,6 +112,99 @@ TEST(Model, add_components) {
...
+
+TEST(Model, contains) {
+    libcellml::Model m;
+    libcellml::Component c1, c2;
+
+    EXPECT_FALSE(m.containsComponent("child1"));
+
+    m.addComponent(c1);
+    m.addComponent(c2);
+    EXPECT_TRUE(m.containsComponent("child2"));
```
nickerso, line 138: don't forget to set the component names :)

src/api/libcellml/model.h
```diff
@@ -67,6 +67,78 @@ class LIBCELLML_EXPORT Model: public Nameable
...
+    Component takeComponentAt(size_t index);
+
+    /**
+     * @brief replace a component at index.
+     * Replaces the component at index with c.  Index must be a valid index
+     * position.
+     * @param index Index of the component to replace.
+     * @param c The component to use for replacement.
+     */
+    void replaceComponent(size_t index, const Component& c);
```
nickerso: all index arguments should be clarified to be zero-based indices.

tests/model/model.cpp
```diff
@@ -112,6 +112,99 @@ TEST(Model, add_components) {
...
+    libcellml::Component c1, c2;
+    c1.setName("child1");
+    c2.setName("child2");
+    m.addComponent(c1);
+    m.addComponent(c2);
+
+    m.removeComponent(0);
+    EXPECT_EQ(1, m.componentCount());
+    EXPECT_THROW(m.removeComponent(1), std::runtime_error);
+
```
nickerso, line 156: should also serialise and check the model is correct.

tests/model/model.cpp
```diff
@@ -112,6 +112,99 @@ TEST(Model, add_components) {
...
+    m.addComponent(c1);
+    m.addComponent(c1);
+    m.removeComponent("child1");
+    EXPECT_EQ(1, m.componentCount());
```
nickerso: I would expect the above `removeComponent` to only remove one of the `c1` copies from the model and therefore the component count would be 2 in this case.

tests/model/model.cpp
```diff
@@ -112,6 +112,99 @@ TEST(Model, add_components) {
...
+    c1.setName("child1");
+    c2.setName("child2");
+    m.addComponent(c1);
+
+    c2 = m.getComponent(0);
+    c2.setName("childA");
+
+    std::string a = m.serialise(libcellml::CELLML_FORMAT_XML);
+    EXPECT_EQ(e, a);
+}
```
nickerso: This test demonstrates behaviour I would expect.

src/api/libcellml/model.h
```diff
@@ -67,6 +67,78 @@ class LIBCELLML_EXPORT Model: public Nameable
...
+
+    /**
+     * @brief take the component at the given index and return it.
+     * Removes the component at the given index position and returns it.
+     * If an invalid index is passed to the method a std::runtime_error
+     * exception is thrown.
+     * @param index the index of the component to take.
+     * @return the component at the given index.
+     */
+    Component takeComponentAt(size_t index);
```
nickerso: following the other `*Component` methods, I would expect to have a `takeComponent(size_t index)` and a `takeComponent(const std::string& name)` method.

tests/model/model.cpp
```diff
@@ -112,6 +112,99 @@ TEST(Model, add_components) {
...
+    c1.setName("child1");
+    c2.setName("child2");
+    m.addComponent(c1);
+    m.addComponent(c2);
+
+    libcellml::Component c = m.takeComponentAt(1);
+    EXPECT_EQ(1, m.componentCount());
+
+    EXPECT_EQ("child2", c.getName());
+}
```
nickerso, line 213: apart from the issue with the `takeComponentAt` method name, this behaves as I would expect.

tests/model/model.cpp
```diff
@@ -112,6 +112,99 @@ TEST(Model, add_components) {
...
+    m.addComponent(c2);
+
+    std::string a = m.serialise(libcellml::CELLML_FORMAT_XML);
+    EXPECT_EQ(e_orig, a);
+
+    m.replaceComponent(1, c3);
+
+    a = m.serialise(libcellml::CELLML_FORMAT_XML);
+    EXPECT_EQ(e_after, a);
+}
```
nickerso, line 319: seems reasonable to me.

hsorby:
When #34  has been merged, this pull request should be reviewed to see if it covers what is described there.  I believe it does.

src/api/libcellml/component.h
```diff
@@ -49,7 +48,109 @@ class LIBCELLML_EXPORT Component: public Nameable
      */
     void addComponent(const Component &c);
```
agarny: The casing for the description of `serialise()` and `addComponent()` is not consistent with the other descriptions.

src/api/libcellml/component.h
```diff
@@ -49,7 +48,109 @@ class LIBCELLML_EXPORT Component: public Nameable
...
+    /**
+     * @brief Get a component at index.
+     * Returns a reference to a component at the index @p index.  If the index
+     * is not valid a std::out_of_range exception is thrown.
+     *
+     * This is an overloaded method.
+     * @param index The index of the Component to return (zero-based).
+     * @return A reference to the Component at the given index.
+     */
+    Component& getComponent(size_t index);
```
agarny: Shouldn't we also have `getComponent()` functions that take a `const std::string &name` as a parameter (i.e. something similar to what has been done for `removeComponent()`, `takeComponent()` and `replaceComponent()`)?

src/api/libcellml/model.h
```diff
@@ -61,13 +61,114 @@ class LIBCELLML_EXPORT Model: public Nameable
     std::string serialise(libcellml::CELLML_FORMATS format) const;
```
agarny: The casing for the description of the `serialise()` method should be corrected.

tests/component/component.cpp
```diff
@@ -45,6 +45,168 @@ TEST(Component, unset_name) {
...
+    libcellml::Component c;
+    libcellml::Component c1, c2;
+    c1.setName("child1");
+    c2.setName("child2");
+
+    EXPECT_FALSE(c.containsComponent("child1"));
+
+    c.addComponent(c1);
+    c.addComponent(c2);
+    EXPECT_TRUE(c.containsComponent("child2"));
```
agarny, line 85: Might also want to add:

```
EXPECT_TRUE(c.containsComponent("child1"));
```

or drop `c2` and do everything with `c1`. (The same goes for `TEST(Model, contains)`.)

tests/model/model.cpp
```diff
@@ -112,6 +112,212 @@ TEST(Model, add_components) {
...
+//TEST(Model, simplified_replace) {
+//    std::vector<structure> ss;
+//    structure s1, s2, s3;
+//    ss.push_back(s1);
+//    ss.push_back(s2);
+//    ss.erase(ss.begin());
+//    ss.insert(ss.begin(), s3);
+
+//}
+
```
agarny, line 289: Shouldn't all that code for `big_and_complicated`, `structure` and `TEST(Model, simplified_replace)` be removed since it's not being used and doesn't seem to be related to the issue at hand?

agarny:
Some comments:
- Common methods between `Model` and `Component`: both classes have quite a few methods in common (e.g. `getComponent()` and `removeComponent()`). So, rather than having them duplicated in both `Model` and `Component`, I would try to have some kind of a class from which `Model` and `Component` would inherit, and that would both define and implement those methods (and, to some extent, tests)? This would certainly avoid all the duplicated code, but also avoid potential problems in the future (e.g. you need to update one of those methods and, by mistake, only do it for one of the two classes).
- Throwing exceptions: I believe this has been quickly discussed at some point, but I am not sure whether agreement has been reached. Whatever the case, I can see that some methods (e.g. `removeComponent()`) can throw exceptions. Personally, I really hate having to rely on `try...catch` statements to do things. Not only that, but the [Google C++ Style Guide](https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Exceptions) recommends not using them.

nickerso:
Regarding exceptions, by returning references (and const references) to objects exceptions is the only way to fail since there is no `NULL` reference and we don't want to be creating a dummy object and returning that. In your own code you could first test for a component's existence and then you can remove it without needing a `try...catch` block. But the library itself needs some way to fail.

So the question isn't really whether to use exceptions or not, but whether we want to return references, pointers, or copies of the objects. My vote is for references, as per containers in the standard library.

agarny:
I understand what you are saying about exceptions being needed when wanting to return a (const) reference.

As for writing my code in such a way as to avoid exceptions, well... I am really not keen on that approach (it would kind of go against the way the API would be designed).

Anyway, personally, I am in favour of returning pointers (and thus not having to rely on exceptions).

src/component.cpp
```diff
@@ -136,5 +138,72 @@ void Component::addComponent(const Component &c)
     mPimpl->mComponents.push_back(c);
 }
 
+void Component::removeComponent(const std::string& name)
+{
+    auto result = mPimpl->findComponent(name);
+    if(result != mPimpl->mComponents.end()) {
+        mPimpl->mComponents.erase(result);
+    } else {
+        throw std::out_of_range("Named component not found.");
```
rchristie: It would be handy if these exception strings had the function name in them... e.g. __func__ or __FUNCTION__, as long as it's portable.

src/api/libcellml/component.h
```diff
@@ -49,7 +48,109 @@ class LIBCELLML_EXPORT Component: public Nameable
...
+    /**
+     * @brief Get a component at index.
+     * Returns a const reference to a component at the index @p index.  If the
+     * index is not valid a std::out_of_range exception is thrown.
+     *
+     * This is an overloaded method.
+     * @param index The index of the Component to return (zero-based).
+     * @return A const reference to the Component at the given index.
+     */
+    const Component& getComponent(size_t index) const;
```
rchristie: Just highlighting what may be obvious: if the client calls any methods on the returned reference /after/ any other code removes the component, it'll crash (although I don't know how deterministic it is, whether it is catchable in a useful way etc.). This is admittedly a dumb thing to do; I'm just pointing out it's possible.

rchristie:
I'm very in favour of APIs that throw exceptions for all bad inputs, unavailability of memory etc. It means clients can just write code and expect it to work. It's the modern way.
There are grey areas though. Checking for existence before getting can halve performance, but this only matters for high performance APIs which libCellML is surely not.

agarny:
I am happy to believe that it's the modern way, but does it necessarily imply that it's the best way to do things? :)

Otherwise, I wonder how "it means clients can just write code and expect it to work"? If clients don't properly handle exceptions, then your application will likely crash (as it will if you don't properly handle an API that returns pointers, I agree).

Indeed, libCellML is not a high performance API, yet the current CellML API is rather 'slow'. I am certainly hoping that libCellML will be lightingly fast in comparison (so far, it looks very good to me).

rchristie:
Code that checks return codes exhaustively is both less efficient and less readable. When return codes are not checked, errors go undetected (and the reputation of the library goes down).
It seems the main argument against exceptions in the google coding standards is to fit in with existing software that hopes to use it, particularly if retrospectively adding it into existing APIs.
It shouldn't be necessary to try-catch exceptions in your use of any API: if you use it exactly as designed you should not get any exceptions unless really unexpected things such as running out of memory occur, in which case ending the program is probably right.

rchristie:
Regarding an earlier statement about the standard library: STL container methods more commonly take or return iterators than references (consider find(), begin(), erase(); references /are/ returned by the operator[] and equivalent at() methods though). Iterators are more like pointers and they can have an invalid value i.e. the container's end() method.

Issue #31, opened 2015-07-01
Make coverage testing availble on OS X


hsorby:
gcov is available on OS X so we should make the coverage testing functionality available on this platform

hsorby:
There is a handy tool called gcovr (http://gcovr.com/) that we should make use of to enable this.  Enable is probably the wrong word, we should use gcovr because it will (I think) work across both GNU/Linux and OS X without having to do platform specific checks. And it also has a handy html output format that would be useful for code review.

hsorby:
This would require some adjustments to the Buildbot setup to get this working.

PR #37, opened 2015-07-09
Prefer gcovr over gcovcoverage.

hsorby:
By using gcovr we can enable coverage testing on OS X.  It will also give us a nicer summary of the coverage and a very nice html page output that we can add to the documentation.

Addresses issue #31


nickerso:
Does this one rely on other changes, such as f6d17b0? I can't get coverage enabled on my OS X build...

hsorby:
The intention was that it would be self contained.  I can't see any reason why it would rely on  f6d17b0.  It works as expected on the build machine.

nickerso:
there is a test in `cmake/environmentchecks.cmake` that will only test for gcov if using a GNU C++ compiler. Do you expect coverage testing to work on OS X with this change? or is this just setting up the linux builds to use gcovr?

hsorby:
That should work better!  Buildbot automatically merges the branch to be tested into the develop branch before running the tests, so if you don't do it you will get different results.

nickerso:
ok, thats much better. Works well and I can get a nice coverage HTML page. I think this change should be merged in.

agarny:
It's all working fine for me too, although I am not sure to what the branches coverage corresponds and whether we should be 'concerned' about those low percentages for `model.cpp` and `component.cpp`:

image

Still, `gcovr` is set up and it's doing its job, so I am going to merge.

hsorby:
A bit of information on the branches displayed here by gcovr (to the best of my current understanding).

First a branch is a transition from one basic block of assembler code to another.  A basic block is a chunk of machine code that once started executing will continue to do so until the end of the basic block.  So in assembler a branch is identified by things like a jump statement.

This then creates quite understandable branches in C code, in C++ the situation is not the same.  For example, in C++ a class constructor can throw an exception, so if the code that is tested doesn't make the class constructor throw an exception then that is a branch that you haven't covered.  In C++ there are a lot of these types of branches that may not be immediately apparent.

For now I think we should take a wait and see approach, hopefully the covered branch percentage rises as we create more code and have a smaller ratio of untriggered exceptions to actual lines of code.  If not, we may have to revisit this.

agarny:
Thanks for the clarification @hsorby.

Issue #39, opened 2015-07-12
Create documentation out-of-source


hsorby:
It would be a good idea to be able to generate the documentation outside of the source directory.  This keeps the source directory clean and keeps to the same philosophy that we have when building the library.

It should also be possible to create the documentation through a command in the build directory like <code>make docs</code>.  Of course, the documentation you can create will be dependent on the tools available, currently we require:
- Sphinx - For the project documentation
- Doxygen- For the API documentation
- gcov and Python - For the coverage documentation

PR #40, opened 2015-07-14
Docs

hsorby:
Create some documentation targets so that the documentation can be built outside of the source tree just like the code.

Addresses #39


mirams:
Looks good to me

Issue #38, opened 2015-07-12
Return references or pointers?


hsorby:
Now that we are a little way through the implementation and have had a look at how the API is starting to form, there appears to be a difference of opinion on what should be returned through the API:
- references, or
- pointers.

This issue is to decide which route we should take.

This issue was raised in pull request #36

hsorby:
references for me.

rchristie:
For me: neither; I prefer safe handles to objects e.g. smart pointers.

Giving references or raw pointers to internal objects is not safe as there is no guarantee on the lifetime of the referred-to object. You can cause memory corruption simply by using the API; I don't expect a high level API such as libCellML to be that unsafe, but YMMV.

The currently proposed APIs for removing components are not very object oriented, and are flawed:
- indexes can become invalid with other re-ordering methods
- names are not necessarily unique

Why I'm mentioning this here is that I'd expect to supply an actual object handle to remove. This can't be done safely with a reference as the caller is left with an invalid reference. Addresses of pointers could be passed and the pointer cleared, but this is inelegant.

If you work with handles at all times you need fewer APIs:
- one API to get by index
- one API to get by name
- one remove API taking a handle
- take = get + remove

nickerso:
My vote is for references.

agarny:
@rchristie makes some good points. I would be curious to see some sample code for the different approaches.

This aside, and FWIW, it looks (from a very quick glance at their source code) like both [libSBML](http://sbml.org/Software/libSBML) and [libSEDML](https://github.com/fbergmann/libSEDML/) return pointers. I know, it doesn't mean that we should do the same, but for someone coming from SBML and interested in CellML, it might make things easier if we did.

Anyway, at this stage, my vote is still for pointers.

nickerso:
@mirams @jonc125 @mtcooling - would be good to get this resolved, so please express an opinion :)

mirams:
Shared pointers would seem to make sense to me, so that things don't disappear if the API goes out of scope.

jonc125:
I'd also prefer safe handles, and smart pointers are the easiest standard solution. It's easy to envisage a user creating a component, adding it to a model, later deleting the model (or even just removing the component from the model), and wondering why accessing the component now causes the program to crash.

nickerso:
[...]

Issue #35, opened 2015-07-04
Use cases upto 1.vii


hsorby:
The use cases we wish to satisfy are:
- manipulation of a model with multi-level component encapsulation hierarchy
  - remove a component
  - change the name of a top-level component
  - change the name of an encapsulated child component
  - replace one component with a new component
  - take a component (remove the component and return it to the user)
  - determine if a component with a given name exists in a model or component
  - determine the number of components encapsulated by a model or component

nickerso:
Resolved by #42.

PR #42, opened 2015-08-04
Implementation for usecases upto 1vii smart

hsorby:
A variation using smart pointers for managing CellML entities.

Alternative option for issue #35 and help on deciding the outcome of issue #38


src/CMakeLists.txt
```diff
@@ -35,15 +35,20 @@ set(SOURCE_FILES
   ${CMAKE_CURRENT_SOURCE_DIR}/version.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/model.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/component.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/componententity.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/namedentity.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/entity.cpp
```
agarny, line 37: I would favour listing files in alphabetical order. It makes it easier to check/review/etc.

src/CMakeLists.txt
```diff
@@ -35,15 +35,20 @@ set(SOURCE_FILES
...
 set(API_HEADER_FILES
   ${CMAKE_CURRENT_SOURCE_DIR}/api/libcellml/version.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/api/libcellml/types.h
   ${CMAKE_CURRENT_SOURCE_DIR}/api/libcellml/model.h
   ${CMAKE_CURRENT_SOURCE_DIR}/api/libcellml/component.h
   ${CMAKE_CURRENT_SOURCE_DIR}/api/libcellml/entity.h
   ${CMAKE_CURRENT_SOURCE_DIR}/api/libcellml/enumerations.h
-  ${CMAKE_CURRENT_SOURCE_DIR}/api/libcellml/nameable.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/api/libcellml/namedentity.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/api/libcellml/componententity.h
```
agarny: Same as above, i.e. list files alphabetically.

src/api/libcellml/component.h
```diff
@@ -1,11 +1,24 @@
...
+limitations under the License.Some license of other
+*/
+
 #ifndef LIBCELLML_LIBCELLML_COMPONENT_H
 #define LIBCELLML_LIBCELLML_COMPONENT_H
 
-#include <string>
-#include <memory>
-
 #include "libcellml/libcellml_export.h"
```
agarny, line 20: Besides listing includes in alphabetical order (and the same throughout the code), I was wondering whether it wouldn't be better to have `export.h` (and `config.h`) rather than `libcellml_export.h` (and `libcellml_config.h`)? I just don't see the point of having `libcellml_` in front, not to mention that using `_` kind of breaks our current file name convention, no?

src/api/libcellml/componententity.h
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * @brief Replace a component with the given name.
+     * Replaces the component with the given name @p name with @p c.  Name must be a valid name
+     * of a component in the Component, if it isn't a std::out_of_range exception is thrown.
+     *
+     * This is an overloaded method.
+     * @param name The name of the Component to replace.
+     * @param c The Component to use for replacement.
+     */
+    void replaceComponent(const std::string &name, const Component_Ptr& c);
```
agarny: `void replaceComponent(const std::string &name, const Component_Ptr &c);`

rather than

`void replaceComponent(const std::string &name, const Component_Ptr& c);`

src/api/libcellml/entity.h
```diff
@@ -25,17 +27,100 @@ namespace libcellml {
...
+
+    /**
+     * @brief Serialise entity to std::string.
+     * @param format The format to serialise the object to.
+     * @return std::string representation of the object.
+     */
+    std::string serialise(libcellml::CELLML_FORMATS format) const
+    {
+        return doSerialisation(format);
+    }
```
agarny: I would prefer the function body to be in a corresponding `.cpp` file. The same goes with `getParent()`, `setParent()`, etc. below.

src/api/libcellml/types.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef LIBCELLML_LIBCELLML_TYPES_H_
+#define LIBCELLML_LIBCELLML_TYPES_H_
+
+#include <memory>
+
+namespace libcellml {
+
+class Model; /**< Forward declaraion of Model class. */
+class Component; /**< Forward declaration of Component class. */
+typedef std::shared_ptr<Component> Component_Ptr; /**< Type definition for shared component pointer. */
```
agarny: Shouldn't we avoid using `_` in class names? In other words, shouldn't we use `ComponentPtr` rather than `Component_Ptr`?

src/componententity.cpp
```diff
@@ -0,0 +1,222 @@
...
+ComponentEntity::ComponentEntity(const ComponentEntity& rhs)
+    : NamedEntity(rhs)
+    , mPimpl(new ComponentEntityImpl())
+{
+    mPimpl->mComponents = rhs.mPimpl->mComponents;
+}
+
+ComponentEntity& ComponentEntity::operator=(ComponentEntity c)
+{
+    NamedEntity::operator =(c);
```
agarny: Why a space between `operator` and `=`?

tests/component/component.cpp
```diff
@@ -45,16 +45,193 @@ TEST(Component, unset_name) {
...
+    c.removeComponent("child1");
+    EXPECT_EQ(2, c.componentCount());
+    a = c.serialise(libcellml::CELLML_FORMAT_XML);
+    EXPECT_EQ(e2, a);
+    EXPECT_THROW(c.removeComponent("child3"), std::out_of_range);
+}
+
+TEST(Component, getcomponent) {
+    const std::string e1 = "<component/><component name=\"childA\"/><encapsulation><component_ref><component_ref component=\"childA\"/></component_ref></encapsulation>";
+    const std::string e2 = "<component name=\"parent\"/><component name=\"gus\"/><component name=\"childB\"/><component name=\"child3\"/><component name=\"gus\"/><component name=\"childB\"/><component name=\"child3\"/><encapsulation><component_ref component=\"parent\"><component_ref component=\"gus\"><component_ref component=\"childB\"/><component_ref component=\"child3\"/></component_ref><component_ref component=\"gus\"><component_ref component=\"childB\"/><component_ref component=\"child3\"/></component_ref></component_ref></encapsulation>";
```
agarny, line 143: Not sure why the name of the components got changed...

tests/component/component.cpp
```diff
@@ -45,16 +45,193 @@ TEST(Component, unset_name) {
...
+    EXPECT_THROW(c.getComponent(4), std::out_of_range);
+
+    libcellml::Component_Ptr cAr = c.getComponent("gus");
+    EXPECT_EQ("gus", cAr->getName());
+
+    // Modify a deeper Component
+    c.setName("parent");
+    c.addComponent(c1);
+    c1->addComponent(c2);
+    c1->addComponent(c3);
```
agarny, line 173: Same here, not sure why `c1` is now added to `c` before `c2` and `c3` are added to `c1`. Just 'cleaning up' the code?

agarny:
This looks good to me. Just some minor comments/corrections.

src/api/libcellml/component.h
hsorby, line 20: Unfortunately this is a by-product of the CMake build system, it is generated for us.  We can change it but the effort is probably not worth the reward.  Also config.h is probably the worst name to give to a header file that is available publicly (even if it is removed behind a libcellml prefix).

src/componententity.cpp
hsorby: Auto complete in action.

tests/component/component.cpp
hsorby, line 173: I was thinking of adding another example of the other way to show behaviour, but (obviously) didn't follow thought with action.

src/api/libcellml/component.h
agarny, line 20: Fair enough, although a quick and dirty way to 'fix' it would be to rename the file after it has been generated by CMake. This would also allow to rename the file, if 'needed'. In other words, we could have a simple CMake macro that is used to generate those configuration files and rename them, if needed?

nickerso, line 20: I'd prefer to leave the default CMake behaviour alone for these sorts of files. Should help ease maintainability in future...

agarny, line 20: What CMake behaviour could break things here? The only thing I can think of is the name of the file it generates, something that would affect whatever solution we go for. Anyway, consistency is my main motivation here, but I am 'fine' either way.

hsorby, line 20: I think adding an issue about the name of the generated file would be appropriate.  At a later time we can revisit this and see if something can be done, at this point in time I think time is better spent progressing along the functionality path set out in the roadmap.

agarny, line 20: Agreed, I am going to create an issue for it, but in the meantime we should indeed focus on making progress.

Issue #44, opened 2015-09-08
Support use case 1.viii: a model with imported components


nickerso:
This issue is for implementing support for use case 1.viii, creating a model with imported components.

hsorby:
I think we might need to change the use-cases a little to fit the direction that we have taken with the codebase.

PR #45, opened 2015-09-10
starting to play around with options for component imports

nickerso:
some ideas on how we might want to think about handling component imports. Just adding some test code that is not actually getting built or tested. Aimed at addressing #44.


hsorby:
Having had a talk to @nickerso about this, I actually think option two is the best option.  We do need to be able to represent imports so that we can manage which component is imported from which model and the separation of the imports.  Consider case A where I want to import two components from the same model but import them independently.  I would describe this as (roughly)

```
<import name=bob src=fred><component name=dave></image>
<import name=alf src=fred><component name=alf></image>
```

alternatively consider case B where I want to import two components from the same model but import them as a whole, thus including their connections (possibly among other things)

```
<import name=bob src=fred><component name=dave><component name=alf></image>
```

To achieve this I think having an object in the model to represent this is appropriate, certainly from a software point of view.

nickerso:
can't help myself, need to correct the CellML code:

```
<import xlink:href=bob.xml><component name=dave component_ref=fred/></import>
<import xlink:href=bob.xml><component name=alf component_ref=andre/></import>
```

and

```
<import xlink:href=bob.xml>
    <component name=dave component_ref=fred/>
    <component name=alf component_ref=andre/>
</import>
```

nickerso:
the other reason why we'd need to use something like option 2 is that we don't want to have to instantiate  and/or load an external model in order to create an imported component in the current model. The external model may not even exist, and its not until we need to instantiate a model to do something with it, such as validation, that we want to have to load/resolve those external model references.

agarny:
I am not sure I understand David's last comment, but this aside isn't the issue discussed by Hugh about the serialisation of the model?

If so, then all the current suggestions could result in the 'correct' serialisation of the model. It's just that, from a software point of view, it would be a bit more work to do so using options 1 or 3.

This aside too, is there anything in the CellML specifications that prevents someone from having something like the first serialisation? I would certainly consider both serialisations to be correct, so once again it would be up to libCellML to realise that the first serialisation refers to two components being imported from the same CellML file.

... or did I miss something?...

hsorby:
I think your understanding is correct, but I'll let @nickerso  enlighten on what he was writing about.  I would just like to comment on the software point of view.

All the current suggestions would result in a valid serialisation but when a user of the library is dealing with libcellml objects it seems to me to be more obvious to deal with import objects to manage imports rather than hiding or dual purposing components.  Of course this is from a software perspective, usually you think of using C++ to model the domain of the problem so it seems to be a better way of thinking about this.

agarny:
If I am thinking of the `CellML Text` view in OpenCOR, then it's true that having an import object (in C++ terms) would make my life much easier, but I thought the idea of libCellML was to 'hide' the underlying CellML format as much as possible (hence my not going for option 2)?

hsorby:
Yes we don't want to provide direct access to the CellML format but that was more to do with not letting people get their hands on the XML DOM.  We want to provide C++ objects to allow users to easily and intuitively manipulate models in software, for this I think it is better to provide an import object.

agarny:
Ok, then I am in favour of an import class and with, therefore, option 2.

nickerso:
I have updated my suggestions to make use of the import (or similar) class. Still at least a couple of different options...

tests/model/component-import.cpp
```diff
@@ -0,0 +1,138 @@
...
+    importedComponent->setName("component_in_this_model");
+    importedComponent->setSourceComponent(imp, "a_component_in_that_model");
+    m.addComponent(importedComponent);
+
+    std::string a = m.serialise(libcellml::CELLML_FORMAT_XML);
+
+   EXPECT_EQ(e, a);
+}
+
+TEST(ComponentImport, multipleImport02) {
```
agarny: Shouldn't this read:

```
TEST(ComponentImport, multipleImport01) {
```

since in this test, you explicitly add the import to the model (unlike in `singleImport02`)?

tests/model/component-import.cpp
```diff
@@ -0,0 +1,138 @@
...
+    c3->setName("c3");
+    c3->setSourceComponent("cc1");
+    imp.addComponent(c3);
+
+    std::string a = m.serialise(libcellml::CELLML_FORMAT_XML);
+
+   EXPECT_EQ(e, a);
+}
+
+TEST(ComponentImport, multipleImport01) {
```
agarny: As a result, shouldn't this read:

```
TEST(ComponentImport, multipleImport02) {
```

?

agarny:
I like the first option best, i.e. the one where the import object is explicitly added to the model object.

tests/model/component-import.cpp
nickerso: thanks @agarny - silly copy-paste error :) sorted now.

hsorby:
I prefer singleImport02 where the import object is just an object to reference components from

nickerso:
I think I also prefer the *02 version. I think it makes more sense to simply use the import object as a source for components and keep the model object as the component container.

Just tagging @mirams @jonc125 and @mtcooling to see if they have any thoughts. It would be good to at least pick a preliminary favourite so that implementation can get underway and find the problems with whichever approach we choose :)

jonc125:
I have a slight preference for the *02 version as well. It looks nicer just to add components to the model, but keep a reference for where they originated - makes imported components more similar to local components.

mirams:
I think I like the '02 version best too. More intuitive to add just the bit you need to the model I think?

agarny:
*01 offers a better mapping with the CellML format, which is the reason I went for *01, but I can see the rationale for going with *02, so I am happy to go with it too.

nickerso:
@agarny it is important to remember that the XML serialisation is just an exchange format, not a blueprint for an object model :) We are kind of hamstrung by the fact that we have all been stuck dealing directly with the XML structure for so long that it has become what we are used to seeing and comfortable dealing with it. But we have to try and forget that and imagine how much better things could be! And then ask @hsorby very nicely to implement our imaginings ;-)

agarny:
Indeed and I welcome (and will keep welcoming) being reminded about it. In fact, that's the approach Frank has taken with libSEDML and the one I am also taking with my (very basic) COMBINE Archive API, and I like it that way. I guess I am just too 'familiar' with raw CellML... :)

mtcooling:
I also prefer 02.

hsorby:
Looking at the failed test output (shown below) the actual value and expected values are functionally the same but the strings, which are used to determine success or failure, are not.

/Users/libcellml/buildslave/Unit_Tests_Builder_Mac_OSX_10_9/libcellml/tests/model/component_import.cpp:143: Failure
Value of: a
Actual: 
`"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\"><import xlink:href=\"some-other-model.xml\" xmlns:xlink=\"http://www.w3.org/1999/xlink\"><component component_ref=\"cc1\" name=\"c1\"/><component component_ref=\"cc2\" name=\"c2\"/></import><import xlink:href=\"some-other-model.xml\" xmlns:xlink=\"http://www.w3.org/1999/xlink\"><component component_ref=\"cc1\" name=\"c3\"/></import></model>"`
Expected: e
Which is: 
`"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/1.2#\"><import xlink:href=\"some-other-model.xml\" xmlns:xlink=\"http://www.w3.org/1999/xlink\"><component component_ref=\"cc1\" name=\"c3\"/></import><import xlink:href=\"some-other-model.xml\" xmlns:xlink=\"http://www.w3.org/1999/xlink\"><component component_ref=\"cc1\" name=\"c1\"/><component component_ref=\"cc2\" name=\"c2\"/></import></model>"`

What to do? Options are:
1. Force the order of the output to match the expected order from inside libCellML.
2. Use an xml parser ???? to compare across elements regardless of order.

Opinions requested (or other options).

agarny:
I have come across the same issue in OpenCOR with tests that need to compare XML outputs. Personally, OpenCOR always sorts the attributes of an XML element (using a small Qt-based function that I wrote), even though I know that the order of attributes doesn't matter in XML, but it does indeed make it easier to compare outputs.

So, either you do something similar (which I wouldn't mind since it would indeed make it easier to compare CellML files) or you can pre-process your XML strings before comparing them in your tests ([here](http://dalelane.co.uk/blog/?p=3225) is some Python code that should help you with that).

mtcooling:
As long as the solution to this doesn't violate:
"
1. Load a CellML 1.2 model, make changes to it and save it.
   1. Ability to preserve the underlying XML structure/ordering for
      documents read in and written out.

"
[http://libcellml.readthedocs.org/en/latest/roadmap.html]
1. sounds much better to me, since it fixes that problem at the source and
   no where else. ...if you can find a library that does it :) Seems like this
   might interact with the tinyxml discussion. I'm afraid I don't know the
   best library to use off the top of my head.

my 2c,

nickerso:
I agree, using an XML tool to compare documents is probably the way to go and something that will be needed anyway. Probably what we can do is normalise the serialised document so that document ordering is deterministic and then compare that to the test string using the standard string comparison. We will need to update the current test strings to match our determined ordering and whitespace normalisation.

One way to achieve this would be to use XSL to normalise whitespace and sort the document. This would add dependencies on an XML parser and a XSLT library (e.g., libxml2/libxslt or Xerces/Xalan). An alternative is to use just an XML parser and probably implement the normalisation ourselves (using an algorithm similar to that Python code linked above). This can probably be done using TinyXML and thus keep our dependencies limited.

hsorby:
Having an internal dependency like tinyxml for me is preferrable and maybe some small changes to the tinyxml2 codebase will normalise the output enough to enable us to do string comparisons again.

Of course, as @mtcooling reminds us, we may not preserve the underlying XML structure in this situation.  [passing a document through libCellML and then using that as your starting point would.]

hsorby:
So I have worked around my issue of not knowing the order that multiple imports are serialized in by considering all possible orders.  As can be seen from the green tick (nickerso@3159525) this has satisfied the tests.

I did as an initial attempt make use of libxml2 canonicalization, but in the end I didn't actually require it so I removed it.  (The canoncialization didn't order the elements in a determinable order.)  You can see my libxml2 implementation in commit nickerso@a51ae28 for the interested.

If someone could review this work that would be great, cheers.

agarny:
Argh, sorry, but I don't like it. :(

I can imagine your approach making things rather difficult in the future (with 'complex' tests). Since we are going to use `libxml2`, what is wrong with implementing some kind of normalisation ourselves (i.e. what @nickerso suggested above)?

nickerso:
@agarny - it wasn't that clear in the comments, but if you look at a51ae28 and c5daaaf you can see where @hsorby tried to use libxml2 to canonize the models but it still didn't help with this issue. Ideally we would use libxml2 to create the canonical XML document, but this still appears to not reproducibly produce the same XML string for test purposes across all three test platforms. Would be great if someone had an idea of how we could make it work!

Regarding the work-around that @hsorby implemented here, it seems like a reasonable approach for now based on the philosophy we have adopted based on getting the tests to pass. This is obviously not going to be a good solution for larger tests, but I think it is sufficient for this particular test.

hsorby:
It really comes down to the fact that CellML is declarative and it really doesn't matter in which order elements appear because of this we cannot determinable say which element should appear before any other.  As it really shouldn't matter which order the elements appear in then this test is exactly what we want, telling us that the order is unimportant.

We could break this down to just look for the elements that we expect to see in the document and not concern ourselves with the whole picture.  A strategy that may serve when the documents get a bit bigger.

agarny:
My bad, I was still focussing on the order of attributes within a CellML element. Now, I understand the problem better and, as can be expected, the [Python script](http://dalelane.co.uk/blog/?p=3225) I mentioned earlier doesn't solve that problem either. However, I guess the problem could be solved by merging identical CellML elements (i.e. the `import` element in the current test) and then ordering its child elements based the value of their respective attributes.

Ok, this is indeed more 'complex' than I had originally anticipated. As a result, I am 'happy' with @hsorby's current 'solution'...

PR #50, opened 2015-11-17
Update Object model images.

hsorby:
Addresses issue #49


nickerso:
Seems good to me and matches the current implementation.

mirams:
closes #49

agarny:
Although it has been merged and #49 closed, I have a few comments:
- `Entity`:
  - There is no `deserialise()` method.
  - `setParent(Component *)` rather than `setParent(ComponentPtr)` and `setParent(Model *)` rather than `setParent(Model)`.
  - `hasParent(Component *)` rather than `hasParent()`.
- `Import`:
  - Missing `mReference`.
  - No space before the closing bracket in `setSource(std::string)`.
  - Missing `getSource() const: std::string`.
- `NamedEntity`:
  - No need to name the parameter (`name`) of `setName()`.
- `ImportedEntity`:
  - `mImportReference` rather than `mImportedReference`.
  - Missing `isImport() const: bool`.
  - `getImport() const: ImportPtr` rather than `getImport(): ImportPtr` and `getImportReference() const: std::string` rather than `getImportReference(): std::string`.
- `ComponentEntity`:
  - Missing `addComponent(ComponentPtr): void`.
  - Missing `removeComonent(std::string): void`.
  - Note: `containsComponent()` should probably be made `const`.
  - Various `getComponent()` and `takeComponent()` methods missing or 'wrong'.
  - Missing `replaceComponent(std::string, ComponentPtr): void`.
  - Missing `componentCount() const: int`.
- `Model`:
  - There is no `addComponent()` method.
- `Component`:
  - There is no `addComponent()` method.
  - Missing `setSourceComponent(ImportPtr, std::string): void`.

A few other comments:
- Not related to this issue _per se_, but if we are to use the pImpl idiom (e.g. for `Import`), then shouldn't we be using it for our other classes (e.g. `Entity`)?
- Wouldn't it be good to specify the type of a class property?
- Surely, there must exist some kind of a tool that can automatically generate this type of diagram from the source code?

Issue #48, opened 2015-11-17
Support use case 1.ix: a model with units


hsorby:
This issue is for implementing support for use case 1.ix, creating a model with units.

PR #52, opened 2015-11-17
Add initial tests for Units, including skeleton Units class.

hsorby:
Starting to work towards addressing issue #48


hsorby:
If people could have a look at the tests I have put together:

https://github.com/hsorby/libcellml/blob/76c7d4bec413e68ea6b99bb391cd685705f5fb09/tests/units/units.cpp

and see what they think (the link above is to the full file of tests the skeleton implementation isn't very important at this stage.)

nickerso:
I would be more inclined to go with the string units name when adding a unit, but could have the list of standard (built-in) units available as string constants rather than an enum.

Otherwise its all looking good to me.

src/api/libcellml/enumerations.h
```diff
@@ -31,6 +31,83 @@ enum CELLML_FORMATS
      CELLML_FORMAT_XML = 0 /**< XML format for CellML files. */
 };
 
+/**
+ * @brief The CELLML_PREFIXES enum.
+ * Enum dexcribing the prefixes that are available for
```
agarny: describing, not dexcribing.

src/api/libcellml/units.h
```diff
@@ -0,0 +1,157 @@
...
+     * Set the base unit flag to signify whether this Units is a base unit
+     * or not.  This flag cannot be set if this Units is a compound unit.
+     * @param state boolean value to set the base unit flag to.
+     */
+    void setBaseUnit(bool state=true);
+
+    /**
+     * @brief Add a unit to this Units.
+     * Add a unit as a child of this Units.  This method takes optional arguments
+     * scale, and offset.
```
agarny: Clearly not important, but why the comma?

src/api/libcellml/units.h
```diff
@@ -0,0 +1,157 @@
...
+     * @param prefix The prefix for the unit, one of CELLML_PREFIXES.
+     * @param exponent The exponent.
+     * @param scale The scale.
+     * @param offset The offset.
+     */
+    void addUnit(const std::string &units, CELLML_PREFIXES prefix, double exponent,
+                 double scale=1.0, double offset=0.0);
+
+    /**
+     * @brief Add a unit to this Units.
```
agarny, line 115: Make it clear that it is a CellML standard unit? (Same in the rest of the class.)

src/api/libcellml/units.h
```diff
@@ -0,0 +1,157 @@
...
+    void setBaseUnit(bool state=true);
+
+    /**
+     * @brief Add a unit to this Units.
+     * Add a unit as a child of this Units.  This method takes optional arguments
+     * scale, and offset.
+     * @param units The name of the unit to add.
+     * @param prefix The prefix for the unit, one of CELLML_PREFIXES.
+     * @param exponent The exponent.
+     * @param scale The scale.
```
agarny: If we respect the CellML lingua, we should use 'multiplier' rather than 'scale'.

tests/units/units.cpp
```diff
@@ -0,0 +1,163 @@
...
+    u->setName("valid_name");
+
+    /* Base unit is false by default */
+    EXPECT_FALSE(u->isBaseUnit());
+    u->setBaseUnit();
+    EXPECT_TRUE(u->isBaseUnit());
+    u->setBaseUnit(false);
+    EXPECT_FALSE(u->isBaseUnit());
+    u->setBaseUnit(true);
+    EXPECT_TRUE(u->isBaseUnit());
```
agarny, line 48: Just to make the test more relevant, I would swap lines 45-46 with lines 47-48.

src/api/libcellml/enumerations.h
```diff
@@ -31,6 +31,83 @@ enum CELLML_FORMATS
...
+ * units.
+ */
+enum CELLML_PREFIXES
```
agarny: Since we are in the `libcellml` namespace, is there really a need to have `CELLML` everywhere? In other words, why not simply have `PREFIXES`, `PREFIX_YOTTA`, `PREFIX_ZETTA`, etc. (Same for the standard units, and the formats, in fact.)

agarny:
Regarding, @nickerso's comment about going with string units name, I agree that it would be nice. In fact, we could have both strings and the enum, and we could also do the same for prefixes.

src/api/libcellml/units.h
hsorby: Artefact from when there were more optional arguments.

src/api/libcellml/units.h
hsorby, line 115: I was thinking that this would allow users to add units that were not standard units like fahrenheit etc. so it is not just for CellML standard units.

tests/units/units.cpp
hsorby, line 48: line 43 sets the base unit flag true, then we turn it off again and check that the flag is false and finally turn the flag on again in a different way so that the unit is output with the base unit flag set.

src/api/libcellml/units.h
nickerso, line 115: we definitely need to be able to addUnit with user defined units.

hsorby:
So I have made some changes, now I have only strings for the units and the prefixes can be defined as either enums or integers.  The reason for this is so that you cannot put a prefix in the place of a unit and a unit in place of a prefix, which is what we would have had if strings had been allowed for both.

tests/units/units.cpp
agarny, line 48: Oops, my bad, I wrongly assumed that line 43 was setting the base unit flag to `false`. I am OK with the test now.

agarny:
Ok, it all looks good to me now.

Now, I don't want to make a mistake, so checking: am I correct in remembering that we agreed that it's 'OK' to merge a 'failing' PR? If so, then whoever confirms this can then merge the PR in my name.

nickerso:
No, I don't think we want to merge until we get the green tick...http://libcellml.readthedocs.org/en/latest/dev_review.html

agarny:
Ok, glad that I asked then (and thanks for reminding me about the review process page)!

agarny:
Side note: reading the review process page... what is the "Documentation Builder link" and what are "entry 2. in step 5." and the like? Then, some nitpicking, e.g. "in its final format" rather than "in it's final format".

nickerso:
> Side note: reading the review process page... what is the "Documentation Builder link" and what are "entry 2. in step 5." and the like? 

If you go to the _details_ link for a given build you can see the described link. This particular one doesn't generate the documentation because the tests fail, but a successful build such as http://autotest.bioeng.auckland.ac.nz/libcellml-buildbot/builders/Tests%20Builder/builds/226 you can see a second entry in step 5 which is a link to the documentation build for that revision and if its successful you'll end up with a link to the `dox` which can then be checked as per the review.

Clear as mud? :)

> Then, some nitpicking, e.g. "in its final format" rather than "in it's final format".

sounds like a separate issue ;-)

agarny:
> If you go to the details link for a given build you can see the described link. This particular one doesn't generate the documentation because the tests fail, but a successful build such as http://autotest.bioeng.auckland.ac.nz/libcellml-buildbot/builders/Tests%20Builder/builds/226 you can see a second entry in step 5 which is a link to the documentation build for that revision and if its successful you'll end up with a link to the `dox` which can then be checked as per the review.

Ok, I must confess that it took me a few seconds, but now I have got it. Wow! :)

hsorby:
At this point we don't want to merge we're just checking that the tests look ok and look like they are testing the functionality we want.  We are free to add more tests if we want at a later date if we miss something now.  Hopefully working in this way we make sure the library is functioning as we expect.

agarny:
> At this point we don't want to merge we're just checking that the tests look ok and look like they are testing the functionality we want. We are free to add more tests if we want at a later date if we miss something now. Hopefully working in this way we make sure the library is functioning as we expect.

I am definitely happy with that way of proceeding.

src/api/libcellml/units.h
```diff
@@ -0,0 +1,182 @@
...
+     * multiplier and offset.
+     *
+     * @overload
+     * @param units The name of the unit to add.
+     * @param prefix The prefix for the unit expressed as an integer.
+     * @param exponent The exponent.
+     * @param multiplier The multiplier.
+     * @param offset The offset.
+     */
+    void addUnit(const std::string &units, int prefix, double exponent,
```
nickerso: could you give `exponent` a default value of `1` and then you wouldn't need the following two methods?

nickerso:
Trivial comment inline, but otherwise this seems sufficient to meet the use cases 1.ix.

nickerso:
Looks good to me, happy to see this one merged in.

agarny:
This all looks fine to me, but... I can't seem to see _1.ix.d_, i.e. multiply 'a single base units with valid name' and 'a units which defines micro-Ampere \* Kelvin / milli-siemens'? (Thinking of it, I am not sure what that test is supposed to test...)

nickerso:
good spotting @agarny, with that test implemented I think we are now all good to merge, so I'll do that.

Issue #54, opened 2015-11-17
Update rough object model diagram to match implementation


hsorby:
@agarny possibly considers the currently accepted rough object model to be too rough.  This issue is to tidy up the diagram and reduce it's roughness.

PR #55, opened 2015-11-17
Add corrections to object model as noted by @agarny.

hsorby:
Addresses issue #54


nickerso:
I'll let @agarny review this one.

agarny:
It all looks good to me. However, I believe I missed something: it isn't `Model` and `Component` that have 0+ `Component`, but `ComponentEntity`. I am happy to merge once that has been fixed.

Otherwise and just in case: regarding `ComponentEntity:: containsComponent()`, I meant making it `const` in the diagram, but also in the source code.

Edit: I have [just seen](https://github.com/cellml/libcellml/pull/52/files#diff-d2ee493885a137515e56ea7e8cf5fa1aL83) that `const` has been added to `ComponentEntity:: containsComponent()`, so please forget about my above remark.

Issue #59, opened 2016-03-30
Update use-case order to Modify before Load


dladd:
This is a proposal to swap out the Load (currently 2) and Modify (currently 3) tasks in [usecases.rst](https://github.com/cellml/libcellml/blob/develop/docs/usecases.rst). 

The thinking here is that it will be a more logically incremental step for development/testing to extend the models from the Create task (1) than to start dealing with loading/importing models at this stage.

jonc125:
Seems sensible to me. I note that task 1.viii appears at least partially to depend on loading models - it might be worth clarifying if what is intended is just to create the import elements, not instantiate the imports?

nickerso:
yeah, all those create ones are just about creating the information in a CellML model, there is no intention in 1.\* to instantiate imports.

PR #60, opened 2016-03-30
Re-order use-cases doc to modify models before loading

dladd:
Updates the usecases.rst doc to modify created models before loading new ones. This only affects documentation.

Addresses #59


nickerso:
Looks good to me.

Issue #61, opened 2016-03-30
Updates to object model for use-cases 1.i-1.xi


dladd:
This covers proposed changes to the rough object model(s) to incorporate use-case 1.xi: creating a model with variables.

PR #62, opened 2016-03-30
Proposed object model for use-cases 1.i-1.xi.

dladd:
Updates the object model to include features from use-case 1.xi. Introduces variables (see item 11 in current CellML 1.2 Specification).

Addresses #61


nickerso:
Not sure if I'm missing something, but the `get/add/removeVariable` methods probably belong on the `Component` class?

jonc125:
Indeed, I was wondering what the `#mVariable` on `Variable` was for. And a component will need a list/set/collection of `Variable`s.

As an aside, is there a reason for having `Component.setSourceComponent` when you could presumably use the base class method `setImport`?

hsorby:
@jonc125  <code>Component.setSourceComponent</code> is a convenience method that sets the import and component reference at the same time.  It does more than just <code>setImport</code> does.

dladd:
Thanks for the comments, @nickerso  and @jonc125.  I (with @hsorby ) have now updated the use-case diagram. Apologies for any confusion while I sort through how everything fits together.

This now:
- adds a `mVariables` attribute and `add/remove/getVariable` operations to the `Component` class
- adds a `mUnits` attribute and `set/getUnits` operations to the `Variable` class

nickerso:
I think that is all looking good now, ready to start implementing :)

agarny:
A couple of comments (sorry @nickerso :)):
- It might be good to rename the file to something like `libCellMLObjectModel.png` (or `libCellMLRoughObjectModel.png`), i.e. a generic name. Indeed, in this PR, the file name has changed from `libCellMLRoughObjectModel-UseCases1-i-1-x.png` to `libCellMLRoughObjectModel-UseCases1-i-1-xi.png`, which makes it difficult (though not impossible, I agree) to compare the two versions (using GitHub). Also, if we were to follow the current logic, it means that we would keep renaming the file as we keep on improving the object model, which is really not ideal (if we want to be able to keep track of incremental changes).
- In `Component`, we probably ought to add `removeVariable(size_t): void`, as well as `getVariable(std::string) const: const VariablePtr&`, `getVariable(size_t): VariablePtr` and `getVariable(size_t) const: const VariablePtr&`. The idea is to be consistent with what has already been done with `ComponentEntity`. This means that we probably ought to add `variableCount() const: size_t` too. However, I get the feeling that `take/contains/replaceVariable()` are not needed here.

Once this is done, I believe 'we' should be ready to start implementing... :)

hsorby:
I don't believe that the removeVariable is required here it should be added once we start to modify the objects in earnest.  The component entity is a bit of a one-off where we have added the modification code, really just to make sure the implementation was possible in that way and not have to redo changes to five or six other objects if it didn't work out.  Certainly the other forms of getVariable should be added and if not in the rough object model definitely into the source code.

The png is considered binary data for git and as such there is no incremental changes.  The file is named that way as that is what gets automatically set from the application from which it is exported.  I prefer to have the reference to the use case that the image is for so that I can be sure that I know what the image should be showing, without the reference I could quite easily start comparing the image from use case 1.xi to code for use case 1.xii.

dladd:
@hsorby re: diagram file naming, I believe @agarny is mainly talking about how Github displays changesets for images. For example, if you compare [my last commit with the same diagram name](https://github.com/cellml/libcellml/pull/62/commits/544e79b8cd5334d0713d413567a839546f62337f) to [my first commit where I renamed the file](https://github.com/cellml/libcellml/pull/62/commits/df44d7980ccecfba03a82f1d35f9641e73fb2990), the former gives a side-by-side diff rather than just showing that the file was deleted in the latter. Even though both effectively replace the binary file, Github appears to be clever enough to render both in the browser if the filename is the same. I could see this as useful, especially when new classes are added. I think the title at the top of the image should sufficiently indicate which use cases the diagram covers.

agarny:
> I don't believe that the removeVariable is required here it should be added once we start to modify the objects in earnest. The component entity is a bit of a one-off where we have added the modification code, really just to make sure the implementation was possible in that way and not have to redo changes to five or six other objects if it didn't work out. Certainly the other forms of getVariable should be added and if not in the rough object model definitely into the source code.

Fair enough regarding `removeVariable()`. As for `getVariable()`, to have all the forms in the diagram does indeed overload the diagram, but it also makes it easier to review things (otherwise, as a reviewer, I could personally easily forget to check what has/hasn't been implemented).

> The png is considered binary data for git and as such there is no incremental changes. The file is named that way as that is what gets automatically set from the application from which it is exported. I prefer to have the reference to the use case that the image is for so that I can be sure that I know what the image should be showing, without the reference I could quite easily start comparing the image from use case 1.xi to code for use case 1.xii.

My reason for renaming the diagram file is indeed the one given by @dladd, not to mention that besides the 2-up view, you also have the swipe and onion skin views. So, yes, it would help to have a generic name for the diagram file.

agarny:
Looks good to me now (@nickerso, @jonc125 and @hsorby?).

jonc125:
+1

Issue #63, opened 2016-03-31
 Support use case 1.xi: a model with variables


dladd:
This issue is for implementing support for use case 1.xi, creating a model with variables. 

This includes:
- [x] **a.** model from _1.iv.a_ and define a variable with a valid name and units dimensionless
  - [x] **1.** with a valid variable initial value of 0.0
- [x] **b.** model from _1.iv.a_ and define a variable with an invalid name and units dimensionless
- [x] **c.** model from _1.iv.a_ and define a variable with a valid name and invalid units name.
- [x] **d.** a model with a single component containing two variables.
  - [x] **1.** with valid variable initial values of 1.0 and -1.0, respectively
  - [x] **2.** one with an initial value of 1.0 and the other with an initial value of the first variable.

dladd:
If there are no objections (@hsorby, @agarny, @nickerso, @jonc125), I'd like to limit this issue to use-cases 1.xi.a - 1.xi.c. This issue/pull will then just cover the reviewed modifications to the object model from #62.

That will then leave 1.xi.d, which deals with interface connections, for the next set of edits to the object model(s), testing, and implementation.

nickerso:
sounds good to me. Will be good to keep the initial implementation of the variable object separate from the issue of dealing with connections.

jonc125:
+1

agarny:
+1

PR #64, opened 2016-04-04
Supporting use-case 1.xi (a-c): a model with variables

dladd:
This request (upon completion) implements tests and code support for use-case 1.xi: a model with variables

Addresses #63


tests/variable/variable.cpp
```diff
@@ -0,0 +1,117 @@
...
+limitations under the License.Some license of other
+*/
+
+#include "gtest/gtest.h"
+
+#include <libcellml>
+
+
+TEST(Variable, serialise) {
+    const std::string e = "";
```
nickerso: as with the other classes, I would expect a new variable to produce at least an empty `<variable/>` element in the XML serialisation.

tests/variable/variable.cpp
```diff
@@ -0,0 +1,117 @@
...
+    u->setName("dimensionless");
+    v->setUnits(u);
+
+    std::string a = m.serialise(libcellml::FORMAT_XML);
+    EXPECT_EQ(e, a);
+    EXPECT_EQ("valid_name", v.getName());
+}
+
+TEST(Variable, invalidVariableName) {
+    const std::string in = "invalid_name";
```
nickerso: I should go back and check the component test, but this should actually be an invalid name like `invalid name` :)

tests/variable/variable.cpp
```diff
@@ -0,0 +1,117 @@
...
+    libcellml::ComponentPtr c = std::make_shared<libcellml::Component>();
+    c->setName(in);
+    m.addComponent(c);
+
+    libcellml::VariablePtr v = std::make_shared<libcellml::Variable>();
+    v->setName(in);
+    c->addVariable(v);
+
+    libcellml::UnitsPtr u = std::make_shared<libcellml::Units>();
+    u->setName("invalid_name");
```
nickerso: again, should be an invalid name

nickerso:
good start. Is the buildbot testing not working currently? shouldn't there be a green tick or red cross on the current state of the pull request? (@hsorby)

hsorby:
It would appear that Buildbot has given up @nickerso I shall investigate.

dladd:
This now passes all tests and adds the features as set out in #63. 

However, there were two applications I implemented aspects of as part of the outlined object model (see #61) that were overlooked in the use-cases (see #63). These were (1) components having multiple variables and (2) variables having initial values.

Adding tests and support for these should be relatively straightforward. I propose:
- **1.xi.a.1** a model from 1.xi.a with a valid variable initial value of 0.0
- **1.xi.d** a model with a single component containing two variables named "variable_a" and "variable_b" with units dimensionless.
  - **1.xi.d.1** a model from 1.xi.d with variable initial values of -1.0 and 1.0, respectively.

The use-case currently listed as 1.xi.d dealing with interface connections could be shifted into 1.xii as discussed in (#63), with the rest of the use-cases to be shifted accordingly

hsorby:
I see some future code is added here (which consequently drops the coverage, http://autotest.bioeng.auckland.ac.nz/libcellml-buildbot/d0a57450d6e63d5e40ce534cd7409d254fde7d9b/dox/coverage/).  Should be removed for now and added back in at the appropriate time.

src/variable.cpp
```diff
@@ -0,0 +1,94 @@
...
+std::string Variable::doSerialisation(FORMATS format) const
+{
+    std::string repr = "";
+    if (format == FORMAT_XML) {
+        repr += "<variable";
+        if (getName().length()) {
+            repr += " name=\"" + getName() + "\"";
+            if (getUnits() != nullptr) {
+                repr += " units=\"" + getUnits()->getName() + "\"";
+            }
```
hsorby: This should be independent of whether that variable has a name or not.

dladd:
Thanks, @hsorby. We should now be back up to 100% coverage on this. 

Tests now also cover multiple variables in a component. That leaves variable initial values to be tacked on to this, unless others think that is better left for a separate pull.

src/variable.cpp
jonc125: And hence a test added where a variable has units but no name.
(When we come to initial values, we should test more combinations.)

jonc125:
I'd put it in a separate pull given the title of this one :)

Haven't checked thoroughly, but apart from the extra test I mentioned this seems sane.

dladd:
Thanks, @jonc125. I've now added the units-only test and will go with a separate pull for initial values (see #67).

PR #66, opened 2016-04-05
The treatment of MathML in libCellML

nickerso:
As we get closer to implementing some of the use-cases involving MathML in libCellML, it would be good to make sure we are all on the same page in terms of how integrated MathML support will be in "core" libCellML. I have added a couple of points to the current thinking document to describe how I see it happening, at least for now, and I think this captures what @hsorby and I have been talking about.

### DIFF
--- a/docs/current_thinking.rst
+++ b/docs/current_thinking.rst
@@ -4,7 +4,7 @@
 Current Thinking for libCellML
 ==============================
 
-This document simply outlines some of the rationale that has an influence on how the codebase is developed.
+This document simply outlines some of the current rationale that has an influence on how the codebase is developed.
 
 - Not dealing with external documents.
  
@@ -15,4 +15,8 @@ This document simply outlines some of the rationale that has an influence on how
 - Serialise and deserialise from a string.
 - Present a useful interface not one tied to XML structure.
 - Validation is going to be quite separate (you are free to make invalid CellML models).
+- Not creating our own MathML object model
 
+  - Treat MathML as a string.
+  - Enables libCellML development to focus on getting core "CellML" support.
+  - Expect to provide another layer that would handle MathML as a separate thing.


jonc125:
OK, that does impact my suggested tests :)

I think my one concern with this would be that there is one CellML-specific aspect embedded within the MathML - the cellml:units attribute. So a future layer that processes the MathML string and does something useful would need some CellML-specific knowledge to handle that and talk to libCellML to get the units definitions.

nickerso:
Would it be a better approach to include a basic abstract syntax tree of some description inside core libCellML? That might be one way to get the pure math into other tools while handling the CellML bits internally...and I'd guess a very basic AST model which doesn't do anything beyond storing the structure of the MathML (but _not_ a DOM representation)...

jonc125:
Yeah, I think that would be more helpful. Then cn elements can have a link to units, and ci elements a link to the variable (if it exists). Agree not DOM though!  Is it worth looking at something similar to libSBML, or extreme cut-down element tree, class for MathElement, subclass for cn & ci, just get/set element name & children? Anything on attributes (beyond id)?

hsorby:
It might be a better approach to include a basic syntax tree but importantly there is no actual need for it at this time.  We can (and should) get by with whatever is required to pass the tests, be that a regular expression based solution or whatever.

The tests that we want to pass now are based around adding maths to the model.  The proposal at this point in time is to have the math passed as strings.  And some other entity/library handling the math, leaving the math until we can actually handle it properly.  When we get to the validation stage then we can come up with the bare minimum to get the tests passing for that stage.  For cellml validation we only want to check the variables exist in the component and that the units are defined.  Which we can do easily with regular expressions.  

Checking for dimensional consistency will take a lot more work and is something that will be done outside core libcellml (I think).  So I'm happy with @nickerso's proposal and think this is the way forward.

agarny:
+1 for a basic AST.

jonc125:
One issue with just treating MathML as a string is that presumably you are using some proper XML parser to parse the full string containing a CellML fragment? (I've lost track of details having not been very involved for a while.) Telling it _not_ to parse embedded MathML as XML is then... not very nice, and possibly not possible depending on the parser.  I do agree that we don't want to handle the MathML as mathematics, or check dimensional consistency, at this stage. But we probably do want to recognise that it is XML.

nickerso:
Currently we have only been concerned with creating CellML models, not loading them. So there is currently no XML parsing in libCellML at all :)

jonc125:
OK, so in that case we could just treat it as an opaque string for now, and not check variable or units refs at all, in the understanding that once we start loading models that'll probably change?

nickerso:
yep, that is what I'm currently leaning toward. Keep needing @hsorby to remind me to not think too far ahead and just focus on the current use-cases and associated tests.

dladd:
+1 for sticking with MathML strings as we work through the Create and Modify use-cases steps.

nickerso:
So I think we are converging on the proposed change being acceptable?

jonc125:
Yep

mirams:
Sounds good!

agarny:
Sounds good to me too.

PR #67, opened 2016-04-10
Update use cases: add variable initial values and modify connections

dladd:
As per discussion on issue #63 and pull #64, this updates the use-cases document to include support for variable initial values and separates the implementation of connections from variables.

If accepted, I plan to add support for initial values as a second pull request under issue #63 and set up connections as its own issue/pull set.

### DIFF
--- a/docs/usecases.rst
+++ b/docs/usecases.rst
@@ -60,12 +60,21 @@ Use-cases for libCellML
    #. a model with variables
    
       a. model from *1.iv.a* and define a variable with a valid name and units dimensionless
+         #. with a valid variable initial value of 0.0
+         #. with an invalid variable initial value
+
       #. model from *1.iv.a* and define a variable with an invalid name and units dimensionless
       #. model from *1.iv.a* and define a variable with a valid name and invalid units name.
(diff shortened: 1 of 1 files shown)


jonc125:
Looks fine to me.

docs/usecases.rst
```diff
@@ -60,12 +60,21 @@ Use-cases for libCellML
...
+
       #. model from *1.iv.a* and define a variable with an invalid name and units dimensionless
       #. model from *1.iv.a* and define a variable with a valid name and invalid units name.
-      #. model from *1.vi.a*, define a single variable in each component with units dimensionless
-      
-         #. private interface in the parent and public interface in the child components and connect the variable in both children to the parent.
-         #. public interface in all components and connect the variables in the children to the parent
+      #. a model with a single component containing two variables.
+
+         #. with valid variable initial values of 1.0 and -1.0, respectively.
```
nickerso, line 70: I would add:
- one with an initial value of 1.0 and the other with an initial value of the first variable.

dladd:
Thanks, @nickerso- updated now with a set initial value by variable reference use-case.

Issue #41, opened 2015-07-16
link documentation on readthedocs to current "develop" doxygen and coverage


nickerso:
Now that the documentation links to the doxygen and coverage generated documentation in addition to the sphinx docs, we need some way to link to a version of that matches the current state of the develop branch (which is what "latest" on readthedocs points at). We can do this by having a step in the buildbot setup that syncs the docs when a merge is being performed, rather than a pull-request being updated.

This issue is mostly here as a reminder that this needs to be implemented. When someone has a chance to do so :)

hsorby:
Closing this as we now have the documentation in place and available on the libcellml.org website.

PR #68, opened 2016-04-11
Supporting variable initial values (part of use-cases 1.xi)

dladd:
Adds tests and support for variable initial values.

Addresses the remaining requirements of issue #63


nickerso:
This all looks great to me.

src/api/libcellml/variable.h
```diff
@@ -55,8 +55,20 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
     UnitsPtr getUnits() const;
 
     /**
+     * @brief Set the initial value for this variable using a string.
+     * Set the initial value for this variable using a string.
+     *
+     * @overload
+     * @sa getInitialValue
+     * @param initialValue the initial value to set.
```
agarny: "The initial..." rather than "the initial..."

Note: there are other similar instances here and there throughout the whole code.

src/api/libcellml/variable.h
```diff
@@ -55,8 +55,20 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
+     */
+    void setInitialValue(const std::string initialValue);
```
agarny: `const std::string &initialValue` and not `const std::string initialValue`.

src/variable.cpp
```diff
@@ -91,4 +96,26 @@ UnitsPtr Variable::getUnits() const
     return mPimpl->mUnits;
 }
 
+void Variable::setInitialValue(const std::string initialValue)
```
agarny: `const std::string &initialValue`

src/variable.cpp
```diff
@@ -91,4 +96,26 @@ UnitsPtr Variable::getUnits() const
...
+}
+
+void Variable::setInitialValue(double initialValue)
+{
+    std::ostringstream strs;
+    strs << initialValue;
+    mPimpl->mInitialValue = strs.str();
+}
+
+void Variable::setInitialValue(const VariablePtr &referenceVariable)
```
agarny: I would personally use simply use `variable` rather than `referenceVariable`.

tests/variable/variable.cpp
```diff
@@ -89,6 +89,47 @@ TEST(Variable, setUnitsAndName) {
     EXPECT_EQ(e, a);
 }
 
+TEST(Variable, setInitialValue) {
```
agarny: Might want to rename the test `setInitialValueString` to make it clear that we are initialising the value using a string (as opposed to using a double).

Otherwise, rather than using variable names such as `e`, `v` and `a`, I would use more explicit names such as `expected`, `variable` and `actual`.

Note: the same is true through the whole code (it makes it easier to read/maintain).

tests/variable/variable.cpp
```diff
@@ -89,6 +89,47 @@ TEST(Variable, setUnitsAndName) {
...
+TEST(Variable, setInitialValueDouble) {
+    const std::string e = "<variable initial_value=\"0\"/>";
+    libcellml::Variable v;
+    double value = 0.0;
+    v.setInitialValue(value);
+    std::string a = v.serialise(libcellml::FORMAT_XML);
+    EXPECT_EQ(e, a);
+}
+
+TEST(Variable, setInitialValueByRef) {
```
agarny: It doesn't harm to spell things out, i.e. `setInitialValueByReference`.

tests/variable/variable.cpp
```diff
@@ -89,6 +89,47 @@ TEST(Variable, setUnitsAndName) {
...
+    const std::string e = "<variable initial_value=\"referencedVariable\"/>";
+    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
+    v1->setInitialValue("1.1");
```
agarny: Not needed for the purpose of this test.

tests/variable/variable.cpp
```diff
@@ -134,6 +175,29 @@ TEST(Variable, addTwoVariables) {
     EXPECT_EQ(e, a);
 }
 
+TEST(Variable, twoInitialisedVariables) {
```
agarny: Might want to rename the test `componentWithTwoInitialisedVariables`.

tests/variable/variable.cpp
```diff
@@ -285,5 +349,67 @@ TEST(Variable, modelInvalidUnitsName) {
     EXPECT_EQ("invalid name", u->getName());
 }
 
+TEST(Variable, modelTwoNamedInitialisedVariables) {
```
agarny: Might want to rename the test `modelWithComponentWithTwoNamedInitialisedVariables`.

tests/variable/variable.cpp
```diff
@@ -285,5 +349,67 @@ TEST(Variable, modelInvalidUnitsName) {
...
+    const std::string in = "valid_name";
```
agarny, line 401: If you are to keep track of the name of the component, then you probably ought to do the same for the name of the variables, as well as for their value.

Note: as a result, something similar ought to be done for the other tests.

tests/variable/variable.cpp
```diff
@@ -285,5 +349,67 @@ TEST(Variable, modelInvalidUnitsName) {
...
+    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
+    v2->setName("variable2");
+    v2->setInitialValue("-1.0");
+    c->addVariable(v2);
+
+    std::string a = m.serialise(libcellml::FORMAT_XML);
+    EXPECT_EQ(e, a);
+}
+
+TEST(Variable, modelTwoNamedInitialisedVariablesOneReferenced) {
```
agarny, line 431: Might want to rename the test `modelWithComponentWithTwoNamedInitialisedVariablesIncludingOneReferenced`.

src/api/libcellml/variable.h
jonc125: It possibly depends on how this gets rendered by the documentation system - for doxygen an initial lower case actually looks better.

agarny:
It all looks good to me too besides the minor comments that I made. (This is the kind of comments I was mentioning [here](https://github.com/cellml/libcellml/commit/7319139319d561ba1e4ef686320b8e8244ea83f0).

src/api/libcellml/variable.h
agarny: Indeed, @jonc125, the actual rendering will have to be checked. It's just that I have seen both "The" and "the", so this particular comment was more about consistency.

nickerso: just for reference, the doxygen is available through the buildbot for each pull request - follow the link to the buildbot results for the pull request, then the documentation builder, and then you should see a link to the "dox". For example, this particular code for this revision is currently available [here](http://autotest.bioeng.auckland.ac.nz/libcellml-buildbot/81db9d9c74d926a0c43d7ba491eb9024c3f9efa4/dox/code/classlibcellml_1_1Variable.html#a4b5f832ae37e025130b8b3880a6a47f8). I think the generated docs get flushed every now and then, so the links are not permanent. #41 is the issue on how these should get integrated into the main project docs on readthedocs.

dladd:
Thanks @agarny for the thorough review. I've now added commits that should address most of your comments. 

An exception is making variable names within tests explicit, which I think should be discussed as part of the overall code tidy-up after this pull.

nickerso:
I'm happy for this to be merged in, but will leave it for @agarny to confirm ;-)

dladd:
Just a heads up: we'll want to merge the bugfix #71 in before this one to avoid conflicts.

src/api/libcellml/variable.h
agarny: Thanks @nickerso. Back to the current issue, I prefer when it's capitalised since there is no ":" between the variable name and its description (see the end of the page mentioned by @nickerso for an exemple of capitalisation).

src/variable.cpp
```diff
@@ -108,9 +108,9 @@ void Variable::setInitialValue(double initialValue)
     mPimpl->mInitialValue = strs.str();
 }
 
-void Variable::setInitialValue(const VariablePtr &referenceVariable)
+void Variable::setInitialValue(const VariablePtr &v)
```
agarny, line 111: If anything, `referenceVariable` should have been replaced with `variable` and not `v`.

agarny:
There are still a couple of very minor issues left, but they can be 'fixed' as part of issue #75.

Issue #69, opened 2016-04-12
Tidy up presentation in Visual Studio


hsorby:
It is nice to group the targets created into folders in Visual Studio.  This makes the code much more manageable and usable.

mirams:
Closed by https://github.com/cellml/libcellml/commit/7a90044c7975977e5769ec165e9f40d433e657b5

PR #70, opened 2016-04-12
Nicer target presentation for Visual Studio

hsorby:
Addresses issue #69


nickerso:
Looks good to me.

jonc125:
Nice, I'll have to remember this trick!

agarny:
Indeed, it's much better that way:

image

cmake/common.cmake
```diff
@@ -45,3 +45,29 @@ function(HIDE_DISTRACTING_VARIABLES)
...
+        set(_F_PREFIX "Generated Header Files\\")
+      endif()
+      get_filename_component(_G ${_L} DIRECTORY)
+      if(_G)
+        string(REPLACE "/" "\\" _G ${_G})
+      endif()
+      source_group("${_F_PREFIX}${_G}" FILES "${_F}")
+    endforeach()
+  endif()
+endfunction()
```
agarny, line 72: I would prefer having 'proper' variable names, i.e. replace `_F` with `FILE`, `_F_PREFIX` with `FILE_PREFIX`, etc.

agarny:
Apart from my minor comment above, it's all fine with me.

cmake/common.cmake
hsorby, line 72: I don't think using FILE would be appropriate, CMake might have something to say about that!  I'll change the variable names to something 'proper' but not dangerous.

agarny, line 72: Actually, CMake is smart enough to know when `FILE` is used as a command or as a variable. This being said, I am happy for you to use something different as long as it is 'proper'. :)

hsorby:
I think this one is ready.

mirams:
closes #69

PR #71, opened 2016-04-12
Bugfix: enable variable serialisation in unnamed component

dladd:
@hsorby highlighted a bug from pull #64 by submitting a failing test. A nested if statement prevented serialisation of variables in unnamed components.

Relevant to issue #63


nickerso:
good catch @hsorby :)

the new tests and fixed implementation look good to me.

tests/variable/variable.cpp
```diff
@@ -134,6 +151,34 @@ TEST(Variable, addTwoVariables) {
     EXPECT_EQ(e, a);
 }
 
+TEST(Variable, addVariables) {
+	const std::string e =
+		"<component>"
+		"<variable name=\"var1\" units=\"dimensionless\"/>"
+		"<variable name=\"var2\"/>"
+		"<variable name=\"var3\"/>"
```
jonc125: It might be worth having one of these with only units, or only initial_value, and no name, so more combinations are checked?

src/component.cpp
```diff
@@ -142,12 +142,12 @@ std::string Component::doSerialisation(FORMATS format) const
...
-                endTag = true;
-                repr += ">";
-                for(size_t i = 0; i < variableCount(); i++) {
-                    repr += getVariable(i)->serialise(format);
-                }
+        }
+        if (variableCount() > 0) {
+            endTag = true;
+            repr += ">";
+            for(size_t i = 0; i < variableCount(); i++) {
```
agarny: Very minor improvement: prefer `++i` to `i++`. Also, I would personally favour something like:

```
for (size_t i = 0, iMax = variableCount(); i < iMax; ++i) {
...
```

Note also the space between `for` and `(`.

tests/variable/variable.cpp
agarny: +1

Otherwise, I would indent the 'variable' lines.

agarny:
Besides the above minor comments and Jonathan's suggestion, it all looks good to me too.

tests/variable/variable.cpp
dladd: Good idea, @jonc125 , as a way to pick up more permutations in a single test. This bugfix pull actually pre-dates the addition of initial values so will add those in under that pull set. :-)

src/component.cpp
dladd: Re: "++i" instead of "i++", the former does indeed seem to be the preferred form in the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) (GCSG) we are using as a guideline. The same goes for the spacing in "`for (`". I can go through and do a formatting fix for instances of this and other things we pick up throughout the code after this bugfix and the initialValues pulls go through.

As for 

`for (size_t i = 0, iMax = variableCount(); i < iMax; ++i) {`

versus

`for (size_t i = 0; i < variableCount(); ++i) {`

I personally prefer the latter for succinctness but don't see any hard and fast rules in the GCSG. They do seem to give some examples [more in line with the latter](https://google.github.io/styleguide/cppguide.html#Implementation_Comments) e.g., 

`for (int i = 0; i < result->size(); i++) {`

But I guess consistency is what really matters- I could be convinced if more prefer the "`iMax`" style. Any opinions @nickerso , @hsorby , or @jonc125 ?

nickerso: I definitely prefer

`for (int i = 0; i < result->size(); i++) {`

much easier to read for me :) and I don't think we are really concerned with the performance hit in evaluating the "size" on each iteration in this code. (Similarly, I'm not worried about the performance hit in the pre- vs post-fix increment for simple integer data types, but can see that it would be best to be consistent and always go with the prefix increment operator)

nickerso: also, I'd suggest it might be best to resolve the current set of pull requests and then have a concerted "code tidy up" - for the code outside the direct scope of this chunk of work. For these specific instances it is best to fix them here and now.

tests/variable/variable.cpp
```diff
@@ -166,14 +167,16 @@ TEST(Variable, addVariables) {
...
+    libcellml::VariablePtr v4 = std::make_shared<libcellml::Variable>();
 	c.addVariable(v1);
 	c.addVariable(v2);
 	c.addVariable(v3);
+    c.addVariable(v4);
 
 	libcellml::UnitsPtr u = std::make_shared<libcellml::Units>();
 	u->setName("dimensionless");
 	v1->setUnits(u);
+    v3->setUnits(u);
```
hsorby, line 180: Tabs instead of spaces issue?

dladd, line 180: Good catch- fixed in d5ddf20da73ca214e74bdc7accfd5a0928f49e34

nickerso:
So I believe all issues here are resolved, so I'll merge it in to avoid conflicts in #68.

Issue #72, opened 2016-04-12
CMake deprecation warning for add_compiler_export_flags() function


agarny:
Using CMake 3.5.1 (i.e. the latest version currently available), we get the following message when doing a simple configuration:

```
 CMake Deprecation Warning at /Applications/CMake.app/Contents/share/cmake-3.5/Modules/GenerateExportHeader.cmake:383 (message):
   The add_compiler_export_flags function is obsolete.  Use the
   CXX_VISIBILITY_PRESET and VISIBILITY_INLINES_HIDDEN target properties
   instead.
 Call Stack (most recent call first):
   src/CMakeLists.txt:68 (add_compiler_export_flags)
```

nickerso:
closed by #73.

PR #73, opened 2016-04-12
CMake: replaced obsolete use of add_compiler_export_flags() function (#72)

agarny:
Addresses issue #72.


src/CMakeLists.txt
```diff
@@ -65,7 +65,8 @@ set(HEADER_FILES
   ${LIBCELLML_CONFIG_H}
 )
 
-add_compiler_export_flags()
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
```
hsorby: These values should be set as properties on the target not through manipulating the default CMake variables.

agarny: Good point. I believe this should now be the case with https://github.com/cellml/libcellml/pull/73/commits/d7aaf9346f7952691709be0aa8571637314f0e59.

src/CMakeLists.txt
```diff
@@ -65,16 +65,17 @@ set(HEADER_FILES
...
   ${SOURCE_FILES}
   ${HEADER_FILES}
   ${MODULE_HEADER_FILES}
   ${API_HEADER_FILES}
 )
 
+set_target_properties(cellml PROPERTIES
+  CXX_VISIBILITY_PRESET hidden
+  VISIBILITY_INLINES_HIDDEN 1)
+
```
hsorby: Last thing and then I'm done, just a little further down in this file there is a set_target_properties command these two occurences should just be one.

agarny: Fair point. Done.

Issue #77, opened 2016-04-15
Add support for variable interfaces


dladd:
Before we move on to connections, we should add support for the variable interface attribute (`public` / `private` / `public_and_private` / `none` ). I'll put in pull requests for the corresponding updates to the use-cases doc, the object model, tests, and implementation under this issue.

Current thinking is to implement the interface attribute as an enum rather than simple strings.

PR #78, opened 2016-04-15
Add interface use-cases to variable use-case item

dladd:
Adds some use-cases for the variable interface attribute within the variable item (1.xi).

Addresses part of issue #77

### DIFF
--- a/docs/usecases.rst
+++ b/docs/usecases.rst
@@ -62,6 +62,7 @@ Use-cases for libCellML
       a. model from *1.iv.a* and define a variable with a valid name and units dimensionless
 
          #. with a valid variable initial value of 0.0
+         #. with a private interface
 
       #. model from *1.iv.a* and define a variable with an invalid name and units dimensionless
       #. model from *1.iv.a* and define a variable with a valid name and invalid units name.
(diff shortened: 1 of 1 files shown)


agarny:
I think we should be a bit more explicit. A public/private interface can have one of different known values, so we ought to have tests that account for all possible cases. In other words, I would expect tests for the following:
- A variable without any public/private interface. This is, in effect, already covered by existing tests.
- A variable with a public interface of `in`, `out` and then `none`.
- A variable with a private interface of `in`, `out` and then `none`.
- A variable with both a public and a private interface of say `none`.

nickerso:
CellML 1.2 has just a single interface property for variables, and that interface is one of `public`, `private`, `public_and_private`, or `none`. If no value is specified, the interface is assumed to be `none`. There is no longer a separate `public_interface` and `private_interface` property for variables.

So I think the suggested use cases cover the range of options, when combined with the existing cases which all have the assumed no interface.

agarny:
Oh, ok, my bad.

Just out of curiosity (and sorry in advance if it has already been asked/answered): how is libCellML going to handle CellML 1.0 and 1.1 files? Or, are we just focussing on CellML 1.2 for now and then addressing CellML 1.0/1.1?

nickerso:
The plan has always been to natively support CellML 1.2 in libCellML. Eventually, earlier versions of CellML will be imported and exported. This currently comes after validating 1.2 models in the [roadmap](http://libcellml.readthedocs.org/en/latest/roadmap.html).

agarny:
Ok, my bad again (what a start to my day!) for not remembering that...

agarny:
Ok, apologies to @dladd. It all looks good to me.

PR #80, opened 2016-04-15
Tests and implementation for variable interfaces

dladd:
This pull will add tests and code to support variable interfaces in libCellML

Addresses the remainder of issue #77


src/variable.cpp
```diff
@@ -118,4 +131,42 @@ std::string Variable::getInitialValue() const
...
+{
+    return mPimpl->mInterfaceType;
+}
+
+std::string interfaceTypeToString(Variable::INTERFACE_TYPES interfaceType)
+{
+    std::string str;
+    switch (interfaceType) {
+    case Variable::INTERFACE_TYPES::INTERFACE_TYPE_UNDEFINED: {
+        str = "";
```
dladd: I seem to be dropping a bit of coverage here since the if statement leading in to `interfaceTypeToString()` excludes the undefined interfaces since we don't want to serialise those.

I'm guessing there's possibly a cleverer C++11 way to do this? If so, I bet @hsorby knows what it is :-)

jonc125: Presumably you could have a test that doesn't serialise, just calls the method directly?

dladd:
This is not quite ready to be pulled in but I thought it would be a good idea to let others see the direction I was headed in before the weekend.

src/variable.cpp
dladd: Good point @jonc125 , I now see that is what has been done for the similar situation with unit prefixes in the coverage tests.

src/api/libcellml/variable.h
```diff
@@ -39,6 +39,20 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
     Variable& operator=(Variable n); /**< Assignment operator */
 
     /**
+     * @brief The INTERFACE_TYPES enum.
+     * Enum describing the interface types that are available for
+     * a given Variable.
```
agarny: Why capitalise Variable? (See other instances in the code.)

src/api/libcellml/variable.h
```diff
@@ -39,6 +39,20 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
+     */
+    enum INTERFACE_TYPES
+    {
+        INTERFACE_TYPE_UNDEFINED,
```
agarny: When would `INTERFACE_TYPE_UNDEFINED` be used? Shouldn't a variable have an interface of `none` (and therefore `INTERFACE_TYPE_NONE`) by default?

src/api/libcellml/variable.h
```diff
@@ -92,6 +106,24 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
     std::string getInitialValue() const;
 
+    /**
+     * @brief Set the interface type for this Variable.
+     * Set the interface type for this variable from the available
+     * options in the INTERFACE_TYPES enum.
+     * @sa getInterfaceType
+     * @param interfaceType The interface type to set for this variable.
+     */
+    void setInterfaceType(const INTERFACE_TYPES interfaceType);
```
agarny: Drop the `const`.

src/api/libcellml/variable.h
```diff
@@ -92,6 +106,24 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
+
+    /**
+     * @brief Get the interface type for this Variable.
+     * Set the interface type for this variable from the available
+     * options in the INTERFACE_TYPES enum.
+     * @sa setInterfaceType
+     * @param interfaceType The interface type to set for this variable.
+     */
```
agarny, line 123: Needs rephrasing after copying/pasting. :)

src/variable.cpp
```diff
@@ -23,12 +23,22 @@ limitations under the License.Some license of other
 namespace libcellml {
 
 /**
+ * @brief Convert a INTERFACE_TYPES into it's string form.
```
agarny: its!!! (Same below.)

src/variable.cpp
```diff
@@ -23,12 +23,22 @@ limitations under the License.Some license of other
...
+std::string interfaceTypeToString(Variable::INTERFACE_TYPES interfaceType);
+
+/**
  * @brief The Variable::VariableImpl struct.
  * The private implementation for the Variable class.
  */
 struct Variable::VariableImpl
 {
     std::string mInitialValue = ""; /**< Initial value for this Variable.*/
+    INTERFACE_TYPES mInterfaceType = INTERFACE_TYPES::INTERFACE_TYPE_UNDEFINED; /**< Interface type for this Variable. Default to undefined.*/
```
agarny: This is where I would use `INTERFACE_TYPE_NONE`.

src/variable.cpp
```diff
@@ -81,6 +91,9 @@ std::string Variable::doSerialisation(FORMATS format) const
         if (getInitialValue().length()) {
             repr += " initial_value=\"" + getInitialValue() + "\"";
         }
+        if (getInterfaceType() != Variable::INTERFACE_TYPES::INTERFACE_TYPE_UNDEFINED) {
```
agarny: I would simply test against `INTERFACE_TYPE_NONE`. Even if the user explicitly specifies an interface of `none`, I think it's fine not to have it in the serialisation.

src/variable.cpp
```diff
@@ -118,4 +131,42 @@ std::string Variable::getInitialValue() const
     return mPimpl->mInitialValue;
 }
 
+void Variable::setInterfaceType(const Variable::INTERFACE_TYPES interfaceType)
```
agarny: Drop the `const`.

src/variable.cpp
```diff
@@ -118,4 +131,42 @@ std::string Variable::getInitialValue() const
...
+}
+
+std::string interfaceTypeToString(Variable::INTERFACE_TYPES interfaceType)
+{
+    std::string str;
+    switch (interfaceType) {
+    case Variable::INTERFACE_TYPES::INTERFACE_TYPE_UNDEFINED: {
+        str = "";
+        break;
+    }
```
agarny: Remove that case.

src/variable.cpp
```diff
@@ -118,4 +131,42 @@ std::string Variable::getInitialValue() const
...
+    case Variable::INTERFACE_TYPES::INTERFACE_TYPE_NONE: {
+        str = "none";
+        break;
```
agarny, line 152: Simply `return "none";` (and the same for the other cases)?

tests/variable/variable.cpp
```diff
@@ -129,6 +129,58 @@ TEST(Variable, getSetInitialValue) {
     EXPECT_EQ(e, a);
 }
 
+TEST(Variable, setInterfaceTypeNone) {
+    const std::string e = "<variable interface=\"none\"/>";
```
agarny: `const std::string e = "<variable/>";`

if we are to take into account my comment above about dropping `INTERFACE_TYPE_UNDEFINED`.

src/variable.cpp
agarny: This might be solved by dropping `INTERFACE_TYPE_UNDEFINED`? (See one of my other comments.)

src/api/libcellml/variable.h
dladd: Thanks @agarny , the undefined type is there to discern between an interface set to `none` and one without any interface specified. According to the 1.2 spec, a given variable doesn't necessarily have an interface.

agarny: Sure, but if it doesn't have an interface, then it means that it effectively has an interface of `none`, no?

Personally, I find that `INTERFACE_TYPE_UNDEFINED` confusing. To me, we can't have an undefined interface, not least since none is specified then it effectively means that we have an interface of `none`.

jonc125: It basically depends on whether you care if the library can round-trip a particular serialisation, or whether a semantically equivalent serialisation is fine. If the latter, then we probably don't need UNDEFINED. If the former, then we definitely do (and we probably will end up needing to care about whitespace etc which could get nasty!).

src/variable.cpp
dladd: My read of the 1.2 Spec suggested that there is an expectation of being able to explicitly specify `interface=none` but if you and others disagree I would be happy to take that into account.

src/api/libcellml/variable.h
hsorby: If I get a vote in this, I would vote for semantically equivalent serialisation.  CellML is for storage and exchange, we just need to the tools to get the same meaning from the same document, for this we only need semantically equivalent serialisation.  

Taking care of whitespace changes .... Urghhhhhh.

dladd: That all sounds fair to me and I'd agree that I don't see a semantic difference between `interface=none` and a non-existent interface. But in that line, I would question why a `none` type even need to be mentioned in the Spec at all, as it would essentially be taking the place of my (possibly misguided) introduction of the ~~dummy~~ artificial `undefined` type.

hsorby: It is still a valid state so you need it, but as it's the default state you don't actually need to serialise it.  You can though but the meaning is no different.  An implementation detail vs the specification detail.

dladd: Ok consider me convinced. Good to know we won't be worrying about whitespace in the future either :-)

nickerso: The current use cases can be met with semantically equivalent serialisation, so I think it is fine for now to have `none` and drop the `undefined`. Some of the latter parts of the roadmap currently require maintaining a particular serialisation, so we might need to revisit this later.

src/variable.cpp
nickerso: Just for clarity, I believe the above discussion resolves this as being fine to not serialise the case when the interface is `none`.

src/variable.cpp
dladd, line 152: The switch is set up this way to avoid a `control reaches end of non-void function` compile error: even if an `enum` is fully covered by the switch, it is possible to define an enumeration that has values not defined by any of its enumerators. So the `return` should be  outside of the enum cases unless we use a `default`.

This also follows existing practice in the codebase (see `prefixToString()` in `units.cpp`).

dladd:
Thanks all for the review and productive discussion. This should now address the comments so far (pending any further thoughts on the way we have our `enum` switches)

nickerso:
Looks good to me. I'm happy with the current handling of the enum switches (consistency is good).

dladd:
Sorry for the late catch but, on the subject of `enum` switches, a couple of lines were dropped in coverage as part of pull #81 over the weekend. I've reverted the changes in commit 9791f79c3f071676b8b2846becc6c7ec4b86fc9b to the 100% coverage state to highlight the issue. A similar problem occurs in the `enum` switch for interfaces.

Basically, the `*ToString` functions are called after a conditional check that excludes an exception case. This causes coverage issues since the exception case lines won't be reached and these functions were designed to be private.

That leaves a few options, depending on how much we prioritise coverage:
1. Go with the reverted state of my last commit, with a `case` without a `break` for the exception.
2. Accept that we won't have 100% coverage and have the uncovered code include the `break`.
3. Try to access the private function through its symbol to reach those few lines.  
4. Export the functions to reach them through the API.

I think I prefer (1) but understand the argument for needing a `break` in every `case` and that coverage shouldn't come at the cost of code quality. (3) is probably the best option but will also require a bit of work to figure out for what is ultimately four lines of code.

src/variable.cpp
```diff
@@ -146,8 +146,8 @@ std::string interfaceTypeToString(Variable::INTERFACE_TYPES interfaceType)
     std::string str;
     switch (interfaceType) {
     case Variable::INTERFACE_TYPES::INTERFACE_TYPE_NONE: {
-        str = "none";
-        break;
+        /* Should not ask for the string version of this.
+        With the current codebase there is no way to trigger this case. */
```
agarny, line 150: I don't like this. Why couldn't I ask for the string version of the none type? In fact, if I was to ask for such a string, I would currently get `private` which is clearly wrong. Personally, I would revert that change (and have something similar to what I have done in for PREFIX_UNIT in commit https://github.com/cellml/libcellml/commit/022e30f4d59ed177372ee5fc604f647482ed5796).

src/variable.cpp
agarny, line 152: Hmm, good point, I have indeed seen that `control reaches end of non-void function` compile error before (only with MSVC I believe), even though everything would have actually been fine here.

agarny:
Ok, I have just seen commit https://github.com/cellml/libcellml/pull/80/commits/9791f79c3f071676b8b2846becc6c7ec4b86fc9b. I really don't like it (see https://github.com/cellml/libcellml/pull/80#discussion_r60063166). As for the fact that it 'breaks' our 100% coverage, well I would just just try to find a way to get that coverage back to 100%. I am not sure what you mean by your third option. As for your fourth option, I guess that "export[ing] the functions to reach them through the API" might make sense at some point (when the need arises), but not right now, I would agree.

dladd:
OK, it sounds like (3) or some alternative testing workaround to reach those lines is the way to go then. The idea for (3) is to access the private function by its address. A bit hacky, but it's probably better for anything hacky to be in the tests rather than the main codebase.

hsorby:
I don't think exporting them through the API is quite the right language for option 4.  What we could do is export the symbol when testing, it wouldn't show up in the API seen through header files.

dladd:
A way around this to keep coverage up could be to define a macro that exports private functions like this while testing. While that may seem overkill for this tiny bit of code, it is probably worth establishing how we handle irreconcilable dropped coverage instances going forward.

hsorby:
I am still of the opinion that if having 100% code coverage means having empty cases in switch statements then this is Ok.  Not having 100% code coverage is highly undesirable.  With 100% coverage we know that if it is not there then it is not required.

src/variable.cpp
```diff
@@ -143,24 +143,25 @@ Variable::INTERFACE_TYPES Variable::getInterfaceType() const
...
+    case Variable::INTERFACE_TYPE_PUBLIC: {
         str = "public";
         break;
     }
-    case Variable::INTERFACE_TYPES::INTERFACE_TYPE_PUBLIC_AND_PRIVATE: {
+    case Variable::INTERFACE_TYPE_PUBLIC_AND_PRIVATE: {
         str = "public_and_private";
         break;
     }
+    case Variable::INTERFACE_TYPE_NONE: {
```
agarny: From your reordering in ```prefixToString()```, I thought you wanted to have everything alphabetically ordered? If so, ```Variable::INTERFACE_TYPE_NONE``` ought to be first.

src/variable.cpp
```diff
@@ -118,4 +131,39 @@ std::string Variable::getInitialValue() const
...
+    }
+    case Variable::INTERFACE_TYPE_PUBLIC: {
+        str = "public";
+        break;
+    }
+    case Variable::INTERFACE_TYPE_PUBLIC_AND_PRIVATE: {
+        str = "public_and_private";
+        break;
+    }
+    case Variable::INTERFACE_TYPE_NONE: {
```
agarny: From your reordering in `prefixToString()`, I thought you wanted to have everything alphabetically ordered? If so, `Variable::INTERFACE_TYPE_NONE` ought to be first.

agarny:
> I am still of the opinion that if having 100% code coverage means having empty cases in switch statements then this is Ok. Not having 100% code coverage is highly undesirable. With 100% coverage we know that if it is not there then it is not required.

Agreed on 100% code coverage being highly desirable. This being said, I would still want the code to be 'correct'. For that reason, I would go for @hsorby's suggestion about "export[ing] the symbol when testing".

src/variable.cpp
dladd: Sorry- I had swapped the excepted cases to the end while we were playing with work-arounds. They should be back to alpha-order now.

agarny:
Regarding @dladd's latest commits (https://github.com/cellml/libcellml/pull/80/commits/7d8396c8317736e2ad16dd1e264b728f1220310c and https://github.com/cellml/libcellml/pull/80/commits/c30e9cb76f736d3aeaa45c86ade866353fef9c6e), am I missing something or shouldn't have buildbot been triggered for them?

hsorby:
Yes, it should have.  The buildmaster has been upgraded recently so something must not be working correctly post updgrade.  I'll investigate tomorrow.

agarny:
Ok, thanks @hsorby.

dladd:
This now uses EXPORT_FOR_TESTING to cover the offending lines in the enum switches and bring our coverage back to 100%. Thanks to @hsorby for helping sort this out.

nickerso:
The CI still not working?

nickerso:
Assuming the testing is all good and coverage is at 100% then I think this is all looking good and all the review comments have been addressed.

hsorby:
> The CI still not working?

Sorry forgot to check while in the office today, I'll write it on the back of my hand for tomorrow.

src/CMakeLists.txt
```diff
@@ -85,6 +85,12 @@ target_include_directories(cellml
   ${CMAKE_CURRENT_BINARY_DIR}
 )
 
+if (LIBCELLML_UNIT_TESTS)
```
agarny: No space between `if` and `(`.

agarny:
Pending the very minor correction above, I am with David: as long as the testing and coverage are all good/100%, then it's all fine with me too.

dladd:
Thanks- I've just fixed that spacing typo. This passed all tests with 100% coverage on my machine but we could double check tomorrow with the buildbot.

hsorby:
CI is back online thanks back of hand (@nickerso)

dladd:
CI did pick up a Windows compilation bug that I've fixed now.

nickerso:
There is still an issue with the generated documentation not being available, but looking at the testing logs the 100% coverage is clear to see (and the tests are all passing), so I'll merge this one in.

Issue #82, opened 2016-04-18
Current thinking on connections


dladd:
The next feature to build on from the use-cases doc is 1.xii: a model with connections. I thought it would be a good idea to put down some of the current thinking on how things might fit together.

Consider a simple serialised model with a connection:

``` xml
    <?xml version=\"1.0\" encoding=\"UTF-8\"?>
    <model xmlns=\"http://www.cellml.org/cellml/1.2#\">
      <component name=\"component1\">
        <variable name=\"variable1\"/>
      </component>
      <component name=\"component2\">
        <variable name=\"variable2\"/>
      </component>
      <connection>
        <map_components component_1=\"component1\" component_2=\"component2\"/>
        <map_variables variable_1=\"variable1\" variable_2=\"variable2\"/>
      </connection>
    </model>
```

To build this up, two options have been discussed so far: (1) allowing direct access to connections through the API and (2) keeping connections "under the hood":
- 1. The connection is built through the API and passed variables to map: e.g.,

``` c++
    libcellml::Model m;
    // Create connection object.
    libcellml::ConnectionPtr connection1 = std::make_shared<libcellml::Connection>(); 
    libcellml::ComponentPtr comp1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp2 = std::make_shared<libcellml::Component>();
    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();

    comp1->setName("component1");
    comp2->setName("component2");
    v1->setName("variable1");
    v2->setName("variable2");

    comp1->addVariable(v1);
    comp2->addVariable(v2);
    m.addComponent(comp1);
    m.addComponent(comp2);
    // Pass the two variables to the connection object
    connection1.mapVariables(v1,v2); 
    // Add the connection to the model
    m.addConnection(connection1); 
```
- 2. The connection is instantiated as a part of the model inside the library: e.g., don't create `connection1` and replace the last two non-comment lines from (1) with:

``` c++
    ...
    // Instantiate the connection; pass v1 and v2 to the connection; add the connection to the parent model
    v1.isEquivalentTo(v2); 
```

In both cases, I am assuming that we get each component for `map_components` as each variable's respective parent component. In (2) I also assume that we can similarly get the parent model from these components to instantiate a new connection. 

[...]

PR #85, opened 2016-04-20
Add connections to use-cases object model

dladd:
This updates the use-cases object model diagram to 1.xii: a model with connections. It builds from the discussion from issue #82 .

In this proposal, variable equivalence is set on variable objects using (for example) `v1->setEquivalentTo(v2)` methods. In this case, if `v2` is not already in the variable attribute set `mEquivalentVariables` on `v1`, `v2` will be added to that set. Multiple equivalent variables may be set on each variable. The reciprocal equivalence (i.e., `v2->setEquivalentTo(v1)`) will also similarly be set automatically.

At serialisation, it becomes necessary to identify the complete set of unique variable equivalences in the model. This is done by constructing an attribute set `mVariableMaps` on the model object. A `VariableMap` is a unique variable equivalence pair (see NOTE 1).  Uniqueness is determined by querying variables and their respective `mEquivalentVariables` in the model against the pairs in `mVariableMaps` as it is built up. 

Once the set `mVariableMaps` is built, connections may be serialised by sequentially serialising out each `VariableMap` consisting of variable pairs with the same parent components.

NOTE 1: I'm open to others' preference for the naming what I am calling a `VariableMap` in this proposal- the variable equivalence pair. They match up with the concept of `map_variables` in the Spec so I could also see `MapVariables`, `VariableEquivalencePair`, or `VariablePair` as other ready choices. It makes sense to me to keep this concept separate from a "connection", which may be composed of many `map_variables` of the same `map_components`.

EDIT: added variable reciprocal equivalence


jonc125:
How does this handle reciprocity of equivalence? What about a situation where we do:

```
v1.setEquivalentTo(v2);
v1.setEquivalentTo(v3);
v4.setEquivalentTo(v5);
v2.setEquivalentTo(v5);
```

?

nickerso:
I like this proposal in terms of implementation, but I'm not sure I see a reason to expose the variable maps in the libCellML API - it really seems to be an internal implementation detail.

> How does this handle reciprocity of equivalence?

I don't think this is a problem as CellML itself is only concerned with connections between pairs of components. And I think this proposal will be sufficient to achieve serialisation. More work will be needed to handle the case when we want to start assembling variable equivalence graphs over the whole model when its instantiated into a procedural code or something, but that will be a separate layer of libCellML. I think :)

dladd:
Just to illustrate,

In @jonc125 's case, this would currently construct `mEquivalentVariables` on each variable:

| v1->mEquivalentVariables | v4->mEquivalentVariables | v2->mEquivalentVariables |
| --- | --- | --- |
| v2 | v5 | v5 |
| v3 |  |  |

And actually wouldn't serialise out anything unless the variables were added to components and models. **If the variables were each in their own component**, we would have:

| mVariableMaps[0] | mVariableMaps[1] | mVariableMaps[2] | mVariableMaps[3] |
| --- | --- | --- | --- |
| v1 | v1 | v4 | v2 |
| v2 | v3 | v5 | v5 |

and the XML connections output would be:

``` xml
      <connection>
        <map_components component_1=\"c1Name\" component_2=\"c2Name\"/>
        <map_variables variable_1=\"v1Name\" variable_2=\"v2Name\"/>
      </connection>
      <connection>
        <map_components component_1=\"c1Name\" component_2=\"c3Name\"/>
        <map_variables variable_1=\"v1Name\" variable_2=\"v3Name\"/>
      </connection>
      <connection>
        <map_components component_1=\"c4Name\" component_2=\"c5Name\"/>
        <map_variables variable_1=\"v4Name\" variable_2=\"v5Name\"/>
      </connection>
      <connection>
        <map_components component_1=\"c2Name\" component_2=\"c5Name\"/>
        <map_variables variable_1=\"v2Name\" variable_2=\"v5Name\"/>
      </connection>
```

But if, for example, v2 and v3 were in the same component (c2),  v1 was in c1, and we drop v4 and v5 for simplicity,

| v1->mEquivalentVariables |
| --- |
| v2 |
| v3 |

| mVariableMaps[0] | mVariableMaps[1] |
| --- | --- |
| v1 | v1 |
| v2 | v3 |

``` xml
      <connection>
        <map_components component_1=\"c1Name\" component_2=\"c2Name\"/>
        <map_variables variable_1=\"v1Name\" variable_2=\"v2Name\"/>
        <map_variables variable_1=\"v1Name\" variable_2=\"v3Name\"/>
      </connection>
```

EDIT: Clarified that the `map_variables` and `map_components` identifiers are the names at serialisation.

dladd:
> I'm not sure I see a reason to expose the variable maps in the libCellML API

That might be a misunderstanding on my part re: the UML class diagram? I envisioned wanting access to functions in the `VariableMap` class during serialisation from the `Model` class, not to open them up to the API.

Also, in the current proposal, the instantiation of every `VariableMap` is a private operation in the `Model` class to happen at serialisation. So a potential user shouldn't be aware of the existence of variable maps or be able to get a handle on them.

nickerso:
OK, seems I've been interpreting this diagram wrong :) I think @hsorby has some ideas on how to clarify the bits that are part of the public API and which are internal implementation, but that is a different issue.

agarny:
This all sounds good to me. Regarding the proposed updated diagram:
- `Model` class:
  - Why have `createVariableMap()`? Couldn't we simply have the body of that function in `addVariable()`, which would mean `addVariable()` getting the signature of `createVariableMap()`?
- `Variable` class:
  - For consistency, `countEquivalentVariables()` ought to be renamed `equivalentVariableCount()`.
  - I am not sure I see the point of having `setEquivalentTo()` anymore since we have `addEquivalentVariable()` (which I actually prefer since its name is more consistent).
  - Despite my previous comment, is the idea of the different signatures for `setEquivalentTo()` to be able to refer another variable by providing a pointer to its corresponding component, as well as the name of that variable or its index in the list of variables? If so, do we really need this? I would think that we could achieve the same result using other methods from the API?
- `VariableMap` class:
  - We know that `mVariables` will always consist of two variables, so why not simply have `mVariable1` and `mVariable2`?
  - As a result of my previous comment, `getVariable()` would be replaced with `getVariable1()` and `getVariable2()` (which means that we wouldn't have to handle things like `getVariable(13)`).

jonc125:
OK, I agree on not making equivalence transitive at this stage. But I think we do still want reciprocity. If as a user of the API I set `v1` equivalent to `v2` I would expect to see `v2` appearing in the set of variables equivalent to (i.e. connected directly to) `v1` **and** vice versa.

dladd:
Solid points, @agarny - I'll re-jig things tomorrow. The only thing I might note is on `setEquivalentTo()` vs. `addEquivalentVariable()`. The thinking there is that when a user calls `v1.setEquivalentTo(v2)` there will be an internal check on whether `v2` is already in `mEquivalentVariables` on `v1`. If not, `addEquivalentVariable(v2)` adds it. If so, this step is skipped to cut down on duplicates building up in `mEquivalentVariables`.

Re: reciprocity @jonc125 , I suppose we could also add the reciprocal equivalence without too much trouble. It will cause twice as much work when we are building the unique set of `VariableMaps`. But I could see how users might expect this and rely on it.

dladd:
Since we're going with reciprocal variable equivalences, we should probably be changing the method name `setEquivalentTo()` to something like `setEquivalence()`. The former reads as more of a uni-directional relationship. 

Or, actually, we should just probably make the uni-directional `setEquivalentTo()` private, to be called by the public `setEquivalence()` to avoid infinite recursion.
I.e., `v1.setEquivalence(v2)` calls `v1.setEquivalentTo(v2)` and `v2.setEquivalentTo(v1)`

dladd:
Re: my second to last comment:

Apologies, @agarny , I realise now that you probably meant "why not bring the functionality of `addEquivalentVariable()` into `setEquivalentVariable()`" rather than asking what the former does. I think that is a fair point- it probably won't be necessary (especially at this planning stage).

dladd:
Thanks all- I've now updated the diagram. This also simplifies the diagram based on feedback from issue #87 . 

I've changed the main public-facing function to `addEquivalence()` per @agarny 's suggestion but kept a private uni-directional function as `setEquivalentTo()` per my comment above.

agarny:
This is starting to look good to me. Just a few more comments:
- Overloaded functions: consider `Component::getVariable(std::string): VariablePtr {...}`. Personally, this gives me the impression that `getVariable()` always has the same signature and that, you telling me it's an overloaded function, it is its implementation that differs (!!). Why not simply have `Component::getVariable(...): VariablePtr` to indicate that a function is overloaded?
- `Variable` class: I can see that you ordered the attributes alphabetically. That is fine by me, but then you might want to do the same for the `Entity` class.
- `VariableMap` class: why have renamed `hasVariable()` to `containsVariable()`, which is not consistent with `Variable::hasEquivalentVariable()` anymore?

dladd:
Thanks again, @agarny , I've now incorporated your most recent points into the diagram. I also reordered the corresponding groups of operations in `Variable` to match the alphabetical attribute order. Re: `has...` vs. `contains...`, it seems like the `contains...` prefix should be left for encapsulation.

nickerso:
seems ready for implementation to me :)

hsorby:
I don't think there will be a link between the variable map and the model classes.  The complete map of connections will need to be constructed on demand I think, much the same as what is done for the imports currently.

dladd:
I see... if we did this more similarly to [how imports are constructed](https://github.com/cellml/libcellml/blob/develop/src/model.cpp#L52), we also might be able to cut out some of the methods on `Model` in favour of standard functions like `make_pair`. We might even be able to do away with the `VariableMap` class itself and instead just `push_back` equivalence pairs into a VariableMap `container`. 

Functionally, everything should work the same as the diagram. Just using standard functions rather than created ones.

Maybe we should go ahead with tests/implementation but leave this pull open until we agree which way is a better approach?

agarny:
It all looks fine to me now, but I shall leave it to @hsorby to merge the PR in case he has more comments to make.

dladd:
I've now updated this to match the implementation from pull #89. This removes the private `VariableMap` class and associated methods on the `Model` class. `Variable::addEquivalence()` is underlined to indicate that it is a static function, following UML-style.

agarny:
Fine by me.

Issue #88, opened 2016-04-26
Support for connections (use-case 1.xii)


dladd:
This issue adds support for connections, building on the discussion from issue #82 and pull #85. 

In addition to tests for expected functionality and full coverage, specific outlined use-cases to test for are:
- [x] **a.**  a model with one component encapsulating two children, each child containing a single variable
  - [x] **1.**  with a private interface in the parent and public interface in the child components and connect the variable in both children to the parent.
  - [x] **2.**  with a public interface in all components and connect the variables in the children to the parent

PR #89, opened 2016-04-26
Add tests and implementation for connections (use-case 1.xii)

dladd:
Addresses issue #88

### DIFF
--- a/src/component.cpp
+++ b/src/component.cpp
@@ -37,8 +37,8 @@ struct Component::ComponentImpl
 
 std::vector<VariablePtr>::iterator Component::ComponentImpl::findVariable(const std::string &name)
 {
-    return std::find_if(mVariables.begin(), mVariables.end(),
-                        [=](const VariablePtr& v) -> bool {return v->getName() == name;});
+    return std::find_if (mVariables.begin(), mVariables.end(),
+                        [=](const VariablePtr& v) -> bool { return v->getName() == name; });
 }
 
 Component::Component()
--- a/src/component.cpp
+++ b/src/component.cpp
@@ -88,6 +88,7 @@ void Component::setSourceComponent(const ImportPtr &imp, const std::string &name
 void Component::addVariable(const VariablePtr &v)
 {
     mPimpl->mVariables.push_back(v);
+    v->setParent(this);
 }
 
 void Component::removeVariable(const std::string &name)
--- a/src/component.cpp
+++ b/src/component.cpp
@@ -37,7 +37,7 @@ struct Component::ComponentImpl
 
 std::vector<VariablePtr>::iterator Component::ComponentImpl::findVariable(const std::string &name)
 {
-    return std::find_if (mVariables.begin(), mVariables.end(),
+    return std::find_if(mVariables.begin(), mVariables.end(),
                         [=](const VariablePtr& v) -> bool { return v->getName() == name; });
 }
 
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -214,9 +214,6 @@ std::string Model::doSerialisation(FORMATS format) const
                 if (currentComponent1 == nextComponent1 && currentComponent2 == nextComponent2) {
                     connection += "<map_variables variable_1=\"" + variablePair2.first->getName() + "\""
                                                 " variable_2=\"" + variablePair2.second->getName() + "\"/>";
-                } else if (currentComponent1 == nextComponent2 && currentComponent2 == nextComponent1) {
-                    connection += "<map_variables variable_1=\"" + variablePair2.second->getName() + "\""
-                                                " variable_2=\"" + variablePair2.first->getName() + "\"/>";
                 }
                 ++componentMapIndex2;
             }
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -22,6 +22,7 @@ limitations under the License.Some license of other
 #include <stack>
 
 #include <libcellml/import.h>
+#include <libcellml/variable.h>
 
 namespace libcellml {
 
@@ -51,10 +52,22 @@ Model& Model::operator=(Model m)
 
 std::string Model::doSerialisation(FORMATS format) const
 {
+    // ImportMap
     typedef std::pair <std::string, std::string> ImportNamePair;
     typedef std::vector<ImportNamePair>::const_iterator VectorPairIterator;
     typedef std::map <ImportPtr, std::vector<ImportNamePair> > ImportMap;
     typedef ImportMap::const_iterator ImportMapIterator;
+    // VariableMap
+    typedef std::pair <VariablePtr, VariablePtr> VariablePair;
+    typedef std::vector<VariablePair>::const_iterator VariableMapIterator;
+    typedef std::vector<VariablePair> VariableMap;
+    VariableMap variableMap;
+    // ComponentMap
+    typedef std::pair <Component*, Component*> ComponentPair;
+    typedef std::vector<ComponentPair>::const_iterator ComponentMapIterator;
+    typedef std::vector<ComponentPair> ComponentMap;
+    ComponentMap componentMap;
+    ComponentMap serialisedComponentMap;
 
     // Gather all imports
     ImportMap importMap;
@@ -86,6 +99,34 @@ std::string Model::doSerialisation(FORMATS format) const
                 comp = comp->getComponent(index);
             } else {
                 incrementComponent = true;
+            }
+
+            // Build unique variable equivalence pairs (VariableMap)
+            for (size_t j = 0; j < comp->variableCount(); ++j) {
+                VariablePtr variable = comp->getVariable(j);
+                if (variable->equivalentVariableCount() > 0) {
+                    for (size_t k = 0; k < variable->equivalentVariableCount(); ++k) {
+                        VariablePtr equivalentVariable = variable->getEquivalentVariable(k);
+                        VariablePair variablePair = std::make_pair(variable, equivalentVariable);
+                        VariablePair reciprocalVariablePair = std::make_pair(equivalentVariable, variable);
+                        bool pairFound = false;
+                        for (VariableMapIterator iter = variableMap.begin(); iter < variableMap.end(); ++iter) {
+                            if (*iter == variablePair || *iter == reciprocalVariablePair) {
+                                pairFound = true;
+                                break;
+                            }
+                        }
+                        if (!pairFound) {
+                            // New unique variable equivalence pair found
+                            variableMap.push_back(variablePair);
+                            // Also create a component map corresponding with pairs of the variable map
+                            Component* component1 = static_cast<Component*>(variable->getParent());
+                            Component* component2 = static_cast<Component*>(equivalentVariable->getParent());
+                            ComponentPair componentPair = std::make_pair(component1, component2);
+                            componentMap.push_back(componentPair);
+                        }
+                    }
+                }
             }
 
             if (incrementComponent) {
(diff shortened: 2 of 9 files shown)


tests/connection/connection.cpp
```diff
@@ -0,0 +1,281 @@
...
+}
+
+TEST(Connection, componentlessVariableInvalidConnection) {
+    const std::string e =
+    "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+    "<model xmlns=\"http://www.cellml.org/cellml/1.2#\">"
+      "<component name=\"component1\">"
+        "<variable name=\"variable1\"/>"
+      "</component>"
+      "<variable name=\"variable2\"/>"
```
nickerso: I wouldn't expect to see this here.

tests/connection/connection.cpp
```diff
@@ -0,0 +1,281 @@
...
+    parent->addVariable(v1);
+    child1->addVariable(v2);
+    child2->addVariable(v3);
+    v1->addEquivalence(v2);
+    v1->addEquivalence(v3);
+    v1.setInterfaceType(libcellml::Variable::INTERFACE_TYPE_PRIVATE);
+    v2.setInterfaceType(libcellml::Variable::INTERFACE_TYPE_PUBLIC);
+    v3.setInterfaceType(libcellml::Variable::INTERFACE_TYPE_PUBLIC);
+
+    std::string a = parent.serialise(libcellml::FORMAT_XML);
```
nickerso: should be `m.serialise(...)` :)

src/variable.cpp
```diff
@@ -77,6 +87,36 @@ void Variable::swap(Variable &rhs)
     std::swap(this->mPimpl, rhs.mPimpl);
 }
 
+void Variable::addEquivalence(const VariablePtr &equivalentVariable)
+{
+    setEquivalentTo(equivalentVariable);
+    VariablePtr *pV = new VariablePtr(this);
```
dladd: For anyone that happens to be looking through this now, I definitely plan to sort out a safer and more C++11 way to do this tomorrow. I was just using this to check that the rest of the current code passes the variable-side connections tests. 

Of course, any suggestions are always welcome :-)

nickerso: I'm starting to wonder if, given the reciprocal nature of this method, perhaps it is better implemented as a static method on the variable class that takes two `VariablePtr` arguments?

dladd: A static method does indeed seem to get around the issue nicely. It also takes some of the possible ambiguity about reciprocity for anyone not digging into the documentation (I would expect reciprocity out of `addEquivalence(v1,v2)` but not necessarily out of `v1.addEquivalence(v2)`)

dladd:
This now passes all tests and proposed use-cases. A few notes:
- Per discussion at the end of #85 , I've simplified the connection serialisation from the original object model proposal with an explicit private `VariableMap` class. Instead, the same functionality is implemented using vectors of c++ `pairs` in `Model::doSerialisation`. This is also more consistent with the way the `ImportMap` is currently constructed.
- `addEquivalence(v1,v2)` is now implemented as a static `Variable` function. This will hopefully make the reciprocity of this method more obvious to the user than the `v1.addEquivalence(v2)` proposal. 

~~Another issue I noticed is that we haven't yet properly tested serialisation of variables inside imported components. If there are no objections, I'll add that as a separate bugfix request pending the review of this.~~

dladd:
Apologies- please ignore the above comment re: imported component variables. @nickerso confirmed for me that we don't expect variables to be serialised from imported components. 

I've also now added a passing test for a connection between imported and non-imported components.

src/model.cpp
```diff
@@ -51,10 +52,22 @@ Model& Model::operator=(Model m)
...
     typedef ImportMap::const_iterator ImportMapIterator;
+    // VariableMap
+    typedef std::pair <VariablePtr, VariablePtr> VariablePair;
+    typedef std::vector<VariablePair>::const_iterator VariableMapIterator;
+    typedef std::vector<VariablePair> VariableMap;
+    VariableMap variableMap;
+    // ComponentMap
+    typedef std::pair <Component*, Component*> ComponentPair;
+    typedef std::vector<ComponentPair>::const_iterator ComponentMapIterator;
+    typedef std::vector<ComponentPair> ComponentMap;
```
nickerso, line 68: I'd probably change the code slightly for these to make it a bit cleaner, maybe :)

``` c
typedef std::pair<Component*, Component*> ComponentPair;
typedef std::vector<ComponentPair> ComponentMap;
typedef ComponentMap::const_iterator ComponentMapIterator;
```

src/model.cpp
```diff
@@ -127,9 +140,87 @@ std::string Model::doSerialisation(FORMATS format) const
 
         repr += serialiseUnits(format);
 
-        // Serialize components of the model, imported components have already been dealt with at this point.
+        // Serialise components of the model, imported components have already been dealt with at this point.
```
agarny, line 143: Nitpicking: sometimes comments finish with `.`, sometimes not...

src/model.cpp
```diff
@@ -127,9 +140,87 @@ std::string Model::doSerialisation(FORMATS format) const
...
         for (size_t i = 0; i < componentCount(); ++i) {
             repr += getComponent(i)->serialise(format);
+            // Build unique variable equivalence pairs (VariableMap) for connections
+            ComponentPtr component = getComponent(i);
+            for (size_t j = 0; j < component->variableCount(); ++j) {
+                VariablePtr variable = component->getVariable(j);
+                if (variable->equivalentVariableCount() > 0) {
```
agarny, line 150: I would drop the `> 0` in the test, but that's just me.

src/model.cpp
```diff
@@ -127,9 +140,87 @@ std::string Model::doSerialisation(FORMATS format) const
...
+                    for (size_t k = 0; k < variable->equivalentVariableCount(); ++k) {
+                        VariablePtr equivalentVariable = variable->getEquivalentVariable(k);
+                        VariablePair variablePair = std::make_pair(variable, equivalentVariable);
+                        VariablePair reciprocalVariablePair = std::make_pair(equivalentVariable, variable);
+                        bool pairFound = false;
+                        for (VariableMapIterator iter = variableMap.begin(); iter < variableMap.end(); ++iter) {
+                            if (*iter == variablePair || *iter == reciprocalVariablePair) {
```
agarny: I haven't checked the coding style, but personally I prefer to make it (very) clear what we are testing. In other words, I would have:

```
...
if ((*iter == variablePair) || (*iter == reciprocalVariablePair)) {
...
```

Same elsewhere in the code (e.g. line 185 here).

src/model.cpp
```diff
@@ -127,9 +140,87 @@ std::string Model::doSerialisation(FORMATS format) const
...
+                }
+            }
+            // Continue to the next component pair if the current pair has already been serialised
+            if (pairFound) {
+                ++componentMapIndex1;
+                continue;
+            }
+            // Serialise out the new connection
+            std::string connection = "<connection><map_components";
+            if (currentComponent1 != nullptr) {
```
agarny: As for the `> 0` above, I would drop the `!= nullptr`, but again that's just me. Similarly, I wouldn't use curly brackets here since the IF statement is a single-line statement, but then again the Google C++ style guide allows it...

src/model.cpp
```diff
@@ -127,9 +140,87 @@ std::string Model::doSerialisation(FORMATS format) const
...
+                connection += " component_1=\"" + currentComponent1->getName() + "\"";
```
agarny, line 198: I wouldn't have spaces around `+` (same elsewhere in the code).

src/model.cpp
```diff
@@ -127,9 +140,87 @@ std::string Model::doSerialisation(FORMATS format) const
...
+                connection += " component_2=\"" + currentComponent2->getName() + "\"";
+            }
+            VariablePair variablePair = variableMap.at(componentMapIndex1);
+            connection += "/><map_variables variable_1=\"" + variablePair.first->getName() + "\""
+                                        + " variable_2=\"" + variablePair.second->getName() + "\"/>";
+            // Check for subsequent variable equivalence pairs with the same parent components
+            int componentMapIndex2 = componentMapIndex1 + 1;
+            ComponentMapIterator duplicateIterPair = iterPair;
+            ++duplicateIterPair;
+            for (ComponentMapIterator iterPair2 = duplicateIterPair; iterPair2 < componentMap.end(); ++iterPair2) {
```
agarny: Can't `duplicateIterPair` be dropped and have `... iterPair2 = iterPair+1; ...` instead?

src/variable.cpp
```diff
@@ -31,17 +33,27 @@ namespace libcellml {
...
+    std::vector<VariableWeakPtr> mEquivalentVariables; /**< Equivalent variables for this Variable.*/
     std::string mInitialValue = ""; /**< Initial value for this Variable.*/
     INTERFACE_TYPES mInterfaceType = INTERFACE_TYPE_NONE; /**< Interface type for this Variable. Default to none.*/
     UnitsPtr mUnits; /**< A pointer to the Units defined for this Variable.*/
 };
 
+std::vector<VariableWeakPtr>::iterator Variable::VariableImpl::findEquivalentVariable(const VariablePtr &equivalentVariable)
+{
+    return std::find_if(mEquivalentVariables.begin(), mEquivalentVariables.end(),
+                        [=](VariableWeakPtr variableWeak) -> bool {return equivalentVariable == variableWeak.lock();});
```
agarny: I would put a space after `{` and before `}`, i.e.

```
[=](VariableWeakPtr variableWeak) -> bool { return equivalentVariable == variableWeak.lock(); });
```

src/variable.cpp
```diff
@@ -77,6 +90,37 @@ void Variable::swap(Variable &rhs)
...
+
+size_t Variable::equivalentVariableCount() const
+{
+    return mPimpl->mEquivalentVariables.size();
+}
+
+bool Variable::hasEquivalentVariable(const VariablePtr &equivalentVariable)
+{
+    auto result = mPimpl->findEquivalentVariable(equivalentVariable);
+    return result != mPimpl->mEquivalentVariables.end();
```
agarny: Why not simply:

```
return mPimpl->findEquivalentVariable(equivalentVariable) != mPimpl->mEquivalentVariables.end(); 
```

?

tests/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ include(component/tests.cmake)
 include(coverage/tests.cmake)
 include(units/tests.cmake)
 include(variable/tests.cmake)
+include(connection/tests.cmake)
```
agarny: Might want to order those includes?

tests/connection/connection.cpp
```diff
@@ -0,0 +1,546 @@
...
+
+#include "gtest/gtest.h"
+
+#include <libcellml>
+
+TEST(Variable, addAndGetEquivalentVariable) {
+    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
+    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
+    libcellml::Variable::addEquivalence(v1,v2);
+    EXPECT_EQ(v2,v1->getEquivalentVariable(0));
```
agarny: Missing space before `v1` (same for `v2` at lines 24, 31 and 32). (Same elsewhere in the code.)

tests/connection/connection.cpp
```diff
@@ -0,0 +1,546 @@
...
+    EXPECT_EQ(v1,v2->getEquivalentVariable(0));
+}
+
+TEST(Variable, addTwoEquivalentVariablesAndCount) {
+    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
+    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
+    libcellml::VariablePtr v3 = std::make_shared<libcellml::Variable>();
+    libcellml::Variable::addEquivalence(v1,v2);
+    libcellml::Variable::addEquivalence(v1,v3);
+    size_t e = 2;
```
agarny: Missing `const`? (Same elsewhere in this code.)

tests/connection/connection.cpp
```diff
@@ -0,0 +1,546 @@
...
+    libcellml::ComponentPtr comp2 = std::make_shared<libcellml::Component>();
+    comp2->setName("component2");
+    libcellml::VariablePtr v11 = std::make_shared<libcellml::Variable>();
+    v11->setName("variable11");
+    libcellml::VariablePtr v12 = std::make_shared<libcellml::Variable>();
+    v12->setName("variable12");
+    libcellml::VariablePtr v21 = std::make_shared<libcellml::Variable>();
+    v21->setName("variable21");
+    libcellml::VariablePtr v22 = std::make_shared<libcellml::Variable>();
+    v22->setName("variable22");
```
agarny, line 166: I would personally declare everything and then set the names.

Otherwise, I like it that there are empty lines here and there in this test (in fact, I would have added a few more). I find it easier to read the code... (not only the tests, but the code in general)

tests/connection/connection.cpp
```diff
@@ -0,0 +1,546 @@
...
+    std::string a = m.serialise(libcellml::FORMAT_XML);
+    EXPECT_EQ(e, a);
+}
+
+TEST(Connection, twoEncapsulatedChildComponentsWithConnectionsAndMixedInterfaces) {
+    const std::string e =
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model xmlns=\"http://www.cellml.org/cellml/1.2#\">"
+
+          "<component name=\"parent_component\">"
```
agarny: Might simply want to call it `parent`? (Same elsewhere in the code.)

agarny:
Quite a bit of nitpicking, sorry...

agarny:
With regards to issue #88, I think I am happy with this PR. I think because there is quite a bit of code to review, so I have just had a very quick glance at it, but it passes the tests, so...

src/model.cpp
jonc125, line 150: I'd keep it here, as you're comparing a count to some value - the method doesn't return a boolean. So this way is semantically clearer.

src/model.cpp
jonc125: In this case I think treating the pointer as a boolean is actually a more normal approach to testing for null with smart pointers, although the pointers here are raw pointers so less clear.

And I'd keep the curly brackets - avoids future mistakes if the if body becomes multi statement :)

src/variable.cpp
hsorby: Line width issue? I think originally this was a hangover from debugging a problem.

src/model.cpp
hsorby, line 198: Always spaces around operators!

src/variable.cpp
dladd: I've swapped in @agarny 's suggestion but you are right, this does violate the Google style guide on [max line lengths of 80 characters](https://google.github.io/styleguide/cppguide.html#Line_Length). We do appear to be +80 characters in many places though so it might be worth deciding if we want to rigidly follow Google on this or add an exception to our coding style.

dladd:
Thanks all- this should now address the above comments.

src/component.cpp
```diff
@@ -37,8 +37,8 @@ struct Component::ComponentImpl
 
 std::vector<VariablePtr>::iterator Component::ComponentImpl::findVariable(const std::string &name)
 {
-    return std::find_if(mVariables.begin(), mVariables.end(),
-                        [=](const VariablePtr& v) -> bool {return v->getName() == name;});
+    return std::find_if (mVariables.begin(), mVariables.end(),
```
agarny: Just for information (i.e. don't make it delay the merge!): `std::find_if()` is a function, not a an `if` statement as such, so I wouldn't expect a space to be 'needed.

agarny:
Fine by me.

src/component.cpp
dladd: Fair enough- I didn't see an instance of find_if in the Google style guide but did see it on [cplusplus.com](http://www.cplusplus.com/reference/algorithm/find_if/) with spacing. But that makes sense and my Stroustrup book has it without the spacing.

Issue #90, opened 2016-04-27
Funding acknowledgement


nickerso:
We need to be sure to include a funding acknowledgement section in the documentation and track all the funding received to support libCellML development. Perhaps even on the front page of the website with logos and links?

agarny:
+1 on having such an acknowledgement in the documentation.

As for the website, well... what website? For the CellML API, I am aware of https://www.cellml.org/tools/api and http://cellml-api.sourceforge.net/, but do we have anything for libCellML per se? (Incidentally, I can't say that I like having something hosted on SourceForge.)

nickerso:
> As for the website, well... what website? 

Well, currently its just http://libcellml.readthedocs.org :) no plans that I'm aware of to make a separate website, although we should really make sure the links from cellml.org are in place.

agarny:
Ok, then +1 on updating cellml.org with the right link to http://libcellml.readthedocs.org/ and having the front page of the libCellML website with logos and links.

nickerso:
just to point out that readthedocs have today changed everything to readthedocs.io, so the website will now be http://libcellml.readthedocs.io - although http://libcellml.readthedocs.org will redirect for the foreseeable future.

dladd:
I've now added this in pull #91 . You can find the [buildbot test page here](http://autotest.bioeng.auckland.ac.nz/libcellml-buildbot/7035f42beaf3f6586d2196c518f5e8f40e5fe079/dox/index.html).

PR #91, opened 2016-04-27
Add funding acknowledgement

dladd:
Adds funding acknowledgement to libCellML ReadTheDocs documentation.

Addresses #90


docs/index.rst
```diff
@@ -39,3 +39,15 @@ Indices and tables
 * :ref:`modindex`
 * :ref:`search`
 
+Supported by:
+------------
+
+.. figure:: docs/images/vpr.png
```
nickerso: should just be `images/vpr.png` I think.

nickerso:
All looks good to me.

agarny:
The rendering looks a bit strange to me:
image

Regarding the VPR logo, there is thin border to the left and that cyan border at the top, which don't look great. As for the ABI logo, I wonder why there is that dotted line at the bottom...?!

dladd:
Thanks @agarny - I've fixed the border issue on the vpr logo ([new version here](http://autotest.bioeng.auckland.ac.nz/libcellml-buildbot/fc2fc9d55fc6aeb9408234a5a766271b98e26fb1/dox/index.html)). I believe the dotted lines are the style for hyperlinks- you'll see they go solid when you mouseover the logo. I could look into adding an exception for images (I assume there is a stylesheet somewhere in our theme) but it just depends how much time we want to spend on this.

nickerso:
The dotted line is fine. No need to tune this any more at this stage :)

Issue #65, opened 2016-04-05
Use cases for math


nickerso:
It would be useful to add use-cases for math to the current list (i.e., 1.xii and 1.xiii in the current list at http://libcellml.readthedocs.org/en/latest/usecases.html)

jonc125:
Where you have just maths & variables, I guess things like:
- a = 1 ms
- a = b
- For each (CellML subset) MathML operator, a separate test case of the simplest possible equation involving it
- A few more complex cases with several operators and levels of nesting
- A simple ODE, e.g. dV/dt = 1 mV/ms
- A simple system of equations, e.g. a=1, a=b+c, b=2

Since we're just checking construction and serialisation at this stage this is perhaps slightly overkill (the last in particular) but they're the sort of thing we'll need when considering units conversions, math evaluation, etc.

With connections, a few of the simple multi-variable cases above, but where the variables come from different components. Again, since we're not yet evaluating, handling source variables several connections away is probably overkill.

dladd:
From the discussion on pull #66 and resulting changes to [the current thinking doc](http://libcellml.readthedocs.io/en/latest/current_thinking.html), the consensus was to move forward with treating MathML as a string.

Some possible testing permutations would then be:
- A model with valid/invalid maths in a component
  - with no/one/multiple connections
    - with valid/invalid/no variable(s)
      - with valid/invalid/dimensionless/no units

I propose modifying the [use-cases document](http://libcellml.readthedocs.io/en/latest/usecases.html) to explicitly capture a few of these (more permutations will also be tested).
- **1.xiii.** a model with maths and variables
  - **a.** model from _1.xi.d.1_ and define invalid maths
  - **b.** model from _1.xi.d.1_ and define valid maths using the two variables with units dimensionless
- **1.xiv.** a model with maths, variables and connections
  - **a.** model with two components, each containing two variables, maths, and one connection 
  - **b.** model from _1.vi.a_ and define valid maths on variables in child and parent components with one connection

nickerso:
seems like a good idea to me.

jonc125:
If we're just treating maths as opaque strings and not resolving variables or units references, those cases seem... difficult to distinguish?  And we're not doing validation at all yet.  Although I agree they're the sort of thing that would be useful longer term!

dladd:
[...]

PR #96, opened 2016-05-10
Add maths use-cases to doc

dladd:
Addresses issue #65.

### DIFF
--- a/docs/usecases.rst
+++ b/docs/usecases.rst
@@ -81,7 +81,7 @@ Use-cases for libCellML
          
    #. a model with maths and variables
    
-      a. model from *1.xi.d.1* and define invalid maths
+      a. model from *1.xi.d.1* and define valid maths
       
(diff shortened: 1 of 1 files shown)


agarny:
Looks good to me.

docs/usecases.rst
```diff
@@ -81,11 +81,11 @@ Use-cases for libCellML
          
    #. a model with maths and variables
    
-      a. … to be added to ...
+      a. model from *1.xi.d.1* and define invalid maths
```
jonc125: Shouldn't this be valid maths? Not that it really matters at this stage, but saying our first use case is invalid maths sounds bizarre :)

agarny: Hmm, indeed, well spotted!

dladd:
Good point @jonc125 - updated now.

Issue #94, opened 2016-05-05
Support for maths (use-cases 1.xiii-xiv)


dladd:
This issue covers adding support for maths, building on the discussion from issue #65 and pull #66.

This includes:
- [x] Updates to the object model
- [x] Tests and implementation that cover expected functionality and specific use-cases:
  - [x] **1.xiii.** a model with maths and variables
    - [x] **a.** model from _1.xi.d.1_ and define invalid maths
  - [x] **1.xiv.** a model with maths, variables and connections
    - [x] **a.** model with two components, each containing two variables, maths, and one connection

PR #97, opened 2016-05-10
Add string-based maths to object model

dladd:
Adds math-as-string attribute (`mMaths`) and operations (`appendMaths()`, `getMaths()`, and `removeMaths()`) to `Component` class.

Addresses part of issue #94.


dladd:
If we plan to keep the maths-as-strings functionality in the long-term, we may want to change attribute and function names to `mMathsString`, `appendMathsString()`, etc. to avoid confusion.

nickerso:
Possibly a difference between en-UK and en-US, but I would go for the singular `math` rather than `maths` ;-)

I don't have a preference as to `math` vs `mathString`, very slightly inclined toward `mathString`...

dladd:
Works for me- updated to mathString now :)

nickerso:
I'm happy with this API.

jonc125:
Indeed, en-GB would prefer maths :) but for an international effort I'm not going to quibble. And it matches the MathML element name too.

API looks good to me.

agarny:
Argh, it just got closed, but still:
- I would, for now at least, stick to `math` rather than go with `mathString`. (As for the tests, I don't think we should anticipate on what we might need, etc.)
- I wonder whether `appendMathString()` is strictly needed. Why not simply have a setter? Sure, if the user effectively wants to add a mathematical equation, then s/he will have to call the getter too, but by having a setter it means that we wouldn't have a need for `removeMathString()` either. In other words, I would simply have `setMath()` and `getMath()`. I like the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle). :)

jonc125:
Whereas I'd go with being explicit about what the methods are doing, and if you just have `getMath` there's ambiguity about what is returned. We may well get rid of the `*String` methods later, but I don't think there's any harm in having the slightly longer names now.

And `append` is a really useful operation for users, so I vote for keeping it.

agarny:
Fair enough on the `append` function. However, to be consistent with the rest of the API, shouldn't it be renamed `addMath[String]()`? I know it's not quite the same meaning, but...

Regarding the `*String` bit, well... if we were to follow that logic, then we probably ought to rename `Variable::getInitialValue()`.

Talking of consistency, shouldn't `ComponentEntity::containsComponent()` be renamed `ComponentEntity::hasComponent()`?

dladd:
> to be consistent with the rest of the API, shouldn't it be renamed addMath[String]?

I would think this would indicate to users that they would be adding a new separate `math[String]` to the collective attribute like `addVariable()` or `addComponent()`. Whereas `append...` indicates that you are appending to a single attribute.

> Regarding the *String bit, well... if we were to follow that logic, then we probably ought to rename Variable::getInitialValue().

Fair point- I'm happy to rename accordingly (or remove `String` accordingly :) )

> Talking of consistency, shouldn't ComponentEntity::containsComponent() be renamed ComponentEntity::hasComponent()?

I believe the current thinking there was to use the `contains` prefix for encapsulation, whereas the `has` prefix was to be for one-way attributes?

agarny:
> I would think this would indicate to users that they would be adding a new separate `math[String]` to the collective attribute like `addVariable()` or `addComponent()`. Whereas `append...` indicates that you are appending to a single attribute.

Good point. I am now fine with it.

> Fair point- I'm happy to rename accordingly (or remove `String` accordingly :) )

I think it's pretty obvious what my preference is on this. :)

> I believe the current thinking there was to use the `contains` prefix for encapsulation, whereas the `has` prefix was to be for one-way attributes?

Hmm, good point too. In that case, it means that `Variable::hasEquivalentVariable()` ought to be renamed to `Variable::containsEquivalentVariable()`? If so, then mea culpa since I believe I originally asked you to rename it from `contains` to `has`.

jonc125:
I don't have a strong preference on the `String` issue.

Re `Variable::hasEquivalentVariable()` this is not encapsulation but referencing - the variable doesn't contain its equivalents, they are contained in their respective components. So I think `has` makes more sense.

PR #101, opened 2016-05-10
Swap method/attribute names from 'mathString' to 'math' in object model

dladd:
Resolves issue under discussion after merge of pull #97 .

Addresses part of issue #94


nickerso:
Seems there is no strong preference for `mathString` and one strong preference for `math`, so I think this is fine to merge in.

hsorby:
I have a strong preference for `math` over `mathString` also, I also think this is fine to merge in.

Issue #86, opened 2016-04-21
Using PREFIX_UNIT to indicate an undefined prefix


dladd:
Currently, unit prefixes are set by default to the [PREFIX_UNIT enumerator](https://github.com/cellml/libcellml/blob/develop/src/api/libcellml/enumerations.h#L51) as a tidy way to indicate that no prefix has been set.

However, it has been suggested that this may not be strictly correct, since a base unit by definition does not have a prefix and therefore may not belong in the PREFIXES enum.  An alternative proposal is to remove `PREFIX_UNIT` and add a `mHasPrefix` Boolean attribute to `Unit` to indicate whether a prefix has been set.

To vote, use a thumbs-up for keeping `PREFIX_UNIT`, a thumbs-down for `mHasPrefix`, or leave a comment for other proposals (or general discussion).

nickerso:
Just to clarify that the actual default value for a unit prefix is the real value `0.0`, not an undefined prefix name string. Following the other discussions regarding `interface=none`, when the prefix has this value it should not be serialised, but in all other cases the prefix should be serialised. To me, this means that if the prefix enum has a value, it should be serialised. Hence my thumbs-down, although perhaps really should be `mHasPrefixEnum` or `mHasPrefixName`.

hsorby:
> Just to clarify that the actual default value for a unit prefix is the real value 0.0

From the spec. libcellml needs to output or validate according to the spec. but isn't tied directly to it.

I would prefer `mPrefixEnumSet` for class attribute.

I would have preferred to have this incorporated in the enum but sadly the people who make definitions down at SI didn't define `unit` or an equivalent, very inconsiderate!

agarny:
I am a bit confused. If we were to have `mHasPrefix` (or whatever name we would end up with) and get rid of `PREFIX_UNIT`, a base unit would still have the attributes `mPrefixInt` and `mPrefixEnum` (like any other unit). So, what value would those attributes have?

Talking of the `mPrefixInt` and `mPrefixEnum` attributes, I had never realised that the prefix information of a unit is stored either as an `int` or an `enum`. Why not simply have a `mPrefix` attribute as an `int` (and convert an `enum` to an `int` whenever necessary)?

jonc125:
Indeed, I hadn't spotted that. Or even use a `union`?

hsorby:
Just as a note that `int` should really be a `double` or `float` it's a mistake that it is an `int`

dladd:
[...]

PR #103, opened 2016-05-12
Changes to unit prefix handling

dladd:
This implements and tests changes to internal handling of units:
- Removes `PREFIX_UNIT`
- Replaces `mPrefixEnum` and `mPrefixInt` with `mPrefix` (string)
- Replaces `switch`-based handling of `prefixToString()` with a `map`
- All `struct Unit` members now stored as strings
- `Unit::mUnits` changed to `Unit::mName` 

Changes since last round of comments:
- Removed prefix exception handling (now serialises all non-zero user-specified prefix strings). This can be re-added when we get to the validation step and formalise our error/exception handling.
- Included an assert check for prefixToString search.

Addresses issue #86

### DIFF
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -39,7 +39,7 @@ std::string prefixToString(PREFIXES prefix);
  */
 struct Unit
 {
-    std::string mUnits; /**< Name for the unit.*/
+    std::string mName; /**< Name for the unit.*/
     std::string mPrefix = "0.0"; /**< String expression of the prefix for the unit.*/
     double mExponent = 1.0; /**< Exponent for the unit.*/
     double mMultiplier = 1.0; /**< Multiplier for the unit.*/
@@ -130,7 +130,7 @@ std::string Units::doSerialisation(FORMATS format) const
                         if (u.mPrefix != "0.0") {
                             repr += " prefix=\"" + u.mPrefix + "\"";
                         }
-                        repr += " units=\"" + u.mUnits + "\"";
+                        repr += " units=\"" + u.mName + "\"";
                         repr += "/>";
                     }
                     repr += "</units>";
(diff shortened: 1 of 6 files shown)


src/units.cpp
```diff
@@ -157,40 +167,65 @@ void Units::setBaseUnit(bool state)
...
+void Units::addUnit(const std::string &name, PREFIXES prefix, double exponent,
              double multiplier, double offset)
 {
-    Unit u;
-    u.mUnits = units;
-    u.mPrefixInt = prefix;
-    u.mExponent = exponent;
-    u.mMultiplier = multiplier;
-    u.mOffset = offset;
+    auto search = prefixToString.find(prefix);
```
nickerso, line 208: not sure which is better, but I commonly use this instead of find:

``` C++
if (prefixToString.count(prefix)) prefixString = prefixToString(prefix);
```

nickerso:
Looking good to me.

src/units.cpp
dladd, line 208: For this to work correctly, would we need to take the initialising integers out of the PREFIXES enum (see [enumerations.h](https://github.com/cellml/libcellml/blob/develop/src/api/libcellml/enumerations.h#L41))? That's fine by me if so and this is preferred.

dladd, line 208: (Re-commenting to trigger notification)

For this to work correctly, would we need to take the initialising integers out of the PREFIXES enum (see [enumerations.h](https://github.com/cellml/libcellml/blob/develop/src/api/libcellml/enumerations.h#L41))? That's fine by me if so and this is preferred.

dladd, line 208: I did a quick Google re: `find()` vs `count()` and it looks like for this case, `find()` is [slightly preferred](http://stackoverflow.com/questions/25490357/checking-for-existence-in-stdmap-count-vs-find). 

The justification is that `count()` is slightly less efficient as it continues iterating until the end of the map range whether or not the key has been found. The intent of `find()` is also arguably clearer (we are trying to find the key in the map).

src/units.cpp
```diff
@@ -15,21 +15,42 @@ limitations under the License.
 */
 #include "libcellml/units.h"
 
-#include <vector>
+#include <assert.h>
```
agarny, line 18: Nitpicking: might prefer `#include <cassert>` to `#include <assert.h>`.

Issue #105, opened 2016-05-13
Code clean-up items before 'Modify' step


dladd:
This issue picks up general items for clean-up before we move on to Modify use-cases. Feel free to comment with anything else you would like to see addressed.
- [x] Change variable interface type handling to be consistent with unit prefixes from #103 
  - Would allow us to get rid of the remaining EXPORT_FOR_TESTING instance
  - If we store the interface type as a string internally, we could:
    - allow the user to specify the interface type by string.
    - allow the user to serialise out `interface_type=none` if they choose to do so.
- [x] Swap out plain `enum`s for `enum class`es where possible
  - More strongly typed and generally preferred in C++11
  - Example usage:

``` c++
libcellml::Prefixes prefix = libcellml::Prefixes::MICRO;
libcellml::Variable::InterfaceTypes interfaceType1 = libcellml::Variable::InterfaceTypes::NONE;
```

vs. the current style:

``` c++
libcellml::PREFIXES prefix = libcellml::PREFIX_MICRO;
libcellml::Variable::INTERFACE_TYPES interfaceType1 = libcellml::Variable::INTERFACE_TYPE_NONE;
```

jonc125:
I haven't looked through the code, but the aims look good to me.

agarny:
Same as @jonc125.

PR #106, opened 2016-05-13
Add documentation for the use of labels

dladd:
Addresses issue #74.

PR #107, opened 2016-05-16
Implement clean-up items before 'Modify' step

dladd:
Addresses #105. 

Please note that this pull builds on the commits from pull #103 and should be merged after it to avoid conflicts. Commits from May 16 onward are unique to this pull request.


dladd:
This now passes all tests and addresses #105. I've also added changes to address the remaining comments from #103 and #106.

nickerso:
Seems all good to me.

src/api/libcellml/variable.h
```diff
@@ -140,28 +140,36 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
-     * options in the INTERFACE_TYPES enum.
+     * options in the InterfaceTypes enum class.
+     *
+     * @overload
      * @sa getInterfaceType
-     * @param interfaceType The interface type to set for this variable.
+     * @param interfaceType The enum interface type to set for this variable.
      */
-    void setInterfaceType(INTERFACE_TYPES interfaceType);
+    void setInterfaceType(InterfaceTypes interfaceType);
```
agarny, line 165: Nitpicking: I really find it odd to have a type called `InterfaceTypes` (rather than `InterfaceType`, i.e. singular) when the argument refers to one and only one interface type... (Same goes with `Formats` rather than `Format`.)

src/variable.cpp
```diff
@@ -15,23 +15,28 @@ limitations under the License.
...
+ * @brief Map to convert an interface type into its string form.
+ * An internal map used to convert a Variable InterfaceTypes enum class member into its string form.
  */
-EXPORT_FOR_TESTING std::string interfaceTypeToString(Variable::INTERFACE_TYPES interfaceType);
+std::map<Variable::InterfaceTypes, std::string> interfaceTypeToString =
+{
+    {Variable::InterfaceTypes::NONE, "none"},
+    {Variable::InterfaceTypes::PRIVATE, "private"},
+    {Variable::InterfaceTypes::PUBLIC, "public"},
+    {Variable::InterfaceTypes::PUBLIC_AND_PRIVATE, "public_and_private"}
```
agarny, line 38: Nitpicking: I would add a space after/before `{`/`}`.

tests/variable/variable.cpp
```diff
@@ -166,19 +182,19 @@ TEST(Variable, setGetInterfaceTypes) {
...
     v4.setInterfaceType(interfaceType4);
 
-    EXPECT_EQ(interfaceType1, v1.getInterfaceType());
-    EXPECT_EQ(interfaceType2, v2.getInterfaceType());
-    EXPECT_EQ(interfaceType3, v3.getInterfaceType());
-    EXPECT_EQ(interfaceType4, v4.getInterfaceType());
+    EXPECT_EQ("none", v1.getInterfaceType());
+    EXPECT_EQ("private", v2.getInterfaceType());
+    EXPECT_EQ("public", v3.getInterfaceType());
+    EXPECT_EQ("public_and_private", v4.getInterfaceType());
```
agarny, line 197: Nitpicking: I reckon `"none"`, `"private"`, etc. should actually be `std::string` variables?...

agarny:
Some nitpicking, but otherwise it looks good to me, so merging.

src/api/libcellml/variable.h
dladd, line 165: OK I'll change this up in the next round of tidy-up (this is a bit far-reaching to sneak into another changeset). It's implied in the examples in the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Enumerator_Names) that they use plural `enum` names but not explicitly stated. I agree that singular reads better :)

src/variable.cpp
dladd, line 38: I think this might go against the [Google C++ style guide on braced initialiser lists](https://google.github.io/styleguide/cppguide.html#Braced_Initializer_List_Format)? (see the last example) But I can add an exception in our coding style and standardise across our braced lists if this is a strong preference.

jonc125, line 38: I think the lack of spaces is fine. I would indent lines 34-39 another 4 spaces though :)

agarny, line 38: Fair enough about the lack of spaces being fine. The fewer exceptions to the Google C++ style guide the better.

Otherwise, personally, I would indent. I prefer it the way it is. Ok, nitpicking over! :)

Issue #104, opened 2016-05-13
Modify use-cases step (2)


dladd:
This issue covers the Modify step as outlined in the [use-cases doc](http://libcellml.readthedocs.io/en/latest/usecases.html). This includes:
- [x] Updates to the object model
- [x] Tests and implementation to modify models from the Create step (1).
  - [x] add {components, units, maths, variables, connections}
  - [x] remove {components, units, maths, variables, connections}
  - [x] update {components, units, maths, model attributes, variables, connections}

dladd:
I think we've already covered the "add" task with the current API. 

For the "remove" task, I believe we'll also need the methods (edited per comment below):
- `Units::removeUnit(std::string): void`
- `Variable::removeEquivalence(VariablePtr, VariablePtr): void`
- `Component::removeVariable(VariablePtr): void`
- `ComponentEntity::removeUnits(UnitsPtr)`

We could also add `removeAll...()` methods to reset collective attributes if that would be useful for a user.

For the "update" task, some more input would be helpful. The only thing that I think might have needed an explicit `update...()` method is a connection. However, connections are currently constructed at-need and not persistently stored. So I think they (and the other elements) should already be in an 'updated' state?

jonc125:
Indeed, I think "update" is implicit in allowing set, remove & add. The `set*` methods probably do what was envisaged here (e.g. update model attributes, i.e. `setName`). Some remove functionality is there already too.

I don't think you need an argument to `Variable::removeUnits` as a variable can only have one set of units associated. You would need the argument for places that stored multiple different units (models, components) but it seems these are handled implicitly in the current object model?

`Component::removeVariable` could do with a variant passing a `VariablePtr` perhaps.

Presumably the `set*` methods on `Variable` would have equivalent behaviour to remove if passed an empty string/pointer? Which suggests that `removeUnits` might be redundant here (or adding for clarity, in which case the same argument applies to initial value and interface type)?

dladd:
Good points, @jonc125. 

> You would need the argument for places that stored multiple different units (models, components) but it seems these are handled implicitly in the current object model?

It looks like we missed `ComponentEntity::addUnits(UnitsPtr)` in the current object model diagram (I'll add it as part of this changeset). So we'll need a corresponding `ComponentEntity::removeUnits(UnitsPtr)`. Then `remove...()` methods will correspond with `add...()` methods. 

[...]

PR #108, opened 2016-05-17
Update use-cases object model to Modify step (2)

dladd:
Adds `remove...()` methods to the object model diagram to satisfy the Modify use-cases step (2). Also adds existing `ComponentEntity` units features to the diagram. Since changes can be a bit difficult to track in an image, I've added a list of the new methods below.

Addresses part of #104 
### Component

| Method | Description |
| --- | --- |
| removeVariable(VariablePtr): void | removes argument from this component's mVariables |
| removeAllVariables(): void | resets this component's mVariables collective attribute |
| setMath(std::string): void | sets/removes the math string; to replace removeMath() |
### ComponentEntity

| Method/Attribute | Description |
| --- | --- |
| mUnits | collective Units attribute; currently exists in code but not object model |
| addUnits(UnitsPtr): void | add argument to this mUnits; currently exists in code but not object model |
| removeUnits(UnitsPtr): void | remove UnitsPtr from this mUnits |
| removeUnits(std::string): void | remove named units from this mUnits |
| removeAllUnits(): void | resets this component entity's mUnits |
| unitsCount(): const size_t | returns the number of units in this component entity |
| removeComponent(ComponentPtr): void | removes component by pointer |
| removeAllComponents(): void | removes all components from this component entity |
### Units

| Method | Description |
| --- | --- |
| removeUnit(std::string): void | removes named unit from this mUnits |
| removeAllUnits(): void | resets this units' mUnits |
| unitCount(): const size_t | returns the number of units that compose this units |
### Variable

| Method | Description |
| --- | --- |
| removeEquivalence(VariablePtr,VariablePtr): void | static; removes equivalence with reciprocity, calls v1->unsetEquivalentTo(v2) and v2->unsetEquivalentTo(v1) |
| removeAllEquivalences(): void | resets this variable's mEquivalentVariables |
| unsetEquivalentTo(VariablePtr): void | private; opposite function to setEquivalentTo(), removes instance of the argument variable in this variable's mEquivalentVariables |


agarny:
For `ComponentEntity::removeUnits()`, we might want to have a version that takes a `std::string` as an input. In other words, to do something similar to what is done/suggested for `Component::removeVariable(...)`.

I would use `Variable::removeAllEquivalences()` rather than `Variable::removeAllEquivalentVariables()`, and thus be more in sync with say `Component::addVariable()`, `Component::removeVariable()` and `Component::removeAllVariables()`.

Otherwise, shouldn't there be a composition link between `ComponentEntity` and `Units`?

dladd:
Thanks- I've now updated this per @agarny 's comment. I also swapped `removeMath()` with `setMath(std::string)` per discussion in the parent issue and updated the variable interface routines per the changes from #105.

nickerso:
looks good to me.

dladd:
I've also now added `ComponentEntity::unitsCount()` and `Units::unitCount()` to this to be consistent with our other collective attributes.

PR #109, opened 2016-05-18
Add tests and implementation for 'Modify' (use-case 2)

dladd:
Addresses #104.


dladd:
This now passes all tests and implements the features discussed so far in #108.

src/api/libcellml/component.h
```diff
@@ -74,16 +76,32 @@ class LIBCELLML_EXPORT Component: public ComponentEntity
...
      * @sa addVariable
-     * @overload
      * @param name The name of the variable to remove.
      */
     void removeVariable(const std::string &name);
 
     /**
+     * @brief Remove the variable by the given @p variable pointer from this component.
+     * Remove the variable with the given pointer from this component. If the
+     * variable is not found then no variable is removed.
```
jonc125: Is there a reason this is different from the "remove by name" case? Seems like it would be sensible to throw for both, as it's an indication of programmer error?

jonc125: And the units case below throws too!

src/api/libcellml/variable.h
```diff
@@ -179,12 +198,23 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
-     * @sa addEquivalence
+     * @sa addEquivalence, unsetEquivalentTo
      * @param equivalentVariable The variable to add to this variable's equivalent
      * variable set if not already present.
      */
     void setEquivalentTo(const VariablePtr &equivalentVariable);
 
+    /**
+     * @brief Private function to remove an equivalent variable from the set for this variable.
+     * Remove the argument equivalent variable from the set of equivalent variables for this
```
jonc125: Nitpick: could use `@p` to make this clearer!

jonc125:
I've gone through this fairly quickly, but it looks good to me.

src/api/libcellml/component.h
```diff
@@ -74,16 +76,32 @@ class LIBCELLML_EXPORT Component: public ComponentEntity
...
+     * @overload
+     * @param variable The pointer to the variable to remove.
+     */
+    void removeVariable(const VariablePtr &variable);
+
+    /**
+     * @brief Remove all variables stored in this component.
+     * Clears all variables that have been added to this component.
+     */
+    void removeAllVariables();
```
hsorby, line 108: What does this do for connections? Are they completely removed or do we have some dangling connections?  Whatever the behaviour it should be documented here.

src/api/libcellml/component.h
```diff
@@ -60,10 +60,12 @@ class LIBCELLML_EXPORT Component: public ComponentEntity
...
     /**
-     * @brief Remove the math string for this component.
-     * Resets the math string for this component to an empty string.
+     * @brief Set the math string for this component.
+     * Sets the math string for this component. If @p math is an empty
+     * string, math will be removed from the component.
+     * @param math The @c std::string to append for this component.
      */
-    void removeMath();
+    void setMath(const std::string &math);
```
agarny, line 68: Couldn't we have `math` defaulting to an empty string?

src/api/libcellml/component.h
```diff
@@ -74,16 +76,32 @@ class LIBCELLML_EXPORT Component: public ComponentEntity
...
      * @param name The name of the variable to remove.
      */
     void removeVariable(const std::string &name);
 
     /**
+     * @brief Remove the variable by the given @p variable pointer from this component.
+     * Remove the variable with the given pointer from this component. If the
+     * variable is not found then no variable is removed.
+     * @sa addVariable
+     *
```
agarny, line 96: Nitpicking: remove this line.

src/api/libcellml/component.h
agarny, line 108: I imagine this applies to both `removeVariable(...)` and `removeAllVariables()`.

src/api/libcellml/componententity.h
```diff
@@ -47,11 +47,41 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
...
+     * @brief Add a child units to this component entity.
+     * Add a copy of the given units as a child units of this component entity.
+     * @param units The units to add.
      */
-    void addUnits(const UnitsPtr &u);
+    void addUnits(const UnitsPtr &units);
+
+    /**
+     * @brief Remove the units with the given name.
+     * Remove the first units found that matches the name @p name.  If the
```
agarny: Can there be more than one units with the same name? I would think not?...

src/api/libcellml/componententity.h
```diff
@@ -47,11 +47,41 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
...
+     * name is not found throw @c std::out_of_range.
+     * @param name The name of the units to remove.
+     */
+    void removeUnits(const std::string &name);
+
+    /**
+     * @brief Remove the units with the given pointer.
+     * Remove the first units found with the pointer @p units.  If the
```
agarny: Same here (can there be more than one units with the same name?), not least because it's a pointer. (Same elsewhere in the code in fact.)

src/api/libcellml/componententity.h
```diff
@@ -47,11 +47,41 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
...
+
+    /**
+     * @brief Remove all units stored in this component entity.
+     * Clears all units that have been added to this component entity.
+     */
+    void removeAllUnits();
+
+    /**
+     * @brief Get the number of units in the component entity.
+     * Returns the number of units the component entity directly contains.
```
agarny: Why _directly_? I would remove it.

src/api/libcellml/componententity.h
```diff
@@ -72,14 +102,30 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
...
      * @overload
      * @param name The name of the component to remove.
      */
     void removeComponent(const std::string &name);
 
     /**
+     * @brief Remove the component with the given pointer.
+     * Remove the first component found with the pointer @p component.  If the
+     * component is not found throw @c std::out_of_range.
+     *
```
agarny, line 116: Nitpicking: remove line.

src/api/libcellml/variable.h
```diff
@@ -65,6 +65,25 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
     static void addEquivalence(const VariablePtr &variable1, const VariablePtr &variable2);
 
     /**
+     * @brief Remove each argument variable to the other's equivalent variable set.
+     * Removes a copy of @p variable1 from the set of equivalent variables for
+     * @p variable2 if present. Also removes a copy of @p variable2 from the
```
agarny: Nitpicking: "... variable2**,** if present..." (comma). (Same on the next line.)

src/model.cpp
```diff
@@ -127,7 +127,11 @@ std::string Model::doSerialisation(Formats format) const
         if (getName().length()) {
             repr += " name=\"" + getName() + "\"";
         }
-        repr += ">";
+        bool endTag = false;
+        if (importMap.size() > 0 || componentCount() > 0 || unitsCount() > 0){
```
agarny: Nitpicking: replace with `if ((importMap.size() > 0) || (componentCount() > 0) || (unitsCount() > 0)) {`.

tests/connection/connection.cpp
```diff
@@ -379,6 +379,104 @@ TEST(Connection, twoValidConnections) {
...
+
+    comp1->addVariable(v1);
+    comp2->addVariable(v2);
+    comp3->addVariable(v3);
+    m.addComponent(comp1);
+    m.addComponent(comp2);
+    m.addComponent(comp3);
+    libcellml::Variable::addEquivalence(v1,v2);
+    libcellml::Variable::addEquivalence(v1,v3);
+    libcellml::Variable::addEquivalence(v2,v3);
```
agarny: Nitpicking: space before the second variable name. Same on lines 470 and 473.

agarny:
Besides my non-nitpicking comments and those of @jonc125 and @hsorby too, it all looks good to me too.

tests/connection/connection.cpp
mirams: ALAN!

src/api/libcellml/component.h
jonc125, line 68: I think I'd prefer having to pass an empty string explicitly to remove.

tests/connection/connection.cpp
agarny: Eh, I did put "Nitpicking" in front, hence this is just for information and this is in no way meant to prevent a PR merge. :) (As it happens, I have merged PRs for which I had made some nitpicky comments. :))

src/api/libcellml/component.h
dladd, line 108: > Are they completely removed or do we have some dangling connections?

There shouldn't be any dangling connections- our `VariableMap` in `Model::doSerialisation` is now only constructed of unique variable equivalence pairs that have equivalence reciprocity (i.e., exist in each other's `mEquivalentVariables`). 

I'll add doc and tests to make this clearer and double-check. :-)

dladd, line 108: Actually, good catch @hsorby - there is indeed a dangling connection bug here now being tested for in a328468126654ca64c6251b66fb124c0cf890b61

src/api/libcellml/component.h
dladd, line 96: I believe the standard in our existing Doxygen documentation is to have a blank line before an "overload"?

src/api/libcellml/componententity.h
nickerso: in a valid model, no - but users should be able to construct invalid models.

src/api/libcellml/componententity.h
nickerso: perhaps making the distinction between imported units and "local" units?

src/api/libcellml/componententity.h
dladd, line 116: As above re: "overload". I can bring this up in the next round of general cleanup :)

dladd:
Many thanks all- this should now address the comments so far and passes all tests. Apologies for leaving some extra nits in this one. 

For connected variables removed from components, I've stuck with the previous behaviour re: connected variables without components. That is: we can still serialise a connection when one variable lacks a parent component but this variable will not have a component in `map_components`. For a test showing this behaviour, please see [removeVariablesFromConnections](https://github.com/cellml/libcellml/pull/109/files#diff-f1c314e34b078cb9060cf8fd0e7a97e1R480).

src/api/libcellml/component.h
agarny, line 96: I don't know, I haven't checked. It's just that in other places, there is no blank line before an `@overload`. Anyway, this shouldn't hold up the merge indeed.

src/api/libcellml/componententity.h
agarny: Oops, good point!

src/api/libcellml/componententity.h
agarny: I was originally thinking about the difference between units defined at the model level as opposed to units defined in a component, but there is indeed also the case of imported units.

agarny:
Looks good to me.

nickerso:
Looking good to me too.

Issue #110, opened 2016-05-19
Code clean-up items before 'Load' step


dladd:
Proposed cleanup items are:
- [x] **(A)** Change `enum class` names from plural to singular. For example, `libcellml::Prefixes::MICRO` would become `libcellml::Prefix::MICRO` (from comments in #107).
- [x] **(B)** ~~Change Doxygen documentation to remove space above `@overload`~~ (_Edit_) Change Doxygen documentation handling of whitespace (TBD below) [Example of current style](https://github.com/cellml/libcellml/blob/develop/src/api/libcellml/units.h#L150) 
- [x] **(C)** Further tests and implementation for C++11 copy/move

dladd:
If I get a vote, I am for (A) but against (B). I think it is useful to have the visual cue of the extra space to indicate that you are looking at a block of overloaded methods when scrolling through a header file.

agarny:
I currently vote for A only. Indeed, regarding B, your example actually highlights a couple of issues for me:
- I don't know the rules when it comes to mentioning `@overload`, but why isn't `@overload` mentioned for the first `addUnit()` method? Imagine if, for some reason or another, you were to re-order the different `addUnit()` methods, then you might accidentally end up with the first `addUnit()` method mentioning `@overload` and one of the subsequent `addUnit()` methods not mentioning it. So, if anything, I would like to have `@overload` mentioned for all `addUnit()` methods.
- I agree that to have a blank line before `@overload` makes it stand out, but I wonder whether we couldn't have a blank line before `@param` too (only if it's not an overloaded method). It would make the parameters stand out and if it was to be an overloaded method, then `@overload` would still stand out.

jonc125:
I almost never say no to extra blank lines :)

nickerso:
Whitespace

good

:)

dladd:
Re:

> I don't know the rules when it comes to mentioning `@overload`, but why isn't `@overload` mentioned for the first addUnit() method?

If you have a look in the [generated Doxygen](http://autotest.bioeng.auckland.ac.nz/libcellml-buildbot/cf1c987e74bfe9efc35ec7811686c8fe719b7410/dox/code/classlibcellml_1_1Units.html#a95299aac24939e24a6a9e8cf4dcfc960), mentioning `@overload` triggers the following note:

_This is an overloaded member function, provided for convenience. It differs from the above function only in what argument(s) it accepts._

So it should probably only be used in overloaded methods following the first instance.

<br />
Re:

[...]

PR #112, opened 2016-05-20
Implement clean-up items before 'Load' step

dladd:
Addresses #110

### DIFF
--- a/tests/coverage/coverage.cpp
+++ b/tests/coverage/coverage.cpp
@@ -55,8 +55,11 @@ TEST(Coverage, entity) {
 }
 
 TEST(Coverage, units) {
-    std::string e = "";
+    std::string e = "<units name=\"dimensionless\" base_unit=\"yes\"/>";
     libcellml::Units u, um;
+
+    u.setName("dimensionless");
+    u.setBaseUnit();
 
     um = std::move(u);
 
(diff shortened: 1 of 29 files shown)


tests/coverage/coverage.cpp
```diff
@@ -139,3 +148,35 @@ TEST(Coverage, variable) {
...
+    libcellml::Component p, pm;
+    libcellml::ComponentPtr child = std::make_shared<libcellml::Component>();
+    p.addComponent(child);
+    EXPECT_EQ(e, p.serialise(libcellml::Format::XML));
+
+    pm = std::move(p);
+    EXPECT_EQ(e, pm.serialise(libcellml::Format::XML));
+
+    libcellml::Component pc(pm);
+    EXPECT_EQ(e, pm.serialise(libcellml::Format::XML));
```
hsorby: Probably need to test against the expected string of pc and not pm.

tests/coverage/coverage.cpp
```diff
@@ -139,3 +148,35 @@ TEST(Coverage, variable) {
...
+    c.addVariable(v);
+    c.setMath("<1+1=2>");
+    EXPECT_EQ(e, c.serialise(libcellml::Format::XML));
+
+    cm = std::move(c);
+    EXPECT_EQ(e, cm.serialise(libcellml::Format::XML));
+
+    // Copy constructor
+    libcellml::Component cc(cm);
+    EXPECT_EQ(e, cm.serialise(libcellml::Format::XML));
```
hsorby: Probably need to test against the expected string of cc and not cm.

agarny:
Besides @hsorby's comments, it all looks good to me.

dladd:
Thanks all- this now addresses comments, passes tests, and fulfils items discussed in #110

Issue #111, opened 2016-05-19
'Load' use-cases step (3)


dladd:
This issue covers the Load step as outlined in the [use-cases doc](http://libcellml.readthedocs.io/en/latest/usecases.html). This currently includes:
- [x] Updates to the object model
- [x] Load each of the models from 1 and 2 (new models can be added for this part if required).
  - [x] a model with imported components
    - [x] a single component
    - [x] a component with a hierarchy
    - [x] a component from a non-existent URL

The major task for this step is adding `deserialise(std::string)` operations to parse out CellML 1.2 objects. ~~using regular expressions.~~

dladd:
In terms of testing, we could (option A) construct another parallel set of deserialisation test cases- for example:

``` c++
TEST(Component, serialiseValidName) {
    const std::string name = "valid_name";
    const std::string e = "<component name=\"" + name + "\"/>";
    libcellml::Component c;
    c.setName(name);
    std::string a = c.serialise(libcellml::Formats::XML);
    EXPECT_EQ(e, a);
}

TEST(Component, deserialiseValidName) {
    const std::string e = "valid_name";
    const std::string input = "<component name=\"" + e + "\"/>";
    libcellml::Component c = libcellml::Component::deserialise(input);
    std::string a = c.getName;
    EXPECT_EQ(e, a);
}
```

Another option (B) is to simply add a deserialisation/reserialisation step on to the existing test cases:

``` c++
TEST(Component, validName) {
    const std::string name = "valid_name";
    const std::string e = "<component name=\"" + name + "\"/>";
    libcellml::Component c;
    c.setName(name);
    std::string a = c.serialise(libcellml::Formats::XML);
    EXPECT_EQ(e, a);
    libcellml::Component deserialisedC = libcellml::Component::deserialise(e);
    std::string a = deserialisedC.serialise(libcellml::Formats::XML);
    EXPECT_EQ(e, a);
}
```

I think it makes sense to do separate test cases (A) for some of the initial tests while we get this up and running. However, for more complex cases, de/reserialisation (B) will allow us to quickly work through our very thorough testing base with minimal reduplication of effort and additional code to maintain.

jonc125:
Regarding the approach to parsing, is using regular expressions really the way we want to go? There are so many good XML parsing libraries out there, which have been thoroughly tested...

agarny:
[...]

PR #113, opened 2016-05-30
Update use-cases diagram with proposed load XML methods

dladd:
This updates the use-cases diagram with features to satisfy the 'Load' use-cases step (3).

This incorporates the use of LibXML2 for parsing XML strings and documents through the wrapper classes `XmlDoc` and `XmlNode`. 

In this proposal, "deserialise" methods are used, consistent with our "serialise" methods. Another strategy that has been mentioned in #111 is the possibility of using a separate parser. In addition to `doDeserialisation()` functions (not explicitly pictured in the diagram) on each of the deserialisation targets, I'm anticipating convenience functions to deserialise connections and encapsulations in `Model`.

Addresses part of #111


agarny:
This seems fine to me, as a starting point.

Just one thing: I am not sure about 'deserialisation'. In the end, I am wondering whether I wouldn't personally prefer to have `import()` and `export()` methods rather than `deserialisation()` and `serialisation()`?

nickerso:
I'd worry about conflicting interpretation of `import` in relation to CellML imports.

agarny:
> I'd worry about conflicting interpretation of `import` in relation to CellML imports.

Hmm, good point! I still don't like 'deserialisation' though... :) (I think I can live with it though. :))

dladd:
We could maybe swap "deserialise" with "load" or "read"? I'm not sure what would be most clear to a user... "Deserialise" is probably closest to what it actually does but is not necessarily a common phrase.

agarny:
No, having 'serialise', it does indeed make sense to have 'deserialise'. (I am sure I will get used to it.)

dladd:
Updated with separate parser class per comments in #111

nickerso:
Just to be clear, I believe for now all we need is the `parseModel` methods.

dladd:
Fair enough- updated now. I think it still makes sense to modularise the `loadA(APtr, XmlNodePtr)` methods as part of this. But we can add `parseA(string)` methods when/if we need them.

agarny:
Looks OK to me.

dladd:
This has now been updated to match the discussion to date and implementation from #114. I've left out the entity error classes (e.g., `VariableError`, `ImportError`, etc) for now as this is getting a bit crowded if we want to keep this to A4. I'm wondering if it might be better to have another sub-diagram for our error handling/validation. Any preferences?

hsorby:
I would put these error classes on the main object model diagram and not this rough object model diagram.  Although I would be tempted to put a little marker down to make people aware of them.

dladd:
Sounds good @hsorby . That diagram is probably due for an update now anyway :)

nickerso:
I think this one is ready to be merged, right?

PR #114, opened 2016-06-04
Add tests and implementation for 'Load` (use-cases step 3)

dladd:
This changeset contains my current work on implementation and tests for our XML parser and will address #111 upon completion. This is still work-in-progress but, with the HARMONY hackathon approaching, I figured I should get this up and a pull request started.

I've gone with string handling of variable units for now but this may change pending further discussion in #111 

Some things that need fixing up and/or filling out:
- [x] Iterate over all attributes rather than checking for expected attributes
- [x] Recursive encapsulation
- [x] Conversion of all math element content to string
- ~~Updating existing model content~~
- ~~Parsing objects other than models~~


dladd:
This now addresses all tasks and discussed features from #111 and passes all tests.

dladd:
I've now also added a hook to feed parsing errors from libxml (for ex. mismatched start/end tags) into our error handler. This still passes all tests.

src/api/libcellml/logger.h
```diff
@@ -0,0 +1,90 @@
...
+    void addError(EntityErrorPtr error);
+
+    /**
+     * @brief Get the number of errors.
+     *
+     * Return the number of errors currently stored in the logger.
+     *
+     * @return The number of errors.
+     */
+    size_t errorCount() const;
```
jonc125, line 65: A minor point, and this may be consistent with other similar cases in the code, but I'd normally favour method names being verb-based, and this just has nouns. Would `getErrorCount()` be better?

src/api/libcellml/parser.h
```diff
@@ -0,0 +1,168 @@
...
+     *
+     * Updates the argument @p model with CellML entities and attributes
+     * from the argument @p input @c std::string. Any entities, objects,
+     * or attributes of with matching names in the model will be
+     * overwritten by those loaded from the @p input @c std::string.
+     *
+     * @param model The @c ModelPtr to update.
+     * @param input The string to parse into the @p model.
+     */
+    void updateModel(const ModelPtr &model, const std::string &input);
```
jonc125: I thought we weren't doing the update case yet?

src/api/libcellml/parser.h
```diff
@@ -0,0 +1,168 @@
...
+     *
+     * Loads the argument @p model with CellML connection information parsed from
+     * the argument XML @p node. If any variable equivalence relationships already
+     * exist in the @p model, the connection information parsed from the XML @p node
+     * will be added to this network.
+     *
+     * @param model The @c ModelPtr to update.
+     * @param node The @c XmlNodePtr to load the connection from.
+     */
+    void loadConnection(const ModelPtr &model, const XmlNodePtr &node);
```
jonc125: The documentation and arguments don't seem to match the method name very clearly?
Perhaps the `@brief` should be something like `Update the @p model by adding a connection parsed from @p node`?
Similarly the sentence starting "Loads the argument" is unclear.

src/api/libcellml/parser.h
```diff
@@ -0,0 +1,168 @@
...
+     *
+     * Loads the argument @p model with CellML encapsulation information parsed from
+     * the argument XML @p parentComponentNode. If any encapsulation relationships already
+     * exist in the @p model, the encapsulation information parsed from the XML @p node
+     * will be added to this network.
+     *
+     * @param model The @c ModelPtr to update.
+     * @param parentComponentNode The parent @c XmlNodePtr of the encapsulation relationship.
+     */
+    void loadEncapsulation(const ModelPtr &model, XmlNodePtr &parentComponentnode);
```
jonc125: Again this is unclearly documented.

src/api/libcellml/types.h
```diff
@@ -31,6 +33,13 @@ typedef std::shared_ptr<Units> UnitsPtr; /**< Type definition for shared units p
 class Variable; /**< Forward declaration of Variable class. */
 typedef std::shared_ptr<Variable> VariablePtr; /**< Type definition for shared variable pointer. */
 
+class XmlAttribute; /**< Forward declaration of XmlAttribute class. */
```
jonc125: Can I check that this is an internal header, since I don't think we want to expose these types in the API?

src/componententity.cpp
```diff
@@ -252,6 +319,12 @@ bool ComponentEntity::containsComponent(const std::string &name) const
     return result != mPimpl->mComponents.end();
 }
 
+bool ComponentEntity::containsComponent(const ComponentPtr &component) const
+{
+    auto result = mPimpl->findComponent(component);
+    return result != mPimpl->mComponents.end();
```
jonc125: Is there a reason this isn't a single line?  (Minor niggle!)

src/model.cpp
```diff
@@ -168,13 +169,12 @@ std::string Model::doSerialisation(Format format) const
                                 // Get parent components.
                                 Component* component1 = static_cast<Component*>(variable->getParent());
                                 Component* component2 = static_cast<Component*>(equivalentVariable->getParent());
-                                // Do not serialise a variable's parent component in a connection if that variable no
-                                // longer exists in that component. Allow serialisation of one componentless variable.
-                                if (component1) {
```
jonc125, line 173: Is there a reason this case appears to have been lost?

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+            err->setDescription("Unexpected XML element type (not a model).");
+            err->setModel(model);
+            addError(err);
+        }
+    }
+}
+
+void Parser::loadModel(const ModelPtr &model, const XmlNodePtr &node)
+{
+    XmlAttributePtr attribute = node->getRootAttribute();
```
jonc125: `getRootAttribute` seems a strange name: it doesn't match `getChild`, and 'root' has special meaning in some XML contexts. Would just `getAttribute` be better? Or equally change both methods to be `getFirstAttribute` and `getFirstChild` for clarity?

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+                                "' found in units '"+ units->getName() + "'.");
+            err->setUnits(units);
+            addError(err);
+        }
+        attribute = attribute->getNext();
+    }
+    XmlNodePtr childNode = node->getChild();
+    while (childNode) {
+        if (childNode->isType("unit")) {
+            std::string name = "";
```
jonc125: I'd be inclined to extract this out to a separate method!

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+            double multiplier = 1.0;
+            double offset = 0.0;
+            attribute = childNode->getRootAttribute();
+            while (attribute) {
+                if (attribute->isType("units")) {
+                    name = attribute->getValue();
+                } else if (attribute->isType("prefix")) {
+                    prefix = attribute->getValue();
+                } else if (attribute->isType("exponent")) {
+                    try
```
jonc125: There's a case for extracting this try-catch construct to a method too.

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+    // Load the connection map_components.
+    XmlNodePtr mapComponentsNode = node->getChild();
+    if (!mapComponentsNode) {
+        ModelErrorPtr err = std::make_shared<ModelError>();
+        err->setDescription("Connection does not contain any child elements.");
+        err->setModel(model);
+        addError(err);
+        return;
+    }
+    if (mapComponentsNode->isType("map_components")) {
```
jonc125: The spec doesn't require the map_components element to be first - child elements can appear in any order.

src/xmlattribute.cpp
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return valueString;
+}
+
+XmlAttributePtr XmlAttribute::getNext()
+{
+    xmlAttrPtr next = mPimpl->mXmlAttributePtr->next;
+    XmlAttributePtr nextHandle = std::make_shared<XmlAttribute>();
+    nextHandle->setXmlAttribute(next);
+    if (next == NULL) {
```
jonc125: Wouldn't it make more sense not to construct the handle if next is NULL?

src/xmlnode.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+    return attributeHandle;
+}
+
+XmlNodePtr XmlNode::getChild()
+{
+    xmlNodePtr child = mPimpl->mXmlNodePtr->children;
+    XmlNodePtr childHandle = std::make_shared<XmlNode>();
+    childHandle->setXmlNode(child);
+    if (child == NULL) {
```
jonc125: Similarly here - why not do as for attributes?

jonc125:
I've not had time to go through all the test cases, but the implementation looks sensible!

src/model.cpp
hsorby, line 173: This is not a case, it is not possible to get here if there is not at least one component.

src/CMakeLists.txt
```diff
@@ -37,11 +37,16 @@ set(SOURCE_FILES
...
+  ${CMAKE_CURRENT_SOURCE_DIR}/logger.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/model.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/namedentity.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/parser.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/units.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/variable.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/version.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/xmldoc.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/xmlnode.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/xmlattribute.cpp
```
agarny: Nitpicking: line 49 should be before line 47.

src/CMakeLists.txt
```diff
@@ -63,6 +71,9 @@ set(API_HEADER_FILES
 
 set(HEADER_FILES
   ${LIBCELLML_CONFIG_H}
+  ${CMAKE_CURRENT_SOURCE_DIR}/xmldoc.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/xmlnode.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/xmlattribute.h
```
agarny: Nitpicking: line 76 should be before line 74.

src/api/libcellml/componententity.h
```diff
@@ -88,6 +101,127 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
     void removeAllUnits();
 
     /**
+     * @brief Tests to see if the units is within this component entity.
+     *
+     * Tests to see if the units with the given name 'name' is contained
```
agarny: Nitpicking: "...with the given @p name..." (i.e. same as line 73)

src/api/libcellml/componententity.h
```diff
@@ -88,6 +101,127 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
...
+     */
+    bool hasUnits(const std::string &name) const;
+
+    /**
+     * @brief Get a units at index.
+     *
+     * Returns a const reference to a units at the index @p index.  If the
+     * index is not valid a @c std::out_of_range exception is thrown.
+     *
+     * @param index The index of the Units to return (zero-based).
```
agarny: Nitpicking: "... of the units to return..." (i.e. as for line 138)

src/api/libcellml/componententity.h
```diff
@@ -88,6 +101,127 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
...
+     * @overload
+     *
+     * @param name The name of the units to return.
+     *
+     * @return A reference to the units with the given name.
+     */
+    UnitsPtr getUnits(const std::string &name);
+
+    /**
+     * @brief Take the units at the given index and return it.
```
agarny: Nitpicking: "... at the given @p index..."

src/api/libcellml/componententity.h
```diff
@@ -88,6 +101,127 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
...
+
+    /**
+     * @brief Replace a units with the given name.
+     *
+     * Replaces the units with the given name @p name with @p units.  Name must be a valid name
+     * of a units in the component entity, if it isn't a @c std::out_of_range exception is thrown.
+     *
+     * @overload
+     *
+     * @param name The name of the Units to replace.
```
agarny: Nitpicking: "... the units..."

src/api/libcellml/logger.h
```diff
@@ -0,0 +1,90 @@
...
+*/
+
+#ifndef LIBCELLML_LIBCELLML_LOGGER_H_
+#define LIBCELLML_LIBCELLML_LOGGER_H_
+
+#include <string>
+
+#include <libcellml/libcellml_export.h>
+#include <libcellml/types.h>
+#include <libcellml/loggererrors.h>
```
agarny: Swap lines 23 and 24.

src/api/libcellml/logger.h
agarny, line 65: I personally prefer it the way it currently is.

src/api/libcellml/logger.h
```diff
@@ -0,0 +1,90 @@
...
+     * @brief Get error at the specified index.
+     *
+     * Returns a reference to an error at the index @p index.  If the  index
+     * is not valid a @c std::out_of_range exception is thrown.
+     *
+     * @param index The index of the error to return (zero-based).
+     *
+     * @return A reference to the error at the given index.
+     */
+    EntityErrorPtr getError(size_t index);
```
agarny: Nitpicking (?): shouldn't a getter be a `const` function? (Same elsewhere in the code.) I.e.

```
EntityErrorPtr getError(size_t index) const;
```

Otherwise, shouldn't we also have a version that returns a `const` reference? I.e.

```
const EntityErrorPtr& getError(size_t index) const;
```

src/api/libcellml/logger.h
```diff
@@ -0,0 +1,90 @@
...
+
+private:
+    void swap(Logger &rhs); /**< Swap method required for C++ 11 move semantics. */
+
+    struct LoggerImpl; /**< Forward declaration for pImpl idiom. */
+    LoggerImpl* mPimpl; /**< Private member to implementation pointer */
```
agarny: Nitpicking: ought to be `LoggerImpl *mPimpl` and not `LoggerImpl* mPimpl` (same elsewhere in the code).

src/api/libcellml/logger.h
```diff
@@ -0,0 +1,90 @@
...
+};
+
+}
+
+#endif /* LIBCELLML_LIBCELLML_LOGGER_H_ */
+
+
```
agarny: Nitpicking: delete line 90.

src/api/libcellml/loggererrors.h
```diff
@@ -0,0 +1,268 @@
...
+{
+public:
+    /**
+     * @brief Set the description for this entity error.
+     *
+     * Set the @c std::string @p description for why this entity error was raised.
+     *
+     * @param description The @c std::string error description to set.
+     */
+    void setDescription(const std::string& description) { mDescription = description; }
```
agarny: I know that the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) allows `inline` functions, but I must confess that I am really not in favour of them (for those curious about `inline` functions, see [here](http://yosefk.com/c++fqa/inline.html)).

src/api/libcellml/loggererrors.h
```diff
@@ -0,0 +1,268 @@
...
+protected:
+    std::string mDescription; /**< The string description for why this entity error raised. */
+};
+
+/**
+ * @brief The NamedEntityError class
+ *
+ * NamedEntityError class for all nameable loggable errors.
+ */
+class NamedEntityError: public EntityError
```
agarny: I imagine that you anticipate this class is going to be useful at some point and therefore decided to go ahead and have it empty for now? Whatever the case, it's not useful at the moment, so I would simply not have it. (Same with `ComponentEntityError`.)

src/api/libcellml/loggererrors.h
```diff
@@ -0,0 +1,268 @@
...
+    void setImport(ImportPtr import) { mImport = import; }
+
+    /**
+     * @brief Get the import for this import error.
+     *
+     * Get the @c ImportPtr import for this import error.
+     *
+     * @return A pointer to the import this import error was raised on.
+     */
+    ImportPtr getImport() const { return mImport; }
```
agarny: We might also want to have a version that returns a `const` reference? (Same for other similar classes.)

src/api/libcellml/loggererrors.h
```diff
@@ -0,0 +1,268 @@
...
+{
+public:
+    /**
+     * @brief Set the units for this units error.
+     *
+     * Set the @p units that this units error is relevant to.
+     *
+     * @param units A pointer to the units this units error is relevant to.
+     */
+    void setUnits(UnitsPtr units) { mUnits = units; }
```
agarny: `UnitsPtr` ought to be a const reference.

src/api/libcellml/module/libcellml
```diff
@@ -24,13 +24,15 @@ limitations under the License.
...
  */
+#include <libcellml/parser.h>
 #include <libcellml/version.h>
 #include <libcellml/model.h>
 #include <libcellml/component.h>
 #include <libcellml/import.h>
 #include <libcellml/units.h>
 #include <libcellml/variable.h>
+#include <libcellml/logger.h>
+#include <libcellml/loggererrors.h>
```
agarny: Nitpicking: headers ought to be listed alphabetically.

src/api/libcellml/parser.h
```diff
@@ -0,0 +1,168 @@
+/*
+Copyright 2015 University of Auckland
```
agarny, line 2: Nitpicking (?): as far as I know, there is still no final decision that has been made regarding the copyright statement that we ought to use, but here we use 2015 while for the `Logger` class we use 2016...?

src/api/libcellml/parser.h
```diff
@@ -0,0 +1,168 @@
...
+
+#ifndef LIBCELLML_LIBCELLML_PARSER_H_
+#define LIBCELLML_LIBCELLML_PARSER_H_
+
+#include <string>
+
+#include "libcellml/model.h"
+#include "libcellml/types.h"
+#include "libcellml/units.h"
+#include "libcellml/logger.h"
```
agarny: Nitpicking: line 25 should be before line 22.

src/api/libcellml/parser.h
agarny: For consistency, I would replace `XmlNodePtr &parentComponentnode` with `const XmlNodePtr &node`. In fact, the node should be a const reference and not a 'simple' reference.

src/api/libcellml/parser.h
```diff
@@ -0,0 +1,168 @@
...
+     * @param variable The @c VariablePtr to update.
+     * @param node The @c XmlNodePtr to load the @p units from.
+     */
+    void loadVariable(const VariablePtr &variable, const XmlNodePtr &node);
+
+    void swap(Parser &rhs); /**< Swap method required for C++ 11 move semantics. */
+
+    struct ParserImpl; /**< Forward declaration for pImpl idiom. */
+    ParserImpl* mPimpl; /**< Private member to implementation pointer. */
+
```
agarny: Nitpicking: delete empty line.

src/api/libcellml/types.h
```diff
@@ -31,6 +33,13 @@ typedef std::shared_ptr<Units> UnitsPtr; /**< Type definition for shared units p
 class Variable; /**< Forward declaration of Variable class. */
 typedef std::shared_ptr<Variable> VariablePtr; /**< Type definition for shared variable pointer. */
```
agarny, line 36: Nitpicking: ought to order those class names / `typedef`s alphabetically.

src/component.cpp
```diff
@@ -21,6 +21,7 @@ limitations under the License.
 #include <vector>
 
 #include "libcellml/variable.h"
+#include "libcellml/units.h"
```
agarny, line 23: Nitpicking: swap lines 23 and 24.

src/componententity.cpp
```diff
@@ -192,6 +201,64 @@ void ComponentEntity::removeAllUnits()
...
+const UnitsPtr& ComponentEntity::getUnits(size_t index) const
+{
+    return mPimpl->mUnits.at(index);
+}
+
+UnitsPtr ComponentEntity::getUnits(const std::string &name)
+{
+    auto result = mPimpl->findUnits(name);
+    size_t index = result - mPimpl->mUnits.begin();
+    return mPimpl->mUnits.at(index);
```
agarny, line 224: It's clearly not (that) important, but why don't we simply have the following? (Same elsewhere in the code.)

```
return mPimpl->mUnits.at(mPimpl->findUnits(name) - mPimpl->mUnits.begin());
```

src/logger.cpp
```diff
@@ -0,0 +1,91 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "libcellml/logger.h"
+
+#include <vector>
+
+#include "libcellml/types.h"
+#include <libcellml/component.h>
```
agarny: Nitpicking: swap lines 21 and 22.

There is an inconsistent use of `"` and `<`/`>` for `#include`s on lines 21 and 22, as well as throughout the code. Personally, I like to use `<`/`>` for headers that don't belong to the project on which I am working and `"` for headers that belong to it.

src/model.cpp
```diff
@@ -23,6 +23,7 @@ limitations under the License.
 
 #include <libcellml/import.h>
 #include <libcellml/variable.h>
+#include <libcellml/component.h>
```
agarny: Nitpicking: sort headers alphabetically.

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+limitations under the License.
+*/
+
+#include "libcellml/parser.h"
+
+#include <libcellml/model.h>
+#include <libcellml/import.h>
+#include <libcellml/variable.h>
+#include <libcellml/component.h>
+#include <libcellml/loggererrors.h>
```
agarny: Nitpicking: headers ordering.

src/parser.cpp
agarny: +1 re `getFirstAttribute()` and `getFirstChild()`.

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+            model->addComponent(component);
+        } else if (childNode->isType("units")) {
+            UnitsPtr units = std::make_shared<Units>();
+            loadUnits(units, childNode);
+            model->addUnits(units);
+        } else if (childNode->isType("import")) {
+            ImportPtr import = std::make_shared<Import>();
+            loadImport(import, model, childNode);
+        } else if (childNode->isType("encapsulation")) {
+            XmlNodePtr componentRefNode = childNode->getChild();
```
agarny: What about the case where `childNode` somehow has more than one child?

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+            if (componentRefNode) {
+                loadEncapsulation(model, componentRefNode);
+            } else {
+                ModelErrorPtr err = std::make_shared<ModelError>();
+                err->setDescription("Encapsulation does not contain any component_ref");
```
agarny: Missing full stop for the description. Otherwise, I couldn't find a test for this case and although the coverage test does indeed cover this case, I would expect a test for this case. (Maybe the same with other similar bits of code?)

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+    }
+}
+
+void Parser::loadConnection(const ModelPtr &model, const XmlNodePtr &node)
+{
+    // Initialise connection components and variables.
+    ComponentPtr component1 = nullptr;
+    ComponentPtr component2 = nullptr;
+    VariablePtr variable1 = nullptr;
+    VariablePtr variable2 = nullptr;
```
agarny: Nitpicking (?): why define those variables here when they may not be used at all? I would define them only when we know they are going to be used.

src/parser.cpp
agarny: Indeed, and this ought to be tested.

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+    XmlNodePtr mapComponentsNode = node->getChild();
+    if (!mapComponentsNode) {
+        ModelErrorPtr err = std::make_shared<ModelError>();
+        err->setDescription("Connection does not contain any child elements.");
+        err->setModel(model);
+        addError(err);
+        return;
+    }
+    if (mapComponentsNode->isType("map_components")) {
+        if (mapComponentsNode->hasAttribute("component_1")) {
```
agarny: What if we have an attribute that is neither `component_1` nor `component_2`? In other words, why not do like what has been done above, i.e. have a `while` loop for attributes?

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+    }
+}
+
+void Parser::loadEncapsulation(const ModelPtr &model, XmlNodePtr &parentComponentNode)
+{
+    while (parentComponentNode) {
+        bool errorOccurred = false;
+        ComponentPtr parentComponent = nullptr;
+        if (parentComponentNode->isType("component_ref")) {
+            if (parentComponentNode->hasAttribute("component")) {
```
agarny: What if there was somehow to be more than one attribute? We ought to have a `while` loop. In fact, I would expect to see a `while` loop whenever we check elements/attributes (as in `loadImport()` below).

src/parser.cpp
```diff
@@ -0,0 +1,637 @@
...
+                        errorOccurred = true;
+                    }
+                    if (!errorOccurred) {
+                        // Set parent/child relationship.
+                        parentComponent->addComponent(childComponent);
+                        // Load any further encapsulated children.
+                        if (childComponentNode->getChild()) {
+                            loadEncapsulation(model, childComponentNode);
+                        }
+                        // Load an encapsulated component only once through its parent.
```
agarny: The comment doesn't match what is done below, or am I missing something?

src/xmlattribute.cpp
```diff
@@ -0,0 +1,97 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include <string>
+#include <libxml/parser.h>
+#include <libxml/tree.h>
```
agarny, line 18: Nitpicking: ordering of headers.

src/xmlattribute.cpp
```diff
@@ -0,0 +1,97 @@
...
+        type = std::string(reinterpret_cast<const char *>(mPimpl->mXmlAttributePtr->name));
+    }
+    return type;
+}
+
+std::string XmlAttribute::getValue() const
+{
+    std::string valueString;
+    if ((mPimpl->mXmlAttributePtr->name) && (mPimpl->mXmlAttributePtr->parent)) {
+        char* value;
```
agarny: Nitpicking: `char *value;` (same elsewhere in the code).

src/xmlattribute.cpp
agarny: Same for `XmlNode::getNext()`.

Nitpicking (?): ought to replace `NULL` with `nullptr` (as well as elsewhere in the code).

src/xmlattribute.h
```diff
@@ -0,0 +1,97 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#ifndef XMLATTRIBUTE_H
+#define XMLATTRIBUTE_H
+
+#include <string>
+#include <libxml/parser.h>
```
agarny, line 20: Nitpicking: insert empty line between lines 21 and 22.

src/xmlattribute.h
```diff
@@ -0,0 +1,97 @@
...
+#include "libcellml/types.h"
+
+namespace libcellml {
+
+/**
+ * @brief The XmlAttribute class.
+ *
+ * The XmlAttribute class is a wrapper class for operations on
+ * xmlAttribute objects from libXML2.
```
agarny: Nitpicking: it's "libxml2" not "libXML2" (same elsewhere in the code).

src/xmldoc.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include <cstring>
+#include <vector>
+#include <libxml/parser.h>
+#include <libxml/tree.h>
+#include <libxml/xmlerror.h>
+
+#include "xmldoc.h"
+#include "xmlnode.h"
+
+
```
agarny: Nitpicking: delete empty line.

src/xmldoc.cpp
```diff
@@ -0,0 +1,110 @@
...
+ * from the libxml2 context parser used to parse this document.
+ *
+ * @param userData Private data type used to store the libxml context.
+ *
+ * @param error The @c xmlErrorPtr to the error raised by libxml.
+ */
+void structuredErrorCallback(void *userData, xmlErrorPtr error)
+{
+  std::string errorString = std::string(error->message);
+  errorString.replace(errorString.end()-1, errorString.end(), ".");
```
agarny: Is the plan really replace the last character of `errorString` with `.`? (I have no idea what libxml2 errors look like.)

src/xmldoc.cpp
```diff
@@ -0,0 +1,110 @@
...
+    }
+    delete mPimpl;
+}
+
+void XmlDoc::parse(const std::string& input)
+{
+    xmlParserCtxtPtr context = xmlNewParserCtxt();
+    context->_private = reinterpret_cast<void *> (this);
+    xmlSetStructuredErrorFunc(context, structuredErrorCallback);
+    mPimpl->mXmlDocPtr = xmlCtxtReadDoc(context, BAD_CAST input.c_str(),"/", NULL, 0);
```
agarny: Nitpicking: need a space before `"/"`.

src/xmldoc.h
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * @brief Add an @p error raised while parsing this XML document.
+     *
+     * Adds the @p error raise while parsing this XML document to the
+     * list of @c mXmlErrors. These errors will pertain to the ability
+     * of libxml to parse a given document.
+     *
+     * @param error The XML error string to add.
+     */
+    void addXmlError(const std::string error);
```
agarny: `const std::string &error` and not `const std::string error`.

src/xmlnode.cpp
```diff
@@ -0,0 +1,133 @@
...
+        found = true;
+    }
+    return found;
+}
+
+std::string XmlNode::getAttribute(const char *attributeName)
+{
+    std::string attributeValueString;
+    if (hasAttribute(attributeName)) {
+        char* attributeValue = (char*)xmlGetProp(mPimpl->mXmlNodePtr, BAD_CAST attributeName);
```
agarny: Nitpicking: `char *` in both cases and a space after `(char *)`.

tests/component/encapsulation.cpp
```diff
@@ -124,3 +125,47 @@ TEST(Encapsulation, hierarchyCircular) {
...
+            "<encapsulation>"
+            "<component_ref component=\"parent_component\">"
+            "<component_ref component=\"child1\">"
+            "<component_ref component=\"child2\">"
+            "<component_ref component=\"child3\"/>"
+            "</component_ref>"
+            "</component_ref>"
+            "</component_ref>"
+            "</encapsulation>"
+            "</model>";
```
agarny, line 146: Nitpicking: might want to indent here and there to make it easier to read. In that context, in some places (in the code), we use 4-space indentation while in other cases we use 2-space indentation. It would be nice to be consistent (and probably use 2-space indentation).

tests/parser/parser.cpp
```diff
@@ -0,0 +1,961 @@
...
+    std::string a = model->serialise(libcellml::Format::XML);
+    EXPECT_EQ(e, a);
+}
+
+TEST(Parser, emptyEncapsulation) {
+    std::string ex =
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model xmlns=\"http://www.cellml.org/cellml/1.2#\" name=\"model_name\">"
+            "<encapsulation>"
+            "</encapsulation>"
```
agarny: Nitpicking: I would personally replace `<encapsulation></encapsulation>` with `<encapsulation/>` (same elsewhere in the code).

agarny:
Same as @jonc125, I didn't get a chance to go through all the test cases in detail, but it seems to look fine.

src/api/libcellml/loggererrors.h
dladd: I was mainly using the empty error classes like `NamedEntityError` as a placeholder to keep inheritance of our error classes consistent with the main object model. For instance, if we take it out, a `VariableError` will need to be an `EntityError`. I'm happy to remove them if that is still preferred but I find the current state less confusing.

src/api/libcellml/parser.h
dladd: Good point- this is currently set up so that `parseModel()` just creates a new model, calls `updateModel()` and returns the model. But we should probably remove `updateModel()` or at least make it private until we want to deal with updates after validation.

src/api/libcellml/loggererrors.h
dladd: Good point- it makes sense for inline-functions be consistent with our inline-if policy.

src/api/libcellml/parser.h
dladd: In the case of encapsulation, the XML `node` won't necessarily be a `const` since we are getting the XML `node` iterator corresponding with an encapsulated `<component_ref ...>` rather than the `<encapsulation>` XML node. So we're calling a `node = node->getNext()` inside the routine. This allows us to easily deal with multiple levels of encapsulation using recursion.

src/api/libcellml/types.h
dladd: Good catch- it looks like I do need to bury those class declarations further to hide them from the public API.

src/componententity.cpp
dladd, line 224: I think I remember @hsorby commenting on why this is done this way quite a while ago but have been unable to dig up the comment thread.

src/logger.cpp
dladd: That sounds like a reasonable distinction to me. I'll try to clean up instances that violate this.

src/xmldoc.cpp
dladd: All libxml2 errors that I have seen have no ending punctuation but always a carriage return. So this just swaps the "\n" for a period to be consistent with our error formatting.

src/api/libcellml/loggererrors.h
agarny: Personally, I don't see any problem with `VariableError` being an `EntityError`. In fact, if it was to be a `NamedEntityError`, I would then expect it to have a name property, which it clearly doesn't.

src/api/libcellml/parser.h
agarny: Yes, I did see that `parentComponentnode = parentComponentnode->getNext()`, but couldn't we still have:

```
void loadEncapsulation(const ModelPtr &model, const XmlNodePtr &node);
```

for consistency (with the rest of the `loadXXX()` methods) and then, within the implementation of `loadEncapsulation()`, have something like:

```
XmlNodePtr parentComponentnode = node;
...
parentComponentnode = parentComponentnode->getNext();
```

It would be nice if that was to be possible. (I like things to be consistent and `loadEncapsulation()`'s signature just isn't consistent with that of the other `loadXXX()` methods.)

src/componententity.cpp
agarny, line 224: Yes, I seem to recall that discussion before, but can't recall the details of it either. If we ever do, it might then be worth keeping track of it somewhere for 'reference'.

src/xmldoc.cpp
agarny: Ok, that's fair enough then, but... I would still check that the last character is indeed `\n`, in case libxml2 was ever to change the formatting of their errors (better be safe than sorry...).

src/parser.cpp
jonc125: That depends on how much this is trying to validate, and how much just do the minimum checking required to load a model. I think we're aiming for the latter; a separate validator would check we don't have extra attributes.

agarny: If so, then the `loadImport()` method (for example) shouldn't generate an error message if an attribute is neither `href` nor `xlink`. I guess it all comes down to (once again) trying to be consistent throughout the code.

src/api/libcellml/types.h
dladd: The main issue with hiding the XML classes and typedefs from the public API is that we still need to handle them from our public-facing `Parser`. I think the best way to get around this is to move the `load...()` methods that handle the XML objects into the parser private implementation. This will mean moving the `load...()` function declarations from the parser header into the `.cpp` file. We'll also need to give the pimpl a pointer to to the parent `Parser` to be able to add errors for the user to recieve. For example (best seen on Github):

``` c++
...
/**
 * @brief The Parser::ParserImpl struct.
 *
 * The private implementation for the Parser class.
 */
struct Parser::ParserImpl
{
public:
    Format mFormat;
    Parser *mParser;

    /**
     * @brief Update the @p model with attributes parsed from a @c std::string.
     *
     * Update the @p model with attributes and entities parsed from
     * the @c std::string @p input. Any entities or attributes in @p model with names
     * matching those in @p input will be overwritten.
     *
     * @param model The @c ModelPtr to update.
     * @param input The string to parse and update the @p model with.
     */
    void loadModel(const ModelPtr &model, const std::string &input);

private:
    /**
     * @brief Update the @p component with attributes parsed from @p node.
     *
     * Update the @p component with attributes and entities parsed from
     * the XML @p node. Any entities or attributes in @p component with names
     * matching those in @p node will be overwritten.
     *
     * @param component The @c ComponentPtr to update.
     * @param node The @c XmlNodePtr to parse and update the @p component with.
     */
    void loadComponent(const ComponentPtr &component, const XmlNodePtr &node);
...
};

Parser::Parser(Format format)
    : mPimpl(new ParserImpl())
{
    mPimpl->mFormat = format;
    mPimpl->mParser = this;
}
...
```

The only drawback I could see is that it might arguably look less consistent to have these functions declared/documented outside of the header but as far as I can tell, this [shouldn't matter to doxygen](https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html#cppblock).

src/parser.cpp
dladd: This should be covered by the [emptyEncapsulation](https://github.com/cellml/libcellml/pull/114/files#diff-94fa719a93f7d6001ad11518352180f4R333) test. ~~But I also got some weird behaviour coming out of the modelWithInvalidElement test that I'll investigate further.~~ EDIT: This seems to just have been an issue with my debugger.

agarny: Ok, I can see that being the case indeed (the fact that it is covered by the [emptyEncapsulation](https://github.com/cellml/libcellml/pull/114/files#diff-94fa719a93f7d6001ad11518352180f4R333) test. However, I hadn't noticed that those tests check for the length of the error message rather than the actual message itself. I must confess that I don't like it. As far as I am concerned, the test should check the actual message to be 100% certain that we have got the expected error message.

src/parser.cpp
dladd: I think we do probably need to iterate over all present XML nodes and attributes with the parser and log any errors that may come up when building up the objects from input strings. My understanding was that the `Validator` could then be applied to the constructed objects to check whether they meet the Spec. 

So the `Validator` wouldn't actually be touching an input string and the `Parser` is the only place to catch something like a spelling error for an input attribute type.

src/parser.cpp
dladd: This is attempting to explain why we would do a `removeComponent()` here.  Maybe something like:

``` c++
// A child component is added through its parent component rather than the model, so remove if it exists.
```

is more clear?

agarny: "..., so remove it if it exist." ?

tests/component/encapsulation.cpp
dladd, line 146: I think standardising indentation of XML in our tests might be better left as a general code-cleanup item after this set of changes. I've now added it to the list in #116.

src/componententity.cpp
dladd, line 224: I had a chat with @hsorby earlier and this and the instance @jonc125 points out below were done to reduce the line length and break the operation up into logical steps to make them easier to read and debug. But if the single line versions are preferred, perhaps we can add this to the clean-up list?

jonc125, line 224: I think sometimes breaking up lines is helpful, but in this case (and possibly similar ones) the single-line version isn't that long.

It is however slightly counter-intuitive to me, since findUnits returns an iterator, that we don't just check it's valid and dereference, instead of computing an offset then using at!

dladd, line 224: Fair enough- I've now added "Tidy-up less efficient multiline methods where appropriate." to the list in #116.

dladd:
This should now address all comments to date and passes all tests. 

For a comparison since the last review, [see here](https://github.com/dladd/libcellml/compare/load@%7B2weeks%7D...load) (though you would probably be better off using your own diff viewer :) ). Major changes since that review:
- Looping over all XML nodes/attributes to flag errors when parsing.
  - Overhauling the `loadConnection()` and `loadEncapsulation()` routines to do this, as well as flagging all errors rather than just the first error.
- Moving the `load..()` routines into the parser private implementation to hide our XML wrappers from users.
- Using a unified error class rather than derived types for each entity. This avoids requiring `dynamic_cast`ing and associated cross-platform issues.

Big thanks to @jonc125 and @agarny for a very thorough and productive review and to @hsorby for his help with the implementation.

nickerso:
In order to really test this out, I think we need to be checking the parsing of actual models. I have started work to add this in dladd/libcellml#5. It currently doesn't seem to successfully parse the model as loaded from a file - still a work in progress :)

agarny:
Besides @nickerso's comment just above, it all looks good to me (at this stage).

src/api/libcellml/module/libcellml
```diff
@@ -24,13 +24,15 @@ limitations under the License.
...
-#include <libcellml/model.h>
 #include <libcellml/component.h>
+#include <libcellml/error.h>
 #include <libcellml/import.h>
+#include <libcellml/logger.h>
+#include <libcellml/model.h>
+#include <libcellml/parser.h>
 #include <libcellml/units.h>
 #include <libcellml/variable.h>
+#include <libcellml/version.h>
```
agarny: Nitpicking: I believe we agreed on using `"` rather than `<`/`>` for in-project headers? (Same in `tests/coverage/coverage.cpp` and `tests/parser/parser.cpp`, although they are tests, so they might actually be fine...?)

dladd: Good catch on the module includes. I would think we would want to be including `<libcellml>` as if it were an external dependency in the tests but am happy to swap to quotes if that is preferred.

nickerso: I think the tests should be treated as "external"

agarny: Yes, I am definitely happy to consider those includes to be treated as 'external'.

dladd:
To add to my recent mistakes, I managed to mix up the commit messages for https://github.com/cellml/libcellml/pull/114/commits/699112af7797d4f0d1dfd4cc84ad56ff6b79f5d8  and https://github.com/cellml/libcellml/pull/114/commits/169eba96affba1efa97862f629038fa8d440d15a . The former message was supposed to correspond with the change to `src/model.cpp` rather than `tests/model/model.cpp`.

Unfortunately, since these have now been pushed, I would need to close this request, rebase the branch, and submit a new pull request to fix the messages. I could also try to squash this pull down into logical chunks where appropriate (we're now at 125 commits in this pull). 

But this will take a bit of time to do carefully so it's a matter whether we'd rather just move forward. Seeing as the plan is to move the serialisation code into a separate `Printer` eventually anyway, I'd be for just pressing on. My apologies for the inconvenience.

nickerso:
+1 to pressing on! :)

agarny:
FWIW, I have just seen that 4e0c66b includes the new `ComponentPtr getEncapsulatedComponent(const std::string &name)` and `bool containsEncapsulatedComponent(const std::string &name)` methods.

However, as mentioned [here](https://github.com/cellml/libcellml/issues/111#issuecomment-232022389), it looks like we agreed that we should be adding an extra argument to `ComponentPtr getComponent(const std::string &name)` (and, by extension, to `const ComponentPtr& getComponent(const std::string &name) const`) and `bool containsComponent(const std::string &name) const` (note that it's a `const` function; and, by extension, to `bool containsComponent(const ComponentPtr &component) const`) rather than having these two new methods (not to mention that they are not on par with existing ones).

In conclusion, commit 4e0c66b needs some reworking, but maybe that's part of that 'pressing on' process? If so, then fine with me.

dladd:
Thanks @agarny - that commit was my proof-of-concept for fetching encapsulated components. This pull request should now be updated with component methods that include a `bool searchEncapsulated` flag per your helpful suggestion ([example](https://github.com/cellml/libcellml/pull/114/files#diff-2e9f65009003feaf7545ce0b9763f922R336), [header](https://github.com/cellml/libcellml/pull/114/files#diff-d2ee493885a137515e56ea7e8cf5fa1aR291)). The previous unencapsulated/local component find methods have also now been made private and renamed like `getComponentInThis()` ([example](https://github.com/cellml/libcellml/pull/114/files#diff-2e9f65009003feaf7545ce0b9763f922R468)).

src/componententity.cpp
```diff
@@ -246,34 +333,83 @@ size_t ComponentEntity::componentCount() const
...
+            foundComponent = getComponent(i)->getComponent(name);
+            if (foundComponent) {
+                break;
+            }
+        }
+    }
+    return foundComponent;
+}
+
+const ComponentPtr& ComponentEntity::getComponent(const std::string &name, bool searchEncapsulated) const
```
dladd: At the moment, I'm not quite sure how to accomplish getting a `const` component reference for an encapsulated component using recursion. The attempt commented out below results in a "returning reference to temporary" error. I'll try to sort this out again tomorrow but any suggestions would be welcome and much appreciated. :-)

src/componententity.cpp
```diff
@@ -246,34 +333,83 @@ size_t ComponentEntity::componentCount() const
...
-ComponentPtr ComponentEntity::getComponent(size_t index)
+bool ComponentEntity::containsComponent(const ComponentPtr &component, bool searchEncapsulated) const
 {
-    return mPimpl->mComponents.at(index);
+    bool result = false;
+    if (containsComponentInThis(component)) {
+        result = true;
+    } else if (searchEncapsulated) {
+        for (size_t i = 0; i < componentCount(); ++i) {
+            result = getComponent(i)->containsComponent(component);
```
jonc125: Don't you need to pass searchEncapsulated to the recursive call (here and elsewhere)?

src/componententity.cpp
jonc125: The issue may be because you're returning a pointer that is const, rather than a pointer to a const component? So when you call getComponent(i) the component that gives isn't const. You may also need to pass ,true to containsComponent & getComponent(name).

src/componententity.cpp
agarny: > Don't you need to pass searchEncapsulated to the recursive call (here and elsewhere)?

`searchEncapsulated` is `true` and by default `containsComponent()` will search encapsulated components, so it's 'fine', although I agree that it would be better (and safer) to pass `searchEncapsulated` (e.g. in case we were ever to have `searchEncapsulated` defaulting to `false`).

src/componententity.cpp
dladd: > The issue may be because you're returning a pointer that is const, rather than a pointer to a const component?

Ah of course- thanks a lot @jonc125 . This probably also applies to other `const` methods in the library, which currently have that same signature. Perhaps another item for cleanup issue #116?

jonc125: Probably best left to that or a separate issue, yes. Const-qualified methods returning pointers are annoying! Especially as if it's a reference to a member pointer you need to make the pointer const too...

src/componententity.cpp
```diff
@@ -262,18 +375,36 @@ const ComponentPtr& ComponentEntity::getComponent(size_t index) const
...
+            return getComponentInThis(name);
+        } else if (searchEncapsulated) {
+            for (size_t i = 0; i < componentCount(); ++i) {
+                if (getComponent(i)->containsComponent(name, searchEncapsulated)) {
+                    return getComponent(i)->getComponent(name, searchEncapsulated);
+                }
+            }
+        }
+    }
+    throw std::out_of_range("Named component not found.");
```
dladd, line 407: Just to update, I'm having trouble getting the Windows build to throw this exception at the moment. I should be able to interactively debug this on Monday through RDP but will work on the validation side in the meantime.

dladd:
OK I think this now passes all tests and addresses all comments that haven't been delegated to another issue. Thanks to @hsorby for getting our CI back online.

nickerso:
OK, as far as I can tell, we have: a good range of tests for model parsing (including some actual CellML 2.0 models); no outstanding comments that need addressing; and the excellent green tick with all tests passing, 100% coverage, and no memory leaks. So I'll merge this in and hope for the best. Great contribution @dladd!

hsorby:
Whoop! Actual CellML 2.0 models great work @dladd.

dladd:
Thanks but credit also goes to @nickerso and @jonc125 for putting cellml1to2 together and giving us something to test! As well as pretty much everyone for their work, comments, and guidance on this pull.

Issue #118, opened 2016-06-23
Current thinking on validation


dladd:
In a meeting at the ABI on Tuesday, there was some discussion about how we might implement validation as the next step in the use-cases doc.

The team here was in favour of **(A)** setting up a `Validator` object, similar to how we deal with the `Parser` in the changeset under review in #114 . Consistent with this, it was also suggested that we should probably move our serialisation routines into a separate `Serialiser` object (we might want to call this `Printer` or something else instead), rather than doing serialisation from the objects themselves. 

For example, a user wishing to check if a model is valid, output it if so or output errors if not might do:

``` c++
    libcellml::Validator validator;
    libcellml::Serialiser serialiser(libcellml::Format::XML);
    validator.validate(model);
    if (validator.hasErrors()) {
        for (size_t i = 0; i < validator.errorCount(); ++i) {
            std::cout << validator.getError(i)->getDescription();
        }
    } else {
        std::cout << serialiser.serialise(model);
    }
```

An alternative **(B)** might be to do validation on entities as we currently do for serialisation, for example:

``` c++
    model->validate();
    if (model->hasErrors()) {
        for (size_t i = 0; i < model->errorCount(); ++i) {
            std::cout << model->getError(i)->getDescription();
        }
    } else { 
        std::cout << model->serialise(libcellml::Format::XML);
    }
```

While **B** is a bit more concise, the thinking is that it is worth the extra line or two to make our reading, writing, and validation error logging consistent. And  `Parser` may need to pick up errors that do not necessarily translate into an Entity (e.g. unrecognised or mismatched XML tags).

<br>
With the validator, there is also the question of when/if to automatically clear errors. For instance, consider if we call the following validation on two very simple models:

``` c++
    libcellml::ModelPtr m1 = std::make_shared<libcellml::Model>();
    libcellml::ModelPtr m2 = std::make_shared<libcellml::Model>();
    libcellml::Validator validator;
    validator.validate(m1);
    validator.validate(m2);
    m1->setName("model1");
    validator.validate(m1);
```

`validator` might now contain the following errors regarding the requirement for valid models to have names:
- **(1)** no errors if we clear validation errors every time `validate()` is called.  
- **(2)** errors from `m1` and `m2` if we allow validation errors to persist until the user calls `validator.clearErrors()`
- **(3)** an error from `m2` only if we remove any previous errors pertaining to the object being validated from the `validator` and add back any new errors from the updated object.  

I don't have a strong preference but would probably go with **3**.

[...]

PR #121, opened 2016-07-07
Update object models for validate step

dladd:
This updates our overview and use-cases object models to cover validation as under discussion in #118.

I've also separated out the use-cases diagram into one covering entities and another for input/output/error handling.


nickerso:
This looks good to me - I like the separation in the diagrams.

agarny:
It all looks fine to me too, so I am going to merge. This being said, I am not sure about having that `libCellMLRoughObjectModel-Overview.png` file. Sure, it's nice to have, but I get the feeling that it could easily become out of sync.

Issue #123, opened 2016-07-11
Introducing XML Document Identifiers ('id' attributes)


dladd:
The cellml1to2 tests that have now been added in #114 have pointed out that we have not yet implemented XML Document Identifiers. This is item 2.5 in the [Specification](https://docs.google.com/document/d/1MleZJF6DYp4FO4vxslAHInrRFMqrSpARa1s_dPhwYU0/edit?usp=sharing):

> Any element information item in the CellML namespace MAY contain an attribute information item with local name id, in the CellML namespace. This attribute information item SHALL be treated as having attribute type ID, as defined in section 3.3.1 of XML 1.1.

We can support this by adding a `mId` string attribute on the Entity class, with associated `get/set()` methods. This will also require changes to our serialisation and parsing routines. Since I've merged the cellml1to2 tests into the pull for #114, the easiest way to move forward is probably to tack implementation and tests for IDs on to #114 . However, since that pull is already somewhat large and unwieldy, I can try reverting the cellml1to2 merge and setting this up as a separate pull if that is preferred.

nickerso:
Might be worth adding implementation and tests for IDs in creation and serialisation as a separate pull request which will presumably be fairly quick to review and then that can be pulled into #114 in terms of loading/parsing.

dladd:
Thanks @nickerso that sounds reasonable. 

However, I'm beginning to realise this might be less trivial than I expected. My read of the above Spec snippet is that we need to be able to track IDs for **every** element information item (EII). This includes EIIs that are outside of our current Entity hierarchy like connections, encapsulations, and component_refs. Is this the correct interpretation of the Spec?

jonc125:
Hmm, good point. Yes, that is the correct interpretation - any CellML element may have an id. They are mainly used as a reference point for metadata annotations, and in practice it'll be variables & models that carry them most often I expect, though I can see that one might want to annotate connections or encapsulations with justifications. My inclination at this point would be to defer that problem for another day - create a separate issue for annotation of CellML elements that don't have a specific representation in the libCellML object model.

PR #125, opened 2016-07-12
Add annotation 'id's for libCellML entities

dladd:
Following the discussion in #123, this adds basic support for `get/setId()` on libCellML entities and serialisation of these entities.

A few serialisation cases have been intentionally left out of this pull- for example setting the `id` on units within components. These rely on features like `ComponentEntity::getUnits()` introduced in #114 and will be picked up there.

### DIFF
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -131,12 +131,21 @@ std::string Units::doSerialisation(Format format) const
     if (format == Format::XML) {
         if (getName().length()) {
             if (isImport()) {
-                repr += "<import xlink:href=\"" + getImport()->getSource() + "\" xmlns:xlink=\"http://www.w3.org/1999/xlink\">"
-                        "<units units_ref=\"" + getImportReference() + "\" name=\"" + getName() + "\"/>"
-                        "</import>";
+                repr += "<import xlink:href=\"" + getImport()->getSource() + "\" xmlns:xlink=\"http://www.w3.org/1999/xlink\"";
+                if (getImport()->getId().length()) {
+                    repr += " id=\"" + getImport()->getId() + "\"";
+                }
+                repr += "><units units_ref=\"" + getImportReference() + "\" name=\"" + getName() + "\"";
+                if (getId().length()) {
+                    repr += " id=\"" + getId() + "\"";
+                }
+                repr += "/></import>";
             } else {
                 bool endTag = false;
                 repr += "<units name=\"" + getName() + "\"";
+                if (getId().length()) {
+                    repr += " id=\"" + getId() + "\"";
+                }
                 if (isBaseUnit()) {
                     repr += " base_unit=\"yes\"";
                 } else if (mPimpl->mUnits.size() > 0) {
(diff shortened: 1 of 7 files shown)


nickerso:
This seems like a good first step for supporting `id`s while allowing #114 to get finished off.

The code all looks good to me.

src/units.cpp
```diff
@@ -131,12 +131,21 @@ std::string Units::doSerialisation(Format format) const
...
+                }
+                repr += "><units units_ref=\"" + getImportReference() + "\" name=\"" + getName() + "\"";
+                if (getId().length()) {
+                    repr += " id=\"" + getId() + "\"";
+                }
+                repr += "/></import>";
             } else {
                 bool endTag = false;
                 repr += "<units name=\"" + getName() + "\"";
+                if (getId().length()) {
```
jonc125, line 146: It's entirely unimportant, but I was wondering whether there was a reason for the order in which attributes were output? :)

tests/model/model.cpp
```diff
@@ -370,3 +377,48 @@ TEST(Model, constructors) {
 
 }
 
+TEST(Model, setAndCheckIdsAllEntities) {
+    const std::string expected =
+            "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+            "<model xmlns=\"http://www.cellml.org/cellml/1.2#\" name=\"mname\" id=\"mid\">"
+              "<import xlink:href=\"some-other-model.xml\" xmlns:xlink=\"http://www.w3.org/1999/xlink\" id=\"iid\">"
+                "<units units_ref=\"a_units_in_that_model\" name=\"u1name\" id=\"u1id\"/>"
+              "</import>"
```
jonc125, line 386: Should we added a check with imported component?

jonc125:
Apart from a test with an imported component, looks good to me too.

src/units.cpp
agarny, line 146: Kind of the same here. For some reasons, I would expect the id to be output first. (Same elsewhere in the code.)

tests/model/model.cpp
agarny, line 386: Definitely.

agarny:
All looks good to me besides @jonc125's comment about testing imported components.

src/units.cpp
nickerso, line 146: I agree that the order is not important. And is likely to be something that will change once we move to the serialiser proposal to match the parser...

dladd:
Ah yes apologies- I must have committed the imported component `id` implementation and testing into my local 'load' branch instead of this one. I will cherry-pick it over tomorrow morning if this is still open :-)

dladd:
To avoid cherry-picking and resolving merge conflicts multiple times, could we pull this in as is and bring in the imported component `id` with the `load` pull? This feature comes in at https://github.com/cellml/libcellml/pull/114/commits/169eba96affba1efa97862f629038fa8d440d15a

nickerso:
seems reasonable to me. I'll close this one in favour of #114.

Issue #122, opened 2016-07-08
Implement 'Validate' use-cases step (4)


dladd:
This issue covers implementation of the Validate step as outlined in the [use-cases doc](http://libcellml.readthedocs.io/en/latest/usecases.html) and based on the requirements from the planning discussion in #118. This currently includes:
- [ ] Create, load, and modify models and then validate them (the test is that the models are correctly identified as valid or invalid, and for the case when they are invalid the correct reason is given, covering each rule in [the specification](https://docs.google.com/document/d/1MleZJF6DYp4FO4vxslAHInrRFMqrSpARa1s_dPhwYU0/edit?usp=sharing)).
- [x] MathML validation to use libxml2 + the [W3C MathML DTD](https://www.w3.org/Math/DTD/Overview.html)

dladd:
The current plan is to use the MathML DTD to validate our math strings. However, Spec 17.8.4-5 (quoted below) requires the introduction of `cellml:units` attributes in `ci` and `cn` elements, which are not part of the standard MathML DTD. I see two possible ways around this:
- **(A)** We add the `cellml:units` attribute to ci/cn elements in the MathML DTD we distribute as part of libCellML.
- **(B)** Since we are only passing the error messages from the DTD-based MathML validation to our own `Validator` error logger, we could filter out the errors associated with these valid (to us) attributes by their message string.

I suppose this is mainly a matter of whether we mind distributing our own modified MathML DTD. I think I am personally in favour of (B) as I'd prefer to keep the MathML DTD standard and modular.

> 17.8. Interpretation of mathematics
> ...
> - 4. The MathML elements requiring additional type information are:
>   - 1. Every MathML cn element
>   - 2. Every ci element that defines a new variable inside an bvar element.
> - 5. Every MathML element requiring additional type information (as defined above) MUST have an attribute information item in the CellML namespace, with local name units. The value of this attribute information item MUST be a valid units reference. The referenced units SHALL NOT affect the mathematical interpretation of the CellML model. However, CellML processing software MAY use this information to assist the user in the detection and correction of units errors in the CellML model.

agarny:
[...]

Issue #116, opened 2016-06-08
Code clean-up items


dladd:
Proposed items for clean-up after the next set of pulls:
- [x] Convert inline "if"s to multiline (issue #117 )
- [x] Namechange from CellML 1.2 to 2.0
- [x] Update copyright statement
- [x] Standardise visual formatting of XML strings in tests with indentation
- [x] Tidy-up less efficient multiline methods where appropriate.

Please comment if there is anything else you would like to see addressed.

agarny:
Regarding the copyright statement, do we absolutely have to provide the year? Also, does it mean that all libCellML files must have their copyright updated every year? If so, it would then mean that we lose the year at which the project started? In the end, couldn't we simply have:

```
Copyright libCellML Contributors
```

dladd:
That works for me- I'm happy to take out the year altogether as it likely [isn't necessary](http://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/). Git is a better method to actually track when a file was modified than a bit of stub header text that we are likely to forget to update each year.

I do think we are required to give attribution to University of Auckland. And maybe VPR or University of Michigan for further support?

jonc125:
The Chaste notice uses "Copyright (c) 2005-2016, University of Oxford" and we update the end date each year. I would be inclined to include an initial year, but the end year could I think be omitted.

Attributing various funders/organisations should probably be done in separate documentation rather than the copyright notice, or it gets too long. If we're going to be formal about it we probably need a contributor licence agreement etc...

agarny:
FWIW, whatever solution we go with, it shouldn't mean updating all files every year.

dladd:
Regarding the copyright issue, it sounds like there is agreement to keep 2015 in all template text as the start year of the project (rather than going with the year a particular file was added). That just leaves whether or  not to change the template text copyright holder from "University of Auckland" to "libCellML Contributors" (attribution of all funders/organisations will be left for the full license/documentation). I.e., options:
- **(A)**/remain: `Copyright 2015 University of Auckland`
- **(B)**/change: `Copyright 2015 libCellML Contributors`

~~I'll add a +1 for **A**.~~

agarny:
Did I agree to keep 2015 in the copyright? Anyway, for me to have the starting year without the current/ending year gives me the impression that the project started and ended in 2015, hence my preference would be for:
- **(C)**: `Copyright University of Auckland`

dladd:
[...]

PR #127, opened 2016-07-19
Add tests and implementation for 'Validate' (use-case 4)

dladd:
This pull covers the introduction of the `Validator` enabler in libCellML.

Development is still ongoing. Upon completion, this changeset will address the implementation outlined in #122 and concepts discussed in #118.


src/validator.cpp
```diff
@@ -0,0 +1,504 @@
...
+                    }
+                    // Check if we already have another import from the same source with the same units_ref.
+                    // (This looks for matching enties at the same position in the source and ref vectors).
+                    if ((unitsImportSources.size() > 0) && (!foundImportError)) {
+                        if ((std::find(unitsImportSources.begin(), unitsImportSources.end(), importSource) - unitsImportSources.begin())
+                         == (std::find(unitsRefs.begin(), unitsRefs.end(), unitsRef) - unitsRefs.begin())){
+                            ErrorPtr err = std::make_shared<Error>();
+                            err->setDescription("Model '" + model->getName() +
+                                                "' contains multiple imported units from '" + importSource +
+                                                "' with the same units_ref attribute '" + unitsRef + "'.");
```
nickerso, line 353: just trying to work through this logic, but I'm pretty sure that it is valid to have multiple imports of the same units, as long as they have different units names.

dladd, line 353: > I'm pretty sure that it is valid to have multiple imports of the same units, as long as they have different units names.

From Spec 6.1.2:

> The value of the units_ref attribute MUST NOT match the value of the units_ref attribute on any sibling import units element.

Suggests to me that we shouldn't have duplicate `units_ref`'s from the same import. But maybe I am misinterpreting this?

nickerso, line 353: hmmmm....ok, that has slipped in since 1.1 :) so don't worry about my comment for now...

src/api/libcellml/validator.h
```diff
@@ -49,10 +49,14 @@ class LIBCELLML_EXPORT Validator: public Logger
...
 
     void validateComponent(const ComponentPtr &component);
-    void validateUnits(const UnitsPtr &units, const std::vector<std::string> unitsNames);
+    void validateUnits(const UnitsPtr &units, const std::vector<std::string> &unitsNames);
+    void validateUnitsUnit(size_t index, const UnitsPtr &units, const std::vector<std::string> &unitsNames);
     void validateVariable(const VariablePtr &variable, std::vector<std::string> &variableNames);
     bool catchDoubleConversionError(const std::string &input);
 
+    bool isStandardUnitName(const std::string &name);
+    bool isStandardPrefixName(const std::string &name);
```
hsorby: All of these private methods above should be implemented in the private implementation so that they are not visible here in the API.

hsorby:
I think we should consider merging this now which would allow others to start working on the cleanup items listed in #116.

nickerso:
I agree this code should now be merged. I don't believe there are any significant outstanding issues that need to be addressed prior to accepting this code.

PR #137, opened 2016-08-23
Hide private

hsorby:
Hiding private methods in the public API header files.  Also adding const to some methods.  Part of tidy up of codebase issue #116.


nickerso:
Looks great to me - cleans up the public API. (apart from the conflicts that have arisen from other merges :)

hsorby:
Is this one good to go now?

PR #138, opened 2016-08-23
Resolve warning from Visual Studio about dll interface for std::string.

hsorby:
Part of tidy up of codebase issue #116. Resolves this warning by switching to the pImpl idiom for the entity and named entity base classes.


nickerso:
These all seems good to me. The only thing missing is perhaps a test that those moved objects are indeed null as the comments suggest?

nickerso:
The updated comment in 7bb9a62 clarifies things nicely, and I'm happy with these changes now.

Issue #139, opened 2016-08-25
Stop returning const references to smart pointers


hsorby:
Whilst investigating some other issues I came across this:

http://stackoverflow.com/questions/10643563/how-to-return-smart-pointers-shared-ptr-by-reference-or-by-value

I think we should follow this advice and stop returning const references to smart pointers in our codebase.

nickerso:
+1

agarny:
+1

dladd:
+1 

Returning smart pointers by value rather than const reference should also address #126.

hsorby:
Resolved.

Issue #140, opened 2016-08-26
Exceptions (again)


hsorby:
In light of issue #139 the main reason for throwing exceptions is now gone.  Do we still want to throw exceptions in libCellML?  Is it Exceptable to throw an exception when a user wants to remove a component in a model by an index that is not valid for instance?

`if (model.componentCount() == 3) {
    model.removeComponent(5);
}`

nickerso:
I have no problem with exceptions. I'm also happy if all those `void` functions returned a success/failure instead of throwing an exception.

Either way, as long as the documentation matches the implementation, I'd be happy :)

hsorby:
I guess then we are not necessarily talking about exceptions but about how the user is notified if things didn't go to plan.  I can think of three scenarios:
1. Exception is thrown.
2. A status is returned (true, false, nullptr).
3. Nothing happens, methods silently fail.

We should probably decide on one of these approaches and be consistent across the codebase.

nickerso:
I'm ok with 1 or 2, but wouldn't like 3 at all. Users can treat 2 as 3 by ignoring the return code if they want to use the API in that way.

hsorby:
For consistency reasons I would prefer 2.  Currently we are returning nullptr if a component is not found instead of throwing an exception.  I think we should extend this so that all methods return a status in some form.

agarny:
Definitely option 2 for me.

jonc125:
I actually prefer to throw exceptions if the code user has done something 'wrong', and trying to remove/retrieve a non-existent component would fall in this category for me. Null return values are often not checked (original programmer "I know this can't be null here", 2 months later...). Null returns feels much more like a C API as well, exceptions seem to be the norm in the STL.

We can have separate methods to test for existence.

mtcooling:
I've been asked to comment since things seem split.

3 is an abomination.

I lean much more to 2 than 1. A better summary of why than I can write off the top of my head can be found here in Joel Spolsky's blog:

http://www.joelonsoftware.com/items/2003/10/13.html

As a 'user', I wouldn't mind as long as it was consistent.

PR #141, opened 2016-08-26
Remove methods that return const references to smart pointers

hsorby:
Part of the codebase tidy up issue #116. Also addresses #139 as it overlaps.


src/utilities.h
```diff
@@ -0,0 +1,53 @@
...
+ * cannot be converted using @c std::stod, catch the exception and return
+ * @c false. If @p input can be converted, return @c true.
+ *
+ * @param candidate The @c std::string value to try to convert to a @c double.
+ *
+ * @return @c false if @p input cannot be converted to a @c double using
+ * @c std::stod and @c true otherwise.
+ *
+ */
+bool EXPORT_FOR_TESTING canConvertToDouble(const std::string &candidate);
```
dladd: Do we need to EXPORT_FOR_TESTING `canConvertToDouble()` and `hasNonWhitespaceCharacters()`? Can we not cover all cases for these by parsing strings with the `Parser`?

src/api/libcellml/component.h
```diff
@@ -127,60 +127,30 @@ class LIBCELLML_EXPORT Component: public ComponentEntity
...
-     * Returns a const reference to a variable with the name @p name for this
-     * component.  If the name is not valid a std::out_of_range exception
-     * is thrown.
-     *
-     * @overload
-     *
-     * @param name The name of the variable to return.
-     *
-     * @return A const reference to the variable with the given name.
+     * @return A reference to the variable at the given index on success, @c nullptr otherwise.
```
agarny, line 137: Question: do we have any kind of rule/agreement on the length of a line (especially for comments)?

src/api/libcellml/component.h
```diff
@@ -127,60 +127,30 @@ class LIBCELLML_EXPORT Component: public ComponentEntity
...
-    const VariablePtr& getVariable(const std::string &name) const;
+    VariablePtr getVariable(size_t index) const;
 
     /**
      * @brief Get a variable with the given name @p name.
      *
      * Returns a reference to a variable with the name @p name for this
-     * component.  If the name is not valid a std::out_of_range exception
-     * is thrown.
+     * component.  If the name is not found the @c nullptr is returned.
```
agarny: Minor: "... a @c nullptr..."

src/component.cpp
```diff
@@ -151,24 +151,21 @@ void Component::removeAllVariables()
...
 
-const VariablePtr& Component::getVariable(const std::string &name) const
+VariablePtr Component::getVariable(const std::string &name) const
 {
-    return mPimpl->mVariables.at(mPimpl->findVariable(name) - mPimpl->mVariables.begin());
+    auto result = mPimpl->findVariable(name);
+    if (result == mPimpl->mVariables.end()) {
+        return nullptr;
+    }
+    return *result;
```
agarny, line 168: Minor: for consistency with some other bits of the code (at least), it might be better to have (same for `UnitsPtr ComponentEntity::getUnits(const std::string &name) const`):

```
auto result = mPimpl->findVariable(name);
if (result != mPimpl->mVariables.end()) {
    return *result;
}
return nullptr;
```

i.e. finish with `return nullptr;`.

src/componententity.cpp
```diff
@@ -265,53 +267,41 @@ void ComponentEntity::doAddComponent(const ComponentPtr &c)
     mPimpl->mComponents.push_back(c);
 }
 
+void ComponentEntity::removeComponent(size_t index)
+{
+    if (index < mPimpl->mComponents.size()) {
+        mPimpl->mComponents.erase(mPimpl->mComponents.begin() + index);
+    } else {
+        throw std::out_of_range("Index out of range.");
```
agarny, line 383: I guess this will be eventually addressed as part of issue #140?

src/parser.cpp
```diff
@@ -498,11 +501,11 @@ void Parser::ParserImpl::loadUnit(const UnitsPtr &units, const XmlNodePtr &node)
...
                 err->setRule(SpecificationRule::UNIT_MULTIPLIER);
                 mParser->addError(err);
-            } else {
-                multiplier = std::stod(attribute->getValue());
             }
         } else if (attribute->isType("offset")) {
-            if (mParser->catchDoubleConversionError(attribute->getValue())) {
+            if (canConvertToDouble(attribute->getValue())) {
+                offset = std::stod(attribute->getValue());
+            } else {
```
agarny: Might want to have something like this instead (see below for the change to `canConvertToDouble()`):

```
if (!canConvertToDouble(attribute->getValue(), offset)) {
```

src/utilities.cpp
```diff
@@ -0,0 +1,45 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#include "utilities.h"
+
+#include <stdexcept>
```
agarny, line 19: Minor: swap lines 17 and 19 (to be consistent with what has been done here and there).

src/utilities.cpp
```diff
@@ -0,0 +1,45 @@
...
+        std::stod(candidate);
+        canConvert = true;
+    } catch (std::invalid_argument) {
+        canConvert = false;
+    } catch (std::out_of_range) {
+        canConvert = false;
+    }
+
+    return canConvert;
+}
```
agarny, line 39: Might want to replace with:

```
bool canConvertToDouble(const std::string &candidate, double &offset)
{
   bool canConvert = false;
    // Try to convert the input string to double.
    try
    {
        offset = std::stod(candidate);
        canConvert = true;
    } catch (...) {
        canConvert = false;
    }

    return canConvert;
}
```

src/utilities.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+bool EXPORT_FOR_TESTING hasNonWhitespaceCharacters(const std::string &input)
```
agarny: Remove `EXPORT_FOR_TESTING`.

src/utilities.h
agarny: Agreed, it would be better if we didn't have to export things for the sole purpose of testings.

tests/coverage/coverage.cpp
```diff
@@ -18,6 +18,9 @@ limitations under the License.
 
 #include <libcellml>
 #include <iostream>
+
+#include <../utilities.h>
```
agarny: Replace with (we agreed to use `"` rather than `<`/`>` for local includes):

```
#include "../utilities.h"
```

src/api/libcellml/component.h
dladd: The [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Line_Length)  recommends 80 characters but their main argument for that that was based on it being the standard for pre-existing code. But I don't believe we've made a rule on this.

agarny: I think it would be nice to follow that recommendation, at least for comments. (I appreciate that for code itself, it can sometimes be a bit difficult to stick to 80 characters.)

src/utilities.cpp
hsorby: This is the way it should be for all implementation files, the implementation header file first and then system, local header files in alphabetical order.

src/parser.cpp
hsorby: If I'm right you're saying that we should convert to double and return the success failure by return value.

src/componententity.cpp
hsorby: Yes, I wanted to keep it as a separate pull request to make the review easier.

src/utilities.cpp
hsorby: This is required if we want to test directly.  If we don't want it we should remove the whole mechanism that allows this.

src/utilities.cpp
hsorby: I would rather think we want to be explicit about what exceptions we are catching, for clarity.

src/utilities.cpp
agarny: I actually agree with you (this is indeed what I do in OpenCOR). It's just that I have mainly seen the other way round so far.

src/component.cpp
hsorby: I will leave this for when #140 is addressed because all this code is related and will make it consistent throughout all objects then.  Just wanting to avoid dealing with merge conflicts.

src/api/libcellml/component.h
hsorby: I feel that 80 characters is a bit archaic, we can probably set this guideline to be a bit wider.

src/utilities.cpp
hsorby: This should be added to the list of items to tidy up in #116 if it is the case.

tests/coverage/coverage.cpp
hsorby: My bad, will fix it.  Although if we ditch the EXPORT_FOR_TESTING this is not useful anymore.

src/parser.cpp
agarny: I am saying that `canConvertToDouble()` tells us whether we can convert to a double by actually trying to do the conversion. So, it's kind of a waste of time to do another conversion when `canConvertToDouble()` returns `true`. Instead, we might as well pass `offset` so that `canConvertToDouble()` can save the result of the conversion it tries to do, and we then can use that information rather than re-computing it.

src/utilities.cpp
agarny: Agreed in principle, it's just that here both types of exception result in the same outcome, i.e. `canConvert = false;`, so I am not sure what the point of being explicit is here.

src/utilities.h
hsorby: We can, all this does is allow for direct testing.

src/utilities.cpp
agarny: Sorry, I meant that `EXPORT_FOR_TESTING` shouldn't be necessary in a `.cpp` file, no?

src/parser.cpp
hsorby: Yes it is a waste but we are not super performance orientated, I'll make the change though.

src/api/libcellml/component.h
agarny: I am certainly happy with something a bit wider.

src/utilities.cpp
hsorby: Doesn't matter where it is if you want to call the function directly from a test you just need it somewhere.

agarny: Ok, it's just that I have always put it in the corresponding header file (as you have done here too), which makes more sense to me. This aside, why not do the same for `canConvertToDouble ()` or am I missing something?

hsorby: I was just being sloppy, I'll get rid of that and just run the tests through the API I don't like the EXPORT_FOR_TESTING mechanism.

src/parser.cpp
nickerso: and presumably change the method name to `convertToDouble`...

hsorby:
I think this one is now done, if it could be reviewed (and merged if appropriate) a lot of this work effects removing exceptions so it is very helpful if this is merged so that work can start on that.

agarny:
Looks good to me.

src/utilities.cpp
dladd: > Agreed in principle, it's just that here both types of exception result in the same outcome, i.e. canConvert = false;, so I am not sure what the point of being explicit is here.

Being consistently explicit can force us to be aware of and check for less-likely error modes that might be overlooked with a `catch (...)`. It might also be useful to someone interactively debugging to see what kind of error was caught.

EDIT: sorry I just saw that this has now been merged as of a couple of minutes ago

PR #142, opened 2016-08-28
Tidy up header file includes to the correct order.

hsorby:
Addressing part of issue #116 (although not in the checkbox list).

Example ordering for header includes:

```
// First header file is the corresponding header file for the current implementation
// file in this case validator.cpp
#include "libcellml/validator.h"

// Second section of header files is a list of all system header files required in alphabetical
// order, compound header files coming at the end.
#include <algorithm>
#include <map>
#include <sstream>
#include <string>
#include <vector>
#include <libxml/parser.h>
#include <libxml/tree.h>

// Third section of header files is a list of all local project header files required in alphabetical 
// order.
#include "libcellml/component.h"
#include "libcellml/import.h"
#include "libcellml/error.h"
#include "libcellml/model.h"
#include "libcellml/units.h"
#include "xmldoc.h"
```

### DIFF
--- a/src/api/libcellml/componententity.h
+++ b/src/api/libcellml/componententity.h
@@ -16,8 +16,8 @@ limitations under the License.
 
 #pragma once
 
-#include "libcellml/importedentity.h"
 #include "libcellml/exportdefinitions.h"
+#include "libcellml/importedentity.h"
 #include "libcellml/types.h"
 
 namespace libcellml {
--- a/src/api/libcellml/units.h
+++ b/src/api/libcellml/units.h
@@ -19,8 +19,8 @@ limitations under the License.
 #include <string>
 #include <vector>
 
-#include "libcellml/importedentity.h"
 #include "libcellml/exportdefinitions.h"
+#include "libcellml/importedentity.h"
 #include "libcellml/types.h"
 
 namespace libcellml { 
--- a/src/error.cpp
+++ b/src/error.cpp
@@ -13,6 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+
 #include "libcellml/error.h"
 
 #include <map>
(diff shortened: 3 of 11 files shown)


agarny:
I realise (from https://github.com/cellml/libcellml/pull/141#discussion-diff-76368003) that I slightly misunderstood the order in which you thought header files should be listed. Taken your example above, I would personally have:

```
// First: local header files
#include "libcellml/component.h"
#include "libcellml/import.h"
#include "libcellml/error.h"
#include "libcellml/model.h"
#include "libcellml/units.h"
#include "libcellml/validator.h"
#include "xmldoc.h"

// Second: system header files
#include <libxml/parser.h>
#include <libxml/tree.h>
#include <algorithm>
#include <map>
#include <sstream>
#include <string>
#include <vector>
```

This being said, I am fine either way, as long as we are consistent. One thing, however, I don't understand why `libxml/parser.h` and libxml/tree.h`are listed last in your second list when`libcellml/component.h` & Co. are listed first in your third list. To me, there is a slight of consistency here.

hsorby:
I was seeing which way people would prefer.  I was wondering if compound includes should be considered as directories or not.  We could just have it pure alphabetical which I think would be ok.

Either way the header file that corresponds to the implementation file should always go first, this ensures that the header file has everything it needs declared in it.  After that I think required system header files should be declared to avoid masking and if you intend to mask a system function or such then it is obvious.  That then leaves the local includes to come last.

hsorby:
Our own [style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) has this to say:

In dir/foo.cc or dir/foo_test.cc, whose main purpose is to implement or test the stuff in dir2/foo2.h, order your includes as follows:
1. dir2/foo2.h.
2. C system files.
3. C++ system files.
4. Other libraries' .h files.
5. Your project's .h files.

hsorby:
In my opinion the includes all now adhere to the style guide, unless I have made a mistake in which case the review will pick it up :).

agarny:
Looks good to me.

PR #143, opened 2016-09-12
No exceptions

hsorby:
Remove use of exceptions from the codebase.  Addresses issue #140.


src/api/libcellml/componententity.h
```diff
@@ -146,35 +143,33 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
...
-     * exception is thrown.
+     * Takes the first occurence of the units with the given name @p name and returns it.
+     * If no units with name @p name is found then a @c nullptr is returned.
      *
      * @overload
      *
      * @param name The name of the units to take.
      *
-     * @return The Units identified with the given @p name.
+     * @return A reference to the units identified with the given @p name if found, @c nullptr otherwise
```
agarny: Minor: missing trailing full stop.

src/api/libcellml/componententity.h
```diff
@@ -146,35 +143,33 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
...
      */
     UnitsPtr takeUnits(const std::string &name);
 
     /**
      * @brief Replace a units at the given @p index.
      *
-     * Replaces the units at the given @p index with @p units.  If an invalid @p index
-     * is passed to the method a @c std::out_of_range exception is thrown.
+     * Replaces the units at the given @p index with @p units.
+     * @p index must be in the range [0, #units).
```
agarny, line 179: Minor: why not mention that information at line 174? (Same elsewhere in the code.)

src/api/libcellml/componententity.h
```diff
@@ -329,12 +321,11 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
      * @brief Take the component at the given @p index and return it.
      *
      * Removes the component at the given @p index position and returns it.
-     * If an invalid @p index is passed to the method a @c std::out_of_range
-     * exception is thrown.
+     * @p index must be in the range [0, #components).
      *
-     * @param index The index of the Component to take (zero-based).
+     * @param index The index of the Component to take.
```
agarny, line 343: Minor: if "(zero-based)" is to be removed, then it ought to be removed elsewhere in the code too.

src/api/libcellml/units.h
```diff
@@ -271,22 +271,66 @@ class LIBCELLML_EXPORT Units: public ImportedEntity
     void getUnit(size_t index, std::string& reference, std::string &prefix, double &exponent, double &multiplier, double &offset) const;
 
     /**
+     * @brief Get the @c unit attributes for the given reference @p reference.
```
agarny: Minor: "... for the given @p reference." (Same below.)

src/api/libcellml/units.h
```diff
@@ -271,22 +271,66 @@ class LIBCELLML_EXPORT Units: public ImportedEntity
...
+     *
+     * @overload
+     *
+     * @param reference The @c std::string reference for the @c unit attributes to get.
+     * @param prefix The prefix for this @c unit. Defaults to empty string.
+     * @param exponent The exponent for this @c unit. Defaults to 1.0.
+     * @param multiplier The multiplier for this @c unit. Defaults to 1.0.
+     * @param offset The offset for this @c unit. Defaults to 0.0.
+     */
+    void getUnit(const std::string &reference, std::string &prefix, double &exponent, double &multiplier, double &offset) const;
```
agarny: I must confess that I got confused by the description of the function and its name. Why not call it `getUnitAttributes()` rather than `getUnit()`? (Same below.)

Otherwise, I know that we haven't agreed on a maximum length for a line, but this one is clearly too for my liking, not least when it comes to reviewing using GitHub. It looks like GitHub considers that a line should be 120 characters long maximum. Maybe we could just do the same? In other words, the above line would actually be written:

```
    void getUnit(const std::string &reference, std::string &prefix, double &exponent, double &multiplier,
                 double &offset) const;
```

src/api/libcellml/units.h
```diff
@@ -271,22 +271,66 @@ class LIBCELLML_EXPORT Units: public ImportedEntity
...
+
+    /**
+     * @brief Get the @c unit attributes for the given reference @p standardRef.
+     *
+     * Get the attributes for the first @c unit that matches the reference @p standardRef. If
```
agarny, line 295: Unless I am mistaken, there is not going to be more than one unit for a given `standardRef`? If so, then we ought to have "... for the @c unit that matches..."

src/api/libcellml/units.h
```diff
@@ -271,22 +271,66 @@ class LIBCELLML_EXPORT Units: public ImportedEntity
...
+     * @param standardRef The @c StandardUnit enum unit reference attributes to get.
+     * @param prefix The prefix for this @c unit. Defaults to empty string.
+     * @param exponent The exponent for this @c unit. Defaults to 1.0.
+     * @param multiplier The multiplier for this @c unit. Defaults to 1.0.
+     * @param offset The offset for this @c unit. Defaults to 0.0.
+     */
+    void getUnit(StandardUnit standardRef, std::string &prefix, double &exponent, double &multiplier, double &offset) const;
+
+    /**
+     * @brief Remove the unit at the given index @p index.
```
agarny: Minor: "... at the given @index."?

src/component.cpp
```diff
@@ -153,19 +149,21 @@ void Component::removeAllVariables()
 
 VariablePtr Component::getVariable(size_t index) const
 {
+    VariablePtr variable = nullptr;
     if (index < mPimpl->mVariables.size()) {
-        return mPimpl->mVariables.at(index);
+        variable = mPimpl->mVariables.at(index);
     }
-    return nullptr;
+    return variable;
```
agarny, line 165: Why overcomplicate things by declaring and using `variable`? (Same below and elsewhere in the code.)

src/componententity.cpp
```diff
@@ -279,8 +276,6 @@ void ComponentEntity::removeUnits(size_t index)
 {
     if (index < mPimpl->mUnits.size()) {
```
agarny, line 173: What if `index` is lower than zero? Also, I thought (from #140) that we had agreed to return a status (`true`, `false` or `nullptr`)? I would therefore expect to have something like:

```
bool ComponentEntity::removeUnits(size_t index)
{
     if ((index >= 0) && (index < mPimpl->mUnits.size())) {
        mPimpl->mUnits.erase(mPimpl->mUnits.begin() + index);

        return true;
    } else {
        return false;
    }
}
```

The same holds for all the functions where we don't throw an exception anymore.

src/componententity.cpp
```diff
@@ -351,8 +343,10 @@ UnitsPtr ComponentEntity::takeUnits(const std::string &name)
 
 void ComponentEntity::replaceUnits(size_t index, const UnitsPtr &units)
 {
-    removeUnits(index);
-    mPimpl->mUnits.insert(mPimpl->mUnits.begin() + index, units);
+    if (index < mPimpl->mUnits.size()) {
+        removeUnits(index);
+        mPimpl->mUnits.insert(mPimpl->mUnits.begin() + index, units);
+    }
```
agarny, line 259: I would instead expect this function now to look something like (if we are to respect what we agreed in #140):

```
bool ComponentEntity::replaceUnits(size_t index, const UnitsPtr &units)
{
    if (removeUnits(index)) {
        mPimpl->mUnits.insert(mPimpl->mUnits.begin() + index, units);

        return true;
    } else {
        return false;
    }
}
```

The same holds for all the functions where we don't (indirectly) throw an exception anymore.

src/componententity.cpp
```diff
@@ -379,8 +373,6 @@ void ComponentEntity::ComponentEntityImpl::removeComponent(size_t index)
 {
     if (index < mComponents.size()) {
```
agarny: Make sure that `index` is `>= 0`, and return `true`/`false`, as needed. (Same elsewhere in the code.)

src/logger.cpp
```diff
@@ -85,7 +85,11 @@ size_t Logger::errorCount() const
 
 ErrorPtr Logger::getError(size_t index) const
 {
-    return mPimpl->mErrors.at(index);
+    ErrorPtr err = nullptr;
+    if (index < mPimpl->mErrors.size()) {
+        err = mPimpl->mErrors.at(index);
+    }
+    return err;
```
agarny, line 92: What if `index` is lower than zero? Why overcomplicate things? Why not simply have:

```
ErrorPtr Logger::getError(size_t index) const
{
    if ((index >= 0) && (index < mPimpl->mErrors.size())) {
        return mPimpl->mErrors.at(index);
    } else {
        return nullptr;
    }
}
```

agarny:
I made some comments and thought I would wait before reviewing further... :)

src/componententity.cpp
hsorby: As index is an unsigned int comparing `index >= 0` is a no-op and produces a warning from at least one compiler, that is the reason we are not comparing for positive integers.

src/logger.cpp
hsorby: This is supposed to be the non-complicated version, only a single return point at the end of the function.

src/api/libcellml/componententity.h
hsorby: I have removed zero-based where I have put in the range, missed some.

src/componententity.cpp
agarny: Oops, I clearly overlooked the type of `index`. My bad!

src/logger.cpp
agarny: Hmm, I guess it's a matter of taste. Actually, now that you pointed out to me that `index` is an `unsigned int`, I would probably have:

```
ErrorPtr Logger::getError(size_t index) const
{
    return (index < mPimpl->mErrors.size()) ? mPimpl->mErrors.at(index) : nullptr;
}
```

Anyway, I am not sure it's worth fighting over.

hsorby: I like that form but it does hide branches from the coverage testing.  As I (hope I) have written all functions to follow this pattern I think we should leave it as is.

src/api/libcellml/units.h
hsorby: No, I think you can quite happily add as many unit as desired with the same `standardRef`.  Maybe I'll add a test to highlight this.

src/logger.cpp
agarny: Thanks for the reminder about branches.

src/api/libcellml/units.h
agarny: Considering the [definition of `StandardUnit`](https://github.com/cellml/libcellml/blob/fba015fee8327abf0a34e86c408ab448cd4c60ec/src/api/libcellml/units.h#L50), I would indeed be interested by such a test.

agarny: Ok, I can see that you have added a [test](https://github.com/cellml/libcellml/pull/143/commits/3175d33a20cdb8d76bbe9d4b5c0a98efbb0167fa) for it. I stand corrected.

src/componententity.cpp
```diff
@@ -349,15 +250,21 @@ UnitsPtr ComponentEntity::takeUnits(const std::string &name)
...
-    mPimpl->mUnits.insert(mPimpl->mUnits.begin() + index, units);
+    bool status = false;
+    if (index < mPimpl->mUnits.size()) {
+        removeUnits(index);
+        mPimpl->mUnits.insert(mPimpl->mUnits.begin() + index, units);
+        status = true;
+    }
+
+    return status;
 }
```
agarny, line 262: Why not simply have the following since `removeUnits()` now returns a boolean?

```
bool ComponentEntity::replaceUnits(size_t index, const UnitsPtr &units)
{
    bool status = false;

    if (removeUnits(index)) {
        mPimpl->mUnits.insert(mPimpl->mUnits.begin() + index, units);

        status = true;
    }

    return status;
}
```

src/componententity.cpp
```diff
@@ -375,257 +282,164 @@ void ComponentEntity::doAddComponent(const ComponentPtr &c)
...
-ComponentPtr ComponentEntity::ComponentEntityImpl::takeComponentInThis(const std::string &name)
-{
-    return takeComponent(findComponent(name) - mComponents.begin());
-}
 
-void ComponentEntity::ComponentEntityImpl::replaceComponentInThis(const std::string &name, const ComponentPtr &component)
-{
-    replaceComponent(findComponent(name) - mComponents.begin(), component);
+    return status;
 }
```
agarny, line 429: Same here, take advantage of the fact that `removeComponent()` now returns a boolean.

src/units.cpp
```diff
@@ -358,20 +366,33 @@ void Units::getUnit(size_t index, std::string &reference, std::string &prefix, d
     }
 }
```
agarny, line 367: Minor: the order in the implementation of the different `getUnitAttributes()` methods is not the same as that of their declaration.

src/variable.cpp
```diff
@@ -188,8 +193,6 @@ void Variable::VariableImpl::unsetEquivalentTo(const VariablePtr &equivalentVari
     auto result = findEquivalentVariable(equivalentVariable);
     if (result != mEquivalentVariables.end()) {
         mEquivalentVariables.erase(result);
-    } else {
-        throw std::out_of_range("Equivalent variable not found.");
```
agarny, line 192: The fact that we used to throw an exception tells me that we should now return a boolean status. For good measure, we should probably do the same for `setEquivalentTo()`. Then, we ought to modify accordingly anything that in/directly uses `un/setEquivalentTo()`.

tests/component/component.cpp
```diff
@@ -183,22 +186,22 @@ TEST(Component, removeComponentMethods) {
     EXPECT_EQ(1, c.componentCount());
     std::string a = c.serialise(libcellml::Format::XML);
     EXPECT_EQ(e1, a);
-    EXPECT_THROW(c.removeComponent(1), std::out_of_range);
+    c.removeComponent(1);
```
agarny: Shouldn't we have:

```
EXPECT_FALSE(c.removeComponent(1));
```

tests/connection/connection.cpp
```diff
@@ -519,7 +519,7 @@ TEST(Connection, removeEquivalentVariableMethods) {
     libcellml::Variable::removeEquivalence(v2, v3);
     a = m.serialise(libcellml::Format::XML);
     EXPECT_EQ(e2, a);
-    EXPECT_THROW(libcellml::Variable::removeEquivalence(v3, v4), std::out_of_range);
+    libcellml::Variable::removeEquivalence(v3, v4);
```
agarny: If anything, we now ought to have:

```
EXPECT_FALSE(libcellml::Variable::removeEquivalence(v3, v4));
```

meaning that `removeEquivalence()` ought to return a boolean status, etc.

tests/model/model.cpp
```diff
@@ -222,19 +222,19 @@ TEST(Model, removeComponent) {
     EXPECT_EQ(1, m.componentCount());
     std::string a = m.serialise(libcellml::Format::XML);
     EXPECT_EQ(e1, a);
-    EXPECT_THROW(m.removeComponent(1), std::out_of_range);
+    m.removeComponent(1);
```
agarny: Ought to have `EXPECT_FALSE(m.removeComponent(1));`.

tests/units/units.cpp
```diff
@@ -392,6 +396,108 @@ TEST(Units, farhenheit) {
     EXPECT_EQ("fahrenheit", u->getName());
 }
 
+TEST(Units, getUnit) {
```
agarny: Shouldn't that test be called `getUnitAttributes`?

agarny:
I think we are getting there. :)

src/units.cpp
hsorby: I'm not doing anything about that, there is a lot of order mismatch throughout the code.  I don't think it is anything to spend time on.

agarny:
Thanks @hsorby, I am now happy (enough :)) with it.

hsorby:
Couldn't have you 100% happy then you wouldn't have anything to complain about!

Issue #120, opened 2016-07-06
Separate serialisation from entities into a Logger class


dladd:
Eventually we will want to separate our serialisation routines into a separate `Printer` class inheriting from `Logger` to flag errors, as done for the  `Parser` and planned for the `Validator`. This isn't an immediate priority since serialisation as currently implemented is meeting our needs.

_(edited for clarity per Hugh's comment below)_

hsorby:
Into a separate `Printer` class inheriting from `Logger` as is done for the `Parser` class.

PR #144, opened 2016-11-24
Printer

hsorby:
Start implementing a Printer class for consistency with Parser.

Addresses issue #120 .


nickerso:
This all looks good to me. The `serialise` tests that were removed are replaced with equivalent `printer` tests and the other tests are still giving the same answers when entities are serialised using the new `printer`.

src/api/libcellml/printer.h
```diff
@@ -0,0 +1,169 @@
...
+
+#pragma once
+
+#include <string>
+
+#include "libcellml/enumerations.h"
+#include "libcellml/logger.h"
+#include "libcellml/types.h"
+
+//! Everything in libCellML is in this namespace.
```
agarny, line 25: Kind of a given, so delete (might want to do the same in `model.h`, `parser.h`, `validator.h` and `version.h`).

src/api/libcellml/printer.h
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * @brief Serialise the model to @c std::string.
+     *
+     * Serialise the given @p model to a std::string.
+     *
+     * @param model The model to serialise.
+     *
+     * @return The @c std::string representation of the model.
+     */
+    std::string printModel(ModelPtr model) const;
```
agarny, line 51: `const ModelPtr &model` rather than `ModelPtr model`. Same with `UnitsPtr`, `VariablePtr`, etc.

src/api/libcellml/printer.h
```diff
@@ -0,0 +1,169 @@
...
+     *
+     * Serialise the given @p model to a std::string.
+     *
+     * @overload
+     *
+     * @param model The model to serialise.
+     *
+     * @return The @c std::string representation of the model.
+     */
+    std::string printModel(Model model) const;
```
agarny, line 64: `const Model &model` rather than `Model model`. Same with `Units`, `Variable`, etc.

src/printer.cpp
```diff
@@ -0,0 +1,486 @@
...
+Printer::Printer(Printer &&rhs)
+    : Logger(std::move(rhs))
+    , mPimpl(rhs.mPimpl)
+{
+    rhs.mPimpl = nullptr;
+}
+
+Printer& Printer::operator=(Printer p)
+{
+    Logger::operator =(p);
```
agarny, line 72: Remove the space between `operator` and `=`.

tests/connection/connection.cpp
```diff
@@ -94,7 +94,9 @@ TEST(Connection, componentlessVariableInvalidConnection) {
     comp1->addVariable(v1);
     m.addComponent(comp1);
     libcellml::Variable::addEquivalence(v1, v2);
-    std::string a = m.serialise(libcellml::Format::XML);
+
+    libcellml::Printer printer(libcellml::Format::XML);
+    const std::string a = printer.printModel(m);
```
agarny, line 99: Simplify things by removing `const` (same elsewhere in the code or then be consistent and do something similar for line 168 and elsewhere in the code).

src/api/libcellml/printer.h
agarny: Otherwise, rather than having `printModel()` to print a model, `printUnits()` to print a units, etc. I would just overload a method called `print()`.

agarny:
I have made various (very) minor comments, but nothing that should block this PR, so merging.

src/api/libcellml/printer.h
hsorby: I did think about that but thought being consistent with parser was more important.  The Parser only takes strings so we can't overload on that.  Thoughts.

tests/connection/connection.cpp
hsorby: I just added the const to make the code faster.  I have only added const where the string is not changed further thus some are const and some are not.

src/api/libcellml/printer.h
hsorby: see issue #147 for further discussion.

Issue #145, opened 2016-11-26
Add CMake project export


hsorby:
When other CMake based projects come to make use of libCellML it is easiest for such projects to be able to import the project settings through the CMake export mechanism.  libCellML needs to export the project settings to enable this.

PR #146, opened 2016-11-27
Add configuration files and implement export configuration.

hsorby:
Addresses issue #145


hsorby:
The implementation requires the user of libCellML to know the directory containing the libcellml-config.cmake file.  The current setting requires a user to set the libCellML_DIR CMake variable to:

 `-DlibCellML_DIR=${PROJECT_BINARY_DIR}/libcellml`

for working with the build tree (obviously that is the PROJECT_BINARY_DIR of libcellml and not the users project), or 

 `-DlibCellML_DIR=${LIBCELLML_INSTALL_PREFIX}/lib/cmake/libcellml`

for working with the install tree.

An alternative is to not use the libcellml directory to collect the libcellml export files in and thus remove this from the libCellML_DIR variable.  This changes the values slightly as CMake will automatically append 'cmake' to any searched directories.

nickerso:
These changes all look good to me. 

I would personally favour removing the final `libcellml` so that users who choose to install all their dependencies into a single location can simply point their `CMAKE_PREFIX_PATH` at a single location to resolve all dependencies (that use this CMake export functionality). But as this can easily be changed later if we decide to, it is not a show stopper for me.

src/CMakeLists.txt
```diff
@@ -172,8 +177,38 @@ install(FILES
...
+configure_file("${PROJECT_SOURCE_DIR}/cmake/libcellml-config.cmake"
+  ${LIBCELLML_CONFIG_CMAKE_FILE}
+  COPYONLY
+)
+list(APPEND CLEAN_FILES ${LIBCELLML_CONFIG_CMAKE_FILE})
+
+set(ConfigPackageLocation lib/cmake/libcellml)
+install(EXPORT libcellml-targets
+    FILE libcellml-targets.cmake
+    DESTINATION ${ConfigPackageLocation})
```
agarny: - 4-space indentation vs. 2-space above (see line 195 for example)?
- Have the closing bracket at the beginning of a new line (see line 197 for example).

src/CMakeLists.txt
```diff
@@ -172,8 +177,38 @@ install(FILES
...
+
+install(
+    FILES
+      ${LIBCELLML_CONFIG_CMAKE_FILE}
+      ${LIBCELLML_CONFIG_VERSION_FILE}
+    DESTINATION ${ConfigPackageLocation}
+    COMPONENT Devel
+    )
```
agarny: See my previous comment at line 203.

agarny:
Looks good to me too, although it currently breaks (probably as a result of merging PR #144), so not merging it just yet.

Otherwise, I would (like @nickerso, if I understood his comment right) put `libcellml-config-version.cmake `, `libcellml-config.cmake` and `libcellml-targets.cmake` in `${PROJECT_BINARY_DIR}` rather than `${PROJECT_BINARY_DIR}/libcellml`.

src/CMakeLists.txt
```diff
@@ -133,7 +136,11 @@ set_target_properties(cellml PROPERTIES
   CXX_STANDARD_REQUIRED ON
   CXX_VISIBILITY_PRESET hidden
   VISIBILITY_INLINES_HIDDEN 1
-  DEBUG_POSTFIX d)
+  DEBUG_POSTFIX d
+  VERSION ${libCellML_VERSION}
+  INTERFACE_libCellML_MAJOR_VERSION ${libCellML_VERSION_MAJOR}
+  COMPATIBLE_INTERFACE_STRING ${libCellML_VERSION_MAJOR})
+)
```
nickerso, line 143: doubling up on the brackets, which looks to be what is breaking the builds...

nickerso:
Now tests are passing and @hsorby has addressed the preference for installing the CMake config files, so merging in.

Issue #153, opened 2017-08-23
Output XML as version 1.1


hsorby:
Currently the output XML is version 1.0 it has been decided to allow for format version 1.1. Part of the work required for issue #150.

hsorby:
It turns out that libxml2 does not support version 1.1 of xml (which is what libcellml currently uses).  It would seem that some people consider version 1.1 as pointless.  This post seems to be a good summary of the changes between version 1.0 and version 1.1:

https://mail.gnome.org/archives/xml/2010-July/msg00017.html

Is it possible to get this change to the specification reversed?

nickerso:
And this follow up from a W3C XML guy: https://mail.gnome.org/archives/xml/2010-July/msg00020.html. Seems reasonable to stick with 1.0 from this point of view, but need to look back and see why we changed...

MichaelClerx:
Just one more :-)
http://www.ibiblio.org/xml/books/effectivexml/chapters/03.html

nickerso:
On implementation it is becoming pretty clear that we should stick with XML 1.0. We can reopen this if needed in future.

PR #157, opened 2017-08-23
Update XML version from 1.0 to 1.1.

hsorby:
Addresses issue #153.


hsorby:
At this point in time we are going to stick with xml version 1.0 so this PR will be closed.

Issue #162, opened 2017-08-25
Does componentCount() include imported components?


nickerso:
The documentation seems to indicate that the `componentCount()` method for the `ComponenentEntity` class will not include imported components. I'm not sure why not and the implementation seems to not match that behaviour.

I'll add some tests to see if we agree on what should be happening.

hsorby:
I think the documentation is incorrect here we should be counting all components in the model.

nickerso:
That is what the current implementation is doing. #163 adds some tests to demonstrate this.

PR #163, opened 2017-08-25
Adding some tests on component counts with imports.

nickerso:
Need to confirm the behaviour of the componentCount() method and make
sure we know what to expect. Initial tests on imported components to
set the current status of the implementation, which disagrees with the
documentation.

Addresses #162


hsorby:
If you correct the documentation to match the behaviour this should be good to merge.

hsorby:
Yup, looks good to me now.

Issue #166, opened 2017-08-27
Tests for resolving imports


nickerso:
We need tests for resolving imports to make sure our current plans work as expected.

MichaelClerx:
Do we have these now @nickerso ?

hsorby:
This can now be closed with the merge of #167.

PR #167, opened 2017-08-28
Tests for resolving imports

nickerso:
Addresses #166.


tests/resolve_imports/tests.cmake
```diff
@@ -0,0 +1,25 @@
...
+# Using absolute path relative to this file
+set(${CURRENT_TEST}_SRCS
+  ${CMAKE_CURRENT_LIST_DIR}/file_parser.cpp
+)
+#set(${CURRENT_TEST}_HDRS
+#  ${CMAKE_CURRENT_LIST_DIR}/<test_header_files.h>
+#)
+
+
+set(CELLML_SINE_MODEL_RESOURCE "${CMAKE_CURRENT_SOURCE_DIR}/resources/sine_approximations.xml")
```
hsorby: We shouldn't be doing this every time, we should have a common resources script that we include from the tests/CMakeLists.txt file.

tests/resolve_imports/file_parser.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <fstream>
+#include <iostream>
+#include <libcellml>
+#include <sstream>
+#include <vector>
+
+// generated with test resource locations
+#include "test_resources.h"
+
+static int countImportedComponents(libcellml::ModelPtr model)
```
hsorby: Is this something that you are thinking of supplying as a convenience function on the class at some point?  I think you can also use the iterate over items and test for isImport method?

hsorby (commented):
A couple of resource files have been added but not referenced yet, presumably this is to come.  (More a reminder to myself to not forget this later.)

tests/resolve_imports/file_parser.cpp
nickerso: not sure yet - I'm just trying out things to see how it works...once I have something working I'll talk to you about how it should really be done ;-)

tests/resolve_imports/tests.cmake
nickerso: addressed by #172.

src/api/libcellml/import.h
```diff
@@ -36,24 +36,55 @@ class LIBCELLML_EXPORT Import: public Entity
...
 
+    /**
+     * @brief Provide the model used to resolve this import.
+     *
+     * Uses the provided model to resolve this import, which should correspond
+     * to the @c source identified by this import.
+     *
+     * @param model The model to use in resolving this import.
+     */
+    void resolveImport(libcellml::ModelPtr model);
```
hsorby, line 65: I think this method should be called 'setResolvingModel'

tests/resolve_imports/file_parser.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+// generated with test resource locations
+#include "test_resources.h"
+
+static std::string resolvePath(const std::string& filename, const std::string& base)
+{
+    // we can be naive here as we know what we are dealing with
+    std::string path = base.substr(0, base.find_last_of('/')+1) + filename;
+    return path;
+}
```
hsorby, line 34: Utility functions should be put under there own directory probably called 'utilities'.

tests/resolve_imports/file_parser.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+static size_t countImportedChildren(libcellml::ComponentPtr parent)
+{
+    size_t numberImportedChildren = 0;
+    for (size_t n = 0; n < parent->componentCount();  ++n)
+    {
+        libcellml::ComponentPtr c = parent->getComponent(n);
+        if (c->isImport()) ++numberImportedChildren;
```
hsorby, line 42: Really want this split into individual lines for our coverage counting.

tests/resolve_imports/file_parser.cpp
```diff
@@ -0,0 +1,209 @@
...
+static size_t countUnresolvedComponents(libcellml::ComponentPtr component);
+static size_t countUnresolvedImportedComponents(libcellml::ModelPtr model);
+
+static size_t countUnresolvedC(libcellml::ComponentPtr component)
+{
+    size_t count = 0;
+    if (component->isImport())
+    {
+        libcellml::ImportPtr imp = component->getImport();
+        if (! imp->isResolved()) ++count;
```
hsorby, line 58: Separate line please, got to count those lines for coverage.

tests/resolve_imports/file_parser.cpp
```diff
@@ -0,0 +1,209 @@
...
+        if (! imp->isResolved())
+        {
+            std::string url = resolvePath(imp->getSource(), baseFile);
+            std::ifstream t(url);
+            std::stringstream buffer;
+            buffer << t.rdbuf();
+            libcellml::Parser p(libcellml::Format::XML);
+            libcellml::ModelPtr model = p.parseModel(buffer.str());
+            EXPECT_EQ(0, p.errorCount());
+            if (model)
```
hsorby, line 110: Why test for model? There should be a test for EXPECT_NEQ(nullptr, model).

tests/resolve_imports/file_parser.cpp
```diff
@@ -0,0 +1,209 @@
...
+    libcellml::Parser p(libcellml::Format::XML);
+    libcellml::ModelPtr model = p.parseModel(buffer.str());
+
+    EXPECT_EQ(0, p.errorCount());
+
+    size_t nImportedComponents = 0;
+    for (size_t n = 0; n < model->componentCount();  ++n)
+    {
+        libcellml::ComponentPtr c = model->getComponent(n);
+        if (c->isImport()) ++nImportedComponents;
```
hsorby, line 155: Lines for coverage.

tests/resolve_imports/file_parser.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    libcellml::Parser p(libcellml::Format::XML);
+    libcellml::ModelPtr model = p.parseModel(buffer.str());
+    EXPECT_EQ(0, p.errorCount());
+
+    size_t nImportedComponents = 0;
+    for (size_t n = 0; n < model->componentCount();  ++n)
+    {
+        libcellml::ComponentPtr c = model->getComponent(n);
+        if (c->isImport()) ++nImportedComponents;
```
hsorby, line 176: Lines for coverage.

tests/resolve_imports/tests.cmake
hsorby: Will be resolved once #172 is done.

agarny:
Might want to merge `develop` into this PR (`src/xmldoc.cpp` is listed as having changes while it shouldn't).

nickerso:
merged :)

src/api/libcellml/import.h
```diff
@@ -36,24 +36,55 @@ class LIBCELLML_EXPORT Import: public Entity
...
+     *
+     * Get the model which has been assigned to resolve this import. If no model
+     * has been assigned then return the @c nullptr.
+     *
+     * @return The model used to resolve this import.
+     */
+    libcellml::ModelPtr getResolvingModel() const;
+
+    /**
+     * @brief Test is this import is resolved.
```
agarny, line 78: Test **if** this import...

src/api/libcellml/import.h
agarny: Why not simply `setSourceModel()` (and, likewise, `getSourceModel()` rather than `getResolvingModel()`)?

tests/resolve_imports/file_parser.cpp
```diff
@@ -0,0 +1,209 @@
...
+    for (size_t n = 0; n < parent->componentCount();  ++n)
+    {
+        libcellml::ComponentPtr c = parent->getComponent(n);
+        if (c->isImport()) ++numberImportedChildren;
+        numberImportedChildren += countImportedChildren(c);
+    }
+    return numberImportedChildren;
+}
+
+static size_t countUnresolvedC(libcellml::ComponentPtr component);
```
agarny, line 48: Delete.

tests/resolve_imports/file_parser.cpp
```diff
@@ -0,0 +1,209 @@
...
+static size_t countUnresolvedComponents(libcellml::ComponentPtr component);
+static size_t countUnresolvedImportedComponents(libcellml::ModelPtr model);
+
+static size_t countUnresolvedC(libcellml::ComponentPtr component)
```
agarny, line 52: Rename to `countUnresolvedComponent()` (i.e. similar for `resolveComponent()` below).

Actually, I would probably have `doCountUnresolvedImportedComponent()`, `doCountUnresolvedImportedComponents()` and `countUnresolvedImportedComponents()` (and something similar for the resolve methods below). Indeed, I originally got confused when I first saw `countUnresolvedC()` and `countUnresolvedComponents()` and then `countUnresolvedImportedComponents()`. For a second, I thought that the first two methods were not dealing with imported components...

agarny:
I haven't checked the test CellML files, but the code that makes use of them seems to be doing what it should.

However, your current implementation relies on the library user implementing his own code for resolving imported components, which you do in your tests (see `resolveImportedComponents()`) and thus set the resolving/source model (see your (current) `getResolvingModel()` method).

I don't understand why this is the case. If anything, I would expect our `Import` class to have a `resolveImports()` method (or something similar) and then do what your `resolveImportedComponents()` method does. This also means that your (current) `resolveImport()` method would become `private`, meaning that a library user would only be able to retrieve the imported model (and not also set it).

nickerso:
The current thinking is that libCellML is simply a data model which captures the CellML 2.0 specification and provides validation. In order to resolve imports, libCellML would need to have some way to access local and/or remote files - which is currently beyond the scope of libCellML (http://libcellml.readthedocs.io/en/latest/current_thinking.html).

These tests are designed to illustrate this and that the libCellML API is sufficient for enabling "applications" to resolve imports. Which can involve some tricky code if models, for example, need authentication with a resource like PMR in order to access private imports. A higher level "libcellml" would provide this kind of functionality and work for general cases (like using libCURL to access remote import sources), but I think you would always want a way for applications to implement their own resolving (i.e., the way is currently done internally within PMR to avoid thread locking). In the same way that handling math is currently out-of-scope for libCellML and would be achieved using a higher level service built on top of libCellML. And then potentially in future we might bundle several such higher level libraries into a more functionally complete "libCellML" package of some sort.

An alternative is providing some kind of callback registry where the calling application provides libCellML a method to call when import sources need to be serialised to a string.

hsorby:
I myself am not in favour with the callback option, my preferred way of dealing with this is to add layers on top of libcellml to handle what might be termed higher level functionality.  It may be the case that we decide to increase the scope of libcellml to handle this internally.  We made the decision to limit the scope so that we could get to the validation stage quickly or rather with the minimal amount of effort.

agarny:
Ok, I take your (@nickerso and @hsorby) points. So, with that in mind, I would agree with @hsorby about not having a callback option and, instead, having a layer on top of libCellML. However, I would like to have that layer available ASAP after we are done with core libCellML (this is essential if we want libCellML to be used by people, including myself).

hsorby:
What is it we are trying to achieve here? We have actual code that is defined in the tests directory which doesn't make it a part of the library.  We also have tests for code that isn't in the library.  Where is this going and do we want it to be a part of the codebase?

agarny:
Personally, and independently of how libCellML is structured underneath (i.e. core + layers), I would like people to be able to use libCellML without having to write extra code for handling 'trivial' things such as imports. Indeed, such extra code is bound to be useful to most libCellML users, so not having it as part of libCellML (in some way or another) would, IMHO, greatly hinder libCellML's uptake (been there, done that...). Indeed, as a 'simple' CellML end-user, if I had to write such extra code, I would probably not bother with it and would instead simply not use imports, which is clearly not the kind of practice we want to advocate.

Issue #170, opened 2017-08-28
Problem parsing encapsulation hierarchy


nickerso:
While working on resolving component imports, I stumbled across an issue with the parsing of encapsulation hierarchies. Will submit a pull request with a new test highlighting the issue.

nickerso:
Sorry, a few more details might help. Given the encapsulation hierarchy:
```
  <encapsulation>
    <component_ref component="root">
      <component_ref component="L1_c1">
        <component_ref component="L1_L2_c1"/>
      </component_ref>
      <component_ref component="L1_c2"/>
      <component_ref component="L1_c3"/>
    </component_ref>
  </encapsulation>
```
I get the error messages:
```
Encapsulation in model 'complex_encapsulation_example' specifies 'L1_c2' as a parent component_ref but it does not have any children.
Encapsulation in model 'complex_encapsulation_example' specifies 'L1_c3' as a parent component_ref but it does not have any children.
```
without the `L1_L2_c1` child in there, it parses without error.

PR #171, opened 2017-08-28
New test case with complex encapsulation hierarchy.

nickerso:
As far as I can tell, this new test should parse fine, but the addition of the `L1_L2_c1` child into the
encapsulation hierarchy causes the parser to give a bogus error. Demonstrates the issue mentioned in #170.

### DIFF
--- a/tests/parser/file_parser.cpp
+++ b/tests/parser/file_parser.cpp
@@ -95,3 +95,18 @@ TEST(Parser, parseOrdModelFromFile) {
     EXPECT_EQ("public", a);
 }
 
+TEST(Parser, parseComplexEncapsulationModelFromFile) {
+    std::ifstream t(TestResources::getLocation(
+                    TestResources::CELLML_COMPLEX_ENCAPSULATION_MODEL_RESOURCE));
+    std::stringstream buffer;
+    buffer << t.rdbuf();
+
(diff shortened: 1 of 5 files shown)


nickerso (commented):
New test case now passing and all old tests still working so I think we can say this one has been resolved. @hsorby can you please give things a final check.

tests/parser/file_parser.cpp
```diff
@@ -95,3 +95,18 @@ TEST(Parser, parseOrdModelFromFile) {
...
+    std::stringstream buffer;
+    buffer << t.rdbuf();
+
+    libcellml::Parser p(libcellml::Format::XML);
+    p.parseModel(buffer.str());
+
+    EXPECT_EQ(0, p.errorCount());
+    for (size_t i = 0; i < p.errorCount(); ++i) {
+        std::cout << p.getError(i)->getDescription() << std::endl;
+    }
```
hsorby: This printing the description should be removed now as it is redundant.

hsorby (commented):
One minor change.

tests/parser/file_parser.cpp
hsorby: We could add a comment referencing the issue that gave rise to this test, I think that could be helpful in the future.

hsorby (approved):

Issue #176, opened 2017-08-31
Add reset object.


hsorby:
The draft specification specifies a Reset object the library needs to support this.  This issue is to implement support for the [reset element in CellML 2.0](https://docs.google.com/document/d/1MleZJF6DYp4FO4vxslAHInrRFMqrSpARa1s_dPhwYU0/edit#heading=h.z0pc1eqvm9bg) (see [here](https://docs.google.com/document/d/1hJmyhiSY8uHlWs7tjZcZyN4RpGeUgrxU2yq1WSIDNKI/edit) for some examples of the `reset` element).

PR #177, opened 2017-08-31
Add resets.

hsorby:
Addresses issue #176 (see [here](https://docs.google.com/document/d/1MleZJF6DYp4FO4vxslAHInrRFMqrSpARa1s_dPhwYU0/edit#heading=h.z0pc1eqvm9bg) for the `reset` element specifications and [here](https://docs.google.com/document/d/1hJmyhiSY8uHlWs7tjZcZyN4RpGeUgrxU2yq1WSIDNKI/edit) for some examples).


hsorby:
These are the classes that I thinking we will need to add to support resets.

hsorby:
In the When information element how do you distinguish between the condition math element and the value math element?

nickerso:
As per 19.11.3 in the draft spec, the condition is the first `math` child and the value is the second `math` child.

hsorby:
Seems a bit undeclarative.

nickerso:
Standard MathML - similar to the way `piecewise` elements work...and `root`, `power`, etc.

MichaelClerx (commented):
Hi Hugh,
Thanks for updating the python bindings as well!

Not sure if everyone would have been willing/able to do that, so just wondering should we come up with some policy regarding this and add some notes to the contributing guidelines? Don't want to deter people from proposing fixes/improvements, but if we end up with several languages I can see it might!

agarny:
+1 re @MichaelClerx's suggestion. However, depending on the binding, someone might or not have the expertise to keep things in sync, so we should keep that in mind and maybe have people "in charge" of a given binding? For example, maybe @MichaelClerx for Python? In other words, although someone should try his best to keep the bindings in sync, the person in charge of a given binding would have to double check that everything is indeed "right", or get / help getting the binding in sync with the main code, if needed.

hsorby:
I would like to see contributors feel no pressure to do bindings that they are not comfortable with.  I would rather have the contribution no matter how small rather than not.  To that end I think it would be a good idea for there to be a person (or persons) who are responsible for an area like the bindings.

This could also be extended to something like having someone responsible for coding standard and the like.

hsorby:
I think this is now ready for review.

MichaelClerx:
How mergeable is this in terms of the FindPython2,3 stuff at the top?

MichaelClerx:
I can't build this, because of the same issue fixed in #231

src/api/libcellml/orderedentity.h
```diff
@@ -0,0 +1,79 @@
...
+{
+public:
+    OrderedEntity(); /**< Constructor */
+    ~OrderedEntity(); /**< Destructor */
+    OrderedEntity(const OrderedEntity &rhs); /**< Copy constructor */
+    OrderedEntity(OrderedEntity &&rhs); /**< Move constructor */
+    OrderedEntity& operator=(OrderedEntity n); /**< Assignment operator */
+
+    /**
+     * @brief Set the order for this reset.
```
MichaelClerx: ...of this entity

src/api/libcellml/orderedentity.h
```diff
@@ -0,0 +1,79 @@
...
+     *
+     * Sets the order for this reset.
```
MichaelClerx: entity

src/api/libcellml/orderedentity.h
```diff
@@ -0,0 +1,79 @@
...
+     *
+     * @param order The order to set.
+     */
+    void setOrder(int order);
+
+    /**
+     * @brief Get the order for this reset.
```
MichaelClerx: entity

src/api/libcellml/printer.h
```diff
@@ -160,6 +182,17 @@ class LIBCELLML_EXPORT Printer: public Logger
      */
     std::string printEncapsulation(ComponentPtr component) const;
 
+    /**
+     * @brief Serialise the when to @c std::string.
```
MichaelClerx: Can we capitalise When or show it's not just the word "when"?

src/api/libcellml/reset.h
```diff
@@ -0,0 +1,171 @@
...
+     */
+    void addWhen(const WhenPtr &when);
+
+    /**
+     * @brief Remove the when at the given @p index.
+     *
+     * Remove the when with the given @p index. @p index must
+     * be in the range [0, \#whens).
+     *
+     * @param index The index of the when to remove.
```
MichaelClerx: again capitalising here would make it more readable :-)

src/api/libcellml/reset.h
```diff
@@ -0,0 +1,171 @@
...
+     * Whens for the specified @p When pointer. Default value is @c true.
+     *
+     * @return True if the when was replaced, false otherwise.
+     */
+    bool removeWhen(const WhenPtr &when);
+
+    /**
+     * @brief Remove all whens stored in this when entity.
+     *
+     * Clears all whens that have been added to this reset.
```
MichaelClerx: all @c Whens that have been added to this @c Reset ?

src/api/libcellml/reset.h
```diff
@@ -0,0 +1,171 @@
...
+     *
+     * @param when The pointer to the when to remove.
+     * Whens for the specified @p When pointer. Default value is @c true.
+     *
+     * @return True if the when was replaced, false otherwise.
+     */
+    bool removeWhen(const WhenPtr &when);
+
+    /**
+     * @brief Remove all whens stored in this when entity.
```
MichaelClerx: in this reset, not a when

src/api/libcellml/reset.h
```diff
@@ -0,0 +1,171 @@
...
+     *
+     * Clears all whens that have been added to this reset.
+     */
+    void removeAllWhens();
+
+    /**
+     * @brief Tests to see if the when pointer is contained within this reset.
```
MichaelClerx: @p when etc.

src/api/libcellml/when.h
```diff
@@ -0,0 +1,85 @@
...
+    void setCondition(const std::string& condition);
+
+    /**
+     * @brief Get the condition term for this when.
+     *
+     * Get the @c std::string MathML string as the condition term for this when.
+     *
+     * @return The @c std::string MathML string of the condition.
+     */
+    std::string getCondition() const;
```
MichaelClerx, line 57: Out of interest: Will be returning objects here later?

src/component.cpp
```diff
@@ -202,5 +214,58 @@ bool Component::hasVariable(const std::string &name) const
...
+}
+
+ResetPtr Component::getReset(size_t index) const
+{
+    ResetPtr reset = nullptr;
+    if (index < mPimpl->mResets.size()) {
+        reset = mPimpl->mResets.at(index);
+    }
+
+    return reset;
```
MichaelClerx, line 257: Are we happy to return null instead of using exceptions?

MichaelClerx (commented):
Some small things in the documentation, but no other objections from me!

src/api/libcellml/when.h
hsorby: It's a possibility :)

src/component.cpp
hsorby: We are not raising any exceptions in libCellML directly.  I think it is something you avoid unless you find yourself in a truly exceptional circumstance.

MichaelClerx:
Working through the examples now :-)
Worth noting they describe an approximation to the behaviour defined by cellml, unless we start taking limits of `dt`. Something to think about for the informative spec...

MichaelClerx:
Hmmmm. Would we have been better off describing these resets as changes to be made upon certain crossings? So "when B rises past 3 then A <-- 3" (like a [signal edge](https://en.wikipedia.org/wiki/Signal_edge) in electronics). It's not really about "when B > 3", because A is allowed change after that...

The "interpretation" section of the spec also has a mistake (I made a comment just now), by stating: "Each reset element describes a change to be applied to the variable referenced by the variable attribute when specified conditions are met". This is not true, the change is to be applied at the instant when a condition changes from false to true, not for the duration of the period that this transition holds!

We clarify this after, but hmmm... Maybe we need to pick our names better while we still can? Since it's confusing even ourselves

src/api/libcellml/component.h
```diff
@@ -16,8 +16,9 @@ limitations under the License.
 
 #pragma once
 
-#include "libcellml/componententity.h"
 #include "libcellml/exportdefinitions.h"
+#include "libcellml/componententity.h"
```
agarny: I believe we did, at some point, agree on listing include files in alphabetical order. So, we should have `componententity.h` back to where it used to be.

src/api/libcellml/component.h
```diff
@@ -210,6 +213,89 @@ class LIBCELLML_EXPORT Component: public ComponentEntity
...
+    /**
+     * @brief Add a reset by reference as part of this component.
+     *
+     * Add a reset by reference as part of the given component.
+     *
+     * @sa removeReset
+     *
+     * @param r The reset to add.
+     */
+    void addReset(const ResetPtr &r);
```
agarny: For consistency, we might want to replace `r` with `reset` (as is the case with `removeReset()` below).

src/api/libcellml/componententity.h
```diff
@@ -61,11 +61,11 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
      * @brief Remove the component at the given @p index.
      *
      * Remove the component with the given @p index. @p index must
-     * be in the range [0, #components).
+     * be in the range [0, \#components).
      *
      * @param index The index of the component to remove.
      *
-     * @return True if the units were replaced, false otherwise.
+     * @return True if the component was replaced, false otherwise.
```
agarny: "removed" rather than "replaced"?

src/api/libcellml/componententity.h
```diff
@@ -82,7 +82,7 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
      * @param searchEncapsulated Boolean flag to indicate whether we should also search encapsulated
      * components for the component with the specified @p name. Default value is @c true.
      *
-     * @return True if the units were replaced, false otherwise.
+     * @return True if the component was replaced, false otherwise.
```
agarny: "removed" rather than "replaced"?

src/api/libcellml/orderedentity.h
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * @brief Mark the order as unset.
+     *
+     * Marks the order value as unset .
+     */
+    void unsetOrder();
+
+    /**
+     * @brief Test to determine of the order is set or unset.
```
agarny: "... determine if the..."

src/api/libcellml/reset.h
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * @brief Remove the @c When with the given pointer.
+     *
+     * Remove the @c When with the pointer @p when.
+     *
+     * @overload
+     *
+     * @param when The pointer to the @c When to remove.
+     * @c Whens for the specified @p when pointer. Default value is @c true.
```
agarny: Not sure I understand that sentence...

src/api/libcellml/reset.h
```diff
@@ -0,0 +1,171 @@
...
+     *
+     * Clears all @c Whens that have been added to this @c Reset.
+     */
+    void removeAllWhens();
+
+    /**
+     * @brief Tests to see if the @c When pointer is contained within this @c Reset.
+     *
+     * Tests to see if the argument @c When pointer @p when is contained
+     * within this @c When entity. Returns @c true if the @c When is in the @c Reset
```
agarny: "within this @c Reset entity."

src/api/libcellml/when.h
```diff
@@ -0,0 +1,85 @@
...
+    When& operator=(When n); /**< Assignment operator */
+
+    /**
+     * @brief Set the condition term for this when.
+     *
+     * Set the @c std::string MathML string as the condition term for this when.
+     *
+     * @param condition The @c std::string MathML string of the condition to set.
+     */
+    void setCondition(const std::string& condition);
```
agarny: Minor: there should be a space before `&` and none after, i.e.

void setCondition(const std::string &condition);

The same is true in various parts of the code in fact (e.g. `setValue()` below).

src/bindings/interface/component.i
```diff
@@ -70,6 +70,30 @@ other variables), this component will not be serialised in the connection.";
 %feature("docstring") libcellml::Component::variableCount
 "Returns the number of variables the component directly contains.";
 
+%feature("docstring") libcellml::Component::addReset
+"Add a reset `r` to this component.";
```
agarny: "Add a reset `reset` to this component.";

src/parser.cpp
```diff
@@ -24,6 +24,8 @@ limitations under the License.
 #include "libcellml/importsource.h"
 #include "libcellml/model.h"
 #include "libcellml/variable.h"
+#include "libcellml/reset.h"
+#include "libcellml/when.h"
```
agarny: Wrong order of includes.

src/parser.cpp
```diff
@@ -1125,4 +1157,216 @@ void Parser::ParserImpl::loadImport(const ImportSourcePtr &importSource, const M
...
+            const std::string variableReference = attribute->getValue();
+            referencedVariable = component->getVariable(variableReference);
+            if (referencedVariable == nullptr) {
+                ErrorPtr err = std::make_shared<Error>();
+                err->setDescription("Reset referencing variable '" + variableReference +
+                                    "' is not a valid reference for a variable in component '" + component->getName() + "'.");
+                err->setReset(reset);
+                err->setKind(Error::Kind::RESET);
+                err->setRule(SpecificationRule::RESET_INVALID_VARIABLE_REFERENCE);
+                mParser->addError(err);
```
agarny, line 1179: I am not sure I like this, i.e. call `setReset()` and then `setKind()`. I mean, we could, in principle, have a call `setKind()`, but then forget to call `setReset()`. In fact, if at all possible, I would have a simple call to `mParser->addError(...);`, to which all the parameters needed to create and customise `ErrorPtr` would be passed. We could, for example, pass `reset`, but not `Error::Kind::RESET` since this would be implied. This means that we would have several signatures for `addError()`, but at least there wouldn't be any risk for errors from a library user. Just a thought...

src/parser.cpp
```diff
@@ -1125,4 +1157,216 @@ void Parser::ParserImpl::loadImport(const ImportSourcePtr &importSource, const M
...
+        } else if (attribute->isType("order")) {
+            if (convertToInt(attribute->getValue(), &order)) {
+                orderDefined = true;
+            } else {
+                if (reset->getVariable() != nullptr) {
+                    variableName = reset->getVariable()->getName();
+                }
+                ErrorPtr err = std::make_shared<Error>();
+                err->setDescription("Reset in component '" + component->getName() +
+                                    "' referencing variable '" + variableName +
```
agarny, line 1192: We ought to handle the case where `variableName` is a null string.

src/parser.cpp
```diff
@@ -1125,4 +1157,216 @@ void Parser::ParserImpl::loadImport(const ImportSourcePtr &importSource, const M
...
+    if (orderDefined) {
+        reset->setOrder(order);
+    } else {
+        ErrorPtr err = std::make_shared<Error>();
+        err->setDescription("Reset in component '" + component->getName() +
+                            "' referencing variable '" + variableName +
+                            "' does not have an order defined.");
+        err->setReset(reset);
+        err->setKind(Error::Kind::RESET);
+        mParser->addError(err);
```
agarny, line 1233: Am I correct in assuming that we will get two error messages if a non-integer value is provided for the order? If so, is that really what we want?

src/parser.cpp
```diff
@@ -1125,4 +1157,216 @@ void Parser::ParserImpl::loadImport(const ImportSourcePtr &importSource, const M
...
+            mParser->addError(err);
+        }
+        childNode = childNode->getNext();
+    }
+}
+
+void Parser::ParserImpl::loadWhen(const WhenPtr &when, const ResetPtr &reset, const XmlNodePtr &node)
+{
+    std::string referencedVariableName = "";
+    VariablePtr referencedVariable =  reset->getVariable();
```
agarny: (Very) minor: replace the double space before `reset` with only one space.

src/parser.cpp
```diff
@@ -1125,4 +1157,216 @@ void Parser::ParserImpl::loadImport(const ImportSourcePtr &importSource, const M
...
+        err->setKind(Error::Kind::WHEN);
+        mParser->addError(err);
+    }
+
+    size_t mathNodeCount = 0;
+    XmlNodePtr childNode = node->getFirstChild();
+    while (childNode) {
+        if (childNode->isType("math")) {
+            // TODO: copy any namespaces declared in parents into the math element
+            //       so math is a valid subdocument.
```
agarny, line 1317: Shouldn't this be done before having this PR merged in?

src/printer.cpp
```diff
@@ -148,11 +150,19 @@ std::string Printer::printComponent(ComponentPtr component) const
         repr += " id=\"" + component->getId() + "\"";
     }
     size_t variable_count = component->variableCount();
-    if ((variable_count > 0) || (component->getMath().length())) {
+    size_t reset_count = component->resetCount();
```
agarny: Minor: for consistency, rename `variable_count` to `variableCount` and `reset_count` to `resetCount`.

src/utilities.cpp
```diff
@@ -20,13 +20,15 @@ limitations under the License.
 #include <iomanip>
 #include <limits>
 #include <sstream>
+#include <algorithm>
+#include <set>
```
agarny: Order of includes.

src/utilities.cpp
```diff
@@ -53,4 +55,41 @@ std::string convertDoubleToString(double value)
     return strs.str();
 }
 
+bool isCellMLIntegerCharacter(char c) {
+    const std::set<char> validIntegerCharacters = {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9'};
+    return validIntegerCharacters.find(c) != validIntegerCharacters.end();
+}
+
+bool convertToInt(const std::string &candidate, int *value)
```
agarny, line 63: I believe we are using C++11? If so, we might want to use [`stoi()`](http://www.cplusplus.com/reference/string/stoi/).

tests/bindings/python/test_reset.py
```diff
@@ -0,0 +1,144 @@
...
+
+        self.assertEqual(1, r.whenCount())
+
+        self.assertTrue(r.removeWhen(0))
+        self.assertEqual(0, r.whenCount())
+
+        r.addWhen(w)
+
+        self.assertTrue(r.removeWhen(w))
+        self.assertEqual(0, r.whenCount())
```
agarny, line 59: What is the rationale for the 3 lines of code above? It looks like a repeat of what is above?

tests/bindings/python/test_reset.py
```diff
@@ -0,0 +1,144 @@
...
+        from libcellml import When
+
+        w1 = When()
+        w2 = When()
+        w3 = When()
+        r = Reset()
+
+        r.addWhen(w1)
+        r.addWhen(w2)
+        r.addWhen(w3)
```
agarny, line 72: Might want to add the following?

```
self.assertEqual(3, r.whenCount())
```

tests/coverage/coverage.cpp
```diff
@@ -64,6 +64,24 @@ TEST(Coverage, units) {
...
 
+TEST(Coverage, when) {
+    const std::string e = "<reset><when/></reset>";
+    libcellml::When w, wm;
+    libcellml::Reset r;
+
+    wm = std::move(w);
+
+    libcellml::When wc(wm);
+
```
agarny, line 75: Delete empty line.

tests/parser/parser.cpp
```diff
@@ -1417,3 +1417,177 @@ TEST(Parser, parseIds) {
...
+                            "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">"
+                                "some value in mathml"
+                            "</math>"
+                            "<variable/>"
+                        "</when>"
+                    "</reset>"
+                "</component>"
+            "</model>";
+
+    const std::vector<std::string> expectedErrors = {
```
agarny, line 1531: As discussed above, it might be nice to handle the case where there is no variable name (or order value) as such. Also, I can't recall whether we discussed it before (and whether it would be easy to do here), but it might be worth specifying the line/column where an error was found?

tests/reset/reset.cpp
```diff
@@ -0,0 +1,467 @@
...
+
+    EXPECT_EQ(nullptr, r.takeWhen(4));
+}
+
+TEST(Reset, replaceWhenMethods) {
+    libcellml::Reset r;
+    libcellml::WhenPtr c1 = std::make_shared<libcellml::When>();
+    libcellml::WhenPtr c2 = std::make_shared<libcellml::When>();
+    libcellml::WhenPtr c3 = std::make_shared<libcellml::When>();
+    libcellml::WhenPtr c4 = std::make_shared<libcellml::When>();
```
agarny: Delete this line since `c4` is not used.

tests/reset/reset.cpp
```diff
@@ -0,0 +1,467 @@
...
+    r2 = std::move(r1);
+    EXPECT_EQ(1, r2.whenCount());
+
+    // Testing move constructor for Reset
+    libcellml::Reset r3 = std::move(r2);
+    EXPECT_EQ(1, r3.whenCount());
+}
+
+TEST(Reset, printResetWithVariable) {
+    std::string e = "<reset variable=\"A\"/>";
```
agarny: `const std::string e` (and the same in a few places below).

tests/reset/reset.cpp
```diff
@@ -0,0 +1,467 @@
...
+                                  "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">"
+                                      "a value set"
+                                  "</math>"
+                              "</when>"
+                          "</reset>";
+
+    libcellml::ResetPtr r = std::make_shared<libcellml::Reset>();
+    libcellml::WhenPtr w = std::make_shared<libcellml::When>();
+
+    w->setValue("<math xmlns=\"http://www.w3.org/1998/Math/MathML\">a value set</math>");
```
agarny, line 263: Am I correct in assuming that all calls to `When::setValue()` will need to include the `math` element? If anything, I would have expected a call like `w->setValue("a value set")`? I mean, the API is supposed to make our life "easy", no?

tests/when/when.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    EXPECT_EQ("<some mathml type string for condition.>", w->getCondition());
+}
+
+TEST(When, value) {
+    libcellml::WhenPtr w = std::make_shared<libcellml::When>();
+
+    w->setCondition("<some mathml type string for value.>");
+
+    EXPECT_EQ("<some mathml type string for value.>", w->getCondition());
```
agarny: I believe we should have a call to `w->setValue()` and `w->getValue()`.

src/api/libcellml/componententity.h
```diff
@@ -98,7 +98,7 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
      * @param searchEncapsulated Boolean flag to indicate whether we should also search encapsulated
      * components for the specified @p component pointer. Default value is @c true.
      *
-     * @return True if the units were replaced, false otherwise.
+     * @return True if the component was replaced, false otherwise.
```
agarny: "removed" rather than "replaced"?

agarny:
Besides my inline comments:
 - In `libCellMLRoughObjectModel-Entities.png`: why doesn't `OrderedEntity` inherit from `Entity`. The `Reset` and `When` classes are the only classes that inherit from `OrderedEntity`, and they also inherit from `Entity`.
 - In `libCellMLRoughObjectModel-Entities.png`: we are missing the `hasReset()` method in the `Component` class.
 - In `libCellMLRoughObjectModel-Entities.png`: we are missing the `isOrderSet()` method in the `OrderedEntity` class.
 - In `libCellMLRoughObjectModel-Entities.png`: we are missing the `setVariable()`, `getVariable()`, method in the `OrderedEntity` class.
 - In `libCellMLRoughObjectModel-Entities.png`: `removeWhen()` in the `Reset` class should read "removeWhen(...): void" since we have two signatures for `removeWhen()`.

Apart from that, it looks fine to me.

src/parser.cpp
hsorby: I have done this in a later pull request (for validation #227) where I have simplified adding errors slightly.  I haven't gone the whole way on this but this can be added easily enough.

src/parser.cpp
hsorby: Yes, I think it is what we want as we are actually violating two rules.

src/parser.cpp
hsorby: Null string or empty string, I'm not clear on what you mean here.

src/parser.cpp
hsorby: I'm not finding any problems with this at this point so I don't have any desire to fix something that is not broken.  If we have an example that we can test against that does cause a problem then I would fix it.

src/utilities.cpp
hsorby: We are using it, we are just wrapping over it so we can do some manipulation checking if required.

tests/bindings/python/test_reset.py
hsorby: Different method used to remove the when one by index the other by class.

tests/reset/reset.cpp
hsorby: At this time I think this is fine, it wouldn't be hard to add the wrapping if not present if that is desired.

src/parser.cpp
agarny: Fair enough, this is not related to this PR, so it can be simplified some other time, I agree.

src/parser.cpp
agarny: Ok, fair enough, I don't (yet) feel too strongly about it.

src/parser.cpp
agarny: By null string, I indeed mean an empty string. `variableName` can be equal to `""`, which means that the error message would be a bit nonsensical.

src/parser.cpp
agarny: Fair enough, it's just that whenever I see a `TODO`, then I would expect it implemented or deleted (i.e. have the code as clean as possible).

src/utilities.cpp
agarny: Oops, my bad, I didn't spot it!

tests/bindings/python/test_reset.py
agarny: Ok. Actually, it might be worth adding a comment about it?

tests/reset/reset.cpp
agarny: Ok for now, but I will certainly like to have such a wrapper.

Issue #181, opened 2017-09-05
Have a way to uninstall libCellML


agarny:
I can build and install libCellML just fine, but in some cases it would be nice to also be able to uninstall it. So, it would be nice to be able to do something like `make uninstall` in the same way that we can do `make install`.

hsorby:
I think it is quite easy to do with CMake there is already a manifest of what was installed so I think you can just read this manifest and remove the files recorded there.

agarny:
Ok, I will have a quick look at it and see whether I can get anywhere with it.

hsorby:
I think all you need to do is follow the instructions here:

https://cmake.org/Wiki/CMake_FAQ#Can_I_do_.22make_uninstall.22_with_CMake.3F

agarny:
Yes, I was just about to look for something like this, so thanks a lot for the head start.

PR #185, opened 2017-09-06
Added an uninstall target (#181)

agarny:
Addresses issue #181 (based on https://cmake.org/Wiki/CMake_FAQ#Can_I_do_.22make_uninstall.22_with_CMake.3F).


nickerso (approved):
Seems to work fine for me. Doesn't stomp random files that get added to the install tree, but cleans up the installed files nicely. At least on Mac OS...using the ninja build tool.

Issue #189, opened 2017-09-13
`import` is a keyword in Python


MichaelClerx:
Hi all!

We have a class `Import` that encapsulates imported models (or at least models of which a component or unit is imported, is that right?).

With the current swig implmentation this means creating a module `import` for the class `Import` to live in, which is not allowed as `import` is a keyword.

Three questions for discussion:

1. Does anyone have a strong opinion on what I should call the `import` module instead?
2. Should this fix be restricted to the Python bindings?
3. What will be our general strategy for these problems. Try to choose the C++-classes so that we never have this problem OR create small differences in the generated wrappers to work around these issues?

hsorby:
Off the top of my head I can't think of any other class names that would clash with a keyword or reserved word in another language.

I would prefer to rename the `Import` class.  We are going to have to rename it for Python so if the new name is good enough for Python could it not also be good enough for C++.

MichaelClerx:
Any ideas @agarny @nickerso ?

MichaelClerx:
External model? Source?

A good choice would also fix #188, I suppose :-)

nickerso:
This is now reminding me why the CellML API has a mix of `CellML*` objects and non-`CellML*` objects :)

I agree that it would be best to rename `Import`. Some alternatives off the top of my head in order of my preference: `ImportSource`, `ModelReference`, `ModelSource`, `_I__mp_Or_t____`

MichaelClerx:
I like `_I__mp_Or_t____` but only if we rename the library `libce1lm1`

hsorby:
Sorry I can't have that I already have a project called `l1bcellml`, `libce1lm1` is too close I would have to call in the trademark lawyers.

hsorby:
`ImportSource` for me.

agarny:
`ImportSource` for me too.

MichaelClerx:
+1

MichaelClerx:
I could do a pull request but might easier for someone with a nice C++ IDE :-D

agarny:
A PR would still be required, but I am happy to do it.

Issue #188, opened 2017-09-12
setImport and getImport.


hsorby:
In issue #187 it was mentioned that the getImport and setImport functions and documentation was unclear as to what was going to happen.  Should we make this clearer?

hsorby:
I think it could make sense to use setImportModel and getImportModel perhaps.  This may then lead us to renaming and Import as ImportModel.  There is already a setImportReference and getImportReference.  (This API is from the ImportedEntity class.)

hsorby:
Or we could add more information into the documentation for the methods that explain what is going on better.

nickerso:
Just to clarify in my own mind: `(set|get)Import` is setting or getting the instance of an imported model to use (i.e., an `Import` not a `Model`); `(set|get)ImportReference` is setting a reference to the name of the entity in that instance of an imported model (typically these are both used via the `Component::setSourceComponent()` or `Units::setSourceUnits()` methods).

The need for having a handle to the `Import` is simply to ensure that we are correctly grouping imported components and units with the correct instance of an imported model. `(set|get)Import` makes sense to me as we are dealing with `Import`s, not models per se.

I don't think the current documentation (see [here](http://autotest.bioeng.auckland.ac.nz/libcellml-buildbot/b88e0a16a54ddb76c6dec855971bdb7d371f8e73/dox/code/classlibcellml_1_1ImportedEntity.html#a2f07fa682bbd7965ce76779228c8d47c) for example) is particularly clear in this regard and my preference would be to clarify the documentation rather than changing the method names.

MichaelClerx:
Is this still relevant @hsorby @nickerso ?

hsorby:
I think this has now been resolved with the merge of #248. Feel free to re-open if this understanding is incorrect.

PR #191, opened 2017-09-14
Rename the `Import` class to `ImportSource`

agarny:
Addresses issue #189.


hsorby:
The OS X build slave appears to be having issues with it's connection to the internet.  This is under investigation and clearly not fixed, I thought this had been resolved but obviously I will need to dig further into why it is having problems.

You can try resending the request again, @nickerso knows how :).

src/api/libcellml/component.h
```diff
@@ -41,10 +41,10 @@ class LIBCELLML_EXPORT Component: public ComponentEntity
      * Make this component an imported component by defining an import model
      * from which to extract the named component from.
      *
-     * @param imp The import from which the named component originates.
+     * @param import The import source from which the named component originates.
      * @param name The name of the component in the imported model to use.
      */
-    void setSourceComponent(const ImportPtr &imp, const std::string &name);
+    void setSourceComponent(const ImportSourcePtr &import, const std::string &name);
```
hsorby: The parameter here should be named importSource.

agarny:
Thanks for the explanation @hsorby. I have no idea how to resend the request again, so will wait for @nickerso.

src/api/libcellml/importsource.h
```diff
@@ -22,23 +22,23 @@ limitations under the License.
 namespace libcellml {
 
 /**
- * @brief The Import class.
+ * @brief The ImportSource class.
  *
- * The import class is for encapsulating an external CellML Model.
+ * The import source class is for encapsulating an external CellML Model.
  */
```
hsorby: We need to add more documentation here, @nickerso and I talked about this today I will try and add our thinking on this tomorrow.  I have forgotten and don't want to write the wrong thinking down.

src/api/libcellml/importsource.h
```diff
@@ -22,23 +22,23 @@ limitations under the License.
...
+    ~ImportSource(); /**< Destructor */
+    ImportSource(const ImportSource &rhs); /**< Copy constructor */
+    ImportSource(ImportSource &&rhs); /**< Move constructor */
+    ImportSource& operator=(ImportSource m); /**< Assignment operator */
 
     /**
      * @brief Set the source Model.
      *
-     * Set the source Model that this import refers to.
+     * Set the source Model that this import source refers to.
```
hsorby: Again we should add more documentation to make this clearer about what the source model is and what we are doing.

hsorby (commented):
Hopefully @nickerso can clarify the points I have been vague on.

agarny:
Hmm... it looks like the macOS slave is now behaving...

hsorby:
Yes it is temperamental it usually works but sometimes the network access seems to fail for reasons unknown.

src/api/libcellml/importsource.h
nickerso: I'm not one to let being wrong get in the way of writing a long comment on github :) hopefully the following is not so wrong that it points people in the wrong direction...@hsorby and I might revise tomorrow.

The basic idea is that the `ImportSource` does two things: the first is it points to a URL (the source) from which a CellML model is expected to be obtainable; the second is that it acts as a container/set/handle for all imported entities which should be obtained from the same *instance* of that source model. This becomes important, for example, when you have sibling components imported from the same instance (variable connections maintained) or different instances (connections severed). As with @hsorby's comment and from #188, I think a bit more of an explanation about what an `ImportSource` is would be useful here. I also now think "encapsulating an external CellML model" is actually incorrect/misleading.

Maybe @hsorby and I can come up with a better docstring tomorrow based on the above and comments on #188...

MichaelClerx:
Can we merge this or are we waiting for better docstrings?

nickerso:
Better docstrings would be good. I'll try to get time with @hsorby today to propose something.

agarny:
@nickerso's changes look good to me. So, @nickerso, if you are OK with my own changes, then feel free to merge this in (or someone else can merge this in if he's happy with all the changes).

nickerso (approved):

PR #199, opened 2017-09-20
Add ComponentEntity.replaceComponent(c1, c2) method

### LINKED ISSUE #198: Missing method `ComponentEntity.replaceComponent(ComponentPtr, ComponentPtr)`
We've currently got these two methods:

    bool replaceComponent(size_t index, const ComponentPtr &c);
    bool replaceComponent(const std::string &name, const ComponentPtr &component, bool searchEncapsulated=true);

but no

    bool replaceComponent(const ComponentPtr &component1, const ComponentPtr &component2, bool searchEncapsulated=true);

MichaelClerx:
This addresses issue #198.

### DIFF
--- a/src/api/libcellml/componententity.h
+++ b/src/api/libcellml/componententity.h
@@ -245,14 +245,14 @@ public:
      *
      * @overload
      *
-     * @param c1 The Component to replace.
-     * @param c2 The Component to use for replacement.
+     * @param oldComponent The Component to replace.
+     * @param newComponent The Component to use for replacement.
      * @param searchEncapsulated Boolean flag to indicate whether we should also search encapsulated
-     * components for @p c1. Default value is @c true.
+     * components for @p oldComponent. Default value is @c true.
      *
      * @return True if the component was replaced, false otherwise.
      */
-    bool replaceComponent(const ComponentPtr &c1, const ComponentPtr &c2, bool searchEncapsulated=true);
+    bool replaceComponent(const ComponentPtr &oldComponent, const ComponentPtr &newComponent, bool searchEncapsulated=true);
 
     /**
      * @brief Get the number of components in the component.
--- a/src/api/libcellml/componententity.h
+++ b/src/api/libcellml/componententity.h
@@ -236,6 +236,25 @@ public:
     bool replaceComponent(const std::string &name, const ComponentPtr &component, bool searchEncapsulated=true);
 
     /**
+     * @brief Replace the given component.
+     *
+     * Replaces the given component @p c1 with @p c2. If @p searchEncapsulated
+     * is @c true (default) this will also search for the component through this component's
+     * encapsulated components. If @p c1 is not found in the components children then no replacement
+     * is made.
+     *
+     * @overload
+     *
+     * @param c1 The Component to replace.
+     * @param c2 The Component to use for replacement.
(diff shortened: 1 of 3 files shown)


src/api/libcellml/componententity.h
```diff
@@ -236,6 +236,25 @@ class LIBCELLML_EXPORT ComponentEntity: public ImportedEntity
...
+     * @overload
+     *
+     * @param c1 The Component to replace.
+     * @param c2 The Component to use for replacement.
+     * @param searchEncapsulated Boolean flag to indicate whether we should also search encapsulated
+     * components for @p c1. Default value is @c true.
+     *
+     * @return True if the component was replaced, false otherwise.
+     */
+    bool replaceComponent(const ComponentPtr &c1, const ComponentPtr &c2, bool searchEncapsulated=true);
```
agarny: I would personally use `oldComponent` and `newComponent` rather than `c1` and `c2`, respectively.

src/componententity.cpp
```diff
@@ -257,4 +257,16 @@ bool ComponentEntity::replaceComponent(const std::string &name, const ComponentP
...
+{
+    bool status = replaceComponent(mPimpl->findComponent(c1) - mPimpl->mComponents.begin(), c2);
+    if (searchEncapsulated && !status) {
+        for (size_t i = 0; i < componentCount() && !status; ++i) {
+            status = getComponent(i)->replaceComponent(c1, c2, searchEncapsulated);
+        }
+    }
+
+    return status;
+}
```
agarny, line 270: Why reinvent the wheel (and potentially make it error prone when it comes to maintenance)? Why not simply:

```
bool ComponentEntity::replaceComponent(const ComponentPtr &c1, const ComponentPtr &c2, bool searchEncapsulated)
{
    return replaceComponent(c1->getName(), c2, searchEncapsulated);
}
```

Then again, this implies that there are no two components with the same name... and I can see that `removeComponent()` also duplicates code... In the end, I am happy with finding a component by pointer (as opposed to by name), but I am really not keen on duplicated code. We really ought to refactor the code to avoid all duplicates.

agarny:
This looks ok to me, except for the fact that we have duplicate code (as is already the case with `removeComponent()`, for example), which I really don't like (too error prone when it comes to code maintenance).

src/api/libcellml/componententity.h
MichaelClerx: Ok! Can I somehow edit this PR?

src/componententity.cpp
MichaelClerx: Not sure how you'd do that here! We could have a find(componentptr) and a find(name) and then template the replaceComponent method I suppose?

src/api/libcellml/componententity.h
nickerso: pushing more commits to your issue198 branch will automatically update the PR.

src/componententity.cpp
nickerso: Would be good to be able to replace unnamed components without having to iterate over a list in user code, but that can always come later if requested.

MichaelClerx (commented):
Updated parameter names!

MichaelClerx:
Can this be merged then?

nickerso (approved):
Looks good to me.

PR #200, opened 2017-09-20
Added Component.removeVariable(size_t) method

MichaelClerx:
We have a getVariable(size_t), and removeComponent(size_t), so adding this for completeness seems a sensible move!


tests/variable/variable.cpp
```diff
@@ -374,12 +374,22 @@ TEST(Variable, removeVariableMethods) {
     std::string a = printer.printComponent(c);
     EXPECT_EQ(e1, a);
     EXPECT_FALSE(c.removeVariable("BAD_NAME"));
-
+
```
agarny: Remove trailing spaces.

agarny:
This looks good to me, although I would make the same (minor) comments as for PR #199.

nickerso (approved):
I've fixed up a couple of minor issues, looks good to me.

PR #210, opened 2017-09-29
Removed Format enum. Closes #209.

### LINKED ISSUE #209: Remove XML format option.
Adding in the single option of the XML format was a bad idea.  This should be removed.  It was done with the hope that other formats would be possible but that hope was forlorn.  The CellML specification is tied to the XML format so there is no other format possible.

On the plus side it will make using libcellml a little easier, always a silver lining.

MichaelClerx:


agarny:
Looks good to me.

PR #226, opened 2018-03-05
[bindings] Split Python tests into multiple functions. Closes #225.

### LINKED ISSUE #225: Split Python tests into separate functions
As per @metatoaster 's suggestion in #205

MichaelClerx:
This splits the Python tests into smaller chunks, removing the need for lots of `del()` statements and improving legibility.
In addition, it tidies up the test code and bit, makes it pep8 compliant, and removes some unnecessary imports.

### DIFF
--- a/tests/bindings/python/test_version.py
+++ b/tests/bindings/python/test_version.py
@@ -1,21 +1,22 @@
 #
 # Tests the version function bindings
 #
-import sys
 import unittest
+
 
 class VersionTestCase(unittest.TestCase):
 
-    def test_validator(self):
+    def test_version(self):
         import libcellml
-        
+
         # unsigned int version()
         self.assertIsInstance(libcellml.version(), int)
         self.assertGreater(libcellml.version(), 0)
-        
+
         # const std::string versionString()
         self.assertIsInstance(libcellml.versionString(), str)
         self.assertNotEqual(libcellml.versionString(), '')
-        
+
+
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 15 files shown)


tests/bindings/python/test_version.py
```diff
@@ -1,21 +1,22 @@
...
 # Tests the version function bindings
 #
-import sys
 import unittest
 
+
 class VersionTestCase(unittest.TestCase):
 
-    def test_validator(self):
+    def test_version(self):
```
nickerso, line 9: So in the C++ tests we check that the correct version is returned - part of the testing when a version increment happens, I think :) Not sure if it matters that the python test also checks or if we're just checking that the method works. More just noting this here cause I saw it and I'm not sure which is better.

nickerso (approved):
Apart from the comment on the version test, this looks good to me.

tests/bindings/python/test_version.py
MichaelClerx: In general* I've only tested if the method can be called without error, and if it accepts/returns the correct types. In other words, I've only checked that the method (seems to) map to the correct C++ function, leaving the testing of the underlying C++ code for the C++ tests!

*with a few exceptions here and there, for example, if I set a name I check if I can then get the name. I think a small bit of this kind of testing is necessary to check that the Python bindings actually have an effect (or if they're doing nothing but not raising errors)

nickerso: Sounds good.

agarny (approved):

agarny (approved):

PR #231, opened 2018-06-18
BUGFIX: CMake 3.11 changed the directory lookup.

rblake-llnl:
This broke the build.

Fixed as per @hsorby

### DIFF
--- a/src/bindings/python/CMakeLists.txt
+++ b/src/bindings/python/CMakeLists.txt
@@ -47,7 +47,7 @@ set(SWIG_INTERFACE_SRCS
 set_source_files_properties(${SWIG_INTERFACE_SRCS} PROPERTIES CPLUSPLUS ON)
 
 # Set output path (relative to current in build dir)
-set(CMAKE_SWIG_OUTDIR ${CMAKE_CFG_INTDIR}/libcellml)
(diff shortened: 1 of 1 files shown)


hsorby:
I think that we are going to have to merge this as our MacOS build slave is now using CMake 3.11.  This means that no knew code can be merged as the CI will always fail.

agarny:
Ok, happy for you, @hsorby, to merge this PR then.

hsorby:
I think @MichaelClerx or @nickerso will have to do it I am powerless in this regard.

agarny:
Ok, I am going to do it then. :)

MichaelClerx:
Uhm, wasn't the build failing?

agarny:
@MichaelClerx, yes, but my understanding from what @hsorby said is that it cannot pass because "the CI will always fail [on macOS]."

MichaelClerx:
Ok. I think we should start using shorter PRs (like this one) for any cmake or configuration changes, merge them with `develop` and then rebase the long-term branches like `resets` onto develop occasionally so that all big branches can benefit from all fixes!
Now we've got a weird situation with requiring merges to be made in a specific order etc., not really getting the benefit of git!

Issue #215, opened 2017-10-03
Missing method `Model.hasUnits(UnitsPtr units)`


MichaelClerx:
UPDATE: Title changed from `Variable.hasUnits` to `Model.hasUnits`

We only have a `hasUnits(string name)`. Other `hasX()` and `containsX()` methods accept either a name or an object.

hsorby:
Units are tricky I have to constantly argue with @nickerso to make sure I get it right.

agarny:
Units can indeed be a pain, still... I would be interested to know what is, here, the argument for not having a version of `hasUnits()` with an object as an argument...

nickerso:
Just to be clear, this is actually talking about `Model::hasVariable()`, right? I'm not sure there is any reason to not have the matching object argument method...probably just an oversight in the initial implementation.

MichaelClerx:
You're right! I did mean `Model::`, not `Variable::` 

as for the rest of the correction... :D

agarny:
Closing as a result of PR #235.

PR #235, opened 2018-06-23
Add Model::hasUnits(UnitsPtr) method

nickerso:
Addresses #215.


agarny:
Looks good to me. However, shouldn't something be done about Python bindings for this?

nickerso:
yep, python bindings should be done to match. But I'll need to work out how or someone else can help... :)

MichaelClerx:
Python bindings worked automagically! But have added a test so ready for review

Issue #236, opened 2018-06-23
CellML namespace: make sure that a model uses the CellML 2.0 namespace


agarny:
The [tests/resources/sine_approximations.xml](https://github.com/cellml/libcellml/blob/develop/tests/resources/sine_approximations.xml) file currently uses the CellML 1.2 namespace while it ought to use the CellML 2.0 namespace.

On a side note: does the CellML namespace get checked as part of the validation process?

hsorby:
No I don't think it does.

agarny:
No to which part of my comment and why?

hsorby:
You only asked one question, does the validation check the namespace? No I don't think it does.  Probably because I didn't think of checking it, the XML checks to make sure the XML is valid the namespace is just some text, libcellml doesn't confirm whether the namespace is the correct one.  Just need to add a test to show the behaviour that we want and if it doesn't pass make the required changes.

agarny:
Ok, I will look into it.

nickerso:
Closing as per discussion on #237.

PR #237, opened 2018-06-25
CellML namespace: make sure that a model uses the CellML 2.0 namespace (#236)

agarny:
This addresses issue #236.

Note that I didn't find any Python-related tests for things like `SpecificationRule::MODEL_ELEMENT`, so I didn't create any for the CellML namespace. However, there are some tests for C++.


agarny:
@hsorby, I wanted to add you as a reviewer, but I couldn't (since you are not yet a member of the [list of CellML editors](https://github.com/orgs/cellml/teams/editors)). Still, it would be nice if you could review this PR.

agarny:
@hsorby, I wanted to add you as a reviewer, but I couldn't (since you are not yet a member of the [list of CellML editors](https://github.com/orgs/cellml/teams/editors)). Still, it would be nice if you could review this PR.

hsorby (changes requested):
The changes look good but I think you should also add a positive test on one of the examples showing that it works for namespaces that are correct.  You can make use of one of the existing examples and just add a quick test that the namespace is correct.

agarny:
Good point, consider it done.

agarny:
Actually, there are already several places where the namespace is indirectly checked. Indeed, if the parser checks for the namespace to be `http://www.cellml.org/cellml/1.2#` rather than `http://www.cellml.org/cellml/2.0#`, then we get the following test output:

```
Running tests...
Test project /Users/Alan/libCellML/build
      Start  1: entities_unit_test_component
 1/32 Test  #1: entities_unit_test_component .....***Failed    0.02 sec
      Start  2: entities_unit_test_connection
 2/32 Test  #2: entities_unit_test_connection ....***Exception: SegFault  0.01 sec
      Start  3: misc_unit_test_coverage
 3/32 Test  #3: misc_unit_test_coverage ..........   Passed    0.02 sec
      Start  4: object_unit_test_error
 4/32 Test  #4: object_unit_test_error ...........   Passed    0.02 sec
      Start  5: entities_unit_test_math
 5/32 Test  #5: entities_unit_test_math ..........***Failed    0.02 sec
      Start  6: entities_unit_test_model
 6/32 Test  #6: entities_unit_test_model .........***Exception: SegFault  0.01 sec
      Start  7: io_unit_test_parser
 7/32 Test  #7: io_unit_test_parser ..............***Exception: SegFault  0.01 sec
      Start  8: io_unit_test_printer
 8/32 Test  #8: io_unit_test_printer .............   Passed    0.02 sec
      Start  9: entities_unit_test_reset
 9/32 Test  #9: entities_unit_test_reset .........   Passed    0.02 sec
      Start 10: entities_unit_test_units
10/32 Test #10: entities_unit_test_units .........***Failed    0.03 sec
      Start 11: io_unit_test_validator
11/32 Test #11: io_unit_test_validator ...........***Failed    0.39 sec
      Start 12: entities_unit_test_variable
12/32 Test #12: entities_unit_test_variable ......***Failed    0.02 sec
      Start 13: api_unit_test_version
13/32 Test #13: api_unit_test_version ............   Passed    0.02 sec
      Start 14: entities_unit_test_when
14/32 Test #14: entities_unit_test_when ..........   Passed    0.02 sec
      Start 15: python_test_component
15/32 Test #15: python_test_component ............   Passed    0.09 sec
      Start 16: python_test_component_entity
16/32 Test #16: python_test_component_entity .....   Passed    0.09 sec
      Start 17: python_test_entity
17/32 Test #17: python_test_entity ...............   Passed    0.09 sec
      Start 18: python_test_error
18/32 Test #18: python_test_error ................   Passed    0.09 sec
      Start 19: python_test_import_source
19/32 Test #19: python_test_import_source ........   Passed    0.09 sec
      Start 20: python_test_imported_entity
20/32 Test #20: python_test_imported_entity ......   Passed    0.09 sec
      Start 21: python_test_logger
21/32 Test #21: python_test_logger ...............   Passed    0.09 sec
      Start 22: python_test_model
22/32 Test #22: python_test_model ................   Passed    0.09 sec
      Start 23: python_test_named_entity
23/32 Test #23: python_test_named_entity .........   Passed    0.09 sec
      Start 24: python_test_ordered_entity
24/32 Test #24: python_test_ordered_entity .......   Passed    0.09 sec
      Start 25: python_test_parser
25/32 Test #25: python_test_parser ...............   Passed    0.09 sec
      Start 26: python_test_printer
26/32 Test #26: python_test_printer ..............   Passed    0.10 sec
      Start 27: python_test_reset
27/32 Test #27: python_test_reset ................   Passed    0.10 sec
      Start 28: python_test_units
28/32 Test #28: python_test_units ................   Passed    0.10 sec
      Start 29: python_test_variable
29/32 Test #29: python_test_variable .............   Passed    0.10 sec
      Start 30: python_test_validator
30/32 Test #30: python_test_validator ............   Passed    0.09 sec
      Start 31: python_test_version
31/32 Test #31: python_test_version ..............   Passed    0.09 sec
      Start 32: python_test_when
32/32 Test #32: python_test_when .................   Passed    0.10 sec

75% tests passed, 8 tests failed out of 32

Total Test time (real) =   2.35 sec

The following tests FAILED:
	  1 - entities_unit_test_component (Failed)
	  2 - entities_unit_test_connection (SEGFAULT)
	  5 - entities_unit_test_math (Failed)
	  6 - entities_unit_test_model (SEGFAULT)
	  7 - io_unit_test_parser (SEGFAULT)
	 10 - entities_unit_test_units (Failed)
	 11 - io_unit_test_validator (Failed)
	 12 - entities_unit_test_variable (Failed)
Errors while running CTest
make: *** [test] Error 8
```

So, I reckon that ought to be enough?

hsorby:
Yup, ought to be.

MichaelClerx (commented):
Why isn't this picked up by the schema validation?
A `<model>` element from a different namespace should be just as invalid as a `<michael>` element

MichaelClerx (changes requested):
Sorry I think this is the wrong way to fix this. It's now written as

1. Get the model
2. Check it's namespace

But that shouldn't be possible. The namespace is part of what defines a model element. It should be simply

1. Get the CellML 2.0 model

agarny:
@MichaelClerx, how would you go about doing what you are saying?

MichaelClerx:
I don't really understand what's going on here!

- We're calling a validation method on our object model. Should that even have a namespace? Surely it's only the XML element that has a namespace, not the C++ model instance?
- If we're validating XML (which I thought we were going to do with a schema), then _all_ elements need to be in the namespace. We don't need to check this explicitly, it's just part of being a valid CellML2.0 XML file that the elements are in the correct namespace

What am I missing?

agarny:
@hsorby will no doubt correct me if I am getting this wrong, but:
> - We're calling a validation method on our object model. Should that even have a namespace? Surely it's only the XML element that has a namespace, not the C++ model instance?

The validation method doesn't check the namespace. The only place where the namespace is checked is when you load a CellML file (see `Parser::ParserImpl::loadModel`). So, yes, it's only the XML element that has a namespace, not the C++ model instance.

> - If we're validating XML (which I thought we were going to do with a schema), then _all_ elements need to be in the namespace. We don't need to check this explicitly, it's just part of being a valid CellML2.0 XML file that the elements are in the correct namespace

As far as I can tell, we are not validating XML, but the C++ model instance. (Now, I must confess that I can't recall what we said/agreed about using a schema to validate XML...)

MichaelClerx:
@nickerso thoughts?

nickerso:
In general, we are not doing a very good job with handling namespaces. See [here](https://github.com/cellml/libcellml/blob/develop/src/parser.cpp#L397) for example.

If we want to start checking namespaces when parsing an XML file, then we should be doing it everywhere - just cause the model element is in the CellML 2.0 namespace doesn't mean all children elements are. So it would be better to simply update those `getType` methods to be namespace aware. As @MichaelClerx said, its either a CellML 2.0 element or not, I don't see any advantage in adding more `SpecificationRule` entries in this case. If we want to go down this route, I would be doing something like
``` 
node->isType("http://www.cellml.org/cellml/2.0#model")
```
and the `getType()` method would return a similarly namespace'd type to make the error message more informative if needed.

We haven't yet defined an XML Schema for CellML 2.0, but once we have that it should be trivial to add a check against that to libCellML. Users can then make use of it if they choose to prior to parsing an XML file.

agarny:
This PR checks that the root node is a `model` element that declares a default CellML 2.0 namespace. My understanding is that if it is the case, then the `model` element is in the CellML 2.0 namespace, and so are all of its child elements (but not their attributes). That is, unless a child element is prefixed, which I wouldn't expect, although I agree that this could theoretically happen, hence we should indeed check that every element is a CellML 2.0 element.

Now, I must confess that my motivation behind this PR is so that I can determine whether I am (supposed to be) dealing with a CellML 1.0, 1.1 or 2.0 file...

MichaelClerx:
@agarny It's a valid concern! I'd like it too. But like you said, children could still be prefixed. It's also possible for children to change the default namespace (putting all their children in that default namespace too). So we really need to check this more thoroughly.
@nickerso Do we have a CellML 1.0 / 1.1 schema? I could have a go at writing a 2.0 one unless you know a reason why we can't (easily) do this...

agarny:
@MichaelClerx: I have never used it myself, but there is https://www.cellml.org/tools/cellml_1_1_schema.

MichaelClerx:
Oh yeah :D 
LibXML should be able to deal with that, right? And we were planning on using extended XML stuff anyway to read CellML1.0 and 1.1 (xslt)?

agarny:
Yes, I believe schema validation using libxml2 should be possible.

I am not sure there have ever been plans to read CellML 1.0/1.1 models with libCellML (even after internal XSL transformation). I believe the plan is just to read CellML 2.0 models with libCellML, which means that existing CellML 1.0/1.1 will have to be converted to CellML 2.0, prior to being read by libCellML.

@hsorby has written a small program to do that using the [XSL file](https://github.com/cellml/cellml1to2/blob/master/cellml1to2.xsl) @jonc125 wrote a couple of years ago. (I also have an OpenCOR branch that does the same.)

nickerso:
I have no problem checking elements are in the CellML 2.0 namespace when parsing an XML file - but if we want to check namespaces it needs to be done for all elements, not just the model. So for this PR to be acceptable (to me), it would need to check all elements (something like my suggestion to update the `XmlNode` type methods to work with fully qualified element names seems like the way to go to me, but perhaps there are other ways).

I'm also happy to not check namespaces at all when parsing and assume that at some point in the future we would be able to do XML Schema validation to check files before parsing them.

jonc125:
I agree with @nickerso here. The intention is that libCellML only handles CellML 2.0 (or newer!) so you can assume when using it you have a CellML 2.0 model. Validation of this should be handled by a schema during parsing, and only checks that can't be done by a schema should appear in libCellML.

agarny:
I agree that only checking the `model` element is in the CellML 2.0 namespace is not enough. To check all the other elements wouldn't be good enough either. We should also check all their attributes.

So, in the end, I agree with @jonc125 that validation "should be handled by a schema during parsing, and only checks that can't be done by a schema should appear in libCellML".

Ok, I am going to close this PR and create a new one just for `tests/resources/sine_approximations.xml` so that it declares the CellML 2.0 namespace rather than the CellML 1.2 namespace.

MichaelClerx:
> We should also check all their attributes.

Yes! But remember they're not in any namespace (the null namespace)

agarny:
> Yes! But remember they're not in any namespace (the null namespace)

Indeed, but is that what we would want? I honestly don't know. Anyway, I imagine that such a check could be part of the CellML 2.0 schema.

MichaelClerx:
Yup! Attributes nearly always live in the null namespace. If you want anything else you have to write `agarny:name="value"` everywhere. So almost everyone sticks attributes in null and this is best practice

Issue #242, opened 2018-06-28
Windows: rename the `cellml.lib`/`cellml.dll` files to `libcellml.lib`/`libcellml.dll`


agarny:
On Windows, we are currently generating a `cellml.dll` file (as well as the corresponding `cellml.lib` file). However, this is also what the CellML API does (as well as generate other `.lib`/`.dll` files), which means that we cannot use both the CellML API and `libCellML` in the same project (yes, some people want or, rather, need to do that... :)).

So, since the library is called `libCellML`, we probably ought to generate a `libcellml.dll` file, and thus avoid conflicts with the CellML API (as well as be in sync with `libSBML`, `libNuML` and `libSEDML` which generate a `libsbml.dll`, `libnuml.dll` and `libsedml.dll` file, respectively).

PR #243, opened 2018-06-28
Windows: renamed the `cellml.lib`/`cellml.dll` files to `libcellml.lib`/`libcellml.dll`

agarny:
Addresses issue #242.


agarny:
Ping @hsorby for review.

hsorby:
I don't think this is the way to go.  We should rename the library output name on windows and not change the target name.

hsorby (changes requested):
This change has further implications for the CMake generated config file.  I don't think this is the way to go.  We should look at renaming the library on windows and leave the rest alone.

I will try to look at this on Monday when I am at my desk in front of a Windows computer.

nickerso (changes requested):
I agree with @hsorby, better to tweak the target output file name if needed and keep the actual target as is. This seems to show how its done: https://stackoverflow.com/questions/31038963/how-do-you-rename-a-library-filename-in-cmake

agarny:
I stand corrected, it's indeed much easier to have a simple call to `set_target_properties(cellml PROPERTIES PREFIX "lib")` on Windows. Thanks @nickerso for the pointer.

agarny:
Ping

hsorby (commented):
If we are going to add the ability to have both a static library and a dynamic library being built at the same time we would need to adjust the name of either the dll import library or the static library.  As on windows these have the same name by default.

We could set the IMPORT_SUFFIX to ".dll.lib" to differentiate the libraries.

As this stands if we built a static library we would not be able to differentiate it from the dll import library.

One would overwrite the other if they were installed to the same place.

agarny:
> If we are going to add the ability to have both a static library and a dynamic library being built at the same time we would need to adjust the name of either the dll import library or the static library.  As on windows these have the same name by default.

Do we really want to bother with this? I mean, in 99.9% of cases, people will want to build one or the other, which the current PR supports. Not to mention that, as far as I can tell, the `develop` branch is no different from this PR in this respect. So, I don't think that the possibility of adding this ability should stop the current PR from being merged.

hsorby:
If we get in the game of supplying binaries that include a static library then this will have to be done.  In part this is a note to make sure this is done if or when we require it.  I don't see it that it blocks the merging of this PR.

agarny:
Good, so I take it that @nickerso can merge this PR then.

nickerso (approved):

Issue #244, opened 2018-06-29
Windows: MSVC may define `PASCAL`, which conflicts with our `StandardUnit` enum


agarny:
Just got bitten by this on Windows using Qt/C++ with MSVC 2015.

At some point, OpenCOR indirectly includes `C:\Program Files (x86)\Windows Kits\8.1\Include\shared\minwindef.h`, which defines `PASCAL` as `__stdcall`, resulting in the following compilation errors:

```
C:\Users\Alan\Desktop\OpenCOR\src\plugins\thirdParty\libCellML\ext\debug\include\libcellml/units.h(75): error C2059: syntax error: '__cdecl'
C:\Users\Alan\Desktop\OpenCOR\src\plugins\thirdParty\libCellML\ext\debug\include\libcellml/units.h(75): error C3805: '__cdecl': unexpected token, expected either '}' or a ','
```

There is an easy way around it, which is to `#undef PASCAL` just before including `libcellml/units.h` (and `#define PASCAL __stdcall` just after, to be on the safe side).

Still, we might want to do something about it?...

nickerso:
Looks like it can just be `#undef`'d and not brought back... https://stackoverflow.com/questions/2774171/what-is-far-pascal

agarny:
Agreed, so we might want to `#undef PASCAL` in `libcellml/units.h`, so people don't come across the same issue and waste time figuring out why on earth `PASCAL` caused MSVC to complain.

PR #245, opened 2018-06-29
Make sure that `PASCAL` is not already defined

agarny:
Addresses issue #244.


agarny:
Ping @hsorby for review.

hsorby:
Looks good to me. But why the space between the comment and the #undef?

MichaelClerx (approved):

agarny:
> Looks good to me. But why the space between the comment and the #undef?

Because that's how I tend to write my code and it didn't think twice about it, but I am happy to remove it if it doesn't "fit" `libCellML`'s style.

Issue #238, opened 2018-06-26
Resolve local imports.


hsorby:
Whilst resolving any import is quite a big chunk of work enabling support for importing resources from the local disk is not.  If we provide support for local imports we can start to encourage modellers to create modular models that will (hopefully) in the future be easily retrievable from anywhere.

Related to issue #166.

nickerso:
This seems like a good starting point.

PR #248, opened 2018-07-17
Imports revisited

hsorby:
Moving @nickerso work from outside the library to inside.  The added import functionality only works for local on disk imports.

Addresses #238.

This PR supersedes #167, should this PR be merged then #167 should be closed and not merged.


src/api/libcellml/import.h
```diff
@@ -0,0 +1,95 @@
...
+    Import& operator=(Import m); /**< Assignment operator */
+
+    /**
+     * @brief Set the source Model's URL.
+     *
+     * Set the source Model's URL that this import refers to.
+     *
+     * @param reference The referenced model's URL.
+     */
+    void setSource(const std::string &reference);
```
nickerso: `reference` argument should be `source`

src/model.cpp
```diff
@@ -214,4 +218,133 @@ size_t Model::unitsCount() const
...
+{
+    if (importedEntity->isImport()) {
+        libcellml::ImportSourcePtr importSource = importedEntity->getImportSource();
+        if (!importSource->hasModel()) {
+            std::string url = resolvePath(importSource->getUrl(), baseFile);
+            std::ifstream t(url);
+            std::stringstream buffer;
+            buffer << t.rdbuf();
+            libcellml::Parser p;
+            libcellml::ModelPtr model = p.parseModel(buffer.str());
```
nickerso: this will always create a model, even when you have an source URL that is not resolvable. This gives the appearance that an import is resolved when it actually isn't. Need to decide how to fix this.

nickerso: actually, the error should be caught before parsing...so there is no problem here.

src/model.cpp
```diff
@@ -214,4 +218,133 @@ size_t Model::unitsCount() const
...
+}
+
+void resolveImport(ImportedEntityPtr importedEntity,
+                   const std::string& baseFile)
+{
+    if (importedEntity->isImport()) {
+        libcellml::ImportSourcePtr importSource = importedEntity->getImportSource();
+        if (!importSource->hasModel()) {
+            std::string url = resolvePath(importSource->getUrl(), baseFile);
+            std::ifstream t(url);
```
nickerso: we need to check that this succeeds before continuing?

nickerso (approved):
This looks like a great start to get the import resolution working for local files. The current tests cover the common use cases, can be refined as we move forward.

src/model.cpp
nickerso: This is now checked.

src/model.cpp
```diff
@@ -219,4 +222,145 @@ size_t Model::unitsCount() const
...
+            if (t.good()) {
+                std::stringstream buffer;
+                buffer << t.rdbuf();
+                libcellml::Parser p;
+                libcellml::ModelPtr model = p.parseModel(buffer.str());
+                if (model) {
```
agarny: From what I can see in `Parser::parseModel()`, we will always have a model, so no point in testing for it?

tests/resolve_imports/file_parser.cpp
```diff
@@ -0,0 +1,124 @@
...
+    libcellml::ModelPtr model = p.parseModel(buffer.str());
+
+    EXPECT_EQ(0u, p.errorCount());
+
+    EXPECT_TRUE(model->hasUnresolvedImports());
+    model->resolveImports(modelLocation);
+    EXPECT_FALSE(model->hasUnresolvedImports());
+}
+
+TEST(ResolveImports, resolveImportsFromFileLevel0Unresolvable) {
```
agarny, line 109: It would be nice if `libCellML` was to provide some information about why imports couldn't be resolved.

agarny:
Apart from a couple of comments that I made above, this looks good to me.

Note that I cleaned up the code a bit. I initially intended to create a new branch from [hsorby:imports_revisited](https://github.com/hsorby/libcellml/tree/imports_revisited), but forgot about it and ended up pushing my changes directly to [hsorby:imports_revisited](https://github.com/hsorby/libcellml/tree/imports_revisited). Sorry about that, my bad!

And, now, I can see that I broke things! Argh!

nickerso:
The introduction of the `enable_share_from_this` is giving me the below warning with VS 2017 every time `model.h` is included.

```
3>c:\users\...\libcellml\src\api\libcellml\model.h(36): warning C4251: 'std::enable_shared_from_this<libcellml::Model>::_Wptr': class 'std::weak_ptr<_Ty>' needs to have dll-interface to be used by clients of class 'std::enable_shared_from_this<libcellml::Model>'
3>        with
3>        [
3>            _Ty=libcellml::Model
3>        ]
3>c:\program files (x86)\microsoft visual studio\2017\community\vc\tools\msvc\14.14.26428\include\memory(2000): note: see declaration of 'std::weak_ptr<_Ty>'
3>        with
3>        [
3>            _Ty=libcellml::Model
3>        ]
```

tests/resolve_imports/file_parser.cpp
nickerso: It would, but I think that is the job here. We need to extend the validator to do more thorough validation through the import hierarchy on demand. I think that is the way to go, keep this functionality just for resolving the imports that can be resolved.

agarny: Ok, agreed.

src/model.cpp
```diff
@@ -219,4 +233,129 @@ size_t Model::unitsCount() const
...
+            std::string url = resolvePath(importSource->getUrl(), baseFile);
+            std::ifstream file(url);
+            if (file.good()) {
+                std::stringstream buffer;
+                buffer << file.rdbuf();
+                libcellml::Parser parser;
+                libcellml::ModelPtr model = parser.parseModel(buffer.str());
+                if (model) {
+                    importSource->setModel(model);
+                    model->resolveImports(url);
```
nickerso: Won't this cause all imports in the imported model to be resolved, regardless of whether they are needed?

nickerso: OK, so we always try to resolve all imports in all imported models, but the `hasUnresolvedImports` method only checks the required imports are resolved. I guess this is more an efficiency thing that an actual problem with the code.

agarny: My point was that, from the code in `Parser::parseModel()`, `model` will never be equal to `nullptr`, so no point for the test.

agarny:
> The introduction of the `enable_share_from_this` is giving me the below warning with VS 2017 every time `model.h` is included.

Could you try to add the following and see whether it makes any difference with VS 2017?

```
...

#include "libcellml/componententity.h"

#ifndef SWIG
template class LIBCELLML_EXPORT std::enable_shared_from_this<libcellml::Model>;
#endif

//! Everything in libCellML is in this namespace.
namespace libcellml {

...
```

src/model.cpp
nickerso: yep - the `if (model)` can be removed.

agarny:
Ok, my bad, I should have read your warning message properly. My idea was correct, except that it should have read:

```
#ifndef SWIG
template class LIBCELLML_EXPORT std::weak_ptr<libcellml::Model>;
#endif
```

So, now, everything should be fine (it is using my local copy of VS 2017).

nickerso (approved):
Phew! Everything is back working with no warnings for me. I'm happy with the changes Alan has made, but would be good for someone with a bit more C++ expertise to confirm the new approach for import recursion makes sense.

MichaelClerx:
Happy to look at the bindings bit for this once you're sure the interface is final!

nickerso:
pretty sure the interface is final, so would be good to look at the bindings.

MichaelClerx (approved):
Happy to merge this now, from a bindings point of view

src/api/libcellml/model.h
```diff
@@ -224,6 +231,26 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
      */
     size_t unitsCount() const;
 
+    /**
+     * @brief Resolve all imports in this model.
+     *
+     * Resolve all @c Component and @c Units imports by loading the models
+     * from local disk through relative urls.  The @p baseFile is used to determine
```
agarny: Might want to replace "urls" with "URLs" to be consistent with the rest of our documentation.

tests/bindings/python/test_error.py
```diff
@@ -269,6 +269,46 @@ def test_get_variable(self):
         self.assertIsInstance(e.getVariable(), Variable)
         self.assertEqual(e.getVariable().getName(), name)
 
+    def test_set_reset(self):
+        from libcellml import Error, Reset
+
+        # void setReset(const ResetPtr &reset);
+        e = Error()
+        e.setReset(Reset())
```
agarny, line 277: Definitely not a blocker, but not sure there is much point in having such a test (and similar ones) since `setReset()` is effectively tested as part of `test_get_reset()`.

tests/bindings/python/test_model.py
```diff
@@ -225,6 +225,24 @@ def test_units_count(self):
...
+        m.addComponent(c)
+        self.assertFalse(m.hasUnresolvedImports())
+        c.setImportSource(ImportSource())
+        self.assertTrue(m.hasUnresolvedImports())
+
+    def test_resolve_imports(self):
+        from libcellml import Model
+
+        m = Model()
+        m.resolveImports('file.txt')
```
agarny, line 244: No tests as such?

agarny:
Apart from a few comments above, it all looks good to me.

tests/bindings/python/test_error.py
MichaelClerx: I think you're right, but I just wanted to be systematic in having a test for every method!

src/api/libcellml/model.h
MichaelClerx: On that note, these methods `setUrl` and `getUrl` and such, should they be capitalised this way? Or `getURL` etc.?
(Personally, my left shoulder likes using the shift key as little as possible)

tests/bindings/python/test_model.py
MichaelClerx: The actual functionality is tested by the C++ tests. Here I just want to check that the method exists and doesn't raise any exceptions!

src/api/libcellml/model.h
agarny: No, `setUrl()` and `getUrl()` should be left as-is. Indeed, we use lower camel case for class methods (see http://libcellml.readthedocs.io/en/latest/dev_coding_standard.html).

tests/bindings/python/test_model.py
agarny: Fair enough.

Issue #249, opened 2018-07-17
gtest 1.7.0 creating warnings with VS 2017


hsorby:
VS 2017 is warning about the use of tr1 namespace from strings.  Other compilers will probably follow suit in this regard at some point, updating gtest to 1.8.0 should resolve this problem.

PR #250, opened 2018-07-17
Gtest upgrade

hsorby:
Upgrading gtest to version 1.8.0 to deal with VS 2017 generating warnings over 1.7.0.

Addresses #249.

PR #251, opened 2018-07-19
Upgraded gtest to version 1.8.0

agarny:
Upgrading gtest to version 1.8.0 to deal with VS 2017 generating warnings over 1.7.0.

Addresses #249.


nickerso (changes requested):
Still getting the tr1 namespace warnings in VS 2017. I'd also prefer to keep the versioned folder name...

agarny:
FTR, this PR uses [gtest 1.8.0](https://github.com/google/googletest/releases/tag/release-1.8.0) while PR #250 was using [gtest's master branch](https://github.com/google/googletest).

Ideally, we would be waiting for gtest 1.9.0 to be released, but... gtest 1.8.0 was released in August 2016 and, prior to that, gtest 1.7.0 was released in September 2013. So, not quite sure when we can expect gtest 1.9.0 to be released...

So... do we want to drop support for VS 2017, use gtest's master branch or... do something else?

nickerso:
the gtest in #250 worked fine with VS 2017 and all tests passed on our build systems, so I think we should go to the current master branch.

hsorby (approved):
Works for me, I think we'll let @nickerso test on his windows machine before merging this.

nickerso (approved):
This is working for me and clears up #249. Tests failing on mac due to known networking issues, will keep trying to get it through to confirm its all good to merge.

nickerso:
Since we have moved to the current master version of gtest, I'm fine to remove the version string from the folder name.

PR #256, opened 2018-10-18
Code generation functionalities

massimiliano-leoni:
I added a main class and several auxiliary classes to support generation of code to be used in solving ODEs.

Addresses issue #257 .


massimiliano-leoni:
I'm having issues making MSVC digest the code, can anybody have a look?

agarny:
> I'm having issues making MSVC digest the code, can anybody have a look?

Ok, will have a look at after our long weekend.

massimiliano-leoni:
> Ok, will have a look at after our long weekend.

Thank you!

agarny:
Ok, I have addressed the various Windows issues that were still pending. The idea was just to get things to pass, nothing more nothing less.

massimiliano-leoni:
Thank you very much Alan!

src/api/libcellml/generator.h
```diff
@@ -0,0 +1,121 @@
...
+#include "libcellml/model.h"
+#include "libcellml/operatorlibrary.h"
+#include "libcellml/variable.h"
+#include "../xmldoc.h"
+#include "../xmlnode.h"
+
+//! Everything in libCellML is in this namespace.
+namespace libcellml {
+
+#ifdef _MSC_VER
```
MichaelClerx: Stuff like these really needs documentation! What does this do? What is warning 4275 and why should it be disabled

src/api/libcellml/operatorlibrary.h
```diff
@@ -0,0 +1,156 @@
+#include <iomanip>
+#include <memory>
+#include <sstream>
+#include <string>
+
+namespace libcellml {
+
+namespace operators {
+
+class Representable
```
MichaelClerx: Great work! Really happy to see this added. Will need lots of documentation though! (Not everything makes immediate sense to me without it: What's a Representable? What's a STDOperator?)

src/api/libcellml/generator.h
agarny: See https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4275 for warning C4275. Basically, "C4275 can be ignored in Visual C++ if you are deriving from a type in the C++ Standard Library, compiling a debug release (/MTd) and where the compiler error message refers to _Container_base." So, I think we can ignore that warning here, not to mention that such a warning doesn't exist on Linux/macOS.

Anyway, at this stage, the idea is really just to make sure that all the tests pass. Nothing more nothing less. There is still a lot work to be done before this PR gets merged in. At this point, we should concentrate on coming up with a good suite of tests for code generation.

src/api/libcellml/operatorlibrary.h
massimiliano-leoni: Thank you :) The documentation will be added soon.
`Representable` is an umbrella class for anything that can be represented by the code generator. Think of it as a "Generatable" class. On C++ side, it's anything that has a `repr` method, which is the code representation of an object.
`STDOperator` is anything whose representation is in the form `std::xxx`, so anything that is provided by the C++ standard library.

src/api/libcellml/generator.h
```diff
@@ -0,0 +1,121 @@
...
+class LIBCELLML_EXPORT Generator : public Logger
+{
+public:
+    Generator();
+    virtual ~Generator();
+
+    std::string generateCode(ModelPtr m);
+    void writeCodeToFile(std::string filename);
+
+private:
```
hsorby, line 35: When using the Pimpl idiom we shouldn't have any private methods visible in a public API header file.  These private methods should all be moved into the pimpl class in the .cpp file.

src/api/libcellml/operatorlibrary.h
hsorby: We definitely need to add the documentation to these classes.

hsorby (changes requested):
The first and most obvious aspect of this submission to fix is moving the private methods into the .cpp file and making use of the pimpl idiom.  This should fix other problems with the code.

massimiliano-leoni:
MSVC is complaining again. @agarny : maybe we need some more of your `pragma magic`?

src/api/libcellml/generator.h
massimiliano-leoni: I think I solved this.

src/api/libcellml/operatorlibrary.h
massimiliano-leoni: Some documentation is now in place.

agarny:
> MSVC is complaining again. @agarny : maybe need some more of your `pragma magic`?

If I remember correctly, some of the MSVC issues we had were related to your use of templates. Personally, I would simply not use templates. Not only I don't think they are critical here, but they are clearly causing us unnecessary cross-platform issues.

hsorby:
Further to @agarny's comment on templates, I think we might have difficulty applying SWIG on the way the templates are being used.  I have only had a cursory look but I think it may be quite difficult to wrap the generator class as it currently stands.

Could we modify the `template <typename L = CXX>
        std::string generateCode(ModelPtr model);` code to be  `std::string generateCode(ModelPtr m, TARGET_LANGUAGE_ENUM targetLanguage);`

and then only use the templates internally in the .cpp file?

hsorby (changes requested):
The documentation looks good.  Do we need to document `class Addition` etc.?  Do they self document or can this be considered a tidy up issue.

I think we can move the operatorlibrary.h header file to an internal spot.  This header file I think doesn't need to be public, certainly not at this point.

agarny:
> Could we modify the `template <typename L = CXX>
        std::string generateCode(ModelPtr model);` code to be  `std::string generateCode(ModelPtr m, TARGET_LANGUAGE_ENUM targetLanguage);`
> 
> and then only use the templates internally in the .cpp file?

If we are to change the signature of `generateCode()` to help things with SWIG, then we might as well get rid of templates altogether.

hsorby:
> Sorry I didn't proofread the documentation before I committed :(
> By the way, the word was actually meant to be "clash" :D

Even if I do proofread I still have typos. Nothing to worry about, we just have a fix them as we see them approach.

hsorby:
> > Could we modify the `template <typename L = CXX>         std::string generateCode(ModelPtr model);` code to be  `std::string generateCode(ModelPtr m, TARGET_LANGUAGE_ENUM targetLanguage);`
> > and then only use the templates internally in the .cpp file?
> 
> If we are to change the signature of `generateCode()` to help things with SWIG, then we might as well get rid of templates altogether.

Yes, but in this case we can maybe just copy and paste the code out of the header file and into the cpp and worry about replacing the templates later.

agarny:
> > If we are to change the signature of `generateCode()` to help things with SWIG, then we might as well get rid of templates altogether.
> 
> Yes, but in this case we can maybe just copy and paste the code out of the header file and into the cpp and worry about replacing the templates later.

Sure, it's just that I was also thinking about the different templates-related warnings (and therefore errors) that we may get on Windows.

agarny:
Ok, this was long overdue (sorry for that!), but I have come up with some test CellML files, which you can find [here](https://github.com/agarny/test-cellml-files). As you will see, they are based on [`test_simple_odes.cellml`](https://github.com/ModellingWebLab/cellmlmanip/blob/develop/tests/cellml_files/test_simple_odes.cellml) from the [`ModellingWebLab`](https://github.com/ModellingWebLab)/[`cellmlmanip`](https://github.com/ModellingWebLab/cellmlmanip) GitHub repository. There are two folders: [`cellml`](https://github.com/agarny/test-cellml-files/tree/master/cellml) for the CellML 1.0/1.1 version of those test files, and [`cellml-2.0`](https://github.com/agarny/test-cellml-files/tree/master/cellml-2.0) for the corresponding CellML 2.0 version.

Note that I have purposely not provided C code of how I think the generated code should look like. In other words, I think we can use whatever @massimiliano-leoni's code currently generate, and we can see from there (agreed @hsorby?).

Something maybe worth pointing out is that dae_pseudo_v1.cellml and dae_pseudo_v2.cellml are effectively the same models. It’s just that v1 describes:

```
d(x)/d(t) = cos(y)
x(0) = 1
sin(x)-y = 0
```

while v2 describes:

```
d(x)/d(t) = cos(y)
x(0) = 1
y = sin(x)
```

and, using the CellML API, v1 is considered to be a DAE model (hence my use of “pseudo” in the filename) while v2 an ODE model (and rightly so). So, it would be good for libCellML to consider those two models as being identical and, thus, generate the exact same code for the two of them. However, I appreciate that this may require some additional work, so maybe something to keep in mind for a latter stage?

Otherwise, and to finish, those are just some starting test files. No doubt that more will come as we think more about it, but hopefully those ones are already a good starting point. Feel free to comment.

agarny:
Ok, FWIW, I have added a one-component version of the relevant CellML 1.0/1.1 and CellML 2.0 files.

When it comes to this PR, it would be good to have those `*_one_component.cellml` files to work before dealing with the ones that rely on several components (and, therefore, connections).

nickerso:
agreed, it would be great if we can get those one-component versions of the CellML 2.0 test models included into the tests for libcellml and make sure we can generate code for all those cases. Later extension to handle multiple components and connections and imports should then be "trivial" :)

MichaelClerx:
> Ok, FWIW, I have added a one-component version of the relevant CellML 1.0/1.1 and CellML 2.0 files.

Link please!

agarny:
> > Ok, FWIW, I have added a one-component version of the relevant CellML 1.0/1.1 and CellML 2.0 files.
> 
> Link please!

It's [above](https://github.com/cellml/libcellml/pull/256#issuecomment-451797852), but I agree that I should have put it back: https://github.com/agarny/test-cellml-files.

agarny:
Ok, now things pass again. Note that I have:
- updated the code from upstream; and
- removed templating from the code generator.

Regarding the second point, templating was responsible for our tests to fail on Windows. I couldn't fix this within a couple of minutes, so I decided to remove it. This, only so that @massimiliano-leoni can resume his work on code generation ASAP. IOW, although I am not in favour of templating, I could be convinced (assuming it doesn't cause problems with SWIG) that it should be put them back (in which case I would spend a bit more than just two minutes to figure out what is wrong on Windows).

MichaelClerx:
I have had some difficulties with templates in the past. I don't see any massive increases in boilerplate in that final change, so I think the current solution is fine!

MichaelClerx:
Don't forget to update the bindings interface for all these! https://github.com/cellml/libcellml/tree/develop/src/bindings/interface

hsorby (approved):

src/api/libcellml/generator.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <string>
+
+#include "libcellml/logger.h"
+#include "libcellml/types.h"
+
+//! Everything in libCellML is in this namespace.
+namespace libcellml {
+
+class LIBCELLML_EXPORT Generator : public Logger
```
hsorby, line 26: We need to add documentation here.

src/api/libcellml/generator.h
```diff
@@ -0,0 +1,41 @@
...
+{
+public:
+    Generator();
+    virtual ~Generator();
+
+    std::string generateCode(ModelPtr m);
```
hsorby: Documentation here too.

src/api/libcellml/generator.h
```diff
@@ -0,0 +1,41 @@
...
+    void writeCodeToFile(std::string filename);
```
hsorby: Documentation here as well.

hsorby (changes requested):
There is a little bit of documentation missing for our public API generator class.

agarny:
Apologies for the rather long comment, but this is the result of my going through this PR. This is clearly open for comments, so feel free to comment back!
- Looking at what OpenCOR (currently) needs when it comes to ODE models (i.e. not DAE models, at least not at this stage), I would say that we need the following functions to be generated:
   - `void initialize(double *constants, double *states);`
     Initialise constant variables and state variables.
     E.g.
     ```
     constants[1] = 7.0;
     states[3] = 9.0;
     ```
   - `void computeComputedConstants(double *constants);`
     Compute some constant variables that depend on other constant variables and *proper* constants.
     E.g.
     ```
     constants[5] = constants[1]+3.0;
     ```
   - `void computeRates(double voi, double *constants, double *rates, double *states, double *algebraic);`
     Compute the algebraic equations on which the different rates depend, and then compute the rates themselves.
     E.g.
     ```
     algebraic[3] = 0.07*exp(states[0]/20.0);
     algebraic[5] = 1.0/(exp((states[0]+30.0)/10.0)+1.0);
     rates[7] = algebraic[3]*(1.0-states[7])-algebraic[5]*states[3];
     ```
   - `void computeVariables(double voi, double *constants, double *rates, double *states, double *algebraic);`
     Compute the remaining algebraic equations (i.e. those that are not needed to compute the different rates).
     E.g.
     ```
     algebraic[1] = 0.1*(states[0]+25.0)/(exp((states[0]+25.0)/10.0)-1.0);
     ```
- I am not sure whether we should have `Generator::writeCodeToFile()` since our `Printer` class doesn't have a similar method. However, if we really want it then:
   - I would have it return a boolean to tell us whether the writing was successful;
   - I would save the file atomically (i.e. save the generated code to a temporary file and, if successful, rename that temporary file to the requested file name); and
   - I wouldn't have it generate an error if no code has been generated prior to calling `Generator::writeCodeToFile()`. If anything, I would expect `Generator::writeCodeToFile()` to call `Generator::generateCode()` and then write the generated code to file.
- For the `Generator` class, I would expect the following public methods:
   - `std::string generateCode(const ModelPtr &model);`
     A returned empty string would mean that something went wrong, in which case the user could check the error message. Right now, that method returns a non-empty string even if something went wrong, which is not what I would expect.
   - `bool generateCode(const ModelPtr &model, const std::string &filename);`
     That is, if we want a way to generate code and save it to a file, which I am not sure we should have. (If we do then we probably ought to do something similar for the `Printer` class.)
- `Generator::GeneratorImpl::parseMathML()` should eventually take a `ModelPtr` object as a parameter rather than a string containing the math. (I appreciate that this is probably done like this for now because we are only dealing with one component.)
- I would get rid of `Generator::GeneratorImpl::findVoi()` (and its helper, `Generator::GeneratorImpl::findVoiHelper()`) and `Generator::GeneratorImpl::findInitialValues()`. For me, these should be done as part of `Generator::GeneratorImpl::parseMathML()`.
- The generator doesn't account for the fact that we might have more than one variable of integration (actually, is it something that is still allowed in CellML 2.0?).
- The generator could, once we handle several components and imports, end up with the same alias for two variables (from different components) with the same name.
- `Generator::GeneratorImpl::parseNode()` doesn't currently account for the fact that the MathML code may actually be invalid. For example, it accepts the following MathML code:
  ```
      <apply>
          <eq/>
          <apply>
              <diff/>
              <bvar>
                  <ci>t</ci>
              </bvar>
              <ci>y</ci>
          </apply>
          <plus/>
          <ci>y</ci>
          <cn cellml:units="dimensionless">4</cn>
      </apply>
  </math>
  ```
  while it clearly shouldn't since it ought to be:
  ```
      <apply>
          <eq/>
          <apply>
              <diff/>
              <bvar>
                  <ci>t</ci>
              </bvar>
              <ci>y</ci>
          </apply>
          <apply>
              <plus/>
              <ci>y</ci>
              <cn cellml:units="dimensionless">4</cn>
          </apply>
      </apply>
  </math>
  ```
  I agree that a model should probably be valid before code for it gets generated, but...
- The generator relies on the `StdOperator` class, which makes it very C++ centric. I had a quick look at what the code used to do when it was still using templates, and I don't believe templates would make a difference when it comes to that particular class?...
- There are a lot of unneeded parentheses and empty lines in the generated code, which is fine for now but will have to go away at some point. People have, in the past, requested *clean* code.
- The generator doesn't generate equations in the right order.
  E.g.
  ```
      <apply>
          <eq/>
          <apply>
              <diff/>
              <bvar>
                  <ci>t</ci>
              </bvar>
              <ci>y</ci>
          </apply>
          <apply>
              <plus/>
              <ci>y</ci>
              <ci>z</ci>
          </apply>
      </apply>
      <apply>
          <eq/>
          <ci>z</ci>
          <cn cellml:units="dimensionless">50</cn>
      </apply>
  </math>
  ```
  will generate something like:
  ```
  Dy = y+z;
  z = 50;
  ```
  while it should obviously be something like:
  ```
  z = 50;
  Dy = y+z;
  ```
  In fact, right now, the generator will crash trying to generate code to initialise `z`, even though it doesn't need initialising as such.
- We might want, at some point, have an option to enable/disable aliases.
- Rather than generating rate aliases of the form `Dxxx`, we might want to have something like `dXXXdYYY`. For example, for the variable `V` and variable of integration `t`, we would have `dVdt`.
- Not sure that `operatorlibrary.cpp/.h` is a good name for that file since they don't only handle operators as such. In fact, I would be inclined to put all of their contents in `generator.cpp` since they are only used there. This would also prevent us from having a special namespace by renaming `operators::XxxYyy` classes to `GeneratorXxxYyy`.
- I am not a fan of the `Representable`-based class approach. I find it unnecessarily complicated to have to create a new class for every possible type of representable object. FWIW, [here](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIExportEngine.pas#L30) is the generic export engine that I had for [COR](http://opencor.ws/cor/index.html), which I could then specialise for [C](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIToCFileEngine.pas#L29), [C++](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIToCPPFileEngine.pas#L29), [Delphi](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIToDelphiForWin32FileEngine.pas#L29), [Fortran 77](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIToFortran77FileEngine.pas#L29), [Java](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIToJavaFileEngine.pas#L29), [MATLAB](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIToMATLABFileEngine.pas#L29), [Microsoft Word](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIToMSWordFileEngine.pas#L29), [Pascal](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIToPascalFileEngine.pas#L29) and [TeX](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIToTeXFileEngine.pas#L29) by modifying some properties (e.g. [`PowerStr`](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIExportEngine.pas#L97) which was used to specify the name of the power function) and, at some point, calling [`GenerateMathMLEquations()`](https://github.com/agarny/COR/blob/master/Src/CellML/CellMLAPIExportEngine.pas#L1193).
- Generator tests: I would have both the input and output stored in files. Unlike other tests, we are bound to test rather complex CellML models at some point and we clearly can't have everything done (mostly) by hand.

agarny:
Closing as it is being replaced by issue #359 and PR #360.

Issue #263, opened 2019-01-07
Parsing should be done on fully qualified element names


agarny:
This is related to issue #236 and PR #237. There were some discussions about validating a CellML file against a CellML 2.0 schema, but... we don't have one (and who knows when we might get one...). So, it might be faster (for now, if anything) to just parse a CellML file for fully qualified element names. As for attributes, we probably just ought to follow common practice, i.e. expect them to be in the null namespace. @hsorby, @jonc125, @MichaelClerx, @nickerso: does that sound reasonable to you guys? If so, I am happy to have a go at it (I just really need to be able to distinguish between CellML 1.0/1.1 and CellML 2.0 files.)

nickerso:
I think this is the way to go, we want to keep validation separate from parsing but do want to make sure that we only parse CellML 2.0 documents rather than any random XML that happens to have matching local names (e.g., CellML 1.x).

agarny:
FWIW, PR #264 implements what I described above. I have then tried it in OpenCOR and it's working as I expect, i.e. OpenCOR can now distinguish between CellML 1.0/1.1 and CellML 2.0 files.

MichaelClerx:
Cool! Were there any major difficulties in creating a CellML 2 schema? Or just lack of time?

I'm helping @jonc125 teach XML next week, so might be a good time to work on that...

nickerso:
I'm pretty sure its just a lack of time to create a schema, but also I do think the schema validation route is not appropriate for straight out parsing of models. Having a CellML 2.0 schema would definitely be a valuable resource if you are volunteering @MichaelClerx!

PR #264, opened 2019-01-07
Parsing should be done on fully qualified element names (#263)

agarny:


nickerso:
Pretty sure we have dropped the `cmeta` namespace in 2.0 - the `@id` should not have a namespace now.

> Any element information item in the CellML namespace MAY contain an unprefixed attribute information item with local name id. This attribute information item SHALL be treated as having attribute type ID, as defined in section 3.3.1 of XML 1.1.

agarny:
> Pretty sure we have dropped the `cmeta` namespace in 2.0 - the `@id` should not have a namespace now.
> 
> > Any element information item in the CellML namespace MAY contain an unprefixed attribute information item with local name id. This attribute information item SHALL be treated as having attribute type ID, as defined in section 3.3.1 of XML 1.1.

Oops, my bad, I must confess that I checked a CellML 1.0/1.1 file to decide which namespace to use for `id`. Ok, I am going to fix that.

src/parser.cpp
```diff
@@ -325,7 +335,7 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
             }
         } else if (childNode->isType("connection")) {
             connectionNodes.push_back(childNode);
-        } else if (childNode->isType("text")) {
+        } else if (childNode->isType(NULL_NS, "text")) {
```
jonc125: How do you tell the difference between an actual text node, and an element named 'text' in the null namespace?

agarny: Good point, although I understand that your question would still be relevant whether this PR is merged in or not. So, maybe something to raise in a new issue?

nickerso: A "simple" fix is to add a special case to check for text nodes directly rather than using the same method used for elements. i.e., `childNode->isTextNode()` and:

```
bool XmlNode::isTextNode()
{
    return mPimpl->mXmlNodePtr->type == XML_TEXT_NODE;
}
```

that should do the trick, right?

agarny: Ok, I clearly misunderstood @jonc125's point. I think I now understand the problem and something like what @nickerso suggests is probably the way to go indeed. In fact, we should do something similar for comment nodes.

nickerso (approved):

agarny:
FWIW, additional commits are about reverting some changes that I made and that were about cleaning up the code (since they are "properly" addressed in PR #267).

src/parser.cpp
```diff
@@ -251,8 +252,17 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
         return;
     } else if (!node->isType("model")) {
         ErrorPtr err = std::make_shared<Error>();
-        err->setDescription("Model root node is of invalid type '" + node->getType() +
-                            "'. A valid CellML root node should be of type 'model'.");
+        if (node->getType() == "model") {
+            std::string nodeNamespace = node->getNamespace();
+            if (nodeNamespace.empty())
+                nodeNamespace = "null";
+            err->setDescription("Model root node is in invalid namespace '" + nodeNamespace +
```
MichaelClerx: Not sure about the terminology "Model root node". Can we change to "Model element"? (We don't need to call it an information item or anything, since we're explicitly parsing xml here)

src/parser.cpp
```diff
@@ -251,8 +252,17 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
         return;
     } else if (!node->isType("model")) {
         ErrorPtr err = std::make_shared<Error>();
-        err->setDescription("Model root node is of invalid type '" + node->getType() +
-                            "'. A valid CellML root node should be of type 'model'.");
+        if (node->getType() == "model") {
```
MichaelClerx, line 255: Should we not have a general check on all CellML information items to see if they're in the right namespace? This particular check is a nice user-friendly one, because it tells you if you forgot to provide a namespace attribute for the model tag. But all the others inherit this namespace so the checks should be `isType(CELLML_2_0_NS, "component")` etc.

I'd even suggest turning this check into `node->isType(CELLML_2_0_NS, "model")` and have a separate bit of code that checks the root node (whatever type it is) is in the CellML NS, and give a user friendly error if not 

(I think this would also help if we were to start parsing partial CellML docs, for example to allow users to replace a component with a new component specified as XML. In that case you could check the namespace of root of the partial document and (1) be happy if it's CellML, (2) set it to CellML if it's undefined, (3) complain to the user if it's anything else. After that you could pass it off to this parser code...)

src/parser.cpp
```diff
@@ -393,12 +403,12 @@ void Parser::ParserImpl::loadComponent(const ComponentPtr &component, const XmlN
             ResetPtr reset = std::make_shared<Reset>();
             loadReset(reset, component, childNode);
             component->addReset(reset);
-        } else if (childNode->isType("math")) {
+        } else if (childNode->isType(MATHML_NS, "math")) {
             // TODO: copy any namespaces declared in parents into the math element
```
MichaelClerx, line 407: Are we still going down this "convert-to-text" route? I note the validator just uses libxml. If we use libxml to parse the mathml this TODO can go...

src/validator.cpp
```diff
@@ -944,7 +946,7 @@ void Validator::ValidatorImpl::validateAndCleanMathCiCnNodes(XmlNodePtr &node, c
                 // Check for a matching standard units.
                 if (!isStandardUnitName(unitsName)) {
                     ErrorPtr err = std::make_shared<Error>();
-                    err->setDescription("Math has a " + nodeType + " element with a cellml:units attribute '" + unitsName +
+                    err->setDescription("Math has a " + node->getType() + " element with a cellml:units attribute '" + unitsName +
```
MichaelClerx, line 949: Does getType() output include a namespace? Doesn't matter right now, but could cause confusion if we ever have overlapping element names

MichaelClerx (changes requested):
Very happy with these changes, but would like to take the namespace checking 1 step further still!

src/parser.cpp
agarny: I merely mimicked what was there before, but I agree with your suggestion. Consider it done.

src/parser.cpp
agarny: I am not sure I am following you. There are two signatures for `XmlNode::isType()`:
 - `XmlNode::isType(const char *elementNamespace, const char *elementName)`; and
 - `XmlNode::isType(const char *elementName)`.

with the latter calling the former with `CELLML_2_0_NS` as its first argument, and we use that second version for everything that is related to CellML itself.

src/parser.cpp
agarny: Could you remind me what this "convert-to-text" route is all about?

nickerso: I'm pretty sure @MichaelClerx means the way we handle math blocks as text strings rather than dealing with them directly. This TODO is still needed, as the validation will fail with a mathml string in which the MathML namespace has been declared on the CellML model (working from a patchy memory here, so I might be confusing myself).

src/parser.cpp
MichaelClerx: Ah, I missed that one, thanks!

MichaelClerx (approved):
Happy to merge

src/parser.cpp
MichaelClerx: Wait hold on, if it already checks that it's in the cellml namespace, the next check can never fail?

agarny: > Wait hold on, if it already checks that it's in the cellml namespace, the next check can never fail?

Getting late here, but... how do you mean?

agarny:
Ping @nickerso @MichaelClerx. One of you could officially merge this?

src/parser.cpp
nickerso: is this resolved? I think the first check is to see if the root node is a `model` in the CellML namespace and then if it is not, it checks if it is a `model` element in any namespace as a potentially more useful error condition than the root node being any arbitrary XML node...

agarny: For me it should be resolved since @MichaelClerx's "issue" is a non-issue for me, and this for the reason @nickerso gave. Ok, so waiting for @MichaelClerx to confirm that he agrees that this should be resolved, so that either @nickerso or @MichaelClerx can merge this PR.

MichaelClerx: It's resolved. Find it very confusing that `isType()` checks a namespace and `getType` does not return one though.

Issue #274, opened 2019-01-15
XmlNode isType includes namespace, getType does not


MichaelClerx:
I find this confusing. In particular, checks like this:

```
if node->isType("X") {
  // Do what you do with a (CellML) X node
} else if node->getType() == "X" {
  // Do what you do with a non-cellml X node
}
```

Doesn't make sense to me to design the API that way

nickerso:
I agree - this API is confusing me. Although its internal only and won't affect users of libCellML, I can foresee a maintenance headache! Now that #264 is merged we can work on "correcting" the `XmlNode` API.

agarny:
Agreed too. Otherwise, note that `XmlAttribute` would have to be changed too.

Anyway, we already have `XmlNode::isTextNode()` and `XmlNode::isCommentNode()`, so for consistency I would rename `XmlNode::isType()` to `XmlNode::isElementNode()`, leaving the current signatures as they are. As for `XmlNode::getType()`, I would probably rename it to `XmlNode::getNodeName()`.

In fact, we are dealing with a node, so maybe we could/should just drop the `Node` bit, i.e. simply have `XmlNode::isElement()`, `XmlNode::isText()`, `XmlNode::isComment()` and `XmlNode::getName()`?

When it comes to `XmlAttribute`, and to be consistent with `XmlNode`, maybe we should rename `XmlAttribute::isType()` to `XmlAttribute::isAttribute()`, and `XmlAttribute::getType()` to `XmlAttribute::getName()`?

MichaelClerx:
> In fact, we are dealing with a node, so maybe we could/should just drop the Node bit, i.e. simply have XmlNode::isElement(), XmlNode::isText(), XmlNode::isComment() and XmlNode::getName()?
>
> When it comes to XmlAttribute, and to be consistent with XmlNode, maybe we should rename XmlAttribute::isType() to XmlAttribute::isAttribute(), and XmlAttribute::getType() to XmlAttribute::getName()?

Sounds good!
Although maybe `hasName` instead of `isElement` and `isAttribute`?

Hold on, that reintroduces the problem :D

Just `is` ?

nickerso:
I like those suggestions for `XmlNode`. Will need to think a bit more about attributes as the type kind of makes sense in that case...

agarny:
[...]

PR #275, opened 2019-01-16
Update the XmlNode and XmlAttribute type methods.

nickerso:
Attempting to clarify the XmlNode and XmlAttribute type methods as the existing API got a bit confusing with the addition of namespace aware methods. See cellml/libcellml#274.

### DIFF
--- a/src/xmlattribute.h
+++ b/src/xmlattribute.h
@@ -66,31 +66,15 @@ public:
      * the given namespace.
      * Returns @ true if so, and @c false otherwise.
      *
-     * @param attributeNamespace The @c char namespace in which the attribute
+     * @param ns The @c char namespace in which the attribute
      * type name is to be defined.
-     * @param attributeName The @c char attribute type name to check for.
+     * @param name The @c char attribute type name to check for.
      *
      * @return @c true if this XmlAttribute is of the attribute type
-     * specified by the @p attributeName in the namespace @attributeNamespace
+     * specified by the @p name in the namespace @p ns
      * and @c false otherwise.
      */
-    bool isType(const char *attributeNamespace, const char *attributeName);
-
-    /**
-     * @brief Check if this XmlAttribute is of the named attribute type in the
-     * null namespace.
-     *
-     * Checks whether this XmlAttribute has the argument attribute type name in
-     * the null namespace.
-     * Returns @ true if so, and @c false otherwise.
-     *
-     * @param attributeName The @c char attribute type name to check for.
-     *
-     * @return @c true if this XmlAttribute is of the attribute type
-     * specified by the @p attributeName in the null namespace and @c false
-     * otherwise.
-     */
-    bool isType(const char *attributeName);
+    bool isType(const char *name, const char *ns = "");
 
     /**
      * @brief Get the name of this XmlAttribute.
--- a/src/xmlattribute.h
+++ b/src/xmlattribute.h
@@ -93,13 +93,13 @@ public:
     bool isType(const char *attributeName);
 
     /**
-     * @brief Get the type of this XmlAttribute.
+     * @brief Get the name of this XmlAttribute.
      *
-     * Gets the type of this XmlAttribute and returns it as a @c std::string.
+     * Gets the name of this XmlAttribute and returns it as a @c std::string.
      *
      * @return The @c std::string corresponding with the name of this XmlAttribute.
      */
-    std::string getType() const;
+    std::string getName() const;
 
     /**
      * @brief Get the value of this XmlAttribute.
--- a/src/xmlnode.h
+++ b/src/xmlnode.h
@@ -68,15 +68,15 @@ public:
 	 * given namespace with the specified local name.
      * Returns @c true if so, and @c false otherwise.
      *
+     * @param name The @c char element name to check for.
      * @param ns The @c char namespace in which the element
      * node is to be defined.
-     * @param name The @c char element name to check for.
      *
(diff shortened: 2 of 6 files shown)


src/xmlattribute.h
```diff
@@ -93,13 +93,13 @@ class XmlAttribute
     bool isType(const char *attributeName);
```
agarny: Might want to replace `attributeName` with just `name`.

Same in the other signature, in addition to replacing `attributeNamespace` with `namespace`.

src/xmlnode.h
```diff
@@ -61,38 +61,38 @@ class XmlNode
...
+     * name is to be defined.
+     * @param elementName The @c char element name to check for.
      *
-     * @return @c true if this @c XmlNode is of the element type
+     * @return @c true if this @c XmlNode is of the element
      * specified by the @p elementName in the namespace @elementNamespace
      * and @c false otherwise.
      */
-    bool isType(const char *elementNamespace, const char *elementName);
+    bool isElement(const char *elementNamespace, const char *elementName);
```
agarny: Replace `elementNamespace` with `namespace`, and `elementName` with `name`.

src/xmlnode.h
```diff
@@ -61,38 +61,38 @@ class XmlNode
...
-     * @param elementName The @c char element type name to check for.
+     * @param elementName The @c char element name to check for.
      *
-     * @return @c true if this @c XmlNode is of the element type
+     * @return @c true if this @c XmlNode is of the element
      * specified by the @p elementName in the CellML 2.0 namespace and
      * @c false otherwise.
      */
-    bool isType(const char *elementName);
+    bool isCellmlElement(const char *elementName);
```
agarny: Replace `elementName` with `name`.

Otherwise, I am not sure about `isCellmlElement()`. I mean, if we are to have `isCellmlElement()` rather than `isElement()` then we ought to replace `XmlAttribute::isType(const char *attributeName)` with something else, so that we are consistent when it comes to not providing a namespace, and I am not sure what would be a good name for it (clearly not `XmlAttribute::isNullType()`!). So, in the end, I would simply stick to `isElement()`, for consistency.

src/xmlnode.h
```diff
@@ -116,13 +116,13 @@ class XmlNode
     bool isCommentNode();
```
agarny: Replace `isTextNode()` with `isText()` and `isCommentNode()` with `isComment()`.

agarny:
This looks good overall, just a few comments... :)

MichaelClerx:
It might help with the naming to know that XML calls the <this_bit> of an element the "local name", while "cellml:this_bit" is a qualified name

agarny:
> It might help with the naming to know that XML calls the <this_bit> of an element the "local name", while "cellml:this_bit" is a qualified name

That was also my original thinking, but then I saw `XmlAttribute::isType(const char *attributeName)` and was wondering what we should be doing about that one. IOW, I am happy with `XmlNode::isCellmlElement()` if we can come up with something similar for `XmlAttribute::isType(const char *attributeName)`.

nickerso:
Remind me again why we don't just have a single `XmlAttribute::isType(const char *name, const char *ns = "")` (or the valid C++ equivalent if I have that wrong).

src/xmlnode.h
nickerso: Given these are actually checking the type of the XML node, I think it is more correct to leave them as `isTextNode()` and `isCommentNode()`. Ideally we would also be checking in `isElement()` that we actually have an XML Element rather than simply grabbing the name of the node.

agarny:
> Remind me again why we don't just have a single `XmlAttribute::isType(const char *name, const char *ns = "")` (or the valid C++ equivalent if I have that wrong).

It was just because I found it more logical to have the namespace as a first parameter. So, no real good reason, I agree.

Ok, all things considered, I am happy with having only `XmlAttribute::isType(const char *name, const char *namespace = nullptr)`, and have something similar for `XmlNode`, i.e. `XmlNode::isElement(const char *name, const char *namespace = nullptr)`.

src/xmlnode.h
agarny: Ok, then I would still have `isText()` and `isComment()` and then have `isElement()` check that we actually have an XML element.

MichaelClerx:
This sounds like it's heading in the right direction!

nickerso:
just to note that `namespace` is a reserved word in C++ so we can't use that directly ;-)

nickerso:
With `XmlAttribute::isType()` makes sense to have a default namespace (empty), whereas for the elements we are always dealing with namespaces, so no need for a default. And the CellML specific one is really just a shortcut rather than a default namespace value.

I still think `isTextNode()` and `isCommentNode()` is less ambiguous than `isText()` and `isComment()` - but could probably be convinced otherwise...

agarny:
It all looks good to me, except for `isTextNode()` and `isCommentNode()`, which I would still replace with `isText()` and `isComment()`, respectively (and have `isElement()` check that we actually have an XML element).

nickerso:
`isElement()` is now checking that the node is an element.

agarny:
> `isElement()` is now checking that the node is an element.

Oops, indeed! Well, one more reason for replacing `isTextNode()` and `isCommentNode()` with `isText()` and `isComment()`. :)

agarny (approved):

MichaelClerx (approved):
Much better

Issue #272, opened 2019-01-14
Update list of contributors


MichaelClerx:
Besides organisations who've signed some sort of agreement, we could also have a list of people who worked on libcellml (presumably in their spare time)

nickerso:
is there a way to generate this from Github?

MichaelClerx:
There is just this I suppose: https://github.com/cellml/libcellml/graphs/contributors

So no need to write it out ourselves then :D

nickerso:
just need to incorporate it into the docs :)

agarny:
FWIW: https://developer.github.com/v3/repos/#list-contributors.

PR #281, opened 2019-03-04
Add list of source code contributors as recorded by GitHub.

hsorby:
Addresses issue #272.

### DIFF
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -15,8 +15,7 @@ List of Funding Organisations
 List of Contributors
 ====================
 
-The following is a list of contributors(in surname alphabetical order) who have contributed lines of
-source code to the libCellML project on or before 2019-03-05.
+The following is a list of contributors (in surname alphabetical order) who have contributed lines of source code to the libCellML project on or before 2019-03-06.
 
  * `Ted Ahmadi <https://github.com/TedAhmadi>`_
  * `Robert Blake <https://github.com/rblake-llnl>`_
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -15,15 +15,17 @@ List of Funding Organisations
 List of Contributors
 ====================
 
-The following is a list of contributors (in surname alphabetical order) who have contributed lines of source code to the project on or before the 5/3/2019.
+The following is a list of contributors(in surname alphabetical order) who have contributed lines of
+source code to the libCellML project on or before 2019-03-05.
 
-* `Robert Blake <https://github.com/rblake-llnl>`_
-* `Michael Clerx <https://github.com/MichaelClerx>`_
-* `Jonathan Cooper <https://github.com/jonc125>`_
-* `Alan Garny <https://github.com/agarny>`_
-* `David Ladd <https://github.com/dladd>`_
-* `Gary Mirams <https://github.com/mirams>`_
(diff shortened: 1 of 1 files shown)


agarny (commented):
The list of contributors is likely to change with time, which means that we will likely have to update it on a regular basis. So, why not simply point to https://github.com/cellml/libcellml/graphs/contributors instead?

hsorby:
Can do, I wanted to try out this to see what people thought. (I think I got carried away with your link to the API.)

agarny:
> Can do, I wanted to try out this to see what people thought.

Ok, let's see what others think.

Otherwise, and FWIW, I am happy with this PR if the list of contributors points to https://github.com/cellml/libcellml/graphs/contributors.

docs/contributors.rst
```diff
@@ -6,8 +6,24 @@ Contributors
...
+The following is a list of contributors (in surname alphabetical order) who have contributed lines of source code to the project on or before the 5/3/2019.
+
+* `Robert Blake <https://github.com/rblake-llnl>`_
+* `Michael Clerx <https://github.com/MichaelClerx>`_
+* `Jonathan Cooper <https://github.com/jonc125>`_
+* `Alan Garny <https://github.com/agarny>`_
+* `David Ladd <https://github.com/dladd>`_
+* `Gary Mirams <https://github.com/mirams>`_
+* `David Nickerson <https://github.com/nickerso>`_
+* `Hugh Sorby <https://github.com/hsorby>`_
```
agarny: What is the point of having that list if we then point to the up-to-date list of contributors? It's bound to be out of date within a few weeks. I would simply have something like:

```
An up-to-date list of contributors can be found at https://github.com/cellml/libcellml/graphs/contributors.
```

hsorby: I like the idea of having the names of contributors available without having to do that extra click.       I am very anti extra mouse clicks.

nickerso: I also slightly prefer being able to see the "current" list immediately and then click through to get all the details. It will take some small effort to keep it from getting too out of date, but manageable...

hsorby: I have created a small Python Script that will generate the list ready for copy and pasting into reStructured text, reducing a little of the burden?

agarny: People tend to forget to update things that can easily get out of date, which is the reason I prefer the link on its own, but nevermind.

agarny (approved):

docs/contributors.rst
```diff
@@ -6,8 +6,26 @@ Contributors
...
+=============================
 
  * `University of Auckland <https://www.auckland.ac.nz/>`_
  * `VPR <http://www.virtualrat.org/>`_
+
+List of Contributors
+====================
+
+The following is a list of contributors(in surname alphabetical order) who have contributed lines of
+source code to the libCellML project on or before 2019-03-05.
```
agarny: "contributors (in" and not "contributors(in".

Also, didn't we agree that we should have one line per sentence? This makes it *much* easier to compare versions.

hsorby: Sorry that was a change because my Python code went over the PEP8 line length guideline.

hsorby:
I have put the line back to one line again now but that means it won't compare too well this time, apologies.

PR #282, opened 2019-03-07
Addresses various minor Clang-Tidy issues

agarny:


src/orderedentity.cpp
```diff
@@ -29,6 +29,7 @@ struct OrderedEntity::OrderedEntityImpl
 {
     int mOrder;
     bool mOrderSet;
+    char padding[3];
```
hsorby, line 32: What is the purpose of this?

src/reset.cpp
```diff
@@ -30,6 +30,7 @@ namespace libcellml {
 struct Reset::ResetImpl
 {
     int mOrder = 0; /**< An integer for determining relative order.*/
+    char padding[4];
```
hsorby, line 33: Not at all keen on these padding items in the structure what exactly is the issue that clang is having here.

hsorby (commented):
Clang seems to be taking a strange stance on some of these warnings.  These are not compiler warnings correct?

On the whole I'm ok with these changes but I'm not keen on the padding.

agarny:
> Clang seems to be taking a strange stance on some of these warnings. These are not compiler warnings correct?

Correct, they are not compiler warnings. They are warnings generated by [Clang-Tidy](https://clang.llvm.org/extra/clang-tidy/). Clang-Tidy could be configured to be much more pedantic, but personally I am using the default settings available in [Qt Creator](https://www.qt.io/qt-features-libraries-apis-tools-and-ide/#ide).

> On the whole I'm ok with these changes but I'm not keen on the padding.

The padding-related warnings are a way for Clang-Tidy to let people know that some padding is going to be done, and that the user might want to check the code in case s/he wants to "control" the padding. For `OrderedEntity::OrderedEntityImpl`, I put the padding at the end because it's fine there. However, when it comes `Reset::ResetImpl`, I put it where it is so that `mVariable` gets properly (memory) aligned.

When it comes to the padding itself, I must confess that, in principle, I am not too keen on it either, but... it's done in the private implementation of a couple of classes, so I am ok with it. As you suggested off-line, we could have a comment and/or have `_padding` (rather than `padding`) to make it clear what the purpose is.

nickerso (changes requested):
Generally good to get these changes, but two main issues that can see. The first is that there seems little point to do a one-off change like this without extending our testing to enforce clang-tidy (or similar) compliance on future pull requests. Is that something that is possible? (This is a separate issue and I don't think it should stop these changes going in.)

The one that I would like to see changed is the padding. This makes no sense to me. This is what compilers do for us and is introducing something that is going to become a potential maintenance headache in future with different compilers and different basic type sizes. I'd like to see this removed and clang-tidy configured to ignore this issue.

agarny:
> Generally good to get these changes, but two main issues that can see. The first is that there seems little point to do a one-off change like this without extending our testing to enforce clang-tidy (or similar) compliance on future pull requests. Is that something that is possible? (This is a separate issue and I don't think it should stop these changes going in.)

Good point and, indeed, this could be done in our CI process. Indeed, Clang-Tidy can be used from the command line, so we could have that be part of our CI process. When it comes to CMake, it could look for Clang-Tidy and if it's present then run it with a set of settings that we are happy with. Ok, I am going to create a GitHub issue for it.

> The one that I would like to see changed is the padding. This makes no sense to me. This is what compilers do for us and is introducing something that is going to become a potential maintenance headache in future with different compilers and different basic type sizes. I'd like to see this removed and clang-tidy configured to ignore this issue.

Padding is indeed done by compilers, but as I mentioned [here](https://github.com/cellml/libcellml/pull/282#issuecomment-470750786), it may not be optimised (from what I have read here and there), hence Clang-Tidy warns us about padding issues so that we can optimise things as we see fit.

Now, I must confess that when it comes to libCellML, this is unlikely to make any difference. So, yes, we could probably ignore that particular warning (if we were to implement your first point).

agarny:
Closing as this has now been superseded by issue #283.

agarny:
Closing as this has now been superseded by issue #283.

Issue #285, opened 2019-03-10
CMake: hide more distracting variables


agarny:
If we configure libCellML using `cmake-gui` on Windows or `ccmake` on Linux/macOS, we will see that a variable named `LibXml2_DIR` on Windows or `pkgcfg_lib_PC_LIBXML_xml2` on Linux/macOS. As for some other variables, it would be good to have it set as advanced so that only libCellML-related variables are visible by default.

nickerso:
I think I would generally expect basic configuration options for dependencies to be visible by default...but I could easily be swayed in this :)

agarny:
I believe that, at some point, we agreed that we should only see libCellML-related settings.

Also, personally, I like things to be clean. So, when configuring libCellML, well... I want to configure libCellML itself, not its dependencies which I assume where properly handled by our CMake file. If that's not the case, then ok I go into advanced mode and deal with them there.

hsorby:
I would like to only see LIBCELLML_ type  variables in my CMake interface.  There is generally not a lot one can configure with the dependencies and it seems best fit as an advanced configuration option to me.

MichaelClerx:
Agreed! Ideally the standard options are just things you can add on / leave out, ways to configure libcellml but not stuff that can break it

nickerso:
The only problem is that this would force Windows users to toggle to the advanced options in order to be able to build libCellML (there is no standard install of a suitable libXML on Windows, right?). Unless we add a `LIBCELML_LIBXML_DIR` variable...

agarny:
I am against having `LIBCELML_LIBXML_DIR` only so that Windows users don't have to toggle the advanced options. On Windows, people will anyway have to get their hands dirty (e.g. have to install our recommended `LibXml2` libraries), so I would rather keep things clean for the majority of people.

nickerso:
Define majority? :)

A single non-libcellml variable is all that is currently needed for a windows user to build libcellml and use it in their project. Another one for SWIG is needed if you want Python bindings, I think, but I'm happy for Python bindings to be an advanced option. 

libxml is a hard dependency for libcellml, you simply can't build the library without it. It is not an advanced option for the user to define where libxml can be found on their system. Basic "get up and running" documentation should not start with "and now toggle the advanced options, don't mess with anything other than ... "

Also, I'm definitely not suggesting that we have a `LIBCELLML_LIBXML_DIR` variable, that should have had the :-) appended!

agarny:
[...]

PR #286, opened 2019-03-10
CMake: hide more distracting variables (#285)

agarny:


cmake/common.cmake
```diff
@@ -45,6 +45,11 @@ function(HIDE_DISTRACTING_VARIABLES)
     mark_as_advanced(CMAKE_OSX_DEPLOYMENT_TARGET)
     mark_as_advanced(CMAKE_OSX_SYSROOT)
   endif()
+  if(WIN32)
+    mark_as_advanced(LibXml2_DIR)
```
hsorby: I think on windows we are going to have to have this variable visible in the standard view.  It is going to be needed when using a CMake GUI configuration tool which is the default way of configuring CMake projects on Windows.

hsorby (commented):
Need to clarify position when configuring on windows.

cmake/common.cmake
agarny: I just want things to be clean and that wouldn't be the case here. Also, let's be serious, to use the advanced mode is really not such a big deal. If people can't be bothered to do that, then they should use the binaries that I imagine we will be making available at some point.

MichaelClerx: If windows users always have to use the advanced mode, then it's never clean for them

agarny: It depends on your definition of clean. For me, it's when I only see things that are directly related to libCellML. So, `LibXml2_DIR` doesn't fall in that category. Also, to have `LibXml2_DIR` in the main mode means that it wouldn't be consistent with what we would be doing on Linux/macOS.

Also, I have just noticed that we actually get a few more variables: `LIBXML2_INCLUDE_DIR`, `LIBXML2_LIBRARY` and `LIBXML2_XMLLINT_EXECUTABLE`.

Image2

So, it actually gets quite busy and somewhat confusing since the `LIBXML2_XXX` variables all have a value of `LIBXML2_XXX-NOTFOUND` even though everything is "properly" configured.

nickerso:
Comments seem to be leaning toward closing this one without merging. Any further comments?

hsorby:
I think this needs some further fixes.  The libXml2_DIR should be removed if the CONFIG mode search fails.  I want some of it?

agarny:
> I think this needs some further fixes. The libXml2_DIR should be removed if the CONFIG mode search fails. I want some of it?

I would also do something about those `LIBXML2_XXX` variables. In my screenshot above, libxml2 is properly configured and yet one could think it's not considering the value of those `LIBXML2_XXX` variables.

agarny:
This PR is ready to be merged in... or not. Either way, it would be nice if we could make a final decision on it. @hsorby, @nickerso?

nickerso:
Still seems to be some outstanding work regarding the handling of failure when finding in the CONFIG mode. 

I'm still happy to close this one and leave the config as is (works fine under windows for me just the way it is).

agarny:
> Still seems to be some outstanding work regarding the handling of failure when finding in the CONFIG mode.

Sure, but this is not the problem this PR was aiming to "fix".

hsorby:
I think this can be merged it doesn't have much of an affect.

hsorby:
Right this should be good to go.

nickerso (approved):
This looks good now.

agarny:
Cool, someone to "officially" merge it in then.

Issue #283, opened 2019-03-08
Add support for Clang-Tidy and additional Clang warnings


agarny:
[Clang-Tidy](https://clang.llvm.org/extra/clang-tidy/) is a tool for "diagnosing and [automatically] fixing typical programming errors". In the case of libCellML, it might be nice to take advantage of such a tool to ensure that our codebase is as "clean" as possible.

We could, in CMake, check whether Clang-Tidy is installed and, if so, run it against our codebase. If anything gets reported then the build would be considered as having failed.

We clearly can't force developers to have Clang-Tidy installed on their system, but our CI system should definitely have it (at least our macOS slave since Clang is the default compiler on that platform).

To start with, we might want to use the Clang-Tidy settings used by Qt Creator and then refine them as we see fit. Those settings are:

```
-checks=-*,bugprone-*,cppcoreguidelines-avoid-goto,cppcoreguidelines-c-copy-assignment-signature,cppcoreguidelines-interfaces-global-init,cppcoreguidelines-narrowing-conversions,cppcoreguidelines-no-malloc,cppcoreguidelines-pro-*,cppcoreguidelines-slicing,cppcoreguidelines-special-member-functions,misc-*,modernize-*,performance-*,readability-avoid-const-params-in-decls,readability-container-size-empty,readability-delete-null-pointer,readability-deleted-default,readability-else-after-return,readability-function-size,readability-identifier-naming,readability-inconsistent-declaration-parameter-name,readability-misleading-indentation,readability-misplaced-array-index,readability-non-const-parameter,readability-redundant-*,readability-simplify-*,readability-static-*,readability-string-compare,readability-uniqueptr-delete-release,readability-rary-objects
```

We might also want to add additional Clang warnings, based on what Qt Creator's Clang Code Model plugin does. Note that this is clearly only relevant if we are building libCellML using Clang, i.e. on macOS for now.

The idea of the above is to catch as many warnings as possible to ensure that our codebase is as good as possible. Also, if libCellML is to be built with warnings as errors, then we should do the same here.

kerimoyle:
Hi Alan
I'm trying to configure my stuff to use clang-tidy but the options above give me "invalid argument" errors (not sure which flag is problematic).  Is it possible to do a -dump-config from your configuration and include the output in a .clang-tidy file in the repo instead?

agarny:
It should all work fine with Clang-Tidy 8.0.0. So:
- Which operating system are you using? macOS, I believe?
[...]

Issue #284, opened 2019-03-08
Add support for Clang-Format


agarny:
In the same way that we probably ought to use [Clang-Tidy](https://clang.llvm.org/extra/clang-tidy/) (see issue #283), we might want to use [Clang-Format](https://clang.llvm.org/docs/ClangFormat.html) to ensure that our codebase looks the same throughout? Not quite sure where this would sit though...

agarny:
[Here](https://clangformat.com/) is an online resource for generating a `.clang-format` file.

PR #288, opened 2019-03-20
Add support for Clang-Tidy and additional Clang warnings

agarny:
Addresses issue #283.

Clang-Tidy and the additional Clang warnings are used for our `src` and `tests` directories. However, when it comes to the `tests` directory, we had to consider `gtest/` as a system header prefix, to prevent Clang from generating an awful lot of warnings. In the same way, we had to disable some warnings in Clang-Tidy, so that we can "safely" use  `TEST()` and `EXPECT_EQ()`. Also, note that in order to use Clang-Tidy, we had to bump our minimum requirement for CMake. So, now, we need at least CMake 3.6 (while 3.5.2 before).

Note that Clang-Tidy and the additional Clang warnings only make sense on our macOS build machine. Actually, Clang-Tidy could be installed on our Windows and Linux build machines too, if we wanted, but Clang is not used on those machines anyway, so probably best to just stick to our macOS build machine to get a full report of what could be improved in our codebase, if anything, not least since our macOS build machine is the fastest of all three.

The bottom line is that we are now checking an awful lot of things, which should make our codebase pretty solid.

Otherwise, I have taken that opportunity to also:
 - clean up various bits of codes here and there;
 - move the search for LibXml2 from `CMakeLists.txt` to `cmake/common.cmake`, which I think is where it really ought to be?
 - upgrade our version of GTest. We are now using version 1.8.1. This wasn't strictly required, but at some point I thought upgrading GTest might prevent the thousands of warnings I was getting, but it didn't make any difference. Yet, I had upgraded, so we might as well use the latest version of GTest.


agarny:
Argh, our Linux build machine only has CMake 3.5.2 installed on it. @hsorby, I thought it was bioeng49, but it clearly isn't since bioeng49 has CMake 3.12.4. So, which machine is used for building libCellML on Linux? In fact, could you upgrade its copy of CMake?

Regarding our macOS build machine, I also thought that it was bn337051, but once again it doesn't seem to be the case since it seems to be running macOS 10.12...? I installed Clang-Tidy on bn337051, but it looks like it might have been unnecessary. Still, for this PR to make any sense, we need Clang-Tidy installed on the macOS build machine used to build libCellML. Again, @hsorby, it would be nice to have that done.

In summary:
 - [ ] Have our Linux build machine use at least CMake 3.6.
 - [ ] Install Clang-Tidy on our macOS build machine (using `brew install llvm` and then symlinking `clang-tidy` (and `clang-format`) to `/usr/local/bin`).
 - [ ] Repeat commit 89cbd45 to make sure that our macOS build does indeed fail.
 - [ ] Make Clang and Clang-Tidy happy by committing something like 029a7e4.

agarny:
Would be nice to have something done about this PR... especially with Keri now working on libCellML. I would also like to have `clang-format` used to ensure that our code is consistent throughout (see issue #284).

nickerso:
Just to be clear, we're waiting on some upgrades to the build machines to get things playing nice with these changes, right?

hsorby:
I believe you are right.

agarny:
That is correct. The version of CMake on the Linux slave is too old for this PR to pass, and on the macOS slave we should have `clang-tidy` installed so that this PR can actually be useful.

I thought that those slaves were `bioeng49` and `bn337051`, but they are not. `bioeng49` runs Ubuntu 16.04 LTS while the Linux slave used by libCellML runs Ubuntu 14.04 LTS (?!). As for the macOS slave used by libCellML, it seems to be running macOS Sierra (10.12) while `bn337051` runs macOS High Sierra (10.13).

So, in the end, I have no idea what/where those slaves are... and therefore can't update them...

hsorby:
The build slaves have now been updated and all pertinent builds have been triggered.  Failures now indicate true failures that need to be fixed.  This (I'm fairly confident) is the only PR needing changes.

agarny:
Hmm, not quite sure how I missed that one, but I can indeed confirm the same build error on my local Ubuntu 18.04 LTS box. Ok, consider it fixed... soon...

nickerso:
This one is good to merge in now, right? is it going to cause major headaches in the other open pull requests (especially #304) ?

agarny:
> This one is good to merge in now, right? is it going to cause major headaches in the other open pull requests (especially #304) ?

As far as I am concerned, it is indeed ready to merge. And, yes, it's likely to cause some headache with all pending PRs...

hsorby:
First in first served?  That is first to get their PR approved and merged passes on the burden of conflict resolution to PRs not yet approved.

nickerso (approved):

cmake/environmentchecks.cmake
```diff
@@ -14,21 +14,54 @@
...
+  # we wish to use Config mode in the first call.  Finding LibXml2 in config mode
+  # is the preferred method so we will try this first quietly.
+  #
+  # This does change how we get information about include paths and such so we
+  # need to track how we found LibXml2.
+  find_package(LibXml2 CONFIG QUIET)
+  if (LibXml2_FOUND)
+    set(HAVE_LIBXML2_CONFIG TRUE)
+  else ()
+    find_package(LibXml2 REQUIRED)
```
hsorby, line 49: Can we add in here:

`unset(LibXml2_DIR CACHE)`

agarny: There are several calls to `find_package(LibXml2...)`. So, are you meaning that we should that `unset()` call between `set(HAVE_LIBXML2_CONFIG FALSE)`and `if(MSVC)`?

agarny:
@hsorby, would you know why the Python tests fail on the macOS slave? FWIW, those tests work on my Mac...

hsorby:
Yes, the macOS slave is running SWIG 4.0.  So we need to look at that.

hsorby:
The new SWIG 4.0 has changed the way it deals with rpath I think so we need to upgrade the handling of the bound dynamic libraries.

hsorby:
I think this one is good to merge now.

agarny:
FWIW, I want to go through this one more time (today). There have been quite a few `git merge`s lately. So, I want to make sure that it all looks good.

agarny:
Actually, @hsorby, don't we want to create a test for Clang-Tidy in the same that you created one for ClangFormat?

hsorby:
The problem with the clang-tidy is that it is a compiler level change.  It is not really external.  The clang tidy then becomes part of the standard warnings and errors that we already deal with.  So I think we just add it in when available the Buildbot testing will do the same and our checks should continue as they currently are with no further changes necessary.

agarny:
> The problem with the clang-tidy is that it is a compiler level change. It is not really external. The clang tidy then becomes part of the standard warnings and errors that we already deal with. So I think we just add it in when available the Buildbot testing will do the same and our checks should continue as they currently are with no further changes necessary.

Ok, fair enough.

Otherwise, I have just gone through this PR one more time and I am happy with it. So, up to the reviewers now... :)

agarny:
From what I can tell, both @hsorby and @nickerso are OK with this PR. So, either of you can merge it in...

Issue #309, opened 2019-04-26
Printer: use indentation and line feeds


agarny:
Right now, our printer always generates two lines. One for the XML version and encoding we want to use, i.e. `<?xml version=\"1.0\" encoding=\"UTF-8\"?>` and another for the CellML model itself, no matter how long it is.

However, I feel like it might be nice to have our printer output use indentation and line feeds, so that things look a bit better (and are easier to `diff`, if needed). This would also prevent us from having "fake" indentations in our tests, e.g.
```
"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
"<model xmlns=\"http://www.cellml.org/cellml/2.0#\">"
    "<component name=\"parent_component\"/>"
    "<component name=\"child1\"/>"
    "<component name=\"child2\"/>"
    "<component name=\"child3\"/>"
    "<encapsulation>"
        "<component_ref component=\"parent_component\">"
            "<component_ref component=\"child1\"/>"
            "<component_ref component=\"child2\"/>"
            "<component_ref component=\"child3\"/>"
        "</component_ref>"
    "</encapsulation>"
"</model>";
```
while I would personally rather have (note the two-space indentation; this is the indentation used by libxml2):
```
"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
"<model xmlns=\"http://www.cellml.org/cellml/2.0#\">\n"
"  <component name=\"parent_component\"/>\n"
"  <component name=\"child1\"/>\n"
"  <component name=\"child2\"/>\n"
"  <component name=\"child3\"/>\n"
"  <encapsulation>\n"
"    <component_ref component=\"parent_component\">\n"
"      <component_ref component=\"child1\"/>\n"
"      <component_ref component=\"child2\"/>\n"
"      <component_ref component=\"child3\"/>\n"
"    </component_ref>\n"
"  </encapsulation>\n"
"</model>\n";
```
This would also make things better when formatting using Clang-Format (see issue #284).

hsorby:
I don't mind this change.  Having given it some thought I don't mind either output.  If I was going to do a `diff` on a CellML file I would use a node based diff and not line based so the indentation wouldn't effect the results either way.

PR #312, opened 2019-04-29
Printer: use indentation and line feeds

agarny:
Addresses issue #309.

There were two ways to approach this issue:
 1) Rely on libxml2 to do the indentation and line feeding for us; and
 2) Do the indentation and line feeding manually.

I gave both approaches a try and the first approach meant creating a dummy XML document (in case we only wanted to print, say, a component since that bit of XML code might have resulted in more than one top XML element, thus making the XML code invalid for libxml2), which in turn meant getting rid of the dummy top element and unindenting its children. In other words, although it was working, it was a bit of a dirty hack. Also, because of the way we handle math, it meant that the `Parser.parseModelWithComponentsWithMultipleMathElements` test would keep failing.

So, in the end, I went with the second approach, which happens to be much cleaner. I do, however, appreciate why indentation and line feeds were not used in the first place! It's a wee bit more involved... :)


src/xmlnode.cpp
```diff
@@ -157,10 +157,13 @@ XmlNodePtr XmlNode::getParent()
     return parentHandle;
 }
 
-std::string XmlNode::convertToString() {
+std::string XmlNode::convertToString(int format) {
```
hsorby: I would have preferred an enumeration here.

tests/math/math.cpp
```diff
@@ -18,36 +18,33 @@ limitations under the License.
 
 #include <libcellml>
 
+static const std::string emptyMath = "<math xmlns=\"http://www.w3.org/1998/Math/MathML\"/>\n";
```
hsorby: Shouldn't this variable be ALL CAPS?

hsorby (commented):
I see you have opted for 2 space indent and not 4.

src/xmlnode.cpp
agarny: I thought I would mimic what `xmlNodeDump()` does, but if anything I would actually use a `bool`.

tests/math/math.cpp
agarny: I don't mind either way. It's just that I have seen something similar in `tests/generator/generator.cpp`.

agarny:
> I see you have opted for 2 space indent and not 4.

I initially wanted to go with a four-space indentation, but libxml2 uses two-space and I need libxml2 (through `XmlNode::convertToString()`) to format a math string, which means two-space indentation.

Hmm... I am going to check whether we cannot ask libxml2 to use four-space indentation...

hsorby:
I am all for going with the default that libxml2 is using that means less work adapting for what I consider zero benefit.

tests/math/math.cpp
hsorby: Don't look in the tests directory for code conformance!

src/xmlnode.cpp
hsorby: I would use `enum` if there was going to be more than two formats otherwise `bool` if there were only going to be two.  Either way not `int`.

agarny:
> I am all for going with the default that libxml2 is using that means less work adapting for what I consider zero benefit.

That's why I hadn't bothered checking, but I have just seen that it's actually very easy to change the number of spaces to be used for indentation. One just needs to set `xmlTreeIndentString` to whatever (literally) we want. So, here, all we need is to have `xmlTreeIndentString = "    ";` before calling `xmlNodeDump()` in `XmlNode::convertToString()`.

I am happy to make that change, if we want...

tests/math/math.cpp
agarny: Ok, I am going to replace `emptyMath` with `EMPTY_MATH`. No problem.

hsorby:
I'm happy with two. So if there is going to be a vote I'm voting two. (I know some people are very particular about indentation :))

src/xmlnode.cpp
agarny: For whatever reason, libxml2 uses an `int`, but there is really only two values: `0` and `1` for no formatting and formatting, respectively.

Ok, I am going to use a `bool`.

agarny:
I prefer four myself, but I am ok with two.

src/xmlnode.cpp
nickerso: libxml2 is a C API which has no native bool type....

nickerso:
This really seems to be introducing a significant maintenance burden for zero gain. If we are going to do this I would definitely prefer to simply let libXml2 to handle the "prettifying" for us.

agarny:
> This really seems to be introducing a significant maintenance burden for zero gain. If we are going to do this I would definitely prefer to simply let libXml2 to handle the "prettifying" for us.

I did give it a try, but this caused problems (see my original message).

hsorby:
> This really seems to be introducing a significant maintenance burden for zero gain. If we are going to do this I would definitely prefer to simply let libXml2 to handle the "prettifying" for us.

It's a problem either way I'm pretty sure I have already been through once (or twice) and reformatted all the expected test strings.  I'm not keen on making this prettifying change but I will be useful to be able to execute simple text diffs on the printed output from libCellML which I think will be a time saver long term.

nickerso:
If we implement #332 then I believe the need to implement our own formatting goes away and we can simply use libXml2 to handle the code formatting, as per the original option 1.

agarny:
> If we implement #332 then I believe the need to implement our own formatting goes away and we can simply use libXml2 to handle the code formatting, as per the original option 1.

I agreed that option 1 should work. Still, nothing that prevents us from accepting this PR until we actually start work on issue #332.

nickerso (approved):

Issue #297, opened 2019-04-09
Update unit::prefix to be integer valued


nickerso:
As per cellml/cellml-specification#18, the unit prefix is now defined as a "integer valued string" in the current 2.0 draft spec. We need to update libCellML implementation to match.

kerimoyle:
Now fixed in my validation branch

kerimoyle:
Scratch that, more work to do tracking down all references to `double prefix` outside the validator.

PR #318, opened 2019-05-08
Units.Prefix changed from double to integer

kerimoyle:
This PR changes the Units.Prefix from a double to an integer, as discussed earlier [here](https://github.com/cellml/libcellml/pull/304#issuecomment-487437331), and to address https://github.com/cellml/libcellml/issues/297 and https://github.com/cellml/cellml-specification/issues/18.

The changes are:
- removed the python bindings for Units::addUnit methods which have a `Prefix `argument by `%ignore`-ing them in the units.i file
- removed any test from the test_units.py file that used `Prefix`es 
- changed units.cpp and units.h and tests to suit.

I wonder if - as a fail safe - we could use signed `enum`s in the C++?  This way their default integer value would be the same as the integer specification, so that any usage which slips through the cracks at least defaults to something reasonable?

```
enum class Prefix
{
    YOTTA=24,
    ZETTA=21,
    EXA=18,
    ...
    ATTO=-18,
    ZEPTO=-21,
    YOCTO=-24
}
```


agarny:
> I wonder if - as a fail safe - we could use signed `enum`s in the C++? This way their default integer value would be the same as the integer specification, so that any usage which slips through the cracks at least defaults to something reasonable?

In principle, I guess we could, but I would feel somewhat uneasy about having such a "fail safe" implementation. To me, to have such an implementation would send the wrong message in my opinion: don't worry, we may be able to cover your back if you mess up, although there is no guarantee... :)

kerimoyle:
> To me, to have such an implementation would send the wrong message in my opinion: don't worry, we may be able to cover your back if you mess up, although there is no guarantee... :)

It's more about making sure that where our own systems and bindings have the potential to fail silently, that they fail to something sensible.  I don't really think that entering 1 instead of 1. deserves to be called "messing up" to the extent of a silently-applied 1e21 factor :O ... I think that one is on us!  And besides that, it would need some digging through the code to find anyway, so it's not like it's a policy statement or anything :)

agarny:
Fair enough. At the end of the day, C++ allows us to specify a value for our `enum`s, so why not take advantage of that feature indeed.

kerimoyle:
All python tests in Mac failing here  too ... :(

kerimoyle:
Closing as present in more recent PR #349

agarny:
> Closing as present in more recent PR #349

?! I am confused. Why close this issue since PR #349 is about addressing this issue and has yet to be merged in...?!

kerimoyle:
It's because I needed to make a separate issue for it after messing up the larger validator PR.  The issue is listed as #318 and addressed in PR #349.  This branch can be deleted.

agarny:
> It's because I needed to make a separate issue for it after messing up the larger validator PR. The issue is listed as #318 and addressed in PR #349. This branch can be deleted.

Ok, it's just that you have been talking about issue #318 and that PR #349 addresses issue #318. However, there is no issue #318 since #318 is actually a PR. So, yes, all rather confusing.

In the end, it looks like PR #349 actually addresses issue #297, and this ought to be reflected in PR #349.

Issue #320, opened 2019-05-10
Entity: some questions regarding [set|get]Parent()


agarny:
1. I am struggling to see why we need to expose the `Entity::setParent()`. Are there use cases where a libCellML user would actually want to set the parent of an entity?
2. I find `Entity::getParent()` rather cumbersome to use. I have a use case where I have an encapsulated component for which I want to retrieve the (parent) model. With `Entity::getParent()`, I can retrieve my encapsulated component's parent component. But then, I have to retrieve that parent component's parent component until I eventually reach the parent model. This is rather tedious. I wish there was an `Entity::getParentComponent()` method that would return the parent component (should there be one) and an `Entity::getParentModel()` method that would return the parent model.
3. I find very risky to have `Entity::clearParent()` as part of the API.
4. `Entity::hasParent()` probably ought to be renamed `Entity::hasParentComponent()`?

agarny:
A quick implementation of `Entity::getParentComponent()` and `Entity::getParentModel()` can be found [here](https://github.com/agarny/libcellml/commit/94d8dbd22230266a5d3d1996853a87c3b9e87e6a).

hsorby:
If we know what we are getting why not return a smart pointer to the object instead?

If I remember correctly the getParent is used to check for circular parentage.  If the tests still pass then I don't mind changing it (never did like what I did there anyway).

I think you have to have `Entity::clearParent()` don't you?

`Entity::hasParent` isn't the same as `Entity::hasParentComponent` the question being asked is not the same.

agarny:
> If we know what we are getting why not return a smart pointer to the object instead?

Agreed, it was just a quick and dirty implementation (at this stage) for issue #257.

> If I remember correctly the getParent is used to check for circular parentage. If the tests still pass then I don't mind changing it (never did like what I did there anyway).

I have just implemented `Entity::getParentComponent()` and `Entity::getParentModel()` using a smart pointer (and had to modify many of our tests!), and it's all working fine except for the `Connection.componentlessVariableInvalidConnectionClearParentCheck` test. Rather than getting:

```
<?xml version="1.0" encoding="UTF-8"?>
<model xmlns="http://www.cellml.org/cellml/2.0#">
   <component name="component2">
      <variable name="variable2" />
   </component>
   <connection component_1="component2">
      <map_variables variable_1="variable2" variable_2="variable1" />
   </connection>
</model>
```

I am getting:

```
<?xml version="1.0" encoding="UTF-8"?>
[...]

Issue #279, opened 2019-02-14
Improve our Pimpl approach?


agarny:
Our current implementation of the Pimpl idiom relies on each of our classes having a raw pointer to its private implementation. This means that our `Component` class has 5 private implementations, one for itself, another for its parent classes, i.e. `ComponentEntity`, `NamedEntity`, `Entity`, and a final one for `ImportedEntity` since `Component` inherits both from `ComponentEntity` and `ImportedEntitiy`.

However, for such a class, we could get away with "only" 2 private impelementations, one for `ComponentEntity`-`NamedEntity`-`Entity` and another for `ImportedEntity`, this by using inheritance.

~~Also, we could use a smart pointer rather than a raw pointer for our private implementations.~~

In other words, we might want to have a look at [what the Qt guys are doing](https://wiki.qt.io/D-Pointer) and probably mimick their implementation of the Pimpl idiom...?

P.S.: we might also want to clean some of our interfaces. E.g. the `private` section of the `Component` class contains the `doAddComponent()` method, which I believe has nothing to do there and could probably be "fixed" by using inheritance for our private implementations.

MichaelClerx:
(I'm not too involved in the C++ side, but googling shows Pimpl is mainly good for fast compilation. Is that the reason we're using it here, or is libcellml too small to worry about that?)

nickerso:
Pretty sure this approach was taken to keep the API clean as the main reason. @hsorby will likely correct me :)

agarny:
I am no Pimpl expert, but I understand that it's also good for binary compatibility, among other things, which is quite important when it comes to APIs such as libCellML. There are clearly advantages and disadvantages to implementing the Pimpl idiom, but I think that's besides the point here. For some reasons (I don't think I was involved in that decision), it was agreed that libCellML would implement the Pimpl idiom, so I am merely pointing out that our implementation could, I think, be improved.

hsorby:
The Pimpl idiom is being used mainly for the point of being able to control the API.  But there are other benefits like faster compilation.

Part of the keep the API clean is for binary compatibility reasons with the Pimpl idiom it is easier to add methods and not break binary compatibility.

The doAddComponent might be a hangover from when the model object had a different inheritance hierarchy.  This is a virtual function that should not to be in the public API.

I think the Qt guys are using raw pointers as well???

agarny:
> I think the Qt guys are using raw pointers as well???

[...]

PR #322, opened 2019-05-13
Entity class: replaced `getParent()` with `getParentModel()` and `getParentComponent()` (#320)

agarny:
Addresses issue #320.


agarny:
At this point, this PR includes:

- Replacement of `Entity::getParent()` with `Entity::getParentModel()` and `Entity::getParentComponent()`.
- A small fix for one of the `ComponentEntity::takeComponent()` methods.
- ~~Fixed one our Connection tests following our use of smart pointers rather than raw pointers.~~ (see commit 1860300)

What might remain to be done:
- Rename `Entity::hasParent()` to `Entity::hasParentComponent()`?
- Remove `Entity::setParent()` and `Entity::clearParent()` from the API? (This would require `Model`, `Component` and `ComponentEntity` to be friends with `Entity`...? I am not in favour of friend classes, but is there another solution?)

agarny:
@hsorby, Buildbot probably ought to check that the `test_clang_format` target exists (for the Clang Format test)?...

agarny:
I have done a bit more thinking with regards to getting rid of `Entity:: setParent()`and `Entity:: clearParent()`, and this could be achieved in a nice way by improving out Pimpl approach (see issue #279).

I mean, `Entity:: setParent()`and `Entity:: clearParent()` clearly have nothing to do in the API. Yet, using our current Pimpl approach, there is no choice but to have them. However, if our Pimpl approach was to use inheritance, we wouldn't need those methods anymore.

So, I am going to implement inheritance in our Pimpl in this PR... (Ideally, I would do this in issue #279, but we already have far too many PRs waiting to be merged in and my libCellML work is getting unnecessarily complicated as a result, so...)

agarny:
> However, if our Pimpl approach was to use inheritance, we wouldn't need those methods anymore.

Scrap that, that's not the case. :(

agarny:
Ok, I am going to leave things as they are... for now. There are several things that I don't like / think are wrong:
- `Variable` indirectly inherits from `Entity`, which means that it could have a `Model` as a parent...!?
- Conversely, because `Units` also inherits from `Entity`, a `Units` could have a `Component` as a parent...!? (It would have been fine in CellML 1.0/1.1, but not in CellML 2.0.)
- I would rename our two `Entity::setParent()` to `Entity::setParentModel()` and `Entity::setParentComponent()`.
- Likewise, I really think that `Entity::hasParent()` ought to be renamed `Entity::hasParentComponent` since this is really what we check.
- Etc.

Bottom line: it would be nice to have this PR merged in... after we have got the Clang Format test disabled when the `test_clang_format` target doesn't exit.

agarny:
FWIW, this PR is ready to be merged in...

agarny:
Ping @nickerso @hsorby

hsorby:
Your request timed out.

agarny:
> Your request timed out.

??

agarny:
@nickerso and @hsorby: coverage is working once again (thanks @hsorby!), so you are (very) kindly invited to review and merge this PR in... :)

nickerso (dismissed):

agarny:
Ping @hsorby.

agarny:
@hsorby, @nickerso and @kerimoyle... feel free to (re)review this PR... :) (@kerimoyle, I can't put you as a reviewer for now, but feel free to add yourself once you have agreed to be part of the "team".)

hsorby (commented):
With all the changes from using `libcellml::Model` to `libcellml::ModelPtr` and the like do we still have the former?  I think we still need to test these ways of creating libCellML objects.

agarny:
Yes, [`coverage/coverage.cpp`](https://github.com/agarny/libcellml/blob/issue320/tests/coverage/coverage.cpp), [`model/units_import.cpp`](https://github.com/agarny/libcellml/blob/issue320/tests/model/units_import.cpp) and [`units/units.cpp`](https://github.com/agarny/libcellml/blob/issue320/tests/units/units.cpp) still make use of `libcellml::Model`.

agarny:
Ok, I have just tested this PR against LLVM coverage and I am not getting 100% (unlike the `develop` branch), even though GCC coverage says everything is fine. So, I am going to fix that...

agarny:
Ok, back to normal! (Oopsie!)

kerimoyle:
Hi Alan
I'm just going through this PR and had a few thoughts ... 

> Variable indirectly inherits from Entity, which means that it could have a Model as a parent...!?
> Conversely, because Units also inherits from Entity, a Units could have a Component as a parent...!? (It would have been fine in CellML 1.0/1.1, but not in CellML 2.0.)

I find the use of an inheritance structure overall a bit contrived - the entities themselves do not so much "inherit" the attributes of their parent, the only have a connection to the parent which (in my head anyway) isn't really the same thing.  So it's to be expected that things like variables with model parents and units with component parents exist - that's part of the inheritance structure we have.  With that in mind, I think that it would make it better that the functions were named more clearly, as you suggest below, even though they're doing the same thing under the hood (ie: setParentModel is doing the same thing as setParentComponent, just pointing to somewhere else).

> I would rename our two Entity::setParent() to Entity::setParentModel() and Entity::setParentComponent().

I think that's a good idea ... it was the first thing I went looking for and would make reading the code cleaner.

> Likewise, I really think that Entity::hasParent() ought to be renamed Entity::hasParentComponent since this is really what we check.

As above.

agarny:
I think we will, at some point, have to review our object diagram. I will have a think about it when I am back from my soon-to-start holiday in early September... :)

agarny:
Ping @hsorby, @nickerso & @kerimoyle.

kerimoyle:
Hi @agarny.  I've had a look, and as in earlier comments think this is a great idea.  I wonder if the same could be extended to the other ambiguous functions?  eg:

```
bool Entity::hasParent(const ComponentPtr &component) const
{
    bool hasParent = false;
    ComponentPtr parentComponent = mPimpl->mParentComponent.lock();
    if (parentComponent == component) {
        hasParent = true;
    } else if (parentComponent) {
        hasParent = parentComponent->hasParent(component);
    }
    return hasParent;
}
```
Do we need to rather call it it `hasParentComponent` and create a similar sister function for the `model` parent?  Or just overload with a different `hasParent` argument of `model`?

nickerso (approved):

nickerso:
> .. Do we need to rather call it it `hasParentComponent` ...

@agarny's comment on #320 state exactly that :) but this can be done as a separate issue/pr in future. Need to keep the focus on addressing problems with the minimal changes needed so we can get to the 1.0 release.

kerimoyle (approved):
Looks fine to me if we're happy to make things more consistent (as in comments) later on.

agarny:
Two people approved, so one of you can feel free to merge this in... :)

kerimoyle:
Ah, sorry, I was leaving you the honours :) All done now.

agarny:
> Ah, sorry, I was leaving you the honours :) All done now.

Very kind of you, but the author of a PR can't merge his/her own work. Would be too easy otherwise... :)

Issue #316, opened 2019-05-08
SWIG 4.0.0 breaks our Python bindings


agarny:
The recent release of [SWIG 4.0.0](https://sourceforge.net/p/swig/news/2019/04/swig-400-released/) breaks our Python bindings. On macOS, I had to revert to a previous version (SWIG 3.04 using Homebrew) to get things to work again.

PR #333, opened 2019-05-29
Fixing Python bindings for SWIG 4.0.0

hsorby:
SWIG 4.0.0 changed the way the Python modules were imported.  This fix addresses that.

Issue #316 .


cmake/finalise_python_package.in.cmake
```diff
@@ -0,0 +1,3 @@
+
```
agarny: Delete line.

src/bindings/python/__init__.py
```diff
@@ -183,4 +181,4 @@ class Object:
     'MAP_VARIABLES_UNIQUE',
     ], new_base =  libcellml)
 del(convert, libcellml)
-%}
+
```
agarny: Delete line.

src/bindings/python/setup.in.py
```diff
@@ -0,0 +1,48 @@
...
+Operating System :: Unix
+Operating System :: MacOS :: MacOS X
+Topic :: Software Development :: Libraries :: Python Modules
+"""
+
+from setuptools import setup
+from setuptools.dist import Distribution
+
+doclines = __doc__.split("\n")
+
```
agarny, line 24: Delete line.

src/bindings/python/setup.in.py
```diff
@@ -0,0 +1,48 @@
...
+
+class BinaryDistribution(Distribution):
+    def is_pure(self):
+        return False
+
+    def has_ext_modules(self):
+        return True
+
```
agarny, line 32: Delete line.

src/bindings/python/setup.in.py
```diff
@@ -0,0 +1,48 @@
...
+
+setup(
+    name='@PYPI_PACKAGE_NAME@',
+    version='@libCellML_VERSION@@LIBCELLML_DEVELOPER_VERSION@',
+    author='H. Sorby',
+    author_email='h.sorby@auckland.ac.nz',
```
agarny: Wouldn't it be best to have a generic email address for libCellML? Anything that would be emailed to it would then be forwarded to some/all of us?

src/bindings/python/setup.in.py
```diff
@@ -0,0 +1,48 @@
...
+    packages=['libcellml'],
+    package_data={'libcellml': [@SETUP_PY_PACKAGE_FILES_STR@]},
+    url='http://cellml.org',
+    license='Apache Software License',
+    description=doclines[0],
+    classifiers = filter(None, classifiers.split("\n")),
+#    long_description=open('README.rst').read(),
```
agarny: Delete line.

agarny:
I must confess that I don't know enough about the intricacies of Swig to tell whether this commit is "right". All I can say is that it passes CI and that, in some way, I am ok with it. Just a few very minor suggestions.

agarny (changes requested):
Besides my very minor editing suggestions, I am OK with this PR.

src/bindings/python/setup.in.py
hsorby: Sure, is there one?

agarny: I am not aware of one for the libCellML project, but that's maybe something we could look into (@nickerso?). However, this should clearly not block this PR.

src/bindings/python/setup.in.py
hsorby: I left this there as a reminder to write the README, I will add this in.

src/bindings/python/setup.in.py
```diff
@@ -0,0 +1,49 @@
...
+    name='@PYPI_PACKAGE_NAME@',
+    version='@libCellML_VERSION@@LIBCELLML_DEVELOPER_VERSION@',
+    author='H. Sorby',
+    author_email='h.sorby@auckland.ac.nz',
+    packages=['libcellml'],
+    package_data={'libcellml': [@SETUP_PY_PACKAGE_FILES_STR@]},
+    url='http://cellml.org',
+    license='Apache Software License',
+    description=doclines[0],
+    classifiers = filter(None, classifiers.split("\n")),
```
agarny: No spaces before/after `=` (to be consistent with the rest of the code)?

src/bindings/python/setup.in.py
agarny: This will be resolved, if at all, as part of issue #336, so consider this conversation as resolved.

agarny (approved):

agarny:
Now that we have the libcellml@googlegroups.com mailing list, maybe we could update `src/bindings/python/setup.in.py`? However, since we have issue #336, I am happy for this PR to be merged in first.

src/bindings/python/setup.in.py
```diff
@@ -34,8 +34,8 @@ def has_ext_modules(self):
 setup(
     name='@PYPI_PACKAGE_NAME@',
     version='@libCellML_VERSION@@LIBCELLML_DEVELOPER_VERSION@',
-    author='H. Sorby',
-    author_email='h.sorby@auckland.ac.nz',
+    author='libCellML Developers',
```
agarny, line 37: I would put `developers` rather than `Developers`, but that's probably just me. Ok, I know it's just me! :)

nickerso (approved):

src/bindings/python/setup.in.py
```diff
@@ -0,0 +1,49 @@
...
+
+setup(
+    name='@PYPI_PACKAGE_NAME@',
+    version='@libCellML_VERSION@@LIBCELLML_DEVELOPER_VERSION@',
+    author='libCellML Developers',
+    author_email='libcellml@googlegroups.com',
+    packages=['libcellml'],
+    package_data={'libcellml': [@SETUP_PY_PACKAGE_FILES_STR@]},
+    url='http://cellml.org',
+    license='Apache Software License',
```
agarny, line 42: Shouldn't this simply read `Apache 2.0` rather than `Apache Software License`?

hsorby: This comes from the [Python trove classifiers](https://pypi.org/classifiers/) we want to follow standards I think.

agarny: Sure, but then how do we distinguish between Apache 1 and Apache 2?...

hsorby: We don't and I don't think it is something that we need to do.

agarny: Ok, fair enough. I guess I am happy with whatever standard is followed by the Python community.

Issue #341, opened 2019-06-09
Use of Shebang line in Python scripts.


hsorby:
When we use the shebang line from the Python scripts we open ourselves up to the situation where the file system is Windows and the execution environment is Unix-based.  This is most readily seen when using the Docker image for documentation on a Windows OS.

A solution to this would be to directly drive the Python script with a specific Python interpreter and thus remove the execution of the shebang.

MichaelClerx:
> When we use the shebang line from the Python scripts we open ourselves up to the situation where the file system is Windows and the execution environment is Unix-based.

Doesn't windows just ignore the shebang as a comment?

> with a specific Python interpreter

Wouldn't make it specific, just `python` !

hsorby:
Yes Windows does but with a Docker image that is based in Linux accessing files stored on a Windows OS where the line ending are `\r\n` the executable the Docker image tries to run is `python\r` which is a problem.

Also you can find that the Python that you are using to run something like the coverage test is not the Python you used to build the Python tests against.

PR #342, opened 2019-06-09
Avoid using shebang in Python scripts. 

hsorby:
Addresses #341.


agarny (approved):
Looks good to me.

nickerso (approved):
I can confirm this fixes the issues I was seeing with a windows file system mounted under the docker images.

Issue #321, opened 2019-05-10
Variable: add an `isEquivalentVariable()` method


agarny:
Right now, we have all sorts of methods dealing with equivalent variables, but there is none to actually checking whether a variable is equivalent to another through several layers of mapping.

For example, in [this model](https://github.com/cellml/libcellml/files/3164762/two_variables_of_integration.txt):
- `main.time` and `sub.time` would be equivalent;
- `main.time` and `sub_sub.time` would be equivalent; and
- `main.time` and `sub_sub_sub.time` would **NOT** be equivalent.

A quick implementation of such a method can be found [here](https://github.com/agarny/libcellml/commit/742be58f2df24803b1112d616651228c3b90b77b).

kerimoyle:
Hi Alan
There's code in the validator (or will be when the 304 PR goes through) to check this - should it be moved? or removed? or something else?

agarny:
If we agree that such a method should be in libCellML (@hsorby, @nickerso, @MichaelClerx?) then I will create a PR for it.

hsorby:
I think that this is a failing of the `hasEquivalentVariable` method, in my opinion this method should return `true` for this model.

agarny:
I wasn't sure whether `hasEquivalentVariable()` was meant to do what it currently does or whether it should have done what "my" `isEquivalentVariable()` method does. Ok, I am going to modify `hasEquivalentVariable()` accordingly and generate a PR.

MichaelClerx:
Is this done then?

hsorby:
Yes I believe it is.  Please do re-open if this is not the case.

PR #343, opened 2019-06-11
General: replace `getXxx()` with `xxx()` (#325)

agarny:
Addresses issue #325.

(Still need to update the doc diagrams.)

PR #344, opened 2019-06-12
Variable: improve our `hasEquivalentVariable()` method (#321)

agarny:
Addresses issue #321.


src/api/libcellml/variable.h
```diff
@@ -208,6 +208,20 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
+     * Tests to see if the argument variable is equivalent to this variable.
+     * Returns @c true if the argument variable is equivalent to this variable
+     * and @c false otherwise.
+     *
+     * @param equivalentVariable The variable to check whether it is equivalent this variable.
+     *
+     * @return @c true if the @p equivalentVariable is equivalent to this variable
+     * and @c false otherwise.
+     */
+    bool isEquivalentVariable(const VariablePtr &equivalentVariable) const;
```
hsorby: This is exactly the same as `hasEquivalentVariable`?? We don't need both just a properly working one.

agarny: Oops, my bad, I indeed completely overlooked your ["old" comment](https://github.com/cellml/libcellml/issues/321#issuecomment-491995223).

Ok, I am going to remove the contents of the "old" `hasEquivalentVariable()` and replace it with my new one.

hsorby: Sounds good.

src/api/libcellml/variable.h
```diff
@@ -205,22 +205,25 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
      *
      * @return @c true if the @p equivalentVariable is in this variable's equivalent variables
      * and @c false otherwise.
+     *
+     * @warning This method is currently needed by our Printer and Validator classes, but this is
+     * not a method that a libCellML user should ever need. It will therefore be removed at some
+     * point.
```
hsorby: I don't think this belongs in the API documentation.  I would prefer it if we marked the method as deprecated instead.

src/component.cpp
```diff
@@ -34,11 +34,12 @@ namespace libcellml {
 struct Component::ComponentImpl
 {
     std::string mMath;
+    std::vector<VariablePtr> mVariables;
+    std::vector<ResetPtr> mResets;
```
hsorby, line 37: resets before variables?

hsorby (commented):
Not keen on removing the documentation from the private methods I think this is worthwhile keeping.

src/api/libcellml/variable.h
agarny: Good point, I hadn't thought of that special command. Consider it done.

src/component.cpp
agarny: Indeed. Consider it corrected.

agarny:
> Not keen on removing the documentation from the private methods I think this is worthwhile keeping.

It's just that we are not being consistent throughout and, I must confess, that I find it to get in the way. So, I thought I would "clean up" all the private implementation declarations by removing their corresponding documentation...?

nickerso:
Definitely wouldn't pollute a PR with all the unrelated changes by removing the implementation documentation. Best to keep the PR focused on the issue. Cleaning up (or not) documentation of the implementation classes should be a separate issue.

agarny:
> Definitely wouldn't pollute a PR with all the unrelated changes by removing the implementation documentation. Best to keep the PR focused on the issue. Cleaning up (or not) documentation of the implementation classes should be a separate issue.

Fair enough. I am going to put the private documentation back in...

agarny:
Ok, I believe this is now ready.

src/api/libcellml/variable.h
```diff
@@ -205,6 +205,22 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
+     * not a method that a libCellML user should ever need. It will therefore be removed at some point.
+     */
+    bool hasDirectEquivalentVariable(const VariablePtr &equivalentVariable) const;
+
+    /**
+     * @brief Test whether the argument variable is equivalent to this variable.
+     *
+     * Tests to see if the argument variable is equivalent to this variable. Returns @c true if
+     * the argument variable is equivalent to this variable and @c false otherwise.
+     *
```
nickerso, line 219: do we need to document what we mean by "equivalent" or add a see-also link to somewhere that is already defined?

nickerso (dismissed):
minor comment on the documentation (which can be addressed separately), but otherwise this looks good and the tests are demonstrating the behaviour that I would expect.

src/api/libcellml/variable.h
nickerso: happy for this to be bumped to a separate issue.

agarny: I would rather have this addressed in a different issue.

agarny:
FWIW, PR #343 should be merged in before this one (I still need to "fix" a couple of things in PR #343 first though).

agarny:
@hsorby, am I correct in assuming that, for some reasons, the coverage test didn't work properly and that I probably ought to retrigger the commit?

hsorby:
Something has gone wrong??? I'll have a look.

agarny:
> Something has gone wrong??? I'll have a look.

Well, yes:

```
------------------------------------------------------------------------------
                           libCellML Code Coverage Report
Directory: /home/libcellml/bbslave/Coverage_Tests_Builder_Ubuntu_18_04_64_Bit/libcellml-build/src/CMakeFiles/cellml.dir
------------------------------------------------------------------------------
File                                       Lines    Exec  Cover   Missing
------------------------------------------------------------------------------
------------------------------------------------------------------------------
TOTAL                                          0       0    --%
------------------------------------------------------------------------------
```

I redelivered the commit earlier, but no difference.

agarny:
@nickerso and @hsorby: coverage is working once again (thanks @hsorby!), so you are (very) kindly invited to review and merge this PR in... :)

agarny:
Ping @hsorby.

tests/variable/variable.cpp
```diff
@@ -386,6 +386,154 @@ TEST(Variable, getNullParentComponent)
...
 
+TEST(Variable, hasDirectEquivalentVariable)
+{
+    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
+    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
+    libcellml::VariablePtr v3 = std::make_shared<libcellml::Variable>();
+    libcellml::VariablePtr v4 = std::make_shared<libcellml::Variable>();
+    libcellml::VariablePtr v5 = std::make_shared<libcellml::Variable>();
+
+    EXPECT_FALSE(v1->hasDirectEquivalentVariable(v1));
```
hsorby, line 395: This test is way to verbose.  It should just do the minimal required to check for the direct equivalence.  This doesn't require 5 variables.

tests/variable/variable.cpp
```diff
@@ -386,6 +386,154 @@ TEST(Variable, getNullParentComponent)
...
+
+    EXPECT_FALSE(v5->hasDirectEquivalentVariable(v1));
+    EXPECT_FALSE(v5->hasDirectEquivalentVariable(v2));
+    EXPECT_FALSE(v5->hasDirectEquivalentVariable(v3));
+    EXPECT_TRUE(v5->hasDirectEquivalentVariable(v4));
+    EXPECT_FALSE(v5->hasDirectEquivalentVariable(v5));
+}
+
+TEST(Variable, hasEquivalentVariable)
+{
```
hsorby, line 424: Same as direct equivalent test too verbose.

tests/variable/variable.cpp
agarny: Agreed, 3 variables are good enough. Addressed in 941c794.

tests/variable/variable.cpp
agarny: Agreed, 3 variables are good enough. Addressed in 941c794.

agarny:
@hsorby, @nickerso and @kerimoyle... feel free to (re)review this PR... :) (@kerimoyle, I can't put you as a reviewer for now, but feel free to add yourself once you have agreed to be part of the "team".)

kerimoyle:
Am I right in thinking that when this goes through the method of variable validation should change? And how does the stuff which @hsorby is working on in #350 fit in with this? Sorry to come late to the conversation, just trying to figure out how all the bits work together ...

agarny:
Yes, basically whenever a PR gets merged in to `develop`, you *must* merge it in your own PRs and make sure that everything is as you think it should be. It's not necessarily only a matter of merging in and resolving conflicts (should there be any). It may also involve making sure that the new code looks and behaves the way you think it should.

So, when it comes to PRs, the first one that gets merged in "wins" and the others have to be updated so that they can be safely merged afterwards.

As it happens, @nickerso modified the repository's settings to enforce that rule.

nickerso (dismissed):
In general it makes it hard to review the actual new code when there are a bunch of unrelated code cleaning commits mixed into a PR like this. Would be good to keep them separated (and help with more speedy reviews).

agarny:
Agreed, I did that sometime ago and, yesterday, I was wondering whether I should have removed it...

kerimoyle:
Sorry @agarny for the delay - my questions wasn't about the mechanics of merging, it was about the intention of this branch and what @hsorby is trying to do on the other variable-related branch at PR #350 ...?

agarny:
Ok, I shall then leave it to him to get back to you although it might probably take some time since we have a big deadline coming up this Friday...

agarny:
Ping @hsorby, @nickerso & @kerimoyle.

kerimoyle:
Hi again @agarny :) I'm still not clear on the changes that @hsorby is making to variables, which is what this PR and the other variable-related ones (#350) are waiting for ...

nickerso (approved):

nickerso:
> Hi again @agarny :) I'm still not clear on the changes that @hsorby is making to variables, which is what this PR and the other variable-related ones (#350) are waiting for ...

This PR is focused (mostly) on addressing the need to distinguish between when a variable is directly connected to another variable versus the case when two variables are in the same set of connected variables (as we do not represent connections in the libCellML object model). Which is different to the issue @hsorby has mentioned on #350, which is a more fundamental issue with the `Variable` class definition itself. I think, its not completely clear to me :)

nickerso:
and maybe worth pointing out that we mark this direct equivalence method as deprecated from the start as we know this is not the ideal way to achieve this, but it is the pragmatic solution to allow libCellML to progress.

kerimoyle:
... wait ... now I'm more confused ... so there's a difference in libcellml interpretation between these two situations?

```
a is equivalent to b
b is equivalent to c
```

and

```
a is equivalent to b
a is equivalent to c
```

If so I need to rethink some stuff ... I've been treating them as all being part of the same set, no matter how/which direction the equivalence was specified from?

nickerso:
No, you are treating them correctly.

The initial issue that brought this up was that previously libCellML would have incorrectly said that `a` is not equivalent to `c` - fixing that raised an edge case in printing and validating models where you actually need to know about the `connection` elements in the CellML XML - which are not captured in the libCellML data model. So the idea of a "direct equivalence" is a hack to know that there is an actual connection between two variables.

In general, it should be sufficient to know that variables are in the same equivalence set regardless of how many layers of connections may exist between them.

kerimoyle:
So the methods below are there to maintain the same functionality as was previously there (albeit under different function names), but are deprecated and not really expected to be used?  ... in which case, why not just remove them completely?

Or will there ever be a situation where the direct connection is important? ... in which case, what is it?

```
bool Variable::hasDirectEquivalentVariable(const VariablePtr &equivalentVariable) const
{
    return mPimpl->hasDirectEquivalentVariable(equivalentVariable);
}

bool Variable::VariableImpl::hasDirectEquivalentVariable(const VariablePtr &equivalentVariable) const
{
    auto it = findEquivalentVariable(equivalentVariable);
    if (it == mEquivalentVariables.end()) {
        return false;
    }
    return !it->expired();
}
```

nickerso:
Sorry, slowly coming back to me :-) the need for the direct connection
method will always exist but it's presence in the public API is deprecated
and shouldn't be used outside libcellml. The current object model forces us
to include this in the public API for now in order to use this method
internally.

The most obvious need is when printing a model to CellML XML. The printer
needs to know how and when to serialise connection elements.

kerimoyle:
OK, got it.  Thanks!

kerimoyle (approved):
Looks ok to me.  (At the lowest possible level of importance, it would be good to decide what structure our comments have, if any.  For eg, there are instances of "Test ... " and "Tests ..." here in the same function definition.  A job for another PR after all the upcoming stuff is finally merged in!)

agarny:
Two people approved, so one of you can feel free to merge this in... :)

Issue #324, opened 2019-05-15
Hardcoded units in another file?


kerimoyle:
I'd like to avoid duplicating hardcoded lists/dictionaries if possible.  Currently there are string vectors in the validator in functions:
- `isStandardPrefixName`
- `isStandardUnitName`
- `unitsAreEquivalent`

I'd like to either move them to another file or be able to share them, but they aren't enums so the enumerations.h file didn't seem like the right place.  Any suggestions?

```
Inside isStandardUnitName(const std::string &name)
    std::vector<std::string> standardUnitNames =  {
        "ampere", "becquerel", "candela", "celsius", "coulomb", "dimensionless", "farad", "gram", "gray",
        "henry", "hertz", "joule", "katal", "kelvin", "kilogram", "liter", "litre", "lumen", "lux",
        "meter", "metre", "mole", "newton", "ohm", "pascal", "radian", "second", "siemens", "sievert",
        "steradian", "tesla", "volt", "watt", "weber"
    };
   ... etc ... 

Inside isStandardPrefixName(const std::string &name)
    std::vector<std::string> prefixNames =
    {
        "atto", "centi", "deca", "deci", "exa", "femto", "giga", "hecto", "kilo", "mega", "micro", "milli",
        "nano", "peta", "pico", "tera", "yocto", "yotta", "zepto", "zetta"
    };
... etc ...

Inside unitsAreEquivalent()

standardList["farad"]["metre"] = -2.0;
standardList["farad"]["kilogram"] = -1.0;
standardList["farad"]["second"] = -4.0;
standardList["farad"]["ampere"] = 2.0;
    ... 
prefixList["yotta"] = 24;
prefixList["zetta"] = 21;
prefixList["exa"] = 18;
prefixList["peta"] = 15;
... etc ...
```

agarny:
What about `utilities.[cpp|h]`?

hsorby:
Or if you don't consider them to be utilities then `common.[cpp|h]`.  I think `utilities` is the best home though.

kerimoyle:
Thanks both, I'll move to utilities :)

kerimoyle:
A couple of things:
- do we want to include "celsius" in the list of standard units?  It (was) in the current `isStandardUnitName `function test but not in the specification list.  And if so, I'll make it equivalent to kelvin?
[...]

PR #348, opened 2019-06-13
Issue 324 Moved hardcoded definitions to utilities.h

kerimoyle:
- hardcoded built-in and standard units, prefixes, and mathml tags are moved to the common.h file
- celsius is removed from builtin units list
- tests updated to remove celsius
- .gitignore changed to exclude VSCode files

Addresses issue #324 and its comments


agarny:
@kerimoyle, it looks like there are some conflicting files (according to GitHub at least!). Would be good for you to resolve them before I review your PR.

kerimoyle:
@agarny Good to go :)

agarny (approved):
@kerimoyle, as you will have seen, I have edited your PR a bit, so that the code looks consistent with the rest of libCellML's code. I also put the contents of `common.h` into `utilities.h`, as originally suggested by @hsorby and myself.

So, now, as this PR is fine for me to be merged in (even though it does contain things that are not currently used...  I am talking about `standardPrefixList`, for example).

nickerso (approved):

kerimoyle:
> So, now, as this PR is fine for me to be merged in (even though it does contain things that are not currently used... I am talking about standardPrefixList, for example).

Thanks @agarny ... but I don't understand what you mean about not being used?  That array is used in the `isStandardPrefixName` function?  Or did that not make it?

```
bool Validator::ValidatorImpl::isStandardPrefixName(const std::string &name)
{
    bool result = false;
    if (standardPrefixList.count(name) != 0) {
        result = true;
    }
    return result;
}
```

agarny:
Sorry, I didn't mean `standardPrefixList`, but `standardMultiplierList`. As far as I can tell, it's not used anywhere, at this stage.

PR #349, opened 2019-06-13
Issue318 units prefix to integer

kerimoyle:
Second PR from large #304 PR.  

Changing the overload of prefix in the units to be integers instead of doubles.  

Python overloading of the Prefix enum with integers have been removed - even though the enum is now safe, it's still ambiguous so have removed the Prefix in favour of the integer as it's more flexible.  

Addresses issue #297.


agarny:
@kerimoyle, it looks like there are some conflicting files (according to GitHub at least!). Would be good for you to resolve them before I review your PR.

kerimoyle:
Done @agarny you should be good to go ... :)

nickerso (approved):
Not quite sure I understand the implications of the python API, but otherwise looking good to me. Could perhaps remove the commented out code rather than leaving it laying around but perhaps in this case it adds value.

kerimoyle:
> Not quite sure I understand the implications of the python API, but otherwise looking good to me. 

The main implication is that Python users will need to specify prefixes either with strings (eg: milli, micro etc) or with the integer that corresponds to the power of 10 (eg: -3, -6 etc).  They won't be able to use the enum Prefix structure.  

> Could perhaps remove the commented out code rather than leaving it laying around but perhaps in this case it adds value.

Happy to remove it - I only left it there for the "before" picture so we could track what had changed until others were happy overall.

agarny (changes requested):
Looks good to me, but... I would remove any code that is commented out. It has nothing to do in the `develop` branch.

kerimoyle:
@agarny @nickerso  Commented out code removed as suggested

agarny:
> @agarny @nickerso Commented out code removed as suggested

There was still some left (I appreciate that it wasn't of the same kind, but it should still not be merged in), which I have removed. So, since @nickerso is happy with this PR and that I only wanted those commented out code to be removed, I am going to merge this in, once I have got the go-ahead from CI.

agarny (approved):

Issue #295, opened 2019-04-02
Equivalent variables: cycles, interfaces, and name convention?


kerimoyle:
See **19.10.5**: No cycles permitted in the variable equivalence network.  Currently the validator does not catch the cyclical definition created in the TEST(Validator, validateInvalidConnections) test:
```
    libcellml::Variable::addEquivalence(v1_1, v2);    // cycle? created between v1_1, v2, v3
    libcellml::Variable::addEquivalence(v1_2, v2);
    libcellml::Variable::addEquivalence(v1_1, v3);    // cycle? created between v1_1, v2, v3
    libcellml::Variable::addEquivalence(v1_1, v4);
    libcellml::Variable::addEquivalence(v2, v3);       // cycle? created between v1_1, v2, v3
    libcellml::Variable::addEquivalence(v1_1, v3);    // duplicate, skips silently
```
I figure we can test this in the Validator::validateModel function, but wondered whether it should also be checked/prevented in the Variable::addEquivalence function too?

The specifications talk about "connections" and "map_variables" but the code talks about "equivalent variables".  I presume that "mapping" a variable to another the same as making them "equivalent", or are there more subtleties there?

nickerso:
We want to ensure that users of the API are able to create the models they need to create without continually ensuring they have a valid model. For example, when modifying a model there could be many times when the model will be invalid, but we only want to check for a valid model when requested.

We have also tried to develop the libCellML data model to align with what we see as the underlying CellML data model, rather than the XML serialisation. So, yes, the variable equivalences are what capture the information also represented by the connections in the XML serialisation. Perhaps the only subtlety is that there is (I think) a bit of code tracking when equivalences originate from the same connection element in parsing the XML, and we need some way to track id's in case users have annotated the connection/map_variable elements.

kerimoyle:
Ok, thanks, I think I've got it.  I'd missed the significance of the bit which said that there was only one connection pair allowed between components - that makes it all a lot simpler.  Getting there (slowly!).  I'll leave this issue open until the cycle validation is added.

kerimoyle:
Equivalent variables unit testing is working, including nested units and arbitrary powers.  It returns the mismatch in base units where they don't match - needs a tidy around error messages and for speed.  
But I noticed that the specs (19.10.6 and .7) say that multipliers don't need to match ... is that true when the model is validated too?  ie: we'll permit models with equivalent variables with different prefixes or multipliers? mm + cm is ok?

[...]

Issue #326, opened 2019-05-21
12.1.1.2 Validator does not check for unique reset order 


kerimoyle:
See 12.1.1.2:  we currently only check that a `Reset` object has an order set as an integer, but we don't check that it is unique among the referenced `Variable`'s equivalent variables.

nickerso:
Until we finalise resets, I wouldn't worry about adding any further validation.

kerimoyle:
OK, understood (but we must be getting close now, @MichaelClerx ?) 

Will just put this here so I'm reminded later ...

Currently this code checks for unique reset orders within a component ... but this doesn't seem to be something dealt with in the specifications?  Inside `void Validator::ValidatorImpl::validateComponent(const ComponentPtr &component):`
```
std::vector<int> resetOrders;
        for (size_t i = 0; i < component->resetCount(); ++i) {
            ResetPtr reset = component->getReset(i);
            int resetOrder = reset->getOrder();
            if (reset->isOrderSet()) {
                if (std::find(resetOrders.begin(), resetOrders.end(), resetOrder) != resetOrders.end()) {
                    ErrorPtr err = std::make_shared<Error>();
                    err->setDescription("Component '" + component->getName() +
                                        "' contains multiple resets with order '" + convertIntToString(resetOrder) + "'.");
                    err->setComponent(component);
                    err->setRule(SpecificationRule::RESET_ORDER);
                    mValidator->addError(err);
                } else {
                    resetOrders.push_back(resetOrder);
                }
            }
        }
```

The specs say:

> The value of the order attribute MUST be an integer string. The value of the order attribute MUST be unique for all reset elements for variables that are in the same **connected variable set** (see Interpretation of map_variables).

I'd thought that the "connected variable set" was the group of equivalent variables ... but is it just the variables inside a component?

kerimoyle:
Reset validation is removed from the #304 PR, but have kept the code in a separate branch here: https://github.com/kerimoyle/libcellml-keri/tree/reset_validation
... which includes validation for the subject of this issue.

PR #350, opened 2019-06-13
Issue295 validate variables and Issue326 reset order uniqueness

kerimoyle:
Third in a series of smaller PRs replacing the previous #304 PR.  

Addresses #295 and #326:
- Cyclic dependencies are tested in variables and their equivalent network
- Reset order is tested within the equivalent variable network.  NB that this will need to be adjusted after #358 is merged in as the test format will change.


kerimoyle:
@agarny @hsorby @nickerso Ready for review please.

kerimoyle:
As per @hsorby's suggestion in #351 about variable equivalence to self, this test currently returns no validation errors.  I think it should because variables are not allowed (see 17.1.2) to be connected to others in the same component ... will change the validator here to adjust that. 

```
TEST(Validator, variableEquivalentToSelf)
{
    libcellml::ModelPtr model = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr component = std::make_shared<libcellml::Component>();
    libcellml::VariablePtr variable = std::make_shared<libcellml::Variable>();
    libcellml::Printer printer;
    variable->setName("variable");
    variable->setUnits("dimensionless");
    component->setName("component");
    component->addVariable(variable);
    model->setName("modelName");
    libcellml::Variable::addEquivalence(variable, variable);
    validator.validateModel(model);  // <<<< no errors found at the moment
    printErrors(validator);  // <<<< no errors reported
    std::cout << printer.printModel(model);
}
```
The model printed above is:
```
<?xml version="1.0" encoding="UTF-8"?>
<model xmlns="http://www.cellml.org/cellml/2.0#" name="modelName">
  <component name="component">
    <variable name="variable" units="dimensionless"/>
  </component>
  <connection component_1="component" component_2="component">
    <map_variables variable_1="variable" variable_2="variable"/>
  </connection>
</model>
```

kerimoyle:
Added the tests and validator changes as requested in #351

kerimoyle (commented):
Initialising with an empty string at the time of declaration gives me a clang_format error ... hence the move to afterwards.  I think by default strings are empty nowadays, so the initialisation is unneeded in any case?

kerimoyle:
HI @hsorby A bunch of changes I made have now been overwritten by your merge :( how best to get them back, or was there something wrong with them ... ?

hsorby:
I did check that change with clang_format and it was happy on this machine.  Can you show me what changes have disappeared?  I can't see what has been lost on this end.

hsorby:
Ahh, clang tidy error not clang format.

tests/validator/validator.cpp
```diff
@@ -897,7 +924,7 @@ static const std::string emptyMath = "<math xmlns=\"http://www.w3.org/1998/Math/
 TEST(Validator, resets)
 {
     const std::vector<std::string> expectedErrors = {
-        "Component 'comp' contains multiple resets with order '300'.",
+        // "Component 'comp' contains multiple resets with order '300'.",
```
kerimoyle: eg here: should be deleted as test is just wrong ... resets must have unique order within variable set, not component.

tests/validator/validator.cpp
```diff
@@ -755,52 +755,79 @@ TEST(Validator, validateInvalidConnections)
     libcellml::ComponentPtr comp2 = std::make_shared<libcellml::Component>();
     libcellml::ComponentPtr comp3 = std::make_shared<libcellml::Component>();
     libcellml::ComponentPtr comp4 = std::make_shared<libcellml::Component>();
+    libcellml::ComponentPtr comp5 = std::make_shared<libcellml::Component>();
```
kerimoyle: also eg here: had simplified this test ?

kerimoyle:
Changes look to be in the test/validator.cpp doc (at least at first glance) from https://github.com/cellml/libcellml/pull/350/files/4fb2bec4809edc4d91c0e589ff93d0fe5c0a935b .  I'd split out some of the tests to make them simpler but now are back as previously (I think).  eg: validateInvalidConnections (had simplified it already) and resets (first line of error should be removed entirely).  Some of the changes from that commit are valid and should be included though, so not sure why some have made it and others not ...

hsorby:
Can I ask why this PR is addressing two separate issues?

kerimoyle:
> Can I ask why this PR is addressing two separate issues?

Because the design of the checks for order relies on the network of equivalent variables, so it seemed better to put them together than to do all kinds of weird workarounds to make them work separately.  You could think of them both in terms of "validating equivalent variables network" if you like ...

hsorby:
I think it would be just fine sorting out the validation of variables and then as a separate task sort out reset order uniqueness.

kerimoyle:
Basically that means removing the order reset testing from here, and
waiting until all the other PRs are merged.  It doesn't make sense to test
for order uniqueness in a variable set until after the variable set is
checked... Is that a better plan?

hsorby:
Not sure,  I really don't like what I have done with enabling shared from this on the variable class and I am trying to find a way around it.  I haven't gotten around to looking at the order uniqueness, I only noticed this recently and wanted to query why it was as it is.

kerimoyle:
Would it be possible to put the variable changes you're working on in another branch instead of here so that this (and then a separate order checking one if needed) can be merged?  I didn't understand the background for your changes so don't really know what their implications are?

hsorby:
I wouldn't merge this in if it didn't at least have the idea that my changes are trying to effect.

kerimoyle:
... but I still have no background for what your changes were trying to do or why? They just appeared out of the blue so I'm a bit confused as to what's happening?

kerimoyle:
@hsorby any updates on this one?

hsorby:
Still haven't solved the issue with avoiding enable from shared.  Trying some out some new theories.

kerimoyle:
I still don't understand why it needs to be in this PR though? Or which
issue it's for? Sorry, a bit confused here ...!

kerimoyle:
@hsorby Just checking in on this one? I'd really like to get this and the reset branch merge in so that we can all move on ...

kerimoyle:
I'm so confused as to what's happening here ... why don't we need to check for cyclical variables any more?  Why are you adding more tests for resets when they are about to be changed in format and the tests deleted?  I don't get it ... :|

hsorby:
We do want to check for cyclical variables but I removed it because some change that has been made (possibly through the changes from the code generation code) has meant that that code has stopped being triggered.  Since it was not being triggered I removed it and tagged it in a message so that it could be put back when and if we decided that it was required.

hsorby:
The reset tests I have added are to do with the order of the reset which is staying so those tests will still need to be there.  I have left in the when stuff as little as possible knowing that it is about to go down the pan.

hsorby:
I'm not convinced that cyclic variables is actually the problem.  I think what we should be thinking about is the variable interfaces and what that does and doesn't allow.

kerimoyle:
OK, that helps ... I think... I suspect that the cycle checking disappeared when your branch was merged on top of my one - it was the whole point of the issue which prompted this branch, so it's a wee bit odd if it's gone altogether!  We can get around the reset format thing if the other outstanding branch #355 is merged in first?

hsorby:
I think we are in trouble either way, I would prefer to get this one out of the way first and then work on #355.

kerimoyle:
They were written to be separate from each other as much as possible ... fingers crossed it's not too bad!

hsorby (dismissed):
I'm approving the contents of this PR, assuming of course that the code passes its checks I can't be bothered waiting for it.

I would like to see the handling of the cyclic nature of variables revisited as I don't believe that is really what the CellML specification says.  Some more discussion on this is needed (for me at least).  

I notice that I have already closed the issue #295 and I'm not sure why I did that, oh well.

hsorby:
I am aware that a few warnings have cropped up from Visual Studio.  I would like to push this to separate issue and raise it around making windows builds flag warnings as errors like UNIX based builds do.

kerimoyle:
I've looked more closely at the spec around variables/interfaces/directionality, in particular 19.10.8 (below), and had these thoughts:

- From what I read there, I think that the variables are an *undirected* graph, but one with further restrictions on which connections are actually allowed (something which we do not yet check as encapsulation validation is not present). 
    - If that's true, then I'd like to re-include the code and tests which were removed from here which checks for cycles, and add a further check that addresses the encapsulation (which we're also not yet validating) and interface validation together - possibly in a separate branch, if that's cleaner?
    - I don't think this branch makes much sense with the cycle checking removed, that was the point of it in the first place ...

- It's possible to have a network of equivalent variables without having a real hierarchy in the encapsulation (or any encapsulation at all?) ... like between sibling components which would mean that all the interfaces were public?  That means that there could be no directional information taken from the interfaces anyway?

- The way which we currently store variable equivalence means that there is no way to recognise directionality ... if there should be a directionality to the wider network, such a network cannot be made of un-directional connections.  If you want to include directionality now, this will affect interpretation of previous models?

> For a given variable, the available interfaces SHALL be determined by the interface attribute information item on the corresponding variable element as follows.
> A value of public specifies that the variable has a public interface.
> A value of private specifies that the variable has a private interface.
> A value of public_and_private specifies that the variable has both a public and a private interface.
> A value of none specifies that the variable has no interface.
> If the interface attribute information item is absent, then the variable has no interface.
> The applicable interfaces for variables A and B SHALL be defined as follows.
> When the parent component element of variable A is in the sibling set of the parent component element of variable B, the applicable interface for both variables A and B SHALL be the public interface.
> When the parent component element of variable A is in the encapsulated set of the parent component element of variable B, the applicable interface for variable A SHALL be the public interface, and the applicable interface for variable B SHALL be the private interface.
> When the parent component element of variable B is in the encapsulated set of the parent component element of variable A, the applicable interface for variable A SHALL be the private interface, and the applicable interface for variable B SHALL be the public interface.
> CellML models MUST NOT contain a map_variables element where the applicable interface of variable A or B is not defined or is not an available interface.

hsorby:
I think you will find that there is no encapsulation validation because it is not possible to create an invalid encapsulation through the API.  Also the parser is not able to re-create an invalid encapsulation from a file.  So there is no need for encapsulation validation in the validator.

hsorby:
I don't think directionality is really what we need to think of here and possibly the wrong term to use, what I think we have is a set of connected variables and since it is a set it has no graph and no network defined on its own.  The directionality comes from the topology of the parent components that the variables in the set belong too so I think directionality is the wrong term.  Possibly the minimally open interface is a better way of thinking about it.  So the task as I see it is to validate whether the connections as defined in the set of connected variables is actual viable given the topology of the parent components.   There is no consideration of directionality there.  Once this is done we can then define the minimally open interface that allows the connected variables to be connected.

kerimoyle:
So was there anything wrong with what I proposed above?  I'm trying to avoid doing work which is not needed or will be deleted so it would be good to know ahead of time.  I am happy to make a separate branch to do my stuff on and leave this one for you if you'd rather, but it's really tricky to know what to do with both of us working here ...

> I think you will find that there is no encapsulation validation because it is not possible to create an invalid encapsulation through the API. Also the parser is not able to re-create an invalid encapsulation from a file. So there is no need for encapsulation validation in the validator.

... that might be true, but it's the encapsulation structure which defines which interface types are available to the mapped variable pairs, and that's what we're not testing at the moment?

hsorby:
I won't promise that this work won't be deleted but I don't think it will be.

There is no need to check for cycles it will not be possible to create one.  The validation that we need is to make sure that the connections that are made can be mapped across the component hierarchy given the interfaces that we have.  I am not working on this as I thought you were.  I am making a mess in other places.

kerimoyle:
> There is no need to check for cycles it will not be possible to create one.

Sorry, I don't get it :(  What - if anything - would you expect the validator to throw in the situation below?  The three components are siblings, and thus have public interfaces with one another?  Or should the variable equivalence be prevented somehow??  Does anyone else have thoughts here? @agarny @MichaelClerx @nickerso 

```
TEST(Validator, validateNoCyclesSimple)
{
    libcellml::Validator v;
    libcellml::ModelPtr m = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr comp1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp2 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp3 = std::make_shared<libcellml::Component>();

    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v3 = std::make_shared<libcellml::Variable>();

    m->setName("modelName");
    comp1->setName("component1");
    comp2->setName("component2");
    comp3->setName("component3");

    v1->setName("variable1");
    v2->setName("variable2");
    v3->setName("variable3");

    v1->setUnits("dimensionless");
    v2->setUnits("dimensionless");
    v3->setUnits("dimensionless");

    comp1->addVariable(v1);
    comp2->addVariable(v2);
    comp3->addVariable(v3);

    m->addComponent(comp1);
    m->addComponent(comp2);
    m->addComponent(comp3);

    libcellml::Variable::addEquivalence(v3, v1);

    v.validateModel(m);

// This is what the previous cycle checking did ... are cycles like this actually allowed?? 
// or should be prevented from forming elsewhere?  And if so, where?

    std::string error = "Cyclic variables exist, 1 loop found (Component, Variable):\n"
                        "('component1', 'variable1') -> ('component2', 'variable2') -> ('component3', 'variable3') -> ('component1', 'variable1')\n";
    EXPECT_EQ(size_t(1), v.errorCount());
    EXPECT_EQ(error, v.error(0)->description());
}
```

And sorry, I misunderstood what was going on - I was waiting for you to finish whatever changes you had to make to the Variable objects earlier in this branch.  If that's done then I'm all good to keep working on it :)

kerimoyle:
ps - I will add the interface type checking to the validator - that's certainly missing at the mo.

hsorby:
There is no way to determine a cycle here, we have a variable set there is no direction there is no cycle.  The network is defined by the topology of the components that the variables in the set belong too.

hsorby:
```
TEST(Validator, validateNoCyclesSimple)
{
    libcellml::Validator v;
    libcellml::ModelPtr m = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr comp1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp2 = std::make_shared<libcellml::Component>();

    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();

    m->setName("modelName");
    comp1->setName("component1");
    comp2->setName("component2");

    v1->setName("variable1");
    v2->setName("variable2");

    v1->setUnits("dimensionless");
    v2->setUnits("dimensionless");

    comp1->addVariable(v1);
    comp2->addVariable(v2);

    m->addComponent(comp1);
    m->addComponent(comp2);

    v.validateModel(m);

    EXPECT_EQ(0, v.errorCount());
}
```

I would expect this to be fine.

hsorby:
```
TEST(Validator, validateNoConnectionSimple)
{
    libcellml::Validator v;
    libcellml::ModelPtr m = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr comp1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp2 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp3 = std::make_shared<libcellml::Component>();

    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v3 = std::make_shared<libcellml::Variable>();

    m->setName("modelName");
    comp1->setName("component1");
    comp2->setName("component2");
    comp3->setName("component3");

    v1->setName("variable1");
    v2->setName("variable2");
    v3->setName("variable3");

    v1->setUnits("dimensionless");
    v2->setUnits("dimensionless");
    v3->setUnits("dimensionless");

    comp1->addVariable(v1);
    comp2->addVariable(v2);
    comp3->addVariable(v3);

    m->addComponent(comp1);
    m->addComponent(comp2);
    comp2->addComponent(comp3);

    libcellml::Variable::addEquivalence(v1, v3);

    v.validateModel(m);

   EXPECT_EQ(size_t(1), v.errorCount());
    EXPECT_EQ("It is not possible to connect this set of connected variables across their network", v.error(0)->description());
}
```

I would expect this to fail because I cannot make a connection between comp1's variables and comp3's variables.

kerimoyle:
OK, so in that case the spec is really quite misleading! How about this situation:

```
TEST(Validator, validateNoCyclicConnectionSimple)
{
    libcellml::Validator v;
    libcellml::ModelPtr m = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr comp1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp2 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp3 = std::make_shared<libcellml::Component>();

    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v3 = std::make_shared<libcellml::Variable>();

    m->setName("modelName");
    comp1->setName("component1");
    comp2->setName("component2");
    comp3->setName("component3");

    v1->setName("variable1");
    v2->setName("variable2");
    v3->setName("variable3");

    v1->setUnits("dimensionless");
    v2->setUnits("dimensionless");
    v3->setUnits("dimensionless");

    comp1->addVariable(v1);
    comp2->addVariable(v2);
    comp3->addVariable(v3);

    m->addComponent(comp1);
    comp1->addComponent(comp2)
    comp2->addComponent(comp3);
    comp3->addComponent(comp1);

    libcellml::Variable::addEquivalence(v1, v2); /// or anything really ... the error is elsewhere?
    libcellml::Variable::addEquivalence(v3, v1); 
    v.validateModel(m);
}
```

nickerso:
That looks valid to me - and is highlighting where the spec gets things confused in terms of "cycles", which we clearly haven't tided up enough following the removal of directionality from connections.

hsorby:
It would fail on the encapsulation of the components and the variable set.  Run it through the validator and see. Of course the variable set fail won't appear until you have written it :)

kerimoyle:
Really?  So cyclical connections between the components are ok?
`comp1 -> comp2 -> comp3 -> comp1`
  In which case there's a *whole* lot of the spec that is very confusing ...

kerimoyle:
> It would fail on the encapsulation of the components and the variable set. Run it through the validator and see. Of course the variable set fail won't appear until you have written it :)

So it is possible to have invalid encapsulations then?  And the earlier checks for cyclical variables would catch it ... before they were removed ... but if it's not the fault of the variables but rather the component_refs inside the encapsulation, then I'll move that code to reference them instead?

I must be missing something here ...

nickerso:
oops, sorry - missed the encapsulation! I agree with @hsorby - that is an invalid encapsulation hierarchy.

hsorby:
It is based on an invalid component encapsulation it's going to be all sort of weird.

kerimoyle:
... but this is the kind of error you're talking about, right?  Where the parent topology is wrong, and it's not really to do with the variables?

hsorby:
I think you will find that is something to worry about after the easier situations are covered.  A solution to this particular problem may present itself once simpler issues are resolved.

kerimoyle:
Are changes to the CI causing these kinds of errors? It all builds and runs fine on my local machine.

```
/Users/libcellml/BuildSlave/Unit_Tests_Builder_Mac_OSX_10_14/libcellml/src/api/libcellml/units.h:117:18: error: declaring a parameter with a default argument is disallowed [fuchsia-default-arguments-declarations,-warnings-as-errors]
                 double multiplier = 1.0, const std::string &id = "");
```
... which will affect how we've created a lot of the unit interaction functions if so ... 

Others I don't quite understand (there are a lot of these ones):
```
/Users/libcellml/BuildSlave/Unit_Tests_Builder_Mac_OSX_10_14/libcellml/src/api/libcellml/component.h:45:16: error: use a trailing return type for this function [modernize-use-trailing-return-type,-warnings-as-errors]
    Component &operator=(Component rhs); /**< Assignment operator */
    ~~~~~~~~~~~^
    auto                                -> Component &
```

tests/validator/validator.cpp
```diff
@@ -1160,6 +1160,36 @@ TEST(Validator, unitAmericanSpellingOfUnitsRemoved)
     };
 
     libcellml::Validator validator;
+    libcellml::ModelPtr model = std::make_shared<libcellml::Model>();
```
hsorby, line 1163: This is not right (the whole change not just this line). There is already a test for variables in the same component.  I'm surprised that it doesn't compile with an error.

kerimoyle: I had to change two of the tests just to make it compile?

kerimoyle:
Putting this on ice until the remove* bug #404 is fixed ... coverage tests will fail without it.

kerimoyle:
Ignore the comment above ... I've now included the few lines which were changed in #406 into this branch so that the testing could go ahead properly.

kerimoyle:
Just want to confirm what is valid in the equivalent variable set.  I think that the below test should throw you-have-cyclical-variable errors from the validator, but @hsorby I think you'd disagree?  What do the others think?  @nickerso @agarny @MichaelClerx ?

```
TEST(Validator, validateCycleSimple)
{
    libcellml::Validator v;
    libcellml::ModelPtr m = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr comp1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp2 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp3 = std::make_shared<libcellml::Component>();

    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v3 = std::make_shared<libcellml::Variable>();

    m->setName("modelName");
    comp1->setName("component1");
    comp2->setName("component2");
    comp3->setName("component3");

    v1->setName("variable1");
    v2->setName("variable2");
    v3->setName("variable3");

    v1->setUnits("dimensionless");
    v2->setUnits("dimensionless");
    v3->setUnits("dimensionless");

    v1->setInterfaceType("public");
    v2->setInterfaceType("public");
    v3->setInterfaceType("public");

    comp1->addVariable(v1);
    comp2->addVariable(v2);
    comp3->addVariable(v3);

    // No encapsulation, all components can access each other
    m->addComponent(comp1);
    m->addComponent(comp2);
    m->addComponent(comp3);

    libcellml::Variable::addEquivalence(v2, v3);  // Model at this point is fine ... 
    libcellml::Variable::addEquivalence(v3, v1);  // v1->v2->v3->v1 ... is this ok or not?

    v.validateModel(m);

    // ?? what should be returned from the validator here?  are the variables cyclic or not?
}
```

agarny:
I don't think we want to allow v1->v2->v3->v1. So, I would expect the validator to complain about this.

nickerso:
I think this is fine. What we are saying is that there is a single variable `v` in the model which has a bunch of aliases in each of the components - but it is the same variable. I can't see why this is a problem.

hsorby:
I think this is fine too.  We are saying that the variables are equivalent so there is no knowledge about a graph or network that the variables form thus no notion of a cycle.

The problem comes when you consider the equivalence set over a component topology.

In @kerimoyle test example here I would expect no validation errors as the model is perfectly fine.

kerimoyle:
Ok, so when the spec talks about not having cycles in the variable network,
that doesn't apply? It's not really a network, just a collection of aliases?

kerimoyle:
Urgh ... build fails on windows with this error:

> D:\buildslave_libcellml\Unit_Tests_Builder_Windows_8_64_Bit\libcellml\src\api\libcellml/variable.h(37): warning C4251: 'std::enable_shared_from_this<libcellml::Variable>::_Wptr': class 'std::weak_ptr<_Ty>' needs to have dll-interface to be used by clients of class 'std::enable_shared_from_this<libcellml::Variable>'

I haven't seen this one before? Is it because the build machine has changed in warnings-as-errors?

hsorby:
No, it should be working this means you need to declare the weak pointer variant for the variable class, I thought that I had done that (maybe another branch)

hsorby:
You need this at the top (outside the namespace)

```
#ifndef SWIG
template class LIBCELLML_EXPORT std::weak_ptr<libcellml::Variable>;
#endif
```

CMakeLists.txt
```diff
@@ -89,10 +89,11 @@ unset(UNIT_TESTS CACHE)
 
 if(LIBCELLML_COVERAGE OR LIBCELLML_MEMCHECK)
   if(NOT LIBCELLML_UNIT_TESTS)
-    message(SEND_ERROR "Configuration confusion:
```
hsorby: This looks like we are going backwards?

CMakeLists.txt
```diff
@@ -113,12 +114,14 @@ if(NOT IS_MULTI_CONFIG)
   endif()
 endif()
 unset(BUILD_TYPE CACHE)
+
 if(LIBCELLML_MEMCHECK)
   if(LIBCELLML_BUILD_TYPE STREQUAL "Release")
-    message(SEND_ERROR "Configuration confusion:
```
hsorby: going backwards here too?

src/api/libcellml/variable.h
```diff
@@ -345,7 +349,7 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
     std::string interfaceType() const;
 
 private:
-    void swap(Variable &rhs); /**< Swap method required for C++ 11 move semantics. */
+    void swap(Variable & rhs); /**< Swap method required for C++ 11 move semantics. */
```
hsorby, line 357: I don't think this will pass clang tidy.

src/component.cpp
```diff
@@ -181,6 +183,62 @@ void Component::removeAllVariables()
...
 
+*/
+
+bool Component::removeVariable(size_t index)
+{
+    if (index < mPimpl->mVariables.size()) {
+        VariablePtr v = mPimpl->mVariables.at(index);
+        ModelPtr model = v->parentModel();
+        v->clearParent();
+        v->setParent(model);
```
hsorby: I cannot understand why you would want to set the parent of the variable to the model. I would have thought clearing the parent would have been what is required.

hsorby: This holds for all `v->setParent(model)` s

src/componententity.cpp
```diff
@@ -105,21 +105,28 @@ bool ComponentEntity::removeComponent(const std::string &name, bool searchEncaps
     bool status = false;
     auto result = mPimpl->findComponent(name);
     if (result != mPimpl->mComponents.end()) {
+        ComponentPtr c = *result;
+        ModelPtr m = c->parentModel();
+        c->clearParent();
+        c->setParent(m);
```
hsorby: Similar comment applies to components as variables.

src/printer.cpp
```diff
@@ -171,13 +171,6 @@ void buildMaps(const ModelPtr &model, ComponentMap &componentMap, VariableMap &v
                             // Get parent components.
                             ComponentPtr component1 = variable->parentComponent();
                             ComponentPtr component2 = equivalentVariable->parentComponent();
-                            // Do not serialise a variable's parent component in a connection if that variable no longer
-                            // exists in that component. Allow serialisation of one componentless variable as an empty component_2.
-                            if (component2 != nullptr) {
-                                if (!component2->hasVariable(equivalentVariable)) {
-                                    component2 = nullptr;
-                                }
-                            }
```
hsorby, line 180: Not sure why this has been removed?

tests/component/encapsulation.cpp
```diff
@@ -96,23 +96,75 @@ TEST(Encapsulation, reparentComponent)
...
+    // model (parent (child1, child2, child3 ) )
+
     libcellml::Printer printer;
     std::string a_parent = printer.printModel(model);
     EXPECT_EQ(e_parent_1, a_parent);
 
-    // what do we expect this to achieve? The addition of child3 to child2
+    // what do we expect this to achieve? The addition of child3 to child2 << KRM swaps parents of child3 but does not update component() array
+
+    std::cout << "parent component has children: \n";
```
hsorby: For all `std::cout` it should be `EXPECT_*`.

tests/connection/connection.cpp
```diff
@@ -59,6 +59,36 @@ TEST(Variable, addDuplicateEquivalentVariablesAndCount)
...
+    libcellml::Variable::addEquivalence(v1, v2);
+
+    v1->removeAllEquivalences();
+    EXPECT_EQ(e, v1->equivalentVariableCount());
+    EXPECT_EQ(e, v2->equivalentVariableCount());
+}
+
+TEST(Variable, removeAllEquivalencesWithTriangleEquivalence)
+{
+    const size_t e0 = 0;
```
hsorby: Not keen on this as we have previously used `size_t(0)` and it's not like we are reusing this many times.

src/validator.cpp
```diff
@@ -27,7 +27,16 @@ limitations under the License.
...
+#include <deque>
+#include <map>
+#include <regex>
+#include <sstream>
 #include <stdexcept>
+#include <string>
+#include <vector>
+
+// KRM Delete me...
+#include <iostream>
```
hsorby: definitely

hsorby (changes requested):
Having a look at some of the easy stuff and had a few changes.  I will look at the `validator` tests separately.

tests/connection/connection.cpp
kerimoyle: ... I didn't write this function at all ... :O Magic ... not sure where it came from!!

kerimoyle:
Thanks @hsorby for having a look.  I've seen the bits that you've commented on, and they're not my changes (except the ones with KRM - def guilty in those parts!).  This must have come from the earlier confusion of you and I both pushing to this branch and getting out of sync?  Not sure whether that stuff can be undone or whether we should just go through everything manually from where the branch is now??

CMakeLists.txt
kerimoyle: Perhaps ... wasn't me as I don't understand enough of the cmake config to touch it ... ?

src/validator.cpp
kerimoyle: Done :)

tests/component/encapsulation.cpp
kerimoyle: Missed this one - will comment it all out for now.  It shows the bug in how we handle parents.

CMakeLists.txt
kerimoyle: as above ...

src/api/libcellml/variable.h
kerimoyle: I ran test_clang_format before pushing ... so everything here passes all the build/test/format on my machine?

hsorby:
I think we will want to reinstate bits a pieces back over the top of this.  I'll do the CMake stuff.  I think some wierdness happens with the git merge at times.

src/component.cpp
kerimoyle: OK, just had another look at this.  It was because the only way to remove the parent **component** was to call the clearParent function, which wipes **both** the component parent as well as the model parent.  I put these setParent lines there to make sure that the model parent wasn't deleted, as that wasn't the point of the function?

src/componententity.cpp
kerimoyle: See comment above.

src/variable.cpp
```diff
@@ -25,6 +25,10 @@ limitations under the License.
 #include <sstream>
 #include <vector>
 
+#ifndef SWIG
+template class LIBCELLML_EXPORT std::weak_ptr<libcellml::Variable>;
+#endif
```
hsorby: This shouldn't be needed here just in the header.

kerimoyle: Yes, see latest push ...

tests/component/encapsulation.cpp
```diff
@@ -22,6 +22,8 @@ limitations under the License.
 
 #include <iostream>
 
+/*
+// TODO Removed this test as doesn't actually test anything ... except show a bug in how we deal with parents
```
hsorby: I actually think this should be here but turned into a proper test, it does show how the library behaves which is important.

kerimoyle: But the issues it raises are being addressed in another branch

kerimoyle: ... so the test should go there instead maybe?

tests/validator/validator.cpp
```diff
@@ -2132,3 +2132,76 @@ TEST(Validator, interfaceTestingNotValid)
     EXPECT_EQ(expectedErrors.size(), validator.errorCount());
     checkExpectedErrors(expectedErrors, validator);
 }
+
+/* 
+// TODO Commenting out for now, but still demonstrates a bug in the way we store equiv variables
+TEST(Validator, fixUpEquivVariables)
```
hsorby, line 2139: Again I think this test should be in it shows behaviour of the library it looks at usage in a different way all of which is important in understanding what the library does and does not do.

kerimoyle: yes, but perhaps not in this branch?

hsorby: I would change the comments to reflect how we are thinking about what is going on though not what alternative way things should be happening. And therefore changing the answers to what the library is actually giving.

src/component.cpp
kerimoyle: (it doesn't mean that the model parent will be used, only that it's not deleted while deleting the parent component)

src/api/libcellml/variable.h
hsorby: Well ... don't do anything then.

src/component.cpp
hsorby: This is part of the problem with this code (solved by my other PR shameless promotion) there should in theory only ever be one parent set so clearing the parent should only effect the one that is active if any.

src/printer.cpp
kerimoyle: Me either ... will put it back and see what happens

kerimoyle: It makes a test segfault ... Connection.componentlessVariableInvalidConnection, will have a better look and see

src/component.cpp
kerimoyle: Either way... whatever is here should be consistent.  Agreed that it can be fixed in another branch, but there's no point doing it in two places, that will just get super confusing!

src/printer.cpp
kerimoyle: OK, resolved now and added tests to give the coverage which was missing before.

tests/validator/validator.cpp
kerimoyle: Done

tests/connection/connection.cpp
kerimoyle: Done

tests/component/encapsulation.cpp
kerimoyle: Have updated the comments, and we can reinstate the test when we're ready to address the parent stuff.

kerimoyle:
Hi @agarny!  I've found some strange behaviour in the generator files while getting this PR up to date with development - it's what's currently causing the coverage failure because the guts of the parentModel() utility function is never called.  When I ran on my machine I found that there were two generator tests failing:

```
TEST(Generator, twoVariablesOfIntegration)
TEST(Generator, nonFirstOrderOdes)
```
They failed initially because of a segfault in this section of code:

```
void Generator::GeneratorImpl::processEquationAst(const GeneratorEquationAstPtr &ast)
{

    ... do some stuff ... 


        } else if ((variable != mVoi)
                   && !variable->hasEquivalentVariable(mVoi)) {
            ComponentPtr voiComponent = std::dynamic_pointer_cast<Component>(mVoi->parent());
            ModelPtr voiModel = std::dynamic_pointer_cast<Model>(voiComponent->parent());
            ComponentPtr component = std::dynamic_pointer_cast<Component>(variable->parent());
// ^^ component is nullptr
            ModelPtr model = parentModel(component);
// ^^model is nullptr
            ComponentPtr c1 = std::dynamic_pointer_cast<Component>(component->parent());
// c1 is nullptr

// Segfault came from accessing ->name() function of null values
            err->setDescription("Variable '" + getEntityName(mVoi)
                                + "' in component '" + getEntityName(voiComponent)
                                + "' of model '" + getEntityName(voiModel)
                                + "' and variable '" + getEntityName(variable)
                                + "' in component '" + getEntityName(component)
                                + "' of model '" + getEntityName(model)
                                + "' cannot both be a variable of integration.");
            err->setKind(Error::Kind::GENERATOR);

            mGenerator->addError(err);
        }
    }

    // Make sure that we only use first-order ODEs.

    if ((ast->mType == GeneratorEquationAst::Type::CN)
        && (astParent != nullptr) && (astParent->mType == GeneratorEquationAst::Type::DEGREE)
        && (astGrandParent != nullptr) && (astGrandParent->mType == GeneratorEquationAst::Type::BVAR)
        && (astGreatGrandParent != nullptr) && (astGreatGrandParent->mType == GeneratorEquationAst::Type::DIFF)) {
        if (convertToDouble(ast->mValue) != 1.0) {
            VariablePtr variable = astGreatGrandParent->mRight->mVariable;
            ComponentPtr component = std::dynamic_pointer_cast<Component>(variable->parent());
            ModelPtr model = parentModel(component);

// segfault here from accessing name() of null values
            err->setDescription("The differential equation for variable '" + getEntityName(variable)
                                + "' in component '" + getEntityName(component)
                                + "' of model '" + getEntityName(model)
                                + "' must be of the first order.");
            err->setKind(Error::Kind::GENERATOR);

            mGenerator->addError(err);
        }
    }
```
... you can see I've changed the tests a little and changed the errors to catch null->name() calls, but the *real* problem is that maybe the hasEquivalentVariable() function has changed?  I'm not confident enough of how the generator works to fix it myself yet - do you have any hints?

agarny:
Hi @kerimoyle, I have just cloned your repo and built your `issue295_validate_variables` branch, and... it's all working fine for me on macOS (both building and testing). In fact, that's what is implied by the Unit Tests since they pass. So, I am not sure what is wrong with the `twoVariablesOfIntegration` and `nonFirstOrderOdes` tests, if anything...!?

The only thing that I can confirm is that coverage is not 100%. A few lines in `src/utilities.cpp` are not covered:

Screen Shot 2019-10-08 at 09 24 39

kerimoyle:
Yes, exactly!  The coverage is coming from the fact that the parentModel
function doesn't enter the while loop, because earlier in the two tests I
mentioned, the parent pointer is not being set.  I've tweaked your tests
and your code so that the tests pass, but the behaviour of the generator is
what is worrying me? I had to add the getEntityName calls throughout (as
above) so that the tests didn't segfault ... but that's not really the
issue.  The *real* issue is why the parents are coming back as nullptr from
the lines indicated in my comments?

agarny:
Ok, I now understand the issue. I am going to look into it, but... we should never change tests that are supposed to pass. So, first thing first, I am going to revert what you did with regards to the Generator code/test, and see from there.

agarny:
Ok, I can now confirm the issue with the `twoVariablesOfIntegration` and `nonFirstOrderOdes` tests, and will have a look at it today.

agarny:
Ok, I put some traces here and there, and the problem is that you have modified `bool ComponentEntity::removeComponent(const ComponentPtr &component, bool searchEncapsulated)`, which now clears the parent of the component.

Anyway, probably best to wait until some other PRs to have been merged in since a few things are now different when it comes to handling parents.

kerimoyle:
I'm closing this PR as it's now about 5 months out of date.  It will be quicker to rewrite it (perhaps third time's the charm) from scratch instead of trying to fix the changes here.

agarny:
> I'm closing this PR as it's now about 5 months out of date. It will be quicker to rewrite it (perhaps third time's the charm) from scratch instead of trying to fix the changes here.

I can sympathise and this is the reason I always update my PRs whenever the `develop` branch is updated.

kerimoyle:
Yes, I have been throughout, but it's a time sink - basically rewriting the same code over and over again :| . Ah well.

Issue #340, opened 2019-06-07
Need to add validation tests for standard units


kerimoyle:
As @agarny suggested, we need to add tests that the hard-coded standard unit composition is correct (eg: that Newtons = kg.m/s^2 etc).

kerimoyle:
Added this test into the units validation PR #351

kerimoyle:
Closed by https://github.com/cellml/libcellml/pull/351

Issue #327, opened 2019-05-28
Remove American spellings from prefixes and units (deka, meter, liter)


kerimoyle:
> I would suggest we start supporting both - especially as we've currently got "deka" in the spec, while "deca" seems to be the standard

_Originally posted by @MichaelClerx in https://github.com/cellml/libcellml/pull/304#issuecomment-486105594_

I've changed a couple of files to see what the effects of this would be.  Since we can duplicate enumerated values, and since it makes sense to use those values as something meaningful (the log of the prefix value), both the "deca" and "deka" prefixes now have the same value (1).  This means that reverse lookups like the `prefixToString` function will only ever return "deca" (and I would have thought that the map condition of unique keys is violated, but somehow it still runs fine ...?).  

```
std::map<Prefix, const std::string> prefixToString =
{
    {Prefix::ATTO, "atto"},
    {Prefix::CENTI, "centi"},
    {Prefix::DECA, "deca"},
    {Prefix::DECI, "deci"},
    {Prefix::DEKA, "deka"},  // Added to support "deka" and "deca" as both are in common use
    {Prefix::EXA, "exa"},
    {Prefix::FEMTO, "femto"},
... etc ... 
```

The consequence of this is that:
- specifying a unit with a prefix via the `Prefix` enum will end up with a printed string of "deca", even if `Prefix::DEKA` is used.
- specifying a unit with a prefix of a string "deka" or "deca" will have this preserved in their printed units.

```
TEST(Coverage, prefixToString) {
    libcellml::Model m;
    libcellml::Printer printer;
  std::vector<std::string> prefix_str =
        {"atto",
         "centi",
         "deca",
         "deci",
         "deka",  // "deka" will be turned into "deca" on conversion through the enum ...
         "exa",
   ... etc ...
        };
    std::vector<libcellml::Prefix> prefix_enum =
        {libcellml::Prefix::ATTO,
         libcellml::Prefix::CENTI,
         libcellml::Prefix::DECA,
         libcellml::Prefix::DECI,
         libcellml::Prefix::DEKA,
         libcellml::Prefix::EXA,
 ... etc ... 
        };
    for (std::vector<std::string>::size_type i = 0; i != prefix_str.size(); ++i) {
        std::string prefix = prefix_str[i];
        libcellml::UnitsPtr u = std::make_shared<libcellml::Units>();
        u->setName("abcdefg");
        u->addUnit("empty", prefix_enum[i]);   // <<<<< This one fails, prefix added as enum
[...]

PR #351, opened 2019-06-13
Issue361 units validation

kerimoyle:
Fourth in the series of smaller pull requests replacing #304 and addressing issue #361.

Unit validation:

- [x] checks for cyclical dependencies of units
- [x] checks that units are the same between equivalent variables
- [x] tests
- [x] coverage


kerimoyle:
I'm expecting the coverage to fail because I can't test the line in the validator where the prefix is out of the integer range (ie, greater than 18446744073709551615).  Clang doesn't let me enter this as an argument though, because its own out-of-range testing is applied at build time.  How can I get the coverage tests if I can't set the argument?

Error:
```
/Users/kmoy001/libCellML/libCellML-src/tests/validator/validator.cpp:1493:25: error: integer literal is too large to be represented in any integer type [clang-diagnostic-error]
    u->addUnit("metre", 18446744073709551616, 1.0, 1.0); // should throw out of integer range error
```

Here's the code inside `validateUnitsUnit()`:
```
                try {
                    int test = std::stoi(prefix);  // coverage cannot get to this line ... 
                    (void)test;
                } catch (std::out_of_range &) {
                    err->setDescription("Prefix '" + prefix + "' of a unit referencing '" + reference + "' in units '" + units->name() + "' is out of the integer range.");
                    err->setUnits(units);
                    err->setRule(SpecificationRule::UNIT_PREFIX);
                    mValidator->addError(err);
                }
```

Here's the test I wanted to use:
```
TEST(Validator, integerPrefixOutOfRange)
{
    // KRM TODO Not sure how to give a coverage test here as clang doesn't let the out-of-range integer be set?
    libcellml::ModelPtr m = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr c = std::make_shared<libcellml::Component>();
    libcellml::VariablePtr v = std::make_shared<libcellml::Variable>();
    libcellml::UnitsPtr u = std::make_shared<libcellml::Units>();
    m->setName("model");
    c->setName("component");
    u->setName("unit");
    u->addUnit("metre", 18446744073709551616, 1.0, 1.0); // should throw out of integer range error
    v->setName("variable");
    v->setUnits(u);
    c->addVariable(v);
    m->addComponent(c);
    validator.validateModel(m);
    printErrors(validator);
}

```

kerimoyle:
@nickerso @hsorby @agarny Ready for review please.

hsorby (dismissed):
I find the tests too long and that they lack focus.  I would like to see tests that start simple and get more complicated.

* First test should be does this variables units match this other variables units.
* We only need to test one cycle if the cycles are not fundamentally different.
* What happens when you test the same variable to itself?

.gitignore
```diff
@@ -16,5 +16,9 @@ CMakeLists.txt.user
 # PyCharm generated files
 .idea/
 
-# VS Code system files
+# OS X hidden system files
+.DS_Store
+
+# VS Code configuration files
 .vscode/
+
```
agarny: FWIW, this file shouldn't be part of this PR. It has already been modified as a result of another PR.

kerimoyle:
> I find the tests too long and that they lack focus. I would like to see tests that start simple and get more complicated.

Is there one in particular you mean?  The cycle network is as simple as they can be to actually test the functionality of the validator.  I've split that test in half now so that one is valid and the other not.

> - First test should be does this variables units match this other variables units.

I've moved `TEST(Validator, variableEquivalentUnits)` which does exactly that to be the first of the bunch. It has a lot of different cases - starting super-simple and getting more interesting - but I've split them out into separate tests which hopefully makes it easier to follow ...

> - We only need to test one cycle if the cycles are not fundamentally different.

Do you mean that units cycles are reported more than once?  There's only one cycle there but I'll see if I can suppress reporting of the same one more than once - thanks for the heads-up!

> - What happens when you test the same variable to itself?

That's a variable validation thing rather than units.  I'll add it in to #350 where the variable network is tested.

I've also added the individual standard units testing that @agarny wanted from #340 there too.

kerimoyle:
Coverage test is failing at 100% again?

kerimoyle:
Removed American spellings of "deka", "liter" and "meter" from the accepted units and prefixes.  They will return a validation error now.  Addresses #327 and https://github.com/cellml/cellml-specification/issues/21

hsorby (commented):
I don't think we need to go around changing the units in the Ohara Rudy model where it isn't necessary.  Names of units can have whatever spelling is desired we only need the base units to be correctly spelt.

kerimoyle:
I just figured it was easier and more consistent to do it here, esp if it's used as an example.  I can change it back.

kerimoyle:
@hsorby Changed back to be correct spellings for only the base unit of metre: others are now centimeter, micrometer etc. @agarny @nickerso Good to go or more needed?

kerimoyle:
Nudge ... @hsorby @nickerso

hsorby:
Looking at this starting now, sorry for taking so long to get here.

hsorby (commented):
I still find the build up of tests missing.  I think the first test should be looking at testing the simplest possibly case, I don't find that with the tests written here.  In [this PR](https://github.com/kerimoyle/libcellml-keri/pull/20) I have removed a chunk of code that had absolutely zero effect on the test in question.  Unit tests are hard enough to read without adding unnecessary statements.

Apart from the above, I think I am likely to approve these changes if the PR above is merged in.

kerimoyle:
Thanks for the review @hsorby .  I agree that the simplest test case was missing, but I do think that it was important to test for cycles in the presence of branches - which was the point of the complicated (zero effect) bit of the last test.  It's more for testing that the valid branches to which the cycle is connected are not included in the reporting.  If you'd rather, I can move this to another test case, leaving both your simpler one and my branched one separate?

hsorby:
Yes the simple test should be there but if you feel you need a test to show that it works with branches the certainly do so in addition to that one.

kerimoyle:
All done :) Can be merged now @hsorby @nickerso @agarny ?

hsorby (dismissed):
I have had a look and a play with this and find that it is doing the job as advertised. So I am approving this PR.

tests/resources/Ohara_Rudy_2011.cellml
```diff
@@ -61,28 +61,28 @@ http://www.cellml.org/
     <units name="microF">
         <unit prefix="micro" units="farad"/>
     </units>
-    <units name="micrometer_3">
+    <units name="micrometre_3">
```
hsorby: We didn't need to change these (and others like it) as they are user defined.  The user is free to use whichever spelling they prefer.

agarny: Agreed (I saw your previous comment about this above), it's just that the name of some units had already been modified (to have "litre" rather than "liter"), so I thought I would just do the same for metre. However, since I agree with your point, I am going to revert the name of all those variables. We indeed only need the value of `units` to use the correct spelling.

src/validator.cpp
```diff
@@ -29,9 +29,7 @@ limitations under the License.
 #include "libcellml/when.h"
 
 #include <cmath>
-
 #include <iostream>
```
hsorby: This include may not be even needed?

agarny: Good point, I haven't checked, but I am going to in a minute.

src/validator.cpp
```diff
@@ -1183,145 +1156,89 @@ bool Validator::ValidatorImpl::unitsAreEquivalent(const ModelPtr &model,
...
     if (model->hasUnits(uName)) {
-        u = model->units(uName);
+        libcellml::UnitsPtr u = model->units(uName);
         if (!u->isBaseUnit()) {
+            std::string myRef;
+            std::string myPre;
+            std::string myId;
+            double myExp;
+            double myMult;
+            double m;
```
hsorby: Not super keen on this varabile name

hsorby (dismissed):
Apart for very minor varialbe name this is still good for me.

tests/validator/validator.cpp
```diff
@@ -1065,3 +1065,726 @@ TEST(Validator, validMathCnElements)
...
+
+    v1->setUnits(u1);
+    v2->setUnits(u2);
+    m->addUnits(u1);
+    m->addUnits(u2);
+
+    const std::vector<std::string> expectedErrors = {
+        "Units reference 'meter' in units 'testunit2' is not a valid reference to a local units or a standard unit type.",
+        "Variable 'tomayto' has units of 'testunit1' and an equivalent variable 'tomahto' with non-matching units of 'testunit2'. The mismatch is: metre^1, ",
+        "Variable 'tomahto' has units of 'testunit2' and an equivalent variable 'tomayto' with non-matching units of 'testunit1'. The mismatch is: metre^-1, "};
```
agarny: The last two error messages are effectively the same. So, only one of them ought to be sufficient.

Also, the last two error messages end up with `, `, which is clearly not what we want. We want a properly formatted error message.

src/validator.cpp
```diff
@@ -231,6 +225,54 @@ struct Validator::ValidatorImpl
      * @return @c true if @node is a supported MathML element and @c false otherwise.
      */
     bool isSupportedMathMLElement(const XmlNodePtr &node);
+
+    /**
+    * @brief Validate that equivalent variable pairs in the @p model
+    * have equivalent units.
+    * @cellml2_19 Validate that equivalent variable pairs in the @p model
+    * have equivalent units.
+    * Any errors will be logged in the @c Validator.
```
agarny, line 232: What is that `@cellml2_19` about?

tests/validator/validator.cpp
```diff
@@ -1065,3 +1063,726 @@ TEST(Validator, validMathCnElements)
...
+
+    // mm^6
+    libcellml::UnitsPtr u3 = std::make_shared<libcellml::Units>();
+    u3->setName("u3");
+    u3->addUnit("u2", 2.0); // standard, exponent.
+
+    // m^6
+    libcellml::UnitsPtr u4 = std::make_shared<libcellml::Units>();
+    u4->setName("u4");
+    u4->addUnit("u3", 15, 1.0); // standard, prefix, exponent.
```
agarny, line 1456: I am confused here... shouldn't it be `18` rather than `15`? Or is the idea just to have variables with equivalent units as opposed to exactly the same unit?

agarny: In fact, the current validation code seems to be checking only for equivalent units (and rightly so when it comes to validation), which means that we could in fact use any prefix we want here.

tests/validator/validator.cpp
```diff
@@ -1065,3 +1063,726 @@ TEST(Validator, validMathCnElements)
...
+        "Variable 'v5' has units of 'metre' and an equivalent variable 'v6' with non-matching units of 'second'. The mismatch is: metre^1, second^-1, ",
+        "Variable 'v8' has units of 'banana' and an equivalent variable 'v7' with non-matching units of 'apple'. The mismatch is: apple^-1, banana^1, ",
+        "Variable 'v3' has units of 'litre' and an equivalent variable 'v4' with non-matching units of 'gram'. The mismatch is: kilogram^-1, metre^3, ",
+        // "Variable 'v3' has units of 'litre' and an equivalent variable 'v9' with non-matching units of 'big_barrel'. The mismatch is: multiplication factor of 10^-3, ", // NOTE: removed until multiplication checking is turned on.
+        "Variable 'v6' has units of 'second' and an equivalent variable 'v5' with non-matching units of 'metre'. The mismatch is: metre^-1, second^1, ",
+        // "Variable 'v9' has units of 'big_barrel' and an equivalent variable 'v3' with non-matching units of 'litre'. The mismatch is: multiplication factor of 10^3, ", // NOTE: removed until multiplication checking is turned on.
```
agarny: Why are there still a couple of expected errors that are commented out? Am I to understand that this PR is not complete in fact? If so, then I guess that commit 14ea753 should be reverted?...

Also, like in another test, the expected errors are effectively duplicated, which is more confusing than anything. I would really have only one expected error between say `v1` and `v2` rather than one between `v1` and `v2` and another between `v2` and `v1`.

agarny (changes requested):
This looks good overall, pending the few issues I have raised.

src/validator.cpp
kerimoyle: Ah, well spotted! That's a leftover from where I was coding all the validator tests to make sure that everything in the spec was covered.

tests/validator/validator.cpp
kerimoyle: Hi @agarny Yes, apparently we don't mind about the multiplier/prefix mismatch of units across equivalent variables.  It will be raised as a "warning" level when these are in and I can finally get to working on the Issues structure for non-fatal stuff.

tests/validator/validator.cpp
kerimoyle: It's just as the comments say ... removed until multiplication checking is turned on.

src/validator.cpp
kerimoyle: The m? or the "my" ones?

kerimoyle (commented):
I'm confused!  I thought that tests had to be as small as possible, even where they were duplicating much the same stuff.  This has been bigger earlier and was then made smaller ... it would be nice to have consensus on what I was aiming for ... ?

tests/validator/validator.cpp
kerimoyle: Done :)

src/validator.cpp
kerimoyle: Done (I think) :)

kerimoyle:
Changes made as requested, should be good to go now? @agarny @hsorby @nickerso

nickerso:
> I'm confused! I thought that tests had to be as small as possible, even where they were duplicating much the same stuff. This has been bigger earlier and was then made smaller ... it would be nice to have consensus on what I was aiming for ... ?

I agree. I think that as the capabilities of libCellML have grown we have not be particularly consistent in the way that we write tests. I'd be inclined to not worry too much at present and once we catch up this kind of discussion can happen as part of a general code review.

nickerso (dismissed):

tests/validator/validator.cpp
agarny: Fair enough, however it might be worth making a small comment about this in the test. It's just that when I saw it, I got confused and couldn't help myself but working out the units equivalence by hand to make sure that it was "correct" (hence wasting a bit of time).

tests/validator/validator.cpp
agarny: Sure, but... why isn't multiplication checking part of this PR?

agarny (commented):
I am not clear why multiplication checking is not part of this PR, but I am going to assume that it's just missing something here. So, apart from that, I am now fine with this PR.

agarny (dismissed):

src/validator.cpp
hsorby: Yes, I don't like the `m` variable it seems it also means the `myMult` is also a bit disingenuous.

agarny: All fine now.

agarny (dismissed):

tests/validator/validator.cpp
agarny: Ok, I imagine this is because we can only generate errors at this stage, as opposed to issues (which may be of either error, warning or hint type). Fair enough, but this means that this piece of commented out code should be removed. Then, if there isn't already one, a GitHub issue should be created to explain the additional work that will need to be done once issues are available.

agarny (approved):

agarny:
@hsorby and @nickerso, you might want to quickly review this PR and, if you are still happy with it, merge it.

hsorby (approved):
Looks good to me.

kerimoyle (commented):
Just a note that we *do* need to re-include the multMap to enable the multiplication checking later ...

agarny:
Indeed and that should be done as part of issue #381.

Issue #355, opened 2019-06-14
Validation for new format of resets


kerimoyle:
Now that the format of the new resets is settled we need to:
- parse them
- validate them 
- remove all references to whens.

The format is:

```
<reset variable=”name_1” test_variable=”name_2” order=”1”>
   <test_value>MATHML</test_value>
   <reset_value>MATHML</reset_value>
</reset>
```

This issue replaces earlier ones regarding resets: #326 and #294

Issue #314, opened 2019-05-02
Replace the `Error` class with the `Issue` class with a type (e.g. error, warning, hint)


kerimoyle:
There are a couple of areas which would benefit from having non-fatal warnings raised (ie without implying an invalid model or stopping the operation):
- multipliers in the units of equivalent variables (currently unchecked)
- multipliers in units in the maths (currently unchecked)
- validation defaulting to 0-depth when there is no file location specified (see #280)
- (perhaps?) where the parser encounters something which will fail later validation.  Would this work as a response to #302 @hsorby? 

It would be nice to be able to return useful warnings to the user which do not affect the final validity of the model, either by mimicking the Error module as a separate Warning module, or by introducing a severity level to the errors (all current ones become fatal by default).

agarny:
> - (perhaps?) where the parser encounters something which will fail later validation.  Would this work as a response to #302 @hsorby? 

I don't think I would want that kind of "warning" when parsing a model. If I want "warnings" that are related to validation, well... I just ask for the model to be validated...

> It would be nice to be able to return useful warnings to the user which do not affect the final validity of the model, either by mimicking the Error module as a separate Warning module, or by introducing a severity level to the errors (all current ones become fatal by default).

Personally, rather than have an `Error` class, I would probably have an `Issue` class and then have a `type` of issue, which could be `warning`, `error`, `fatal` or whatever we deem necessary.

hsorby:
Are you thinking this is purely to do with the parser?

hsorby:
#302 is reminding us that we don't want any validation code in the parser so a big no to doing validation in the parser.

hsorby:
I like the idea of having an `Issue` class that can have levels `warning` or `error` or whatever but I don't think this should be work done that appears in the  validation PR.

agarny:
> I like the idea of having an `Issue` class that can have levels `warning` or `error` or whatever but I don't think this should be work done that appears in the validation PR.

Agreed. If anything, that work ought to be merged in before validation does.

hsorby:
At this point due to the large and significant changes that the validation PR is going to have it might be prudent to do this after the merge.

kerimoyle:
It's a bit chicken and egg, but either option (upgrade to Issue before or after changes to validation) isn't a deal breaker either way - I guess it depends on when you want the validation branch to merge.

> #302 is reminding us that we don't want any validation code in the parser so a big no to doing validation in the parser.

[...]

Issue #294, opened 2019-04-01
Validator::validateModel doesn't catch resets with variables outside the parent component


kerimoyle:
**Re: 12.1.1.1** "The value of the variable must be a variable reference to a variable defined within the component element parent of the reset element".  Should the validateModel function catch this (added to `TEST(Validator, resets) `)
```
// adding reset to second component which does not contain the variable needed

	libcellml::ModelPtr m2 = std::make_shared<libcellml::Model>();
	libcellml::ComponentPtr c2 = std::make_shared<libcellml::Component>();
	libcellml::VariablePtr var2 = std::make_shared<libcellml::Variable>();
	libcellml::ResetPtr r8 = std::make_shared<libcellml::Reset>();
	libcellml::WhenPtr w3 = std::make_shared<libcellml::When>();

	w3->setOrder(20);
	w3->setCondition("<math xmlns=\"http://www.w3.org/1998/Math/MathML\"></math>");
	w3->setValue("<math xmlns=\"http://www.w3.org/1998/Math/MathML\"></math>");

	var2->setName("var2");
	var2->setUnits("metre");

	r8->setOrder(20);
	r8->addWhen(w3);
// 'var' is a valid variable defined earlier in this test, but is just not inside this component ...
	r8->setVariable(var); 
// the only variable inside this component is 'var2'
	c2->addVariable(var2);
	c2->addReset(r8); 
	c2->setName("comp2");
	m2->setName("main2");
	m2->addComponent(c2);

libcellml::Validator v2;
	v2.validateModel(m2);
	EXPECT_EQ(1u, v2.errorCount());
```

Currently the call to validate the m2 model returns no errors, but I would have expected something to come back?  Have I missed something?

hsorby:
A contextual note here would be that resets and their implementation are still under discussion so the reset validation hasn't been pushed very far.

Yes I realise that I haven't answered the question.

hsorby:
To actually answer the question, yes it should raise an error.

kerimoyle:
OK, thanks Hugh :) Might put this one further down the list until discussions are done

hsorby:
Definitely a good choice there, I have pretty much left resets as is for a while.

kerimoyle:
Just wanted to check in on this? How are the discussions about resets going ...?

nickerso:
[...]

PR #358, opened 2019-06-18
Issue355 new resets

kerimoyle:
Updating all parts of the code reflect the changes to the `reset` format and the removal of `when` elements (issue #355).

- [x] reset description
- [x] parser
- [x] printer
- [x] validator
- [x] coverage

NB: Reset order uniqueness (as discussed in #326) will be done in a separate PR


kerimoyle:
**1. Reset order specification**
Currently we have code in the validator which checks for resets with non-unique order values within a component.  It was my understanding that the order should be unique within the connected variable set, not the component?  

```
if (component->resetCount() > 0) {
        // Check for duplicate order values in resets
        std::vector<int> resetOrders;
        for (size_t i = 0; i < component->resetCount(); ++i) {
            ResetPtr reset = component->reset(i);
            int resetOrder = reset->order();
            if (reset->isOrderSet()) {
                if (std::find(resetOrders.begin(), resetOrders.end(), resetOrder) != resetOrders.end()) {
                    err->setDescription("Component '" + component->name() + "' contains multiple resets with order '" + convertIntToString(resetOrder) + "'.");
                    err->setComponent(component);
                    err->setRule(SpecificationRule::RESET_ORDER);
                    mValidator->addError(err);
                } else {
                    resetOrders.push_back(resetOrder);
                }
            }
        }
        for (size_t i = 0; i < component->resetCount(); ++i) {
            ResetPtr reset = component->reset(i);
            validateReset(reset, component);
        }
    }
```

**2. test_variable component?**
In the old spec document it said that the `variable` needed to be present within the parent `component`. Is this also true for the `test_variable`, or can that be in another `component`?

**3. MathML stuff**
There is a function `validator->validateMath(const std::string &input, const ComponentPtr &component)` which tests the mathML against the `component` specified.  Do the contents of the `test_value` and `reset_value` blocks need to be specific to the parent component or can it refer to variables in other components?

MichaelClerx:
For 2&3: Like all cellml constructs, it refers to variables in the local component only (although they may be defined in other components and then "connected" to the local versions)

What I _wanted_ to say in the earlier version, was that they had to be _defined_ (i.e. with mathml or an initial value) in the local component. But that's not allowed so we're not doing that!

kerimoyle:
> that's not allowed

??

MichaelClerx:
> > that's not allowed
> 
> ??

Eh, we've had discussions about this. Basically we're saying you can write algebraic ODEs etc. in CellML, so component A can say "X + Y  = 1 / Z", and then an entirely different model can have components defining 2 out of 3 of those, and that's all valid and something tools will have to put up with.

In practice most models and tools will prefer models written as assignments only, but for the standard it's easy enough to allow [DAEs](https://en.wikipedia.org/wiki/Differential-algebraic_system_of_equations), which have some important applications

kerimoyle:
OK, that makes sense, thanks.  Any thoughts on the scope where `order`s must be unique?  Component-wide or equivalent-variable-set wide?

MichaelClerx:
equivalent variable set. Are we unclear about this in the spec? We'll need to change that then...

kerimoyle:
The spec seemed clear to me, but the existing tests reported something completely different (as in code above).  I think it would make more sense to move the testing for order uniqueness into the equivalent variable testing routines (like where units are tested etc) instead of just looking for different resets within a component with the same order...?

MichaelClerx:
That sounds sensible to me. You need to have the DAG worked out before you can validate it

kerimoyle:
DAG? Dependency Adjacency Graph? Or ...?

MichaelClerx:
Directed acyclic graph, sorry! The variable equivalence graph is the name we were using I think

kerimoyle:
@nickerso @agarny @hsorby Ready for review please.

src/printer.cpp
```diff
@@ -139,6 +137,8 @@ std::string printConnections(const ComponentMap &componentMap, const VariableMap
 
 std::string printMath(const std::string &math, const std::string &indent)
 {
+    // TODO This becomes unstable: need to remove the automatic addition of the indent so that mathml which
+    // has been parsed from text which already has whitespace in it doesn't get out of control
```
nickerso: good to create an issue for this and list the issue here. Don't think this needs to hold up this PR.

src/printer.cpp
```diff
@@ -351,56 +351,50 @@ std::string Printer::PrinterImpl::printVariable(const VariablePtr &variable, con
...
+    if (!s.empty()) {
         repr += ">\n";
-        repr += printMath(condition, indent + tabIndent);
-    }
-    std::string value = when->value();
-    bool hasValue = !value.empty();
-    if (hasValue) {
-        if (!hasCondition) {
+        // TODO Not happy with how this works - should be more consistent with other calls to printMath but that function
+        // adds extra spaces somehow ... ?
```
nickerso: as above, good to link to the issue for future reference.

src/xmlattribute.cpp
```diff
@@ -62,6 +62,7 @@ std::string XmlAttribute::namespaceUri() const
 bool XmlAttribute::isType(const char *name, const char *ns)
 {
     bool found = false;
+
```
nickerso: seems like a random change that should be taken out of this PR.

src/xmlnode.cpp
```diff
@@ -19,6 +19,7 @@ limitations under the License.
 #include "xmlnode.h"
 
 #include <algorithm>
+#include <iostream>
```
nickerso: not clear why this is here? should remove from PR.

src/xmlnode.cpp
```diff
@@ -68,6 +69,7 @@ bool XmlNode::isElement(const char *name, const char *ns)
         && (xmlStrcmp(reinterpret_cast<const xmlChar *>(namespaceUri().c_str()), reinterpret_cast<const xmlChar *>(ns)) == 0)
         && (xmlStrcmp(mPimpl->mXmlNodePtr->name, reinterpret_cast<const xmlChar *>(name)) == 0)) {
         found = true;
+    } else {
```
nickerso: and this one.

tests/parser/file_parser.cpp
```diff
@@ -25,140 +25,140 @@ limitations under the License.
...
-    libcellml::ModelPtr model = p.parseModel(buffer.str());
-    EXPECT_EQ(size_t(0), p.errorCount());
-
-    std::string a = model->component("c1")->math();
-    EXPECT_EQ(e1, a);
-
-    a = model->component("c2")->math();
-    EXPECT_EQ(e2, a);
-}
+// TEST(Parser, parseSineModelFromFile)
```
nickerso: Why are all these tests being commented out?

tests/parser/libxml_user.cpp
```diff
@@ -24,41 +24,41 @@ limitations under the License.
 
 #include <libxml/parser.h>
 
-TEST(Parser, parseValidXmlDirectlyUsingLibxml)
-{
-    const std::string e =
-        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
-        "<model xmlns=\"http://www.cellml.org/cellml/2.0#\"/>\n";
+// TEST(Parser, parseValidXmlDirectlyUsingLibxml)
```
nickerso: more tests being commented out??

tests/parser/parser.cpp
```diff
@@ -1471,23 +1470,26 @@ TEST(Parser, parseIdsOnEverything)
         "  <units name=\"units3\" id=\"u3id\"/>\n"
         "  <component name=\"component2\" id=\"c2id\">\n"
         "    <variable name=\"variable1\" units=\"blob\" id=\"v1id\"/>\n"
-        "    <reset variable=\"variable1\" order=\"1\" id=\"r1id\">\n"
-        "      <when order=\"5\" id=\"w1id\">\n"
+        "    <variable name=\"variable2\" units=\"blob\" id=\"v2id\"/>\n"
+        "    <reset variable=\"variable1\" test_variable=\"variable2\" order=\"1\" id=\"r1id\">\n"
+        "      <test_value>\n"
```
nickerso: Not sure if I have missed something in the discussion, but I'm expecting the test_value and rest_value to be an expression not an equation. i.e.,
```
<test_value>
   <math>
     <cn cellml:units="variable1units">3.4</cn>
   </math>
</test_value>
```

tests/printer/printer.cpp
```diff
@@ -117,6 +117,7 @@ TEST(Printer, printEmptyReset)
 
     libcellml::Printer printer;
     const std::string a = printer.printModel(m);
+
```
nickerso: not sure why this is here? another random whitespace addition to the PR that should be removed...

tests/reset/reset.cpp
```diff
@@ -41,138 +41,103 @@ TEST(Reset, order)
...
 }
 
-TEST(Reset, addAndCountChildren)
-{
-    libcellml::Reset r;
-    libcellml::WhenPtr child1 = std::make_shared<libcellml::When>();
-    libcellml::WhenPtr child2 = std::make_shared<libcellml::When>();
-    libcellml::WhenPtr child3 = std::make_shared<libcellml::When>();
-    libcellml::WhenPtr child4 = std::make_shared<libcellml::When>();
+// TEST(Reset, removeWhenMethods)
```
nickerso: better to remove old code than to comment out.

tests/reset/reset.cpp
```diff
@@ -201,203 +166,69 @@ TEST(Reset, printResetWithVariable)
...
-        "</math>\n";
 
     libcellml::ModelPtr m = createModelWithComponent();
     libcellml::ComponentPtr c = m->component(0);
-    libcellml::ResetPtr r = std::make_shared<libcellml::Reset>();
-    libcellml::WhenPtr w = std::make_shared<libcellml::When>();
+    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
+    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
+    libcellml::ResetPtr r1 = std::make_shared<libcellml::Reset>();
+    libcellml::ResetPtr r2 = std::make_shared<libcellml::Reset>();
```
nickerso: unused?

tests/reset/reset.cpp
```diff
@@ -201,203 +166,69 @@ TEST(Reset, printResetWithVariable)
...
+    libcellml::ResetPtr r3 = std::make_shared<libcellml::Reset>();
```
nickerso: unused?

tests/reset/reset.cpp
```diff
@@ -476,72 +313,65 @@ TEST(Reset, addRemoveResetFromComponentMethods)
...
-        "  </component>\n"
-        "</model>\n";
     libcellml::ModelPtr m = createModelWithComponent();
     libcellml::ComponentPtr c = m->component(0);
-    libcellml::VariablePtr v = std::make_shared<libcellml::Variable>();
+    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
+    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
     libcellml::ResetPtr r1 = std::make_shared<libcellml::Reset>();
     libcellml::ResetPtr r2 = std::make_shared<libcellml::Reset>();
     libcellml::ResetPtr r3 = std::make_shared<libcellml::Reset>();
```
nickerso, line 264: unused?

tests/reset/reset.cpp
```diff
@@ -201,203 +166,69 @@ TEST(Reset, printResetWithVariable)
...
+    c->addReset(r1);
+    c->addVariable(v1);
+    c->addVariable(v2);
+
+    a = printer.printModel(m);
     EXPECT_EQ(e, a);
 }
 
-TEST(Reset, printResetWithMultipleWhensWithValues)
+TEST(Reset, printResetWithOrder)
```
nickerso, line 134: would be good to add tests with negative orders.

tests/validator/validator.cpp
```diff
@@ -852,143 +857,118 @@ TEST(Validator, integerStrings)
...
+    r6->setOrder(6);
+    // r6->setResetValue(goodMath);
+    r6->setTestValue(goodMath);
+
+    // No test_value
+    r7->setVariable(v1);
+    r7->setTestVariable(v2);
+    r7->setOrder(7);
+    r7->setResetValue(goodMath);
+    // r7->setTestValue(emptyMath);
```
nickerso: would be good to test both options of either not setting the math or setting with an empty string.

nickerso (dismissed):
Looking good. A bit more work to clarify things, test some more combinations, and put the parsing tests back in...

src/printer.cpp
kerimoyle: Done, issue #364

src/printer.cpp
kerimoyle: Done, issue #364

src/xmlattribute.cpp
kerimoyle: Done

src/xmlnode.cpp
kerimoyle: Done (debugging leftovers!)

src/xmlnode.cpp
kerimoyle: Done (debugging leftovers)

tests/parser/parser.cpp
kerimoyle: @nickerso I haven't seen anything either way, other than the contents of test_value and reset_value was to be mathml.  Is there a new version of the spec around resets yet? I've just been going from the use case document?  I'd imagine that the test_value needs to be an expression though? and that there could be cases where you'd want to reset to something related to another variable, so that the reset_value needs to be an expression too?  @MichaelClerx what are your thoughts?

tests/printer/printer.cpp
kerimoyle: Done

tests/reset/reset.cpp
kerimoyle: Done

tests/reset/reset.cpp
kerimoyle: Done

tests/reset/reset.cpp
kerimoyle: Done

tests/reset/reset.cpp
kerimoyle: Umm .. no?

tests/parser/file_parser.cpp
kerimoyle: Back in now

tests/parser/libxml_user.cpp
kerimoyle: Back in now

tests/reset/reset.cpp
kerimoyle: There will be more tests around orders to come when the order checking is in properly - at the moment in this PR its scope is not correct. Related to issue #326 (and discussion earlier in this PR).

tests/validator/validator.cpp
kerimoyle: @nickerso  I've added a test for whitespace only, which returns an equivalent error to reset-value-not-set etc ... but am confused as to whether empty math blocks should be valid or not?  See issue #365 @MichaelClerx

kerimoyle:
> Directed acyclic graph, sorry! The variable equivalence graph is the name we were using I think

@MichaelClerx Is it directed?  Ooops, I'd thought not ... ? Might need to adjust validation checks if so!

tests/reset/reset.cpp
kerimoyle: Order validation is now present in the other PR #350.  Tests will need to be tweaked when this is merged in, so if it's ok will add more tests when they don't need rewriting?

kerimoyle:
Added validation for reset variable and test_variable being inside the same component, and the tests to cover them. Addresses issue #294.

tests/validator/validator.cpp
kerimoyle: Done, including empty maths as per discussion in #365.

tests/reset/reset.cpp
kerimoyle: I've added a test with a negative order to the validator in this branch.

hsorby (changes requested):
TEST(Validator, resets) is way too big and it needs to be broken down into smaller tests.

kerimoyle:
> TEST(Validator, resets) is way too big and it needs to be broken down into smaller tests.

OK, will do.  Micro-tests coming up ... ;)

kerimoyle:
@hsorby @agarny @nickerso I think I've addressed everything requested ... Good to go?

hsorby (changes requested):
We also need to have some printer tests for making sure we print out resets as we expect.

hsorby:
The broken down tests are much better for me I find them much easier to read and make sense of.

kerimoyle:
Not at all DRY though :\

hsorby:
Nope.  But it's testing so essentially you want to repeat yourself.

kerimoyle:
> We also need to have some printer tests for making sure we print out resets as we expect.

We already test printing in the tests/reset.cpp for:

- printResetWithVariable
- printResetWithoutTestValue
- printResetWithOrder
- printResetWithNegativeOrder
- printResetWithOrderAndVariable
- addRemoveResetFromComponentMethods

... and in tests/printer.cpp for:
- printEmptyReset

I just adjusted the existing ones to the new format from what was there previously.  Are there tests that need to be added to these, or should the reset ones be in the printer.cpp file?

hsorby:
There is no right place for them I was just trying to find them and thought they would be in the `tests/printer/printer.cpp` file but I didn't see any changes there.  I'll have another look.

tests/parser/parser.cpp
```diff
@@ -1531,23 +1532,18 @@ TEST(Parser, parseResets)
         "<model xmlns=\"http://www.cellml.org/cellml/2.0#\" id=\"mid\">\n"
         "  <component name=\"component2\" id=\"c2id\">\n"
         "    <variable name=\"variable1\" id=\"vid\"/>\n"
-        "    <reset order=\"1\" id=\"rid\">\n"
-        "      <when order=\"5\">\n"
+        "    <variable name=\"variable2\" id=\"vid2\"/>\n"
+        "    <reset order=\"1\" id=\"rid\" variable=\"variable1\" test_variable=\"variable2\">\n"
+        "      <test_value>\n"
```
hsorby, line 1538: Everything in this test should have an id, this doesn't.

tests/parser/parser.cpp
```diff
@@ -1531,23 +1532,18 @@ TEST(Parser, parseResets)
...
         "        <math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
         "          some condition in mathml\n"
         "        </math>\n"
+        "      </test_value>\n"
+        "      <reset_value>\n"
```
hsorby, line 1543: This should also have an id.

tests/parser/parser.cpp
```diff
@@ -1560,152 +1556,175 @@ TEST(Parser, parseResets)
...
+        "          some value in mathml\n"
+        "        </math>\n";
+    EXPECT_EQ(rt, resetValueString);
 }
 
-TEST(Parser, parseResetsWithNumerousErrors)
+TEST(Parser, parseResetsWithErrors)
 {
-    const std::string input =
+    /* const std::string input =
```
hsorby: This shouldn't be here if it's just commented out.

tests/parser/parser.cpp
```diff
@@ -1560,152 +1556,175 @@ TEST(Parser, parseResets)
...
-    for (size_t i = 0; i < parser.errorCount(); ++i) {
-        EXPECT_EQ(expectedErrors.at(i), parser.error(i)->description());
+    EXPECT_EQ(expectedErrors.size(), p.errorCount());
+    for (size_t i = 0; i < p.errorCount(); ++i) {
+        EXPECT_EQ(expectedErrors.at(i), p.error(i)->description());
     }
 
-    EXPECT_EQ(resetExpected, parser.error(0)->reset());
-    EXPECT_EQ(whenExpected, parser.error(1)->when());
+    std::cout << printer.printModel(model);
```
hsorby: This has to go.

tests/parser/parser.cpp
```diff
@@ -1785,3 +1806,131 @@ TEST(Parser, xmlComments)
...
+    libcellml::ModelPtr model2 = parser.parseModel(output1);
+    std::string output2 = printer.printModel(model2);
+
+    libcellml::ModelPtr model3 = parser.parseModel(output2);
+    std::string output3 = printer.printModel(model3);
+
+    EXPECT_EQ(input, output3);
+}
+
+// TODO Removed so that math print/parse behaviour can be investigated independently through #364
```
hsorby: Not a great job of removing???

hsorby (changes requested):

hsorby:
Trying to find resets API tests `appendResetValue` and so on but haven't managed it can someone point them out to me?

hsorby:
Also do we want to have a `clearResetValue` API etc?

kerimoyle:
> Also do we want to have a `clearResetValue` API etc?

I took the current implementation from what was there previously (see component's mathml functions?).   The same behaviour is found by calling `setResetValue("");` . Easy enough to add, but then will also be needed throughout as well?

kerimoyle:
> Trying to find resets API tests `appendResetValue` and so on but haven't managed it can someone point them out to me?

TEST(Parser, repeatedMathParsePrintBehaviourWithReset)  calls code that uses them.  Does it need to be separately done?

tests/parser/parser.cpp
kerimoyle: True dat ... I left it commented out because of the weirdness I was getting in parsing maths between two branches.

tests/parser/parser.cpp
kerimoyle: Is that a general thing for cellml stuff everywhere?  I thought ids were optional?

kerimoyle: Ignore me, just read the name of the test it's in ... :)

tests/parser/parser.cpp
kerimoyle: Yep

hsorby:
> > Also do we want to have a `clearResetValue` API etc?
> 
> I took the current implementation from what was there previously (see component's mathml functions?). The same behaviour is found by calling `setResetValue("");` . Easy enough to add, but then will also be needed throughout as well?

No I'm pretty sure we decided that setting empty is clear.  That is what everything else does.  Just reminding myself of that.

hsorby:
> TEST(Parser, repeatedMathParsePrintBehaviourWithReset) calls code that uses them. Does it need to be separately done?

I would have that done they are unit tests each part should be tested on it's own.

tests/parser/parser.cpp
kerimoyle: Deleted now that I know that the mathml printing/parsing is inconsistent for invalid maths.

tests/parser/parser.cpp
kerimoyle: Done

tests/parser/parser.cpp
kerimoyle: Done

kerimoyle:
> I would have that done they are unit tests each part should be tested on it's own.

@hsorby Have added four new tests into tests/reset.cpp which deal with each bit individually: 
- testValueSetClear
- testValueAppend
- resetValueSetClear
- resetValueAppend

What's next ;) ?

tests/reset/reset.cpp
```diff
@@ -360,3 +360,266 @@ TEST(Reset, hasResetFromComponentMethod)
...
+    r->appendTestValue(firstMaths);
+    r->appendTestValue(secondMaths);
+
+    const std::string out = p.printModel(m);
+    EXPECT_EQ(test, out);
+}
+
+TEST(Reset, resetValueSetClear)
+{
+    static const std::string emptyMath =
```
hsorby: This should be moved into the test_utils.h file and then used throughout the tests.

tests/reset/reset.cpp
```diff
@@ -360,3 +360,266 @@ TEST(Reset, hasResetFromComponentMethod)
...
+        "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
+        "</math>\n";
+
+    const std::string in1 =
```
hsorby: Not used so delete

tests/reset/reset.cpp
```diff
@@ -360,3 +360,266 @@ TEST(Reset, hasResetFromComponentMethod)
...
+    const std::string out3 = p.printModel(m);
+    EXPECT_EQ(test1, out3);
+}
+
+TEST(Reset, resetValueAppend)
+{
+    static const std::string firstMaths = "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n";
+    static const std::string secondMaths = "</math>\n";
+
+    const std::string in =
```
hsorby: Not used so delete.

tests/reset/reset.cpp
```diff
@@ -360,3 +360,266 @@ TEST(Reset, hasResetFromComponentMethod)
...
+
+    // Test clearing of reset_value block
+    r->setResetValue("");
+    const std::string out3 = p.printModel(m);
+    EXPECT_EQ(test1, out3);
+}
+
+TEST(Reset, resetValueAppend)
+{
+    static const std::string firstMaths = "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n";
```
hsorby, line 545: Could also choose to append two empty maths blocks.

tests/reset/reset.cpp
```diff
@@ -360,3 +360,266 @@ TEST(Reset, hasResetFromComponentMethod)
...
+    EXPECT_EQ(e, a);
+}
+
+TEST(Reset, testValueSetClear)
+{
+    static const std::string emptyMath =
+        "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
+        "</math>\n";
+
+    const std::string in1 =
```
hsorby: Not used so delete

hsorby (changes requested):
A few things to tidy up.

tests/reset/reset.cpp
kerimoyle: Oh, I could have chosen many, many other things ... :)

tests/reset/reset.cpp
kerimoyle: Good plan.

kerimoyle:
I'm missing loads of variables which are declared but unused because there used to be a warning for this which I relied on ... why was it removed?  Seems a more useful warning than the your-test-failed-because-you-missed-a-single-space-tests. Or does anyone know how I do that locally?

tests/reset/reset.cpp
kerimoyle: Done

tests/reset/reset.cpp
kerimoyle: Done

tests/reset/reset.cpp
kerimoyle: Done

kerimoyle:
Good to go? @hsorby @nickerso @agarny

src/api/libcellml/reset.h
```diff
@@ -57,107 +57,112 @@ class LIBCELLML_EXPORT Reset: public OrderedEntity
     VariablePtr variable() const;
 
     /**
-     * @brief Add a child when to this @c Reset.
+     * @brief Set the @c TestVariable for this @c Reset.
      *
-     * Add a copy of the given @p when as a child @c When of this @c Reset.
+     * Sets the @c TestVariable for this @c Reset.
```
agarny: Use the infinitive, i.e. "Set" rather than "Sets". (Check the rest of the PR for similar occurrences.)

kerimoyle: Done :)

src/api/libcellml/reset.h
```diff
@@ -57,107 +57,112 @@ class LIBCELLML_EXPORT Reset: public OrderedEntity
...
+     * @return The @c TestVariable.
      */
-    bool removeWhen(size_t index);
+    VariablePtr testVariable() const;
 
     /**
-     * @brief Remove the @c When with the given pointer.
-     *
-     * Remove the @c When with the pointer @p when.
+     * @brief Appends the argument to the testValue string for this reset.
```
agarny: Replace with something like "@brief Append @p math to the existing testValue string for this @c Reset." (i.e. try to be consistent with what has been done for the `Component` class.) The same for the other API documentation.

kerimoyle: I think we should do this en masse for all of the documentation everywhere once all the pull requests have been merged.  There are lots of inconsistencies everywhere ... !

agarny: Agreed, ignore my comment for now.

kerimoyle:
Nudge ... @hsorby @nickerso

hsorby (commented):
Looks good to me I have made some documentation changes and made a pull request [here](https://github.com/kerimoyle/libcellml-keri/pull/19) to (maybe) address issues I have with the code.

hsorby:
Sorry it took so long to get here.

kerimoyle:
I've merged in @hsorby 's  changes now - all around comments.  Once all the outstanding PRs are into development it would be good to go through the comments again (see @agarny 's earlier comment here ...)

kerimoyle:
So this one is good to merge now :) @hsorby @nickerso @agarny 
ps - not sure why the earlier reviews are still showing as requested, all items are finished.

hsorby (dismissed):

kerimoyle:
Help ... not sure what's going on here ... I've merged the updated development branch into this one just fine by following the instructions here, but now I get a unit test failure on the mergability test?  All tests are passing on my machine and my local branches merged without a hitch.  What does that mean? @hsorby @agarny

agarny:
You might want to try to merge one more time. From experience, GitHug can sometimes report conflicts when there are actually none. It happened to me earlier today with [@hsorby's PR](https://github.com/agarny/libcellml/pull/5).

kerimoyle:
Whew.  Should be good to merge now? @agarny @hsorby @nickerso

kerimoyle:
Nudge nudge ... @hsorby @agarny @nickerso

hsorby:
Looking at it now. :)

hsorby:
There are some TODO's in the parser that mention removing becasuse they are validation.  I think they should definitely be removed and other cases also considered for removal.

kerimoyle:
Thanks @hsorby I'll have another look.  There was a PR at some stage which removed these but I can't find it anymore ... :|

kerimoyle:
... and which other cases are you talking about?

hsorby:
Well ... there are some errors produced from invalid numbers for instance.  I can't decide if this is a more of a validation problem or something we should report that we have changed that may be a surprise to the user thus worthy of informing.

kerimoyle:
I'm just going through the current `parser.cpp `file now and will remove anything that's not related to parsing.  In designing the Issue class I've split up the ideas based on whether people would get an informative error message from the validator alone - there were places in the parser where it made more sense to flag errors as they don't get propagated in a helpful way later.  I'll comment out all the tests and lines which are currently throwing errors here just to check that we're on the same page, then I'll remove them properly.  But .. this should all be part of the bigger discussion about what kind of things and where we want the `Issue` class to report ...?

kerimoyle:
So I've gone back and checked the other messages, in particular around the number formatting stuff.  Even though it's technically validation to check the number format (see below), since the numbers are only _saved_ if they're valid, the validator has no way to access whatever the wrong value was since it wouldn't be saved.  This is one of those examples where once we have the `Issue` class it will make more sense - we can raise a warning in the parser instead of ignoring it altogether ... 

```
        if (attribute->isType("units")) {
            reference = attribute->value();
        } else if (attribute->isType("prefix")) {
            prefix = attribute->value();
        } else if (attribute->isType("exponent")) {
            if (isCellMLReal(attribute->value())) {
                exponent = convertToDouble(attribute->value());
            } else {
                err->setDescription("Unit referencing '" + node->attribute("units") + "' in units '" + units->name() + "' has an exponent with the value '" + attribute->value() + "' that is not a representation of a CellML real valued number.");
                err->setUnits(units);
                err->setRule(SpecificationRule::UNIT_EXPONENT);
                mParser->addError(err);
            }
        } else if (attribute->isType("multiplier")) {
            if (isCellMLReal(attribute->value())) {
                multiplier = convertToDouble(attribute->value());
            } else {
                err->setDescription("Unit referencing '" + node->attribute("units") + "' in units '" + units->name() + "' has a multiplier with the value '" + attribute->value() + "' that is not a representation of a CellML real valued number.");
                err->setUnits(units);
                err->setRule(SpecificationRule::UNIT_MULTIPLIER);
                mParser->addError(err);
            }
```

kerimoyle:
So I've come full circle!  There are a lot of things throughout the parser which we need to figure out how to handle, and this can only be done in a helpful way once the `Issue` class is done.  I think they should stay as-is for the minute and be addressed separately so that it's not confused with the reset format which is the point of this PR?

kerimoyle:
I think it's ready now.  Here's the long story:
- I've put back the previously commented-out error creation in the parser which deal with encapsulations as this PR should only deal with the resets.  The others will be dealt with in #314 and #368 
- I've removed the check for the order in a parsed reset being set, as this is purely validation and can be picked up later
- I've reinstated the checks for validation errors when the **real** cause of that error cannot be tracked downstream, ie: when an invalid piece of text is not saved, so the error that the validator can report is a lot less informative than that which the parser can report.  Things in this category are the number errors (the validator can only say that they're missing, not what's wrong with them, as their value is not saved).
- the other reset errors in the parser follow the same pattern as the parser errors elsewhere, most closely the parsing of encapsulations.  If we want to remove them from parsing the resets, we should consider removing them from parsing the encapsulations ... but that's a story for another PR like the one above.

Short story ... I think it's good to go now?  @hsorby @agarny @nickerso

kerimoyle:
Nudge, nudge ... @hsorby @agarny @nickerso I think this one should go in before the #360 code generation one so that the generated code can be the right format?

kerimoyle:
Pretty pretty please can someone review this ... @hsorby @agarny @nickerso

Issue #359, opened 2019-06-19
Code generation


agarny:
The aim is to come up with a general framework for the generation of code for CellML models.

As a starting point, we want to be able to generate code for any CellML 1.0/1.1 model that has been converted to CellML 2.0 (excluding DAE models).

By default, the framework is to generate C code, but it should be fully configurable so that it can be used to generate other formats, e.g. Python.

**Note:** this issue replaces issue #257.

nickerso:
restricting the scope of this to be code generation for non-DAE CellML 1.0 models, as supported by #360. Extension to this will be worked on in future issues.

PR #360, opened 2019-06-19
Code generation (#359)

agarny:
Addresses issue #359.

Note that this first version of our code generation has the following limitations:
 - It only supports models that have no imports (since the full instantation of models is not yet supported in libCellML). This means that `Generator::GeneratorImpl::compareVariablesByName()` doesn't have its final implementation.
 - It doesn't convert units for connected variables (should some be needed).
 - DAE models are not supported.
 - Resets are not supported.
 - It uses generic names for states, rates and variables, i.e. something like `states[0]` in place of `V`, `rates[0]` in place of `dV_dt`, `variables[0]` in place of `Cm` (cell capacitance).
 - Our `Generator::processModel()` method should start by validating the model and only process it for code generation if there are no validation issues. However, validation has not yet been merged in to the `develop` branch, so this is currently commented out.

There may be other limitations that I can't think of for now, but all of those limitations will obviously be implemented at some point. The idea, at this stage, is to have something that can already be used to generate compilable code for both `C` and `Python`, and this for 99.9% of the models in PMR, which should be the case here.

Here are some gists that result from the `C` and `Python` code generation of some models, together with some very basic numerical integration (forward Euler) code:
 - Fabbri _et al._: [C](https://gist.github.com/agarny/ade4fcea45832ac60e6c0f2e94895f6b) and [Python](https://gist.github.com/agarny/9239d12eeacba594954021ff750c54ae). Note that this model actually contains a (real) division by zero, which is "fine" in C, but not in Python, hence the Python doesn't work out of the box. However, this can be "fixed" by replacing:
```
variables[136] = (3.5988-0.025641)/(1.0+0.0000012155/pow(1.0*variables[45], 1.6951))+0.025641
```
with
```
variables[136] = 0.025641
```
 - HH52: [C](https://gist.github.com/agarny/9187181902d2b736e21665381419bb3f) and [Python](https://gist.github.com/agarny/25538228db9467473df54ef7ff875690).
 - N62: [C](https://gist.github.com/agarny/26dc5b93de7c2c965bfa33f9306fcecb) and [Python](https://gist.github.com/agarny/4962aa3f27c193667c1dde145da20178).
 - Zhang _et al._: [C](https://gist.github.com/agarny/ccc18a30fcb587dd45418e0591eee893) and [Python](https://gist.github.com/agarny/8f2c9e84d4eb4375adecb142d5fa5121).


agarny:
@nickerso, @hsorby, @kerimoyle and @MichaelClerx: I believe I am done with (this first version of) code generation, besides the fact that the Python bindings (and tests) still need to be added, which I will do this when I am back from my holiday on 1 September.

Note that this PR requires other PRs to be merged in first. In other words, this PR contains the code of some other PRs on which it depends, hence 135 modified files while it should clearly be much less! Still, if you click on the `Files changed` and then on `Jump to...` and filter the list using `generator`, you will see the files that are really part of this PR.

(This is a rather big PR and no real way to break it down, so hopefully it's not too buggy...)

hsorby (commented):
I have started going through this PR and have some comments.

1. It seems to work
2. I would change the name of the function 'initializeConstants' to 'initializeVectors'
3. I would like to see Python PEP8 compliant names for the generated code interface functions
4. I would like to see more boiler plate code in the generated code to make the generated code more portable.  For instance I have added the following code to a Python form of the generated code for the HH squid axon model

```
def createStateVector():
    return [nan]*4

def createRateVector():
    return [nan]*4

def createVariableVector():
    return [nan]*18

```

This allows me to readily compute the (forward) Euler solution to the system, as can be seen [here](https://github.com/hsorby/cellsolver)

agarny:
Thanks for the feedback @hsorby, much appreciated!

> 2. I would change the name of the function 'initializeConstants' to 'initializeVectors'

I wouldn't because that method doesn't 1) initialise *all* the vectors (e.g. the `rates` array is not initialised) nor does it 2) initialise all the elements of a given array. It is really *only* about initialising array elements with a constant value.

However, thinking a bit more about it, I would have to agree that `initializeConstants()` is not accurate enough. It is fine for the `variables` array (since its elements that get initialised are really constants), but not for the `states` array (since its elements will likely have their values updated, as a result of the ODE equations being computed).

So, I would probably go for `initializeStatesAndConstants()` instead?

> 4. I would like to see more boiler plate code in the generated code to make the generated code more portable.  For instance I have added the following code to a Python form of the generated code for the HH squid axon model

I wouldn't want to see this kind of code in the generated code. Not only is this specific to Python, but it really has nothing to do with initialising/computing the model. As far as I am concerned, code generation should really just be about generating code for initialising/computing model, nothing else. The API allows us to retrieve the number of states/rates and variables, which should be good enough.

nickerso:
Regarding @hsorby's (4), we do currently have this (or similar) in the code generated by the CellML API. It is quite useful as it allows us to pick up the generated code and use it directly without needing to touch libCellML or the source CellML model - e.g., from exposure pages in PMR.

I think it would be useful if the generated code was portable in this manner, so that a given tool could export code and hand it over to another tool that doesn't use libCellML at all - this is actually a fairly common use case at present.

agarny:
Fair enough, but this kind of boiler plate code should be made configurable through our `GeneratorProfile` class, and shouldn't be there by default. Maybe something could be done after this PR has been merged in?

hsorby:
I don'y see a problem with having it there now, if the tests pass and it works then we should leave it in.  We can add the optional default out after it has been merged.

agarny:
I have had a very quick look at your PR agarny#5 and I can see some of that boiler plate code might be useful, but... before I can properly comment on that PR, it would help if it didn't include the code for PR agarny#3 and PR agarny#4. Not only that, but PR agarny#5 also contains things that have nothing with this issue (e.g. there is some pure CMake-related stuff in there). Whatever the case, it would really help if PR agarny#5 was to be "cleaned up" so that I can really appreciate what it adds. (The same goes with PR agarny#6, PR agarny#7 and PR agarny#8, although those should probably be "cleaned up" once PR agarny#5 has been addressed.)

hsorby:
I built the later PR off earlier ones so it hopefully is an easier task to get them updated.

agarny:
Sure, but I would still expect them to be updated as I merge previous PRs in? In fact, right now, PR agarny#5 has conflicts and lists 66 files as having been changed...

agarny:
@hsorby, your [latest commit](https://github.com/agarny/libcellml/pull/5/commits/237540abb4c9a465f52d89df85a49b9b8f5d9ba8) is wrong. For example, `mEqEqFunctionString` should start with `\n` and end with only one `\n`, not two of them. This change (and other similar ones) ensures that our Python generated code doesn't end with a trailing blank line...

hsorby:
I don't agree with that.  Python functions require two new lines after the function that is why I have the new line at the end of the method declaration.  Only the last function doesn't require the new line.

If the Python generated code is wrong then the test answers are wrong?

hsorby:
> Sure, but I would still expect them to be updated as I merge previous PRs in? In fact, right now, PR [agarny#5](https://github.com/agarny/libcellml/pull/5) has conflicts and lists 66 files as having been changed...

I think this is because you have made changes to the code?

agarny:
> I don't agree with that. Python functions require two new lines after the function that is why I have the new line at the end of the method declaration. Only the last function doesn't require the new line.

That is indeed consistent with the warnings I got from PyCharm with regards to the trailing blank line and which I fixed in commit 5434a3b.

> If the Python generated code is wrong then the test answers are wrong?

Yes, they were. I opened the test answers in PyCharm and got a warning about a trailing blank line (hence commit 5434a3b).



> > Sure, but I would still expect them to be updated as I merge previous PRs in? In fact, right now, PR [agarny#5](https://github.com/agarny/libcellml/pull/5) has conflicts and lists 66 files as having been changed...
> I think this is because you have made changes to the code?

I haven't made any chances to my [issue359 branch](https://github.com/agarny/libcellml/tree/issue359). The last changes I made where related to your PR agarny#4.

hsorby:
When I open the code.py files in my PyCharm from [agarny#5](https://github.com/agarny/libcellml/pull/5) I only get errors about unused parameters.   Nothing on a blank line at end of file.

When I open a code.py file from commit [5434a3b](https://github.com/cellml/libcellml/commit/5434a3b377cc4615af3719d53d9f4030e7d07551) in [pep8online](http://pep8online.com/) I get 'no newline at end of file' amongst other warnings.

I admit I'm confused as to why we a getting different results on this.

agarny:
Now, I am the one who is confused. I have just tried a `code.py` file from agarny#5, commit 5434a3b and my [issue359 branch](https://github.com/agarny/libcellml/tree/issue359) using both PyCharm and [pep8online](http://pep8online.com/), and everything is fine with the end of the file, i.e. no `no newline at end of file` or a message about a trailing blank line.

Ok, maybe best to look into this when you are next in the office.

agarny:
FTR, the reason for the discrepancy between @hsorby and I was because @hsorby was copying/pasting from GitHub, and GitHub doesn't render the trailing `CR` character.

agarny:
@hsorby, @nickerso & @kerimoyle: coming back from my holiday, I saw that @hsorby reviewed this PR, but... it relies on PRs #322 and #344 to be merged in first. Indeed, this PR contains the code of those two other PRs (it made things easier to get this PR ready).

So, before reviewing/accepting this PR (which anyway needs a bit more work since @hsorby has created some PRs for my [issue359 branch](https://github.com/agarny/libcellml/tree/issue359)), we should first review (and merge in) issues #320 (i.e. PR #322) and #321 (i.e. PR #344).

agarny:
@hsorby, @kerimoyle and @nickerso: FWIW, this PR is now ready for (re)reviewing unless we want to add something more to it (e.g. a more accurate versioning information?).

hsorby (commented):
I'm liking this, I need to spend a bit more time going over it to see if there is anything I find distasteful.  

I strongly think we need to implement a versioning that will allow us to determine the state of the generated code.  I think a version for the profile that generated the code and an indication of if it has been modified by an external actor or not.  This would satisfy the need I foresee for applications using generated code.

kerimoyle:
> and an indication of if it has been modified by an external actor or not 

^^ yes to this!
Going through the rest of it now ...

kerimoyle:
So I've had a quick look and will go through in more detail soon.  Sorry to not use the "review" kit here - with so many files and changes it gets really hard to navigate!  Most of my thoughts were coming from a place of ignorance and trying to figure out what was going on, and the things that made this more tricky were:
- not knowing whether variable/function names were acronyms or words, eg: Ode, Ast, Voi etc.  I'm still not sure what Ast means!  Does our clang format stuff complain if we use ODE, VOI etc?  It would make it easier to read and understand?
- should `hasValidModel` be `hasValidModelType`?  I thought it was going to call the validator ... ;)
- should the tagline at the top of the generated file be grammatical?  contents were or content was instead of contents was?
- ditto the function name `isEqual` ... if it's `areEqual` it becomes different from the other ones, if it's `isEqualTo` it's clumsier ... 
- search in generator.cpp for "wich" -> "which"
- I didn't fully understand the implication of the `compareVariablesByName` function comment?

agarny:
Ok, two things:
 - Versioning: right now, we tell which version of libCellML we used to generate a given file. So, to address @hsorby point, I think we could just add something about whether the generated code was done using the (official and therefore unmodified) C profile, the (official and therefore unmodified) Python profile or a customised profiler. In other words, right now we are generating:

```
The contents of this file were generated from version 0.2.0 of libCellML.
```

  while we could now generate one of the following:

```
The contents of this file were generated using the C profile of libCellML 0.2.0.
The contents of this file were generated using the Python profile of libCellML 0.2.0.
The contents of this file were generated using a customized profile and libCellML 0.2.0.
```

  actually, are we using British English or American English?
 - We generate some information for a variable, namely its name, its units and the component from which it comes. However, there is an additional piece of information missing: the type of that variable. It can indeed be useful (for GUI purposes for example) to know whether a variable is a constant, computed constant or algebraic variable. The "problem" though is that right now we are generating code for `VOI_INFO`, `STATE_INFO` and `VARIABLE_INFO`, which matches what we are doing in the Generator API. So, if we want to stick with that (which I think we probably should), it will mean that only `VARIABLE_INFO` will need to have the type of the viarable added to it. Agreed?

agarny:
> * not knowing whether variable/function names were acronyms or words, eg: Ode, Ast, Voi etc.  I'm still not sure what Ast means!

Yes, a few acronyms are used to make some function/variable names quite a bit shorter:
 - Ode: I think you know that one (or should know it! :))
 - Ast: [abstract syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree)
 - Voi: variable of integration

> Does our clang format stuff complain if we use ODE, VOI etc?  It would make it easier to read and understand?

No, it wouldn't (I have just tried `ninja format_code` after having renamed the `Generator::voi()` method to `Generator::VOI()` and it didn't complain about anything), but it wouldn't follow our naming convention, hence `Voi`, etc.

> * should `hasValidModel` be `hasValidModelType`?  I thought it was going to call the validator ... ;)

I would stick to `hasValidModel()`. We want to know whether our generator holds a valid model and it happens that we rely on the model type to tells us that. That's all.

Otherwise, the validator tells us whether a CellML file is valid, but it doesn't and will never tell us whether we are dealing with a valid model, i.e. an algebraic or an ODE (or even a DAE, in the future) model. So, "valid" for the generator doesn't have the same meaning as for the validator.

> * should the tagline at the top of the generated file be grammatical?  contents were or content was instead of contents was?

I am all about grammatical, so thanks for pointing that one out. I think it should be "contents" rather than "content" (I am clearly not a native English speaker, so happy to be corrected!), so we should have "The contents of this file were generated...".

> * ditto the function name `isEqual` ... if it's `areEqual` it becomes different from the other ones, if it's `isEqualTo` it's clumsier ...

I would go with `areEqual()`. So, consider it fixed.

> * search in generator.cpp for "wich" -> "which"

Well spotted. Fixed.

> * I didn't fully understand the implication of the `compareVariablesByName` function comment?

Right now, `compareVariablesByName()` sorts variables based on the name of the component from which they come and, if it's the same component, on the name of the variable. This is fine for now, but at some point our code generator will support imports. So, when it does, we will first need to sort variables based on the model from which they come. To be honest, I don't know whether this will be needed/useful, but I thought I would add a `TODO` just in case.

agarny:
I don't know why, but "The contents of this file were generated..." really sounds weird to me. I would almost be inclined to replace it with "The content of this file was generated...", but I feel like it's not correct ("contents" vs. "content"). Views?

agarny:
Ok, I think I have implemented @hsorby and @kerimoyle's comments. I have also added the type of the variables listed in the `VARIABLE_INFO` array. A few things to keep in mind in that respect (besides the fact that the generated code still needs to be properly tested):
 - **C profile:** we now have a `typedef` for `VariableType`, `VariableInfo` (used by `VOI_INFO` and `STATE_INFO`) and `VariableInfoWithType` (used by `VARIABLE_INFO`). I am not super happy about having to define both `VariableInfo` and `VariableInfoWithType` (since `VariableInfoWithType` is the same as `VariableInfo`, except that it also includes a variable of type `VariableType`), but I can't think of a better way to do this for now.
 - **Python profile:** we create an [enum-based type](https://docs.python.org/3/library/enum.html), which means that, to execute the generated Python code, we need at least Python 3.4. Personally, I am ok with that "constraint".

Finally, the generated C code cannot be used out of the box (unlike the Python code). It does indeed require a header file. I will therefore implement that feature tomorrow...

hsorby:
I would like to see the name of the function `initialize_constants` changed, I myself would prefer `set_initial_conditions`.

hsorby:
I also want to see a version of the profile marked in the generated code and a way to determine if it has been modified from the form defined by the version number.

agarny:
> I would like to see the name of the function `initialize_constants` changed, I myself would prefer `set_initial_conditions`.

Oops, I indeed forgot to deal with that. However, I would rename it to `initializeStatesAndConstants()`, which is really what it is about. For me, to set the initial conditions is what initialising states is about, not initialising constants (subjective, I know). Also, `initializeStatesAndConstants()` would be more consistent with our other methods (`computeComputedConstants()`, `computeRates()` and `computeVariables()`) in the sense that we refer to states, constants, rates, etc.

> I also want to see a version of the profile marked in the generated code 

I understand this bit...

> and a way to determine if it has been modified from the form defined by the version number.

... but not this one...

hsorby:
> > and a way to determine if it has been modified from the form defined by the version number.
> 
> ... but not this one...

It is possible to take our existing profiles and modify them and generate code from this modified profile.  I would like to know if this has been done as there is no knowing what changes have been made and if they would be compatible with my applications code that can work with specific versions of the profile.

agarny:
> > > and a way to determine if it has been modified from the form defined by the version number.
> > 
> > ... but not this one...
> 
> It is possible to take our existing profiles and modify them and generate code from this modified profile. I would like to know if this has been done as there is no knowing what changes have been made and if they would be compatible with my applications code that can work with specific versions of the profile.

Hmm... I feel like this might be quite a bit of work to implement? Could you provide some generated code that you think would be nice to have and that would address your requirements? As you know, right now, we tell in the origin comment whether the file was generated using the C profile, the Python profile or a customised profile. Isn't that enough? (Granted, you can't retrieve that information from some boolean or something.)

tests/generator/generator.cpp
```diff
@@ -0,0 +1,1230 @@
...
+
+    EXPECT_EQ(nullptr, generator.voi());
+    EXPECT_EQ(nullptr, generator.state(0));
+    EXPECT_EQ(nullptr, generator.variable(0));
+
+    EXPECT_EQ(EMPTY_STRING, generator.interfaceCode());
+    EXPECT_EQ(EMPTY_STRING, generator.implementationCode());
+}
+
+TEST(Generator, initialized_variable_of_integration)
```
hsorby: Shouldn't have underscores in these test names.

hsorby (dismissed):
Some of the tests in generator.cpp have used underscores to separate words breaking with other tests.  Test names should be changed to be consistent with existing tests.

tests/generator/generator.cpp
agarny: Good point (that's what happens when you copy/paste). Consider it fixed in a commit soon to be pushed...

agarny:
@hsorby: any idea about [this](https://autotest.bioeng.auckland.ac.nz/buildbot/libcellml/builders/Unit%20Tests%20Builder%20Windows%208%2064%20Bit/builds/1342/steps/RemoveDirectory) (following commit 65e1fa9)?

hsorby:
> @hsorby: any idea about [this](https://autotest.bioeng.auckland.ac.nz/buildbot/libcellml/builders/Unit%20Tests%20Builder%20Windows%208%2064%20Bit/builds/1342/steps/RemoveDirectory) (following commit [65e1fa9](https://github.com/cellml/libcellml/commit/65e1fa9e93b94122eee5a57285c74ac0f0ae61aa))?

It can be Windows not having a handle on the state of the drive, I will re-send the notification and see if the behaviour changes.

agarny:
FWIW, right now the `Generator::processModel()` method looks as follows:

```
void Generator::processModel(const ModelPtr &model)
{
    // Make sure that the model is valid before processing it.

    /*TODO: enable the below code once validation is known to work fine.

    validator.validateModel(model);

    if (validator.errorCount() > 0) {
        // The model is not valid, so retrieve the validation errors and make
        // them our own.

        for (size_t i = 0; i < validator.errorCount(); ++i) {
            addError(validator.error(i));
        }

        return;
    }
*/

    // Process the model.

    mPimpl->processModel(model);
}
```

As you can see, we first check that a model is valid before processing it as such. However, until now, that check was commented out since I was waiting for @kerimoyle to be done with her validation work. So, now that she is done with it, I thought I would uncomment that code and see what happens, and... a lot of errors get generated. For example, for the HH52 model, we get the following errors:

```
Error #1: Namespace prefix cellml for units on cn is not defined. [component: n/a] [variable: n/a]
Error #2: Namespace prefix cellml for units on cn is not defined. [component: n/a] [variable: n/a]
Error #3: Namespace prefix cellml for units on cn is not defined. [component: n/a] [variable: n/a]
Error #4: Namespace prefix cellml for units on cn is not defined. [component: n/a] [variable: n/a]
Error #5: Math in component 'membrane' contains 'time' as a bvar ci element but it is already a variable name. [component: membrane] [variable: n/a]
Error #6: Math cn element has an invalid attribute type 'cellml:units' in the cellml namespace. [component: membrane] [variable: n/a]
Error #7: CellML identifiers must contain one or more basic Latin alphabetic characters. [component: n/a] [variable: n/a]
Error #8: Math cn element with the value '20' does not have a valid cellml:units attribute. [component: membrane] [variable: n/a]
Error #9: Math cn element has an invalid attribute type 'cellml:units' in the cellml namespace. [component: membrane] [variable: n/a]
Error #10: CellML identifiers must contain one or more basic Latin alphabetic characters. [component: n/a] [variable: n/a]
Error #11: Math cn element with the value '10' does not have a valid cellml:units attribute. [component: membrane] [variable: n/a]
Error #12: Math cn element has an invalid attribute type 'cellml:units' in the cellml namespace. [component: membrane] [variable: n/a]
Error #13: CellML identifiers must contain one or more basic Latin alphabetic characters. [component: n/a] [variable: n/a]
Error #14: Math cn element with the value '10.5' does not have a valid cellml:units attribute. [component: membrane] [variable: n/a]
Error #15: Math cn element has an invalid attribute type 'cellml:units' in the cellml namespace. [component: membrane] [variable: n/a]
Error #16: CellML identifiers must contain one or more basic Latin alphabetic characters. [component: n/a] [variable: n/a]
Error #17: Math cn element with the value '0' does not have a valid cellml:units attribute. [component: membrane] [variable: n/a]
Error #18: Namespace prefix cellml for units on cn is not defined. [component: membrane] [variable: n/a]
Error #19: No declaration for attribute cellml:units of element cn. [component: membrane] [variable: n/a]
Error #20: Namespace prefix cellml for units on cn is not defined. [component: membrane] [variable: n/a]
Error #21: No declaration for attribute cellml:units of element cn. [component: membrane] [variable: n/a]
Error #22: Namespace prefix cellml for units on cn is not defined. [component: membrane] [variable: n/a]
Error #23: No declaration for attribute cellml:units of element cn. [component: membrane] [variable: n/a]
Error #24: Namespace prefix cellml for units on cn is not defined. [component: membrane] [variable: n/a]
Error #25: No declaration for attribute cellml:units of element cn. [component: membrane] [variable: n/a]
Error #26: Namespace prefix cellml for units on cn is not defined. [component: n/a] [variable: n/a]
Error #27: Math cn element has an invalid attribute type 'cellml:units' in the cellml namespace. [component: leakage_current] [variable: n/a]
Error #28: CellML identifiers must contain one or more basic Latin alphabetic characters. [component: n/a] [variable: n/a]
Error #29: Math cn element with the value '10.613' does not have a valid cellml:units attribute. [component: leakage_current] [variable: n/a]
Error #30: Namespace prefix cellml for units on cn is not defined. [component: leakage_current] [variable: n/a]
Error #31: No declaration for attribute cellml:units of element cn. [component: leakage_current] [variable: n/a]
Error #32: Namespace prefix cellml for units on cn is not defined. [component: n/a] [variable: n/a]
Error #33: Namespace prefix cellml for units on cn is not defined. [component: n/a] [variable: n/a]
Error #34: Math cn element has an invalid attribute type 'cellml:units' in the cellml namespace. [component: sodium_channel] [variable: n/a]
Error #35: CellML identifiers must contain one or more basic Latin alphabetic characters. [component: n/a] [variable: n/a]
Error #36: Math cn element with the value '115' does not have a valid cellml:units attribute. [component: sodium_channel] [variable: n/a]
Error #37: Math cn element has an invalid attribute type 'cellml:units' in the cellml namespace. [component: sodium_channel] [variable: n/a]
Error #38: CellML identifiers must contain one or more basic Latin alphabetic characters. [component: n/a] [variable: n/a]
Error #39: Math cn element with the value '3' does not have a valid cellml:units attribute. [component: sodium_channel] [variable: n/a]
Error #40: Namespace prefix cellml for units on cn is not defined. [component: sodium_channel] [variable: n/a]
Error #41: No declaration for attribute cellml:units of element cn. [component: sodium_channel] [variable: n/a]
Error #42: Namespace prefix cellml for units on cn is not defined. [component: sodium_channel] [variable: n/a]
Error #43: No declaration for attribute cellml:units of element cn. [component: sodium_channel] [variable: n/a]
Error #44: Namespace prefix cellml for units on cn is not defined. [component: n/a] [variable: n/a]
Error #45: Namespace prefix cellml for units on cn is not defined. [component: n/a] [variable: n/a]
Error #46: Math cn element has an invalid attribute type 'cellml:units' in the cellml namespace. [component: potassium_channel] [variable: n/a]
Error #47: CellML identifiers must contain one or more basic Latin alphabetic characters. [component: n/a] [variable: n/a]
Error #48: Math cn element with the value '12' does not have a valid cellml:units attribute. [component: potassium_channel] [variable: n/a]
Error #49: Math cn element has an invalid attribute type 'cellml:units' in the cellml namespace. [component: potassium_channel] [variable: n/a]
Error #50: CellML identifiers must contain one or more basic Latin alphabetic characters. [component: n/a] [variable: n/a]
Error #51: Math cn element with the value '4' does not have a valid cellml:units attribute. [component: potassium_channel] [variable: n/a]
Error #52: Namespace prefix cellml for units on cn is not defined. [component: potassium_channel] [variable: n/a]
Error #53: No declaration for attribute cellml:units of element cn. [component: potassium_channel] [variable: n/a]
Error #54: Namespace prefix cellml for units on cn is not defined. [component: potassium_channel] [variable: n/a]
Error #55: No declaration for attribute cellml:units of element cn. [component: potassium_channel] [variable: n/a]
```

Some of them are clearly "expected" (the ones about the CellML namespace, although they should clearly be fixed), but others are not (e.g. the fifth error).

So, for now, I am leaving that piece of code commented out, and we will likely have to merge this PR in as-is. However, we should clearly do something about it at some point, i.e. once we have fixed the CellML namespace issue and established what is "wrong" with the other errors.

hsorby:
I think resolution of #175 should fix a lot of those error s we see there.

kerimoyle:
A couple of things ... firstly, I'm not done with validation at all, there are still three ( #358 and #350 .. then I'll finally be able to finish #368) PRs which have been waiting for weeks to be merged in :( ...  That means that there are some fairly fundamental bits missing from the validator at the moment! Secondly I'm not sure whether you mean that the errors are wrong - are they false positives?  Or just ones you didn't expect? Or something else?
I'm glad that there is so much progress on the code generation ... and it would be lovely to be able to also keep moving on the validation too ... nudge nudge ... ;)

agarny:
Thanks, @kerimoyle, for reminding me about validation not being complete yet. Ok, let's wait until it's done before drawing any final conclusion. :)

src/generator.cpp
```diff
@@ -1672,20 +1945,39 @@ void Generator::GeneratorImpl::addImplementationHeaderCode(std::string &code)
...
+
+    if ((interface && !mProfile->interfaceVersionString().empty())
+        || (!interface && !mProfile->implementationVersionString().empty())) {
+        if (interface) {
+            versionAndLibcellmlCode += mProfile->interfaceVersionString();
+        } else {
+            if (modifiedProfile()) {
+                std::regex regEx("([0-9]+\\.[0-9]+\\.[0-9]+)");
+
+                versionAndLibcellmlCode += std::regex_replace(mProfile->implementationVersionString(), regEx, "$1.modified");
```
hsorby: I originally used `post0` because it is part of the definition of version numbering that is understood by Python.  I haven't checked to see if it is part of the semantic definition.  Appending `modified` is not part of that standard.

agarny: Ok, am going to revert that bit, no problem.

agarny: Done in commit b1186ee.

agarny:
Ok, the SHA-1 code needs a bit of reworking to please Clang, so will look into it tomorrow...

agarny:
@hsorby, @kerimoyle & @nickerso: ok, it looks like we might be on track for this PR to be (re)reviewed... 🤔

hsorby (changes requested):
The classes `Generator`, `GeneratorVariable`, and `GeneratorProfile` require class level documentation.

src/api/libcellml/generatorprofile.h
```diff
@@ -18,6 +18,8 @@ limitations under the License.
 
 #include "libcellml/exportdefinitions.h"
 
+#include <string>
```
agarny: This is not needed since it's already included by `libcellml/exportdefinitions.h`.

agarny:
Regarding commit 961edb8, where is this coming from? I mean, how did the C++11 version of the SHA-1 code get lost in the first place?

Otherwise, we seem to be having some issues with Python binding. Will see what I can do about it...

src/api/libcellml/generatorprofile.h
hsorby: Sorry, I guess I figured I was having a problem with std::string in the Python bindings.  Whilst it is actually std::shared_ptr.

agarny:
> The classes `Generator`, `GeneratorVariable`, and `GeneratorProfile` require class level documentation.

This should be addressed in commit 9c3152f.

agarny:
I guess I should now update the documentation and diagrams...

agarny:
I wanted to update the object models using LucidChart, but I only have a free account and can thus only edit things existing objects, not add new ones, copy/paste, etc. In other words, I cannot add the `Generator`, `GeneratorVariable` and `GeneratorProfile` classes to our object models.

src/api/libcellml/generator.h
```diff
@@ -67,6 +74,11 @@ class LIBCELLML_EXPORT GeneratorVariable
     GeneratorVariableImpl *mPimpl;
 };
 
+/**
+ * @brief The Generator class.
+ *
+ * The Generator class is for representing a CellML Generator.
```
hsorby, line 80: I don't really think this documentation helps anyone at all.

hsorby (approved):
I have mis-givings about the documentation (or lack of) and barring the addition of the update to the diagram (which is something that I think could be added as another issue) I am going to approve this PR and seek to fill any holes with future work.

src/api/libcellml/generator.h
agarny: Agreed, I was merely mimicking what is already in the codebase...

kerimoyle:
I'm still not a fan of the function name `hasValidModel`, as it's testing the `modelType`, and everywhere else in the code it's referred to as `modelType` too ... I just find it inconsistent.  But will bow to popular opinion if others are happy with it :)

kerimoyle (approved):

hsorby:
> I'm still not a fan of the function name `hasValidModel`, as it's testing the `modelType`, and everywhere else in the code it's referred to as `modelType` too ... I just find it inconsistent. But will bow to popular opinion if others are happy with it :)

`hasValidModel` is a private implementation only method and thus has no bearing on the API.  We don't need to put too much weight on internal methods like this.  We need to focus our attention on the API and the tests, those are the aspects of the library that we need to have at the highest level we can manage.

Issue #366, opened 2019-06-21
Use llvm coverage tool


hsorby:
llvm has a coverage tool that we could make use of as a further check on our coverage test.

agarny:
That would be nice since we couldn't then test coverage on any platform, not just on Linux.

PR #367, opened 2019-06-21
Llvm coverage

hsorby:
Adding in targets that take advantage of llvm coverage if available.

Addresses #366.


CMakeLists.txt
```diff
@@ -45,6 +45,18 @@ elseif(COVERAGE)
 endif()
 unset(COVERAGE CACHE)
 
+# LLVM_COVERAGE ==> LIBCELLML_LLVM_COVERAGE
+set(_PARAM_ANNOTATION "Enable OSX coverage testing.")
```
agarny: Why OSX? I would use macOS. Also, why only have LLVM-based coverage on macOS? Surely, it should work on all platforms?

hsorby: Search and replace has failed me :(

CMakeLists.txt
```diff
@@ -46,14 +46,14 @@ endif()
 unset(COVERAGE CACHE)
 
 # LLVM_COVERAGE ==> LIBCELLML_LLVM_COVERAGE
-set(_PARAM_ANNOTATION "Enable OSX coverage testing.")
+set(_PARAM_ANNOTATION "Enable llvm coverage testing.")
```
agarny: FWIW, it ought to be `LLVM` and not `llvm`.

hsorby: Could have saved some typing if you had just made the change :)

agarny: Would be too easy. :)

tests/coverage/coverage.cpp
```diff
@@ -33,9 +33,7 @@ TEST(Coverage, connectionComment)
         "  </component>\n"
         "  <connection component_1=\"component1\">\n"
         "    <!-- Comment on connection. -->\n"
-        "    <map_variables variable_1=\"variable1\" variable_2=\"variable2\">"
-            "<!-- Comment in a map variables. -->"
-            "</map_variables>"
+        "    <map_variables variable_1=\"variable1\" variable_2=\"variable2\"><!-- Comment in a map variables. --></map_variables>\n"
```
agarny: Shouldn't we rather have:

```
        "    <map_variables variable_1=\"variable1\" variable_2=\"variable2\">\n"
        "        <!-- Comment in a map variables. -->\n"
        "    </map_variables>\n"
```

So that it's easier to read...?

tests/coverage/coverage.cpp
```diff
@@ -80,7 +78,7 @@ TEST(Coverage, importWithNonHrefXlink)
     libcellml::Parser parser;
     parser.parseModel(e);
     EXPECT_EQ(size_t(0), parser.errorCount());
-//    EXPECT_EQ("bob", parser.error(0)->description());
+    //    EXPECT_EQ("bob", parser.error(0)->description());
```
agarny: Is that something that you actually want to merge in? I am definitely not in favour of merging code that has been commented out.

tests/coverage/coverage.cpp
hsorby: Can't it doesn't have the same meaning.

tests/coverage/coverage.cpp
hsorby: Yes ... and no, I left it in to remind myself of the error that gets generated when testing the attribute.  I'll take it out.

tests/coverage/coverage.cpp
agarny: (Obviously) agreed that it doesn't have the same meaning. It's just that it would be better if it could read that way unless the test really doesn't make sense with those `\n` characters.

hsorby: This test is for the purpose of satisfying llvm coverage reporting.  This line is the tidiest I have for triggering the code path.

I'll have another look at the code in question and see if changes there would work better.  This is a very edge case which probably could be re-worked into a better code path.

hsorby:
This should be good to go now.

cmake/environmentchecks.cmake
```diff
@@ -12,12 +12,16 @@
...
+include(CheckCXXCompilerFlag)
+
 get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
 
 find_package(Python ${PREFERRED_PYTHON_VERSION} COMPONENTS Interpreter Development)
 
 find_program(CLANG_FORMAT_EXE NAMES ${PREFERRED_CLANG_FORMAT_NAMES} clang-format)
 find_program(CLANG_TIDY_EXE NAMES ${PREFERRED_CLANG_TIDY_NAMES} clang-tidy)
+find_program(LLVM_PROFDATA_EXE NAMES ${PREFERRED_LLVM_PROFDATA_NAMES} llvm-profdata HINTS /Library/Developer/CommandLineTools/usr/bin/)
+find_program(LLVM_COV_EXE NAMES ${PREFERRED_LLVM_COV_NAMES} llvm-cov HINTS /Library/Developer/CommandLineTools/usr/bin/)
```
agarny: 1. The programs to be found are currently listed alphabetically, so it might be best to keep it that way?
2. Are the hints really needed? Also, they seem to be tageted at macOS? What about Windows and Linux?

cmake/environmentchecks.cmake
```diff
@@ -56,29 +70,39 @@ else()
...
+  CLANG_FORMAT_EXE
+  GIT_EXE
+  FIND_EXE
+  GCOV_EXE
+  SWIG_EXECUTABLE
+  VALGRIND_EXE
+  LLVM_COV_EXE
+  LLVM_PROFDATA_EXE
+  LLVM_COVERAGE_COMPILER_FLAGS_OK
+  GCC_COVERAGE_COMPILER_FLAGS_OK
```
agarny: Might want to keep that list alphabetically sorted?

src/parser.cpp
```diff
@@ -668,7 +668,7 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr
                     err->setKind(Error::Kind::CONNECTION);
                     mParser->addError(err);
                 }
-            } else if (childNode->isComment()) {
+            } else if (grandchildNode->isComment()) {
```
agarny, line 671: Nice catch!

src/xmlattribute.h
```diff
@@ -55,16 +55,28 @@ class XmlAttribute
...
+    /**
+     * @brief Test if this XmlAttribute is in the given namespace.
+     *
+     * Test if this XmlAttribute is in the given namespace. Return @c true
+     * if this XmlAttribute is in the given namespace and @c false otherwise.
+     *
+     * @param ns The @c char namespace in which the attribute is in.
+     *
+     * @return @c true if this XmlAttribute is in the namespace
+     * specified by the @p ns and @c false otherwise.
```
agarny: `by @p ns` rather than `by the @p ns`?

agarny:
This looks good to me, pending some minor issues above.

cmake/environmentchecks.cmake
hsorby: 1. Ok
2. Yes. I haven't tested on Windows and Linux that have llvm, a case of leave it until actually required.  I was thinking about adding in a variable to add to the hints so a user could add a configuration parameter that would enable finding of llvm in non-standard locations.

cmake/environmentchecks.cmake
hsorby: Ok

cmake/environmentchecks.cmake
agarny: I have LLVM installed on my Windows, Linux and macOS machines, so I guess I will eventually find out how that behaves... :)

hsorby: Excellent.

tests/coverage/coverage.cpp
hsorby: This has been addressed with changes to handling grandchildren of connections.

nickerso (approved):

agarny (approved):

Issue #302, opened 2019-04-10
Remove validation code form parser.


hsorby:
Currently the parser has taken upon itself to do some validation.  I think the parser needs to concentrate more on parsing the xml and leaving the validation to the validator.

agarny:
Fully agree with that!

kerimoyle:
It sounds like you have something specific in mind ... ?

hsorby:
Nope, I just want the parser to concentrate on parsing and leaving validation to the validator.  Should be an easy fix.

kerimoyle:
You'll know from my previous comments elsewhere that I don't yet have a good feeling for what should be a parse error and what should be a validation error ... so ... 

- Should it be possible to parse a file which contains a valid model _and_ a bunch of unrecognised rubbish, and print a valid model from it without encountering an error?  Essentially a cleaner for the input?
- Should it be possible to parse an incomplete file without error? By incomplete I mean by cellml standards, not xml (eg empty encapsulation)

Reading through the parser there are several different kinds of things which create an error, but which could perhaps become a warning instead (if we make warnings possible, see #314):
- presence of unrecognised attributes
- presence of nonwhitespace text (which has no other meaning)
- presence of invalid children of nodes

... and finally, is there consensus on how error messages are constructed?  Some of the time there are single quotes around things specific to the message (eg: component 'myComponent' has a problem) and sometimes no quotes used?

Thanks :)
https://github.com/kerimoyle/libcellml-keri/tree/remove_validation_from_parser

agarny:
> * Should it be possible to parse a file which contains a valid model _and_ a bunch of unrecognised rubbish, and print a valid model from it without encountering an error?  Essentially a cleaner for the input?

For me, a parser must ensure that a CellML file's syntax is fine while a validator ensures that its semantics is. So, if a CellML file contains unrecognisable rubbish then I would expect the parser to report it.

> * Should it be possible to parse an incomplete file without error? By incomplete I mean by cellml standards, not xml (eg empty encapsulation)

Based on what I have just said above: yes.

> Reading through the parser there are several different kinds of things which create an error, but which could perhaps become a warning instead (if we make warnings possible, see #314):
> 
> * presence of unrecognised attributes
> * presence of nonwhitespace text (which has no other meaning)
> * presence of invalid children of nodes

Agreed, this is actually the approach I took in my own "old" CellML API.

[...]

Issue #280, opened 2019-02-27
Validation should check for recursive imports


agarny:
Using OpenCOR (and, for now, the CellML API), @s-fong ended up with a model that was recursively importing a CellML file, resulting in OpenCOR hanging up and memory usage increasing by the second.

So, this issue is mainly so that we don't forget about checking for recursive imports during code generation.

nickerso:
Should we only try to generate code for "valid" models ? that would avoid this kind of issue...

agarny:
It will be very easy to check for recursive imports, so I would definitely have it in... otherwise people who end up in that situation will blame the tool that uses libCellML and may wrongly assume that it's buggy, decide not to use it anymore, etc.

nickerso:
sorry, meant that when you ask for code to be generated libCellML would first check for a valid model before trying to generate code.

Would it ever make sense to try and generate code for an invalid model?

agarny:
> sorry, meant that when you ask for code to be generated libCellML would first check for a valid model before trying to generate code.

Oh, ok. Yes, I would indeed check that a file is valid before actually generating code for it (this is what COR does for example). Ok, I am going to change the subject to make it clear that the check for recursive imports should be done at the validation stage.

> Would it ever make sense to try and generate code for an invalid model?

It wouldn't in my opinion, but then again the CellML API does... :)

kerimoyle:
Hi all
Just checking in on this one. I had some discussion in #293 with @hsorby that we should be able to import and export invalid models, so wanted to know what the consensus was?  
Secondly, do I have it right that forbidding the import of self is *only* to prevent recursion; that duplicate identical models are permitted, as long as they don't import themselves?

MichaelClerx:
> Just checking in on this one. I had some discussion in #293 with @hsorby that we should be able to import and export invalid models, so wanted to know what the consensus was?

If you mean loading and saving then yes! This is something editors will want (so that we can fix the model without manually fixing the XML)

Secondly, do I have it right that forbidding the import of self is only to prevent recursion; that duplicate identical models are permitted, as long as they don't import themselves?

Yup! For example, a model might have 3 copies of some imported component

kerimoyle:
OK, thanks.  So it should be sufficient to test names and/or ids of the imports against the parent? That is, their content could be identical as long as it comes from a model with a different name that's ok?

agarny:
[...]

Issue #361, opened 2019-06-19
Validator: Check for cyclic unit definitions


kerimoyle:
Validator needs to check that:

- unit definitions are not cyclic
- units are the same between equivalent variables
- tests
- coverage

NB no checks for multiplier or prefix equivalence, only base units.

kerimoyle:
Addressed in #351

PR #368, opened 2019-06-21
Issue314: Issue class for nonfatal warnings

kerimoyle:
As discussed in Issue #314 I'd like to create an `Issue` class.  This will replace the `Error` class, but include levels of severity to mimic the current `Error` behaviour, as well as allowing non-fatal warnings as well.

- [ ] merge currently outstanding PRs involving parsing, printing and validation into develop
- [x] add structure for replacement class
- [ ] add non-fatal warnings as needed
- [ ] tests
- [x] coverage

Things that could be warnings include:
- multiplier mismatch in equivalent units
- multiplier mismatch in maths
- units mismatch in maths
- parsed model parts which cause validation errors (?) as in #302 
- cyclic or recursive imports (?) as in #280 
- exponents with units (should be dimensionless) as in #509 
- maths statements which are not assignments (ie: valid MathML/CellML but not runable)
- anything which will cause issues in the Generator, as in #492, #493, #499, 
- ... anything else ... ?


src/api/libcellml/error.h
```diff
@@ -122,6 +122,20 @@ class LIBCELLML_EXPORT Error
...
+     * 
+     * Enum to describe the level of the error
+     */
+    enum class Level
+    {
+        UNDEFINED,
+        FATAL,
+        WARNING,
+        FUTURE_ERROR,
+        INFORMATION
```
agarny: 1. I guess the class is going to be renamed from `Error` to `Issue` at some point, and that the `.cpp` and `.h` files are going to be renamed too?
2. Can we and do we really want an `UNDEFINED` issue? Also, what is a `FUTURE_ERROR`? I would just have the type (not level) of issue that we think we should have in the end, which I imagine are: `ERROR`, `WARNING` and `HINT`?

hsorby:
The whole point to TDD is to write the tests first to show how the code would look and act not start with the implementation.  This allows people to either agree or disagree with how it will be used, the implementation is irrelevant to the user.  It may be that you write some skeleton implementation that does nothing and therefore fails at least some of the tests but here we have no tests written.  But this then allows you to at least compile the code and not get distracted with all the prompting that IDEs show developers.

src/api/libcellml/error.h
kerimoyle: 1. Agreed ... but I've shelved this until all the other outstanding PRs are merged in as there's just no point in doing it all twice!
2. Agreed as well - the default is `FATAL` (at present) to be backward consistent.  It was there because there was an `UNDEFINED` in the other structures and I was following along ... 
FUTURE_ERROR was something I was wondering about - we could use it two ways.  A lot of packages have deprecation warnings, so it could either be that.  But what I was actually heading for was "you have something which is allowed in the parser but will not be valid" or "you have something which is valid but which will not actually run or mean anything", ie: something which conforms to the specs and rules but will create a problem further down the pipeline.  Not much difference from a warning, really ...

agarny: Others may chime in, but:

1. How do you mean no point in doing it all twice? Are you referring to the fact that other PRs currently rely on the `Error` class? Either way, this issue is about reworking the `Error` class. So, I would expect this PR to be a full rework, not a half-baked one.
2. Why should we be backward compatible? There is no official release of libCellML, so it's not like we are going to break something. Otherwise, I am not convinced by `FATAL` and `UNDEFINED`. `FATAL` is subjective, so will leave the others to comment, but `UNDEFINED` is just wrong to me. libCellML should know what kind of issue it reports. As for `FUTURE_ERROR`, your first use could be addressed with a `WARNING` issue. As for your second possible use, I don't think the parser should worry about what is allowed, but is not valid. This will be the job of the model analyser to determine whether a valid model (in the parser sense) makes sense or not.

kerimoyle:
> The whole point to TDD is to write the tests first to show how the code would look and act not start with the implementation. This allows people to either agree or disagree with how it will be used, the implementation is irrelevant to the user. It may be that you write some skeleton implementation that does nothing and therefore fails at least some of the tests but here we have no tests written. But this then allows you to at least compile the code and not get distracted with all the prompting that IDEs show developers.

Agreed!  To be clear ... all this commit did was get the ball rolling by adding a level indicator to the error class.  I'm also aware that we have different backgrounds/thoughts on how TDD works, and I'm really struggling to think of a suitable test for an error message which doesn't exist.  If you can give me an example of what kind of test you would have expected as a starting point in this situation that would be really useful for me to learn from :)

src/api/libcellml/error.h
kerimoyle: > "Undefined is wrong to me" 

What's the rationale of having it elsewhere?  I was just following what's already there.  Do we need to remove `UNDEFINED` in general then?

> Fatal is subjective

I don't understand ... Currently all our errors are effectively fatal errors hence use of that as a default?

> This will be the job of the model analyser to determine whether a valid model (in the parser sense) makes sense or not.

Maybe I'm not understanding the purpose of this class the same way everyone else is - I'd imagined that it was to better support model analysis and debugging by returning to the user (when asked) where possible glitches and errors might occur, based on information that libcellml currently knows (like stuff from the parser/validator/imports) but doesn't currently report?

> How do you mean no point in doing it all twice? Are you referring to the fact that other PRs currently rely on the Error class? Either way, this issue is about reworking the Error class. So, I would expect this PR to be a full rework, not a half-baked one.

Fully-baked is intended :) ... but it's also (nudge nudge) been my experience that PRs take rather a long time to be merged in, which means that their fix-merge-conflicts-and-design-reworks-factor-to-accommodate-work-done-since-writing becomes pretty high.  There's no point - for example - in going through the current errors in development regarding resets, validation, parsing, whens, etc as these won't exist as soon as the outstanding PRs are merged in - it's wasted work.  (If that sounds overly narky I apologise - today, the sarcasm is strong with this one!)

kerimoyle:
Only pushing here so I can see the python tests that are failing ...

src/api/libcellml/error.h
agarny: > > "Undefined is wrong to me"
> What's the rationale of having it elsewhere? I was just following what's already there. Do we need to remove `UNDEFINED` in general then?

When it comes to a libCellML issue, I would remove the `UNDEFINED` type since an undefined issue doesn't make sense to me. There may be other cases where an `UNDEFINED` type might be necessary. For example, we might need such a type when trying to determine the type of a variable in a model (i.e. state, constant, computed constant, etc. or undefined indeed, if we couldn't determine the type of the variable, e.g. the model is under/overconstrained).

> > Fatal is subjective
> I don't understand ... Currently all our errors are effectively fatal errors hence use of that as a default?

Sorry, I meant that the naming of the type itself is subjective. IOW, I would personally have `ERROR` rather than `FATAL`.

> > This will be the job of the model analyser to determine whether a valid model (in the parser sense) makes sense or not.
> 
> Maybe I'm not understanding the purpose of this class the same way everyone else is - I'd imagined that it was to better support model analysis and debugging by returning to the user (when asked) where possible glitches and errors might occur, based on information that libcellml currently knows (like stuff from the parser/validator/imports) but doesn't currently report?

For me (others may have a different view), the `Issue` class is to be used to report any kind of issue that we come across, be it either as a result of parsing a model, validating a model, analysing a model, etc. In that context, I feel like we need to be able to specify whether an issue is an error, a warning or a hint (any other relevant type?).

> > How do you mean no point in doing it all twice? Are you referring to the fact that other PRs currently rely on the Error class? Either way, this issue is about reworking the Error class. So, I would expect this PR to be a full rework, not a half-baked one.
> 
> Fully-baked is intended :) ... but it's also (nudge nudge) been my experience that PRs take rather a long time to be merged in, which means that their fix-merge-conflicts-and-design-reworks-factor-to-accommodate-work-done-since-writing becomes pretty high. There's no point - for example - in going through the current errors in development regarding resets, validation, parsing, whens, etc as these won't exist as soon as the outstanding PRs are merged in - it's wasted work. (If that sounds overly narky I apologise - today, the sarcasm is strong with this one!)

Trust me, I do sympathise with your sentiment with regards to the merging of PRs... :)

kerimoyle: > There may be other cases where an UNDEFINED type might be necessary.

I'm not talking about in the maths part of the code - it's in the existing `Error::Kind` enum (see below).  Personally I think that having a meaningful default (probably fatal/error/you-shall-not-pass/whatever) is better than undefined, but was trying to be consistent ;) . BTW the fact that there's a "kind" means I don't really want to have a "type" as well - they're too close in implication!  I think "level" or "severity" or something which implies effect or outcome (cf "kind" which implies cause) is clearer? 

```
    /**
     * @brief The error Kind enum class.
     *
     * Enum to describe the kind of error a given error is.
     */
    enum class Kind
    {
        COMPONENT,
            ... 
        RESET,
        UNDEFINED,  // <<<<<<<<<
        UNITS,
        VARIABLE,
        WHEN,
        XML
    };

// also in specificationrules.h:

/**
 * @brief The SpecificationRule enum class.
 *
 * Enum describing the specification rules that are used
 * for validation.
 */
enum class SpecificationRule
{
    UNDEFINED,  // <<<<<<<<<<<<<<<<<
    DATA_REPR_IDENTIFIER_UNICODE,
    DATA_REPR_IDENTIFIER_LATIN_ALPHANUM,
    DATA_REPR_IDENTIFIER_AT_LEAST_ONE_ALPHANUM,
        ...
```

> Sorry, I meant that the naming of the type itself is subjective. IOW, I would personally have ERROR rather than FATAL.

That's easy enough :)

kerimoyle:
Hi all
I hope what's here makes sense.  I've had a go at writing two tests for how I could imagine the `Issue` class being used, based on how the `Error` class has worked and introducing a `Issue::Level` indicator enum (`ERROR`, `WARNING`, `HINT`).  
- `ERROR`s are things currently reported that are contrary to the spec.
- `WARNING`s are things which, while valid against the spec will be problematic later, eg:
     - recursive imports (not currently validated)
     - empty MathML blocks in the resets for reset_value and test_value (valid, but meaningless)
     - mismatched variable units inside the MathML (not currently validated)
     - etc ...
- `HINT`s are things which are valid against the spec and may be intentional, eg:
     - mismatch prefix/multiplier in units between equivalent variables (metres in one variable, micrometres in the other)
     - imports from the same source more than once (may be intended, may be not)
     - resets which get stuck at a floor/ceiling condition
     - etc ... 

In terms of implementation, how about:
- by default, calls without a level argument should retrieve `ERROR`s
- overloading the current `Validator::errorCount()` and `Validator::error(int)` functions to include a Issue::Level argument:
```
errorCount() = issueCount(Issue::Level::ERROR) = issueCount() (default is ERROR)
warningCount() = issueCount(Issue::Level::WARNING) 
hintCount() = issueCount(Issue::Level::HINT) 

error(i) = issue(Issue::Level::ERROR, i) // not sure about having tuple argument or chaining it?
warning(i) = issue(Issue::Level::WARNING, i)
hint(i) = issue(Issue::Level::HINT, i)

```

It would also be lovely to be able to put section references on the warnings and hints (the way we do with the specification headings, see below), but with reference to a more human-readable document when/if one exists?
  
```
static const std::map<SpecificationRule, const std::string> ruleToHeading = {
    {SpecificationRule::UNDEFINED, ""},
    {SpecificationRule::DATA_REPR_IDENTIFIER_UNICODE, "3.1.1"},
    {SpecificationRule::DATA_REPR_IDENTIFIER_LATIN_ALPHANUM, "3.1.2"},
    {SpecificationRule::DATA_REPR_IDENTIFIER_AT_LEAST_ONE_ALPHANUM, "3.1.3"},
```
Thoughts? @agarny @hsorby @nickerso @MichaelClerx

agarny:
Thanks @kerimoyle, overall it looks good to me. Just a few comments/FWIW's:
- I would still prefer `Issue::Type` rather than `Issue::Level`, but maybe that's because I use that terminoloy in OpenCOR;
- I wouldn't and don't want to have `issueCount() = errorCount()`. For me, `issueCount()` should be equal to `errorCount()+warningCount()+hintCount()`. This also means that issue we shouldn't have methods with an optional issue type/level. If we have an issue type/level argument, then the end user should have to make it clear which type/level of issue s/he is interested in.
- In the same way that you are suggesting to have `error(i)`, `warning(i)` and `hint(i)`, I would also have a method to retrieve the ith issue, independent of its type/level. Indeed, this is what I will want and need in OpenCOR, i.e. report all the issues independent of their type/level.
- +1 with the section references.

kerimoyle:
> * I would still prefer `Issue::Type` rather than `Issue::Level`, but maybe that's because I use that terminoloy in OpenCOR;

I'm happy with either as long as the combination is clear.  To me, `Issue::Type` and `Issue::Kind` suggest the same thing - the name gives me no clue as to what they are talking about.  If we have `Issue::Type` instead of `Issue::Level` then I'd want to change `Issue::Kind` into something else so that the combination of the two is not so opaque.  We could even be blunt and use `Issue::Cause` instead of kind and `Issue::Effect` instead of level :)

> * I wouldn't and don't want to have `issueCount() = errorCount()`. For me, `issueCount()` should be equal to `errorCount()+warningCount()+hintCount()`. 

That's fine.

> This also means that issue we shouldn't have methods with an optional issue type/level. If we have an issue type/level argument, then the end user should have to make it clear which type/level of issue s/he is interested in.

That's fine too (though seems to go against what you say in the next paragraph where you say you want them independent of levels?)

> * In the same way that you are suggesting to have `error(i)`, `warning(i)` and `hint(i)`, I would also have a method to retrieve the ith issue, independent of its type/level. Indeed, this is what I will want and need in OpenCOR, i.e. report all the issues independent of their type/level.

Users will always have the option to concatenate the types of issues and report them all, or we could do it for them I guess.  But ... what does the index refer to? 
- how does the user want to retrieve them: 
     - show me all issues with a warning level or higher in the order in which they occur?  or, 
     - show me all errors and then also show me all warnings after that?
Or both?  If chronology between the levels is important (as it might be if you're debugging something and want all the issues associated with a single aspect of your model at once) then there's obviously a different implementation needed than if we just want to return all the issues in a concatenated list.

> * +1 with the section references.

Someone has to write the sections first ;) @MichaelClerx, would that be your user-friendly docs?

agarny:
> I'm happy with either as long as the combination is clear. To me, `Issue::Type` and `Issue::Kind` suggest the same thing - the name gives me no clue as to what they are talking about. If we have `Issue::Type` instead of `Issue::Level` then I'd want to change `Issue::Kind` into something else so that the combination of the two is not so opaque. We could even be blunt and use `Issue::Cause` instead of kind and `Issue::Effect` instead of level :)

I woudl simply go with `Issue::Type` instead of level and `Issue::About` instead of kind. At least, we now have several options to choose from!

> > This also means that issue we shouldn't have methods with an optional issue type/level. If we have an issue type/level argument, then the end user should have to make it clear which type/level of issue s/he is interested in.
> 
> That's fine too (though seems to go against what you say in the next paragraph where you say you want them independent of levels?)

I don't think so. Two cases:
1. If we think a method should have an issue type/level argument then I wouldn't want that argument to be optional.
2. If we have no issue type/level argument then I would expect all types/levels of issues to be reported, not only error ones.

> > * In the same way that you are suggesting to have `error(i)`, `warning(i)` and `hint(i)`, I would also have a method to retrieve the ith issue, independent of its type/level. Indeed, this is what I will want and need in OpenCOR, i.e. report all the issues independent of their type/level.
> 
> Users will always have the option to concatenate the types of issues and report them all, or we could do it for them I guess. But ... what does the index refer to?

?? For `error(i)`, `i` means the ith error. Same for `warning(i)` and `hint(i)`. For `issue(i)`, it's the ith issue, independent of its type.

For OpenCOR, say that that we have the following issues (line number | type of issue):

- Line 3 | Error
- Line 3 | Hint
- Line 4 | Warning
- Line 5 | Error
- Line 7 | Warning
- Line 9 | Hint

I would then want those issues to be listed in the above order while if I was to concatenate the types of issues, e.g. errors, then warnings and finally hints, I would end up with:

- Line 3 | Error
- Line 5 | Error
- Line 4 | Warning
- Line 7 | Warning
- Line 3 | Hint
- Line 9 | Hint

which, from a user viewpoint, is not at all ideal (especially if you have tens of issues).

Otherwise, a more generic approach might be to have `issue(i[, type])` with `type` being a concatenation of the different types of issues in which we are interested (and where chronological order is kept) and `i` the index of an issue within the subset of issues of interest. `type` would be optional and would default to `ERROR+WARNING+HINT`. This means that we could easily expand the range of types in the future, if ever needed.

In the same way, we would have `issueCount([type])`. So, no need for `error(i)`, `warning(i)`, `hint(i)`, and `xxxCount(i)`.

kerimoyle:
> Otherwise, a more generic approach might be to have issue(i[, type]) with type being a concatenation of the different types of issues in which we are interested (and where chronological order is kept) and i the index of an issue within the subset of issues of interest. type would be optional and would default to ERROR+WARNING+HINT. This means that we could easily expand the range of types in the future, if ever needed.

That's a much cleaner way to do it. I like :)  It was just hard to figure out how users would want it!  I'm working on my other job for a few days now so will wait until others have said their 2c then write it up next week.  Any other thoughts @nickerso @hsorby @MichaelClerx ?

kerimoyle:
I get a python error I don't understand (see below).  It says I'm missing docstrings for the `libcellml.Logger.error()` function which no longer exists (changed to `issue()` instead)?  How do I stop SWIG building the `error` files (instead or as well as the new `issue` files)?

```
Start 34: python_test_docstrings

34: Test command: /usr/local/Frameworks/Python.framework/Versions/3.7/bin/python3.7 "test_docstrings.py"
34: Environment variables: 
34:  PYTHONPATH=/Users/libcellml/BuildSlave/Unit_Tests_Builder_Mac_OSX_10_14/libcellml-build/src/bindings/python
34: Test timeout computed to be: 10000000
34: E
34: ======================================================================
34: ERROR: test_docstrings (__main__.DocstringTestCase)
34: ----------------------------------------------------------------------
34: Traceback (most recent call last):
34:   File "test_docstrings.py", line 32, in test_docstrings
34:     + ', '.join(missing))
34: Exception: Missing (1) docstrings, for: libcellml.Logger.error <<<<<<<<<<<<<<<<<< ??
34: 
34: ----------------------------------------------------------------------
34: Ran 1 test in 0.031s
34: 
34: FAILED (errors=1)
34/34 Test #34: python_test_docstrings ...........***Failed    0.11 sec
```

hsorby:
It is because you still have the following in the logger interface file:

```
#if defined(SWIGPYTHON)
    // Treat negative size_t as invalid index (instead of unknown method) TODO Check why this didn't work?
    %extend libcellml::Logger {
        IssuePtr error(long index) {
            if (index < 0) return nullptr;
            return $self->issue(size_t(index));
        }
    }
#endif
```

agarny:
@hsorby just beat me to it. Must be because he's working from home... :)

hsorby:
Just warming up for the commute on my work vehicle.

kerimoyle:
Thanks both!  Solves two things at once :)

kerimoyle:
It seems to be working fine, but I'm not entirely happy with the implementation of this function:

```
IssuePtr Logger::issue(std::vector<Issue::Type> &types, size_t index) const
{
    // TODO This seems like a super inefficient way to do it ... a lot of manipulation to retrieve one item :(
    IssuePtr err = nullptr;
    if (index < issueCount(types)) {
        std::vector<int> get_issues(mPimpl->mIssues.size(), -1);
        for (auto type : types) {
            auto t = static_cast<size_t>(type);
            for (auto const &i : mPimpl->mTypeIndex.at(t)) {
                get_issues.at(i) = static_cast<int>(i);
            }
        }
        get_issues.erase(std::remove(get_issues.begin(), get_issues.end(), -1), get_issues.end());
        err = mPimpl->mIssues.at(static_cast<size_t>(get_issues.at(index)));
    }
    return err;
}
```
Getting one kind of issue at a time is very easy as their order is stored in the mTypeIndex array.  I could either also store the combination of types (eg ERROR+WARNING) as well so that it was directly accessible too (avoiding the manipulation as in the function above) but that seemed clumsy too.  Ideas welcome ... :)

I can't do much more which is useful on this branch until the others (#355, #361, #358) are merged in ... nudge nudge ...

agarny:
> Ideas welcome ... :)

Don't worry, I am not happy with that implementation either. :) However, I am not sure you should worry about it too much at this stage. If you can improve your implementation then great otherwise wait for the review process to kick in. (To comment on a small piece of code is not ideal. It's better to see in its general context, i.e. at review time.)

> I can't do much more which is useful on this branch until the others (#355, #361, #358) are merged in ... nudge nudge ...

I am kind of in the same boat. I mean that in PR #360, I have merged the code of PRs #322 and #344. Not great, but... I couldn't afford to wait...

kerimoyle:
Gah, merge issues happening here too :(

agarny:
Yes, it was bound to happen...

kerimoyle:
Replaced by #521

Issue #370, opened 2019-06-22
Upgrade gcovr


hsorby:
We are having some issues with gcovr and gcc-8.  This issue is to investigate if upgrading gcovr is an easy path to take.

PR #371, opened 2019-06-22
Upgrade gcovr

hsorby:
Trying out an upgraded gcovr in a zip file.  Addresses issue #370.


hsorby:
This solution works for me in fixing our gcovr problem.

agarny (approved):
I didn't know that Python can execute a `.zip` file. Nice! It certainly looks neater to have a `.zip` file than a bunch of Python files for gcovr 4.1.1.

Otherwise, it might be worth creating a CMake macro for gcovr. I mean, to have three very similar looking calls to gcovr is not "neat".

nickerso (approved):

PR #382, opened 2019-09-09
take#2 issue355 new reset validation

kerimoyle:
Something is weird on the previous branch for resets (#358) so have rewound this to a previous version to see if I can sort it out.


kerimoyle:
Still failing here, deleting this PR

PR #383, opened 2019-09-09
Resets again 9sept

kerimoyle:
I'll come back and explain it properly if this works ...


kerimoyle:
Woo! Passing here, will try and merge the same code into the other #358 and see if it works ...

Issue #399, opened 2019-09-25
addComponent is not consistent?


kerimoyle:
I'm really confused by what's intended here:

Set up a boring model encapsulation:
   ```
 libcellml::ModelPtr model = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr parent = std::make_shared<libcellml::Component>();
    parent->setName("parent_component");
    libcellml::ComponentPtr child1 = std::make_shared<libcellml::Component>();
    child1->setName("child1");
    libcellml::ComponentPtr child2 = std::make_shared<libcellml::Component>();
    child2->setName("child2");
    libcellml::ComponentPtr child3 = std::make_shared<libcellml::Component>();
    child3->setName("child3");
    parent->addComponent(child1);
    parent->addComponent(child2);
    parent->addComponent(child3);
    model->addComponent(parent);
```
Print it out ... 
```
    std::cout << "parent component has children: \n";
    for (size_t c = 0; c < parent->componentCount(); c++) {
        std::cout << " - " << parent->component(c)->name() << "\n";
    }
    std::cout << "Child3 has parent: " << child3->parentComponent()->name() << "\n";
```

Prints: ----------------
    parent component has children:
     - child1
     - child2
     - child3
     Child3 has parent: parent_component

Now add child3 to child2 and print again:

```
    child2->addComponent(child3);
    std::cout << "parent component has children: \n";
    for (size_t c = 0; c < parent->componentCount(); c++) {
        std::cout << " - " << parent->component(c)->name() << "\n";
    }
    std::cout << "Child3 has parent: " << child3->parentComponent()->name() << "\n"; 
```

Prints: ----------------
parent component has children: 
    - child1
    - child2 
    - **child3 << why is this still there?? shouldn't it no longer be on the child list??**
Child3 has parent: child2

Now see what happens when we add child2 back to the parent component:

```
    std::cout << parent->componentCount() << "\n"; // prints 3
    parent->addComponent(child2);
    std::cout << parent->componentCount() << "\n"; // prints 4 << why is this allowed ?? why not prevented in the same way as if it was in the parent list?
```

Print out the whole hierarchy now:

```
    for (size_t c = 0; c < model->componentCount(); c++) {
        libcellml::ComponentPtr comp = model->component(c);
        std::cout << comp->name() << "\n";
        for (size_t d = 0; d < comp->componentCount(); d++) {
            libcellml::ComponentPtr comp2 = comp->component(d);
[...]

PR #400, opened 2019-09-25
Fix multi parent component bug.

hsorby:
Addresses issue #399.


tests/component/component.cpp
```diff
@@ -499,3 +499,30 @@ TEST(Component, constructors)
     libcellml::ComponentPtr c3 = std::move(c2);
     EXPECT_EQ("my_name", c3->name());
 }
+
+TEST(Component, multiParentWithAddComponentBugIssue399)
```
nickerso: This test looks good to me and would give the behaviour I would expect.

hsorby: Sadly, it doesn't look good to me.  We should be able to add `child3` to `child2` and have `child3` as . a child of both `parent` and `child2`.

hsorby: Our difficulty will come if we ask `child3` who its parent is.

kerimoyle: Three options?
- allow more than one parent until validation, so the user gets the full picture
- update parent and child at the same time, preventing errors but also giving network dependent on the order in which items were added
- forbid changing the network if it will break like this, and alert the user

kerimoyle:
Whoops, posted this in the wrong place ... 
So if we're going to allow the component to be added to the "wrong" place, it will affect how we store the relationship between parents and children. Since at present we can only have one parent, this is how the network gets messed up ... I think we have three options:

1. Require children which are being added to have a NULL parent before they're added, and warn the user otherwise. ie addComponent(component) will return an error when component already has any parent.

- This will prevent some use cases but at least the error messages returned can be done at the appropriate time.
- It goes against allowing invalid model creation though.
- We can keep the same storage classes as we have at the moment.

2. Change the storage class to allow for more than one parent.

- this will allow invalid models to be created, errors only at validation
- the change will propagate through a whole lotta code ...

3. Make sure that the child list of the ex-parent is updated when addComponent is called.

- will get around the bug we currently have, but ...
- does not address the issue of the user ending up with a network which depends on the order in which it was created.
- validation errors may not be able to pinpoint where the error originated from ... only what it is.

All in all, I'd probably vote for 1 as the least bad option ...?

hsorby:
I can only see option 2 as our way forward.   We need to be able to add the same component in multiple places.  We will not have a 1:1 relationship when traversing a component tree.  We will have to document that component traversal should be done from the model down if it is to be q unique traversal.

kerimoyle:
>> I can only see option 2 as our way forward. 

This means that functions such as parentComponent() will now need an index? 

>> We need to be able to add the same component in multiple places

Is that from a modelling point of view (in which case I think it breaks the encapsulation spec) or just from the philosophy of allowing people to do what they like and only checking at validation time?

nickerso:
Now I'm confused. By definition, CellML components can only have one parent. Why would we ever want to support them having more than one parent? Option 3 is clearly the way to go for me.

@hsorby, why do we need to support adding the same component in more than one place? It is not something we should ever support, in the same way that we don't support a variable being added to a units... If we're parsing a CellML XML document that happens to have an encapsulation hierarchy that is broken, we can either report that broken hierarchy and fail to parse or simply make a new instance of the component and flag that as an error in parsing but still give a "usable" model.

I'm starting to think I'm missing something here...

kerimoyle:
I don't think it's because we actually support them having more than one parent, it's just that the way we currently store the parent/child information doesn't really help the user in trying to figure out what's wrong with their model.  

Say a user adds a component to more than one parent. If we did actually store both, we'd be able to tell the user exactly what was wrong.  At the moment, we just switch over *some* of that relationship, depending on the order in which the components were added and where.  This is harder to pick apart and find the real cause later.

I think either we prevent it (and only add components which have a null parent) or we store all parents and fail the validation.

hsorby:
We want to re-use components, like an m or h gate in multiple places.  In this case we only have one component that is added in multiple places a change to this component is then made to all of those components.

agarny:
> We want to re-use **components**, like an m or h gate in multiple places. In this case we only have one component that is added in multiple places a change to this component is then made to all of those components.

Did you really mean **components** or variables?

Anyway, I would expect a variable to be associated with only component and that is it. In other words, if we have a variable `V` and two components `C1` and `C2`, and we do `C1->addVariable(V)` and then `C2->addVariable(V)`, then I would expect `V->parentComponent()` to return `C2`, and that is it, unless I am also missing something...?

nickerso:
> We want to re-use components, like an m or h gate in multiple places. In this case we only have one component that is added in multiple places a change to this component is then made to all of those components.

It is not possible to reuse a component like this in CellML. You need to import a component multiple times and therefore you have multiple components that are each used once. If we are really talking about the library automatically building a model import hierarchy based on the desired "model" - then that is a separate issue.

hsorby:
This is not really about the CellML that is the underlying exchange format, we are talking about the way libCellML is working in C++.  Imports are different they are done as you have explained.

nickerso:
Reuse, in CellML, is supported through the use of importing existing components as new instances. Each reuse requires a new instance. This has nothing to do with the XML serialisation, and everything to do with CellML - which is what libCellML is trying to support, right?

But your mention this being a C++ issue is really making me think I'm missing the point of this whole discussion :)

hsorby:
After having another think about this (and reading the documentation):
```
* Add a copy of the given component as a child component of this component entity.
```
I think we should do as the documentation says.  Further the `addComponent` method should be modified to return the added component.  If the added component does not yet have a parent then simply add that component and return it otherwise make of copy of the component add that copied component and return that.  Then improve the documentation detail exactly that.

hsorby:
So after discussion with @agarny and @nickerso the consensus is that @kerimoyle's option 3 is a good way forward.  However, the documentation on `addComponent` will have to be changed!

Additionally  we could also look at adding an `append*` family of methods that have a different behaviour.  The behaviour of `append*` methods would take a copy of the object being appended so multiple appendages would result in multiple new entities.  Any changes to the entity held would not update any of the entities already added.

nickerso:
I talked with @hsorby and I now understand what he was getting at, and just for the record the use case he was getting at is code like the following:

```
libcellml::ComponentPtr parent = std::make_shared<libcellml::Component>();
libcellml::ComponentPtr child = std::make_shared<libcellml::Component>();
child->setName("child1");

// do some complex and interest stuff to populate child with lots of variables, math, etc.

parent->addChild(child);

child->getVariableByName("interesting_rate_constant")->setInitialValue(30);
parent.addChild(child);

child->getVariableByName("interesting_rate_constant")->setInitialValue(60);
parent.addChild(child);
```

which is where the proposed `append*` methods would be able to kick in and enable this style of code.

kerimoyle:
```
libcellml::ComponentPtr parent = std::make_shared<libcellml::Component>();
libcellml::ComponentPtr child = std::make_shared<libcellml::Component>();
child->setName("child1");

// do some complex and interest stuff to populate child with lots of variables, math, etc.

parent->addChild(child);

child->getVariableByName("interesting_rate_constant")->setInitialValue(30);
parent.addChild(child);

child->getVariableByName("interesting_rate_constant")->setInitialValue(60);
parent.addChild(child);
```
^^ From @nickerso 's last comment. 

I can see some troubles here, partially with the naming, partially with the way the add functions have been used throughout the code.  It's not intuitive to expect that adding an item will create and return a new duplicate of that item, iff it has a parent, unless we also use that idiom to create all our child items.  What do youse fullas think about a "standard" practice becoming:

```
libcellml::ComponentPtr parent = std::make_shared<libcellml::Component>();

// only the declaration of type, and any minimal stuff.  Want it as bare as possible.
libcellml::ComponentPtr child=nullptr;

// child is not allocated, has no attributes, needs to be created.  
// the "addChild" function can be overloaded and thought of as addChildBasedOn(existing entity) or 
// allocate empty if no argument given
child = parent->addChild();  // returns allocated pointer to new creation

// do all the normal stuff ... 
child->setName("child1");

// do some complex and interest stuff to populate child with lots of variables, math, etc.

// Don't need to do this step any more, is already linked to parent
// parent->addChild(child);

child->getVariableByName("interesting_rate_constant")->setInitialValue(30);

// now, I presume, we want to duplicate our existing child, but because we're now used to the idea of addChild returning a value (having used it for all of our inititialisation stuff) we know to assign it to something
child2 = parent.addChild(child);

// this continues to reference the first child, its address has *not* changed
child->getVariableByName("interesting_rate_constant")->setInitialValue(60);

// we can reference the second child as you'd expect
child2->getVariableByName("interesting_rate_constant")->setInitialValue(990);

```

hsorby:
Yes it is another way of doing it but then you need the owning object to exist first, which is one of the reasons I don't like this approach (probably the main reason, I wanted to be able to create objects on their own).

Zinc has this and I find it annoying that I have to always start creating from the ground up.

If we were going to go this way I would call the initial method `createChild` most likely.

agarny:
I don't like the fact that `parentModel()` and `parentComponent()` have been replaced with `parent()`. We are back to having to cast things to know whether we are dealing with a parent model or a parent component, which is one of the things issue #320 aimed at avoiding.

Something that recent changes have done is to introduce some inconsistency in the way we handle header files:
 - Looking at `src/component.cpp`, I can see that we now include `libcellml/component.h` and then the other libCellML-related header files. This makes things unnecessarily "complicated". I would really stick to just having one block of libCellML header files as we used to have, not to mention that not all other `.cpp` files follow that "new" way of doing things.
 - `utilities.h` is not always included in the same place. Sometimes, it's included *before* the other libCellML-related header files (e.g. `src/utilities.cpp`) while sometimes it's included *after* (e.g. `src/component.cpp`).

hsorby:
> I don't like the fact that `parentModel()` and `parentComponent()` have been replaced with `parent()`. We are back to having to cast things to know whether we are dealing with a parent model or a parent component, which is one of the things issue #320 aimed at avoiding.

This is really the way it should be done.  We should be casting it to what we want and the Entity class shouldn't know anything about any of its derived classes.


> * Looking at `src/component.cpp`, I can see that we now include `libcellml/component.h` and then the other libCellML-related header files. This makes things unnecessarily "complicated". I would really stick to just having one block of libCellML header files as we used to have, not to mention that not all other `.cpp` files follow that "new" way of doing things.

This is definitely how it should be it's a mistake that the header file for the implementation file is not the first header file included.

> * `utilities.h` is not always included in the same place. Sometimes, it's included _before_ the other libCellML-related header files (e.g. `src/utilities.cpp`) while sometimes it's included _after_ (e.g. `src/component.cpp`).

`utilities.h` should be included after external API files but before system files.

agarny:
> > I don't like the fact that `parentModel()` and `parentComponent()` have been replaced with `parent()`. We are back to having to cast things to know whether we are dealing with a parent model or a parent component, which is one of the things issue #320 aimed at avoiding.
> 
> This is really the way it should be done. We should be casting it to what we want and the Entity class shouldn't know anything about any of its derived classes.

Agreed with the fact that the `Entity` class shouldn't know anything about its derived classes, but that's not the issue I am raising here. As an end user of the library, you are making my life more "difficult" by forcing to me to cast a parent so that I can tell whether it's a parent model or a parent component. The API should make my life easier, not more "difficult".

> > * Looking at `src/component.cpp`, I can see that we now include `libcellml/component.h` and then the other libCellML-related header files. This makes things unnecessarily "complicated". I would really stick to just having one block of libCellML header files as we used to have, not to mention that not all other `.cpp` files follow that "new" way of doing things.
> 
> This is definitely how it should be it's a mistake that the header file for the implementation file is not the first header file included.

Who said this is the way it should be done? I guess it comes from the [Google guidelines](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)? If so, then we are still doing things wrong. This aside, I don't like it (as said before: unnecessarily "complicated").

> > * `utilities.h` is not always included in the same place. Sometimes, it's included _before_ the other libCellML-related header files (e.g. `src/utilities.cpp`) while sometimes it's included _after_ (e.g. `src/component.cpp`).
> 
> `utilities.h` should be included after external API files but before system files.

That's not what the Google guidelines say. Bottom line: we are completely inconsistent at the moment.

hsorby:
You're correct that is not what the guidelines say, I thought we had some of it correct but apparently not.  

The first include has to be the corresponding header file this is to ensure that the header file includes everything needed to compile and not rely on the user including other header files first.  This is very important.  I don't know how we ever got away from that.

If we need to add functions we can add utility functions like `owningModel` to the API.  This doesn't mean that the `Entity` class should have this capability.

agarny:
> The first include has to be the corresponding header file this is to ensure that the header file includes everything needed to compile and not rely on the user including other header files first. This is very important. I don't know how we ever got away from that.

Clearly, the corresponding header file should be included, but I must confess that I don't see why it should be the first one...!?

> If we need to add functions we can add utility functions like `owningModel` to the API. This doesn't mean that the `Entity` class should have this capability.

Oops, I had completely forgotten that `parentModel()` and `parentComponent()` were part of the `Entity` class. Ok, to replace them with `parent()` makes sense then. However, this goes back to what I mentioned at some point, which is that our current object model is wrong.

Indeed, right now we have:

```
Entity
 |
 +--- OrderedEntity
 |     |
 |     +--- Reset
 |     |
 |     +--- When
 |
 +--- NamedEntity
 |     |
 |     +--- ComponentEntity
 |     |     |
 |     |     +--- Component
 |     |     |
 |     |     +--- Model
 |     |
 |     +--- Units
 |     |
 |     +--- Variable
 |
 +--- ImportSource
```

So, `Reset`, `When`, `Component`, `Model`, `Units`, `Variable` and `ImportSource` can have a parent entity, which in principle could be of any of the aforementioned class types. Thus, an `ImportSource` could have a `Variable` as a parent, or a `Model` a `Component` as a parent. So, sure, we want to allow the creation of "invalid" models, but... maybe our API should prevent nonsensical models where possible?

At the end of the day, I think `Component` should either have a parent model or a parent component. When it comes to `Reset` and `When`, I guess they should have a parent component, as should `Variable`. However, when it comes to `Units` and `ImportSource`, they should have a parent model. Finally, `Model`, shouldn't have any parent?

So, yes, I find `Entity` to be rather confusing...

kerimoyle:
I've just had a quick look through the code because I wanted to know how the inherited parent pointer was actually used throughout the different entity types.  Turns out we mostly* use it when referring to components and their encapsulations (there is one function, Units::clearParent which uses it too.) Of course I don't know how people outside the library might use these functions, but if they're not being set by us at parse time or used later on then maybe they're not really needed?

My point is this - if the usefulness of storing parents to all of our entity types is not consistent, then there's no reason to argue for a consistent parent implementation (by which I mean inheriting a parent pointer and then having to cast it).  I think it makes a whole lot more sense to add in the parents where they're needed (ie: the components) and be clear and clean about what we're doing.  In other words, do not have an inherited parent component, just add it to the concrete implementations in whatever way makes the most sense to those items. 

From what I've needed while coding in this project, this structure has the most use for me:
component -> parentModel *and* component -> parentComponent as they're used for different things!
units -> parentModel 

Thoughts?

hsorby:
I think `variable`s will need to also have some notion of a parent which would require them to have `parentComponent`.  I'm not keen on this and I would like to see some of this detail moved inside to  private headers where users wouldn't be able to manage some of the aspects of this parenting question negating some of this argument. 

I think I would go for a more `ParentableEntity` class approach and then have the `parent` method return the parent of the Entity in a fuller sense.  Maybe use `ModelParentableEntity` and `ComponentParentableEntity` to be specific.

hsorby:
All of this though I think is worth putting down into another issue and subsequent pull request as it seems to me not something to block this pull request on. 

As far as this pull request is concerned is there anything about it that cannot be merged into the codebase given that the functionality hasn't changed other than what this pull request is addressing.

nickerso (dismissed):
This appears to address #399 and I agree with the new test results.

agarny:
@hsorby: PR #405 has been merged in, so need to resolve the conflicts.

hsorby:
Done.

tests/component/component.cpp
```diff
@@ -324,11 +318,11 @@ TEST(Component, componentMethods)
 
     // Modify a deeper Component
     c->setName("parent");
-    c->addComponent(c1);
+    //c->addComponent(c1);
```
agarny: Shouldn't this be removed?

tests/component/component.cpp
```diff
@@ -499,3 +493,30 @@ TEST(Component, constructors)
     libcellml::ComponentPtr c3 = std::move(c2);
     EXPECT_EQ("my_name", c3->name());
 }
+
+TEST(Component, multiParentWithAddComponentBugIssue399)
+{
+    // Addressing Issue 399.
```
agarny: Should remove this comment?

tests/component/encapsulation.cpp
```diff
@@ -282,7 +283,14 @@ TEST(Encapsulation, parseAlternateFormHierarchy)
     libcellml::ModelPtr model = parser.parseModel(input);
 
     EXPECT_EQ(size_t(0), parser.errorCount());
+    printErrors(parser);
```
agarny: What's the point of printing the errors if we don't check what is being printed?
```suggestion
    printErrors(parser);
```
```suggestion
    printErrors(parser);
```

tests/parser/file_parser.cpp
```diff
@@ -31,6 +31,8 @@ TEST(Parser, parseSineModelFromFile)
     p.parseModel(fileContents("sine_approximations.xml"));
 
     EXPECT_EQ(size_t(0), p.errorCount());
+
+    printErrors(p);
```
agarny: What's the point of printing the errors if we don't check what is being printed?

tests/parser/parser.cpp
```diff
@@ -653,25 +653,32 @@ TEST(Parser, encapsulationWithNoComponentAttribute)
 
 TEST(Parser, encapsulationWithNoComponentRef)
 {
+    const std::vector<std::string> expectedErrors = {
+        "Encapsulation in model 'model_name' has an invalid child element 'component_free'.",
+    };
+
```
agarny, line 738: This ought to be replaced with:

```
        "Encapsulation in model 'model_name' has an invalid child element 'component_free'."};
```

which is what we do elsewhere in the code.

tests/parser/parser.cpp
```diff
@@ -653,25 +653,32 @@ TEST(Parser, encapsulationWithNoComponentAttribute)
...
+    expectEqualErrors(expectedErrors, p);
 }
 
 TEST(Parser, encapsulationWithNoComponent)
 {
+    const std::vector<std::string> expectedErrors = {
+        "Encapsulation in model 'model_name' specifies 'bob' as a component in a component_ref but it does not exist in the model.",
+        "Encapsulation in model 'model_name' does not have a valid component attribute in a component_ref element.",
+        "Encapsulation in model 'model_name' specifies an invalid parent component_ref that also does not have any children.",
+    };
```
agarny, line 759: Replace with:

```
        "Encapsulation in model 'model_name' specifies 'bob' as a component in a component_ref but it does not exist in the model.",
        "Encapsulation in model 'model_name' does not have a valid component attribute in a component_ref element.",
        "Encapsulation in model 'model_name' specifies an invalid parent component_ref that also does not have any children."};
```

tests/test_utils.cpp
```diff
@@ -52,6 +54,14 @@ void printErrors(const libcellml::Parser &p)
     }
 }
 
+void expectEqualErrors(const std::vector<std::string> &errors, const libcellml::Logger &logger)
+{
+    EXPECT_EQ(errors.size(), logger.errorCount());
+    for (size_t i = 0; i < logger.errorCount() && i < errors.size(); ++i) {
+        EXPECT_EQ(errors.at(i), logger.error(i)->description());
+    }
+}
```
agarny, line 63: This is a useful test utility, but we should then make use of it everywhere we check for expected errors. For example, `tests/validator/validator.cpp` does something similar (e.g. `invalidCellMLIdentifiersWithSpecificationHeading` although this particular test also checks for the specification heading, something that we might want to add).

agarny (commented):
Looks good, just some minor suggestions.

tests/parser/parser.cpp
hsorby: We have a mix of this stuff.  I prefer the what is there that is why I have done it this way and clang-tidy is happy with this.

tests/test_utils.cpp
hsorby: It is a useful utility although there is a downside.  When an error is reported by gtest it gives the line location of the utility file and not from the line which is inside the test.  Small downside I think though.

tests/parser/parser.cpp
agarny: I don't like trailing commas, but I am not happy either with the way it gets formatted by ClangFormat (when there is no trailing comma). So, I am happy with either in the end. Just want consistency. :)

tests/test_utils.cpp
agarny: Hmm, I hadn't thought of that, but as long as we can trace back the failing test, then I am ok with that small downside.

hsorby: Yes we have the name of the test which is failing, you just can't use the line number to make the jump.

We could add a default parameter to this function to also check other pieces of information.

agarny: > Yes we have the name of the test which is failing, you just can't use the line number to make the jump.

Ok, that's better than nothing although if that utility is used many times in a given test then it might be tricky to determine what is wrong. Hmm, although handy, I am wondering whether it's such a good idea to rely on that utility. What about a macro? I don't like the idea of macros, but... it could work here...?

> We could add a default parameter to this function to also check other pieces of information.

If we end up using that utility, in some way or another, then yes that would be good to have.

tests/component/encapsulation.cpp
hsorby: That was a debug statement that I didn't remove, removing it now.

tests/component/component.cpp
hsorby: Will do

tests/parser/parser.cpp
hsorby: I'm getting used to them, as I prefer them over the alternative.

tests/test_utils.cpp
hsorby: We could use a macro and it would make sense but I think I would have to kill myself.

hsorby: Not that I have strong feelings about macros or anything.

tests/parser/parser.cpp
agarny: Fine by me. Will just need to be consistent (added a bullet point to that respect to issue #402).

tests/test_utils.cpp
agarny: Ok, I leave it to you to convert it to a macro... and try not to kill yourself as a result of it. We still need you for a few things...

src/component.cpp
```diff
@@ -112,10 +114,18 @@ void Component::swap(Component &rhs)
 
 void Component::doAddComponent(const ComponentPtr &component)
```
kerimoyle: This is where the bool return is going?

src/model.cpp
```diff
@@ -99,11 +101,12 @@ void Model::swap(Model &rhs)
 
 void Model::doAddComponent(const ComponentPtr &component)
 {
```
kerimoyle, line 103: ... and here also

src/utilities.h
```diff
@@ -346,6 +346,19 @@ std::string getEntityName(const EntityPtr &entity);
...
-ModelPtr parentModel(const EntityPtr &entity);
+ModelPtr owningModel(const EntityPtr &entity);
+
+/**
+ * @brief Remove the given component for the given entity.
+ *
+ * The entity is expected to be derived from @c ComponentEntity and the
+ * component is expected to be a direct child of that entity.  The component
+ * will only be removed from the entity itself the component hierarchy of the
+ * entity will not be searched for the location of the component.
```
kerimoyle: ... I'm not sure this makes sense? "The component will only be removed from the entity itself the component hierarchy of the entity will not be searched for the location of the component" ... and even if you add an "if" into the middle, I still don't really follow it :|

tests/component/component.cpp
```diff
@@ -499,3 +493,29 @@ TEST(Component, constructors)
...
+
+    parent->addComponent(child1);
+    parent->addComponent(child2);
+    parent->addComponent(child3);
+    model->addComponent(parent);
+
+    EXPECT_EQ(size_t(3), parent->componentCount());
+    EXPECT_EQ(size_t(0), child2->componentCount());
+
+    child2->addComponent(child3);
```
kerimoyle, line 527: Do we test anywhere what happens if a component is added as a child of itself?
`c->addComponent(c); ` ?

tests/component/encapsulation.cpp
```diff
@@ -194,20 +180,35 @@ TEST(Encapsulation, hierarchyCircular)
...
+    // three steps.
     child1->addComponent(child2);
     a_parent = printer.printModel(model);
     EXPECT_EQ(e_parent_2, a_parent);
 
+    // Try to make a circular hierarchy but we will not succeed as this is not
+    // allowed.  The model will stay as it is.
     child2->addComponent(parent);
+    EXPECT_FALSE(parent->hasAncestor(child2));
     a_parent = printer.printModel(model);
```
kerimoyle, line 212: Are we happy what happens when parsing a model with an invalid encapsulation structure cf creating it through the API?  This change will not allow it to be created, which is different to how we're dealing with all the other kinds of errors

src/component.cpp
hsorby: And also in `addComponent`.

tests/component/component.cpp
hsorby: Not sure easy enough to add that test, I would expect that to return `false`.

tests/component/encapsulation.cpp
hsorby: No it won't but it can be logged.

tests/test_utils.h
```diff
@@ -28,3 +28,5 @@ void expectEqualErrors(const std::vector<std::string> &errors, const libcellml::
 
 libcellml::ModelPtr createModel(const std::string &name = "");
 libcellml::ModelPtr createModelWithComponent(const std::string &name = "");
+
+#define EXPECT_EQ_ERRORS(errors, logger) {SCOPED_TRACE("Error occured here");expectEqualErrors(errors, logger);}
```
agarny: Nice one, I like it! :) Might just want to add a full stop in the trace message?

tests/component/encapsulation.cpp
kerimoyle: OK, good :)

tests/component/component.cpp
kerimoyle: Thanks

kerimoyle (commented):
It's probably worth having a look through the validator changes in #350 as the code there changed how connections and components were looked at ... that will all have to be redone now :|

hsorby:
This pull request is ready for another review, I believe I have addressed all comments now.

src/api/libcellml/componententity.h
```diff
@@ -311,8 +313,10 @@ class LIBCELLML_EXPORT ComponentEntity: public NamedEntity
      * implement their own versions.
      *
      * @param component The ComponentPtr to add to the list of components.
+     *
+     * @return True if the component was added, false otherwise.
      */
-    virtual void doAddComponent(const ComponentPtr &component);
+    virtual bool doAddComponent(const ComponentPtr &component);
```
agarny, line 319: I can't recall when `doAddComponent ()` was added to `ComponentEntity`, but it has nothing to do here and should be moved to `ComponentEntity`'s private implementation. (The same holds true with `doAddComponent()` in `Model` and `Component`.)

hsorby: `doAddComponent` is the private virtual function for `addComponent` this is done to allow us to add API without breaking the ABI.  So no these methods cannot go into the private implementation class (until there is a whole private inheritance structure in place).

agarny: Cool, one more reason for doing things "right" when it comes to our private inheritance structure. :)

src/utilities.cpp
```diff
@@ -420,4 +420,10 @@ ModelPtr parentModel(const EntityPtr &entity)
     return model;
 }
 
+void removeComponentFromEntity(const EntityPtr &entity, const ComponentPtr &component)
+{
+    auto componentEntity = std::dynamic_pointer_cast<ComponentEntity>(entity);
+    componentEntity->removeComponent(component, false);
```
nickerso, line 426: Do we not need to worry about the cast failing because this is internal code so we know it will only be called with a `ComponentEntity` entity?

nickerso (dismissed):

src/utilities.cpp
hsorby: That's the idea, but if you can come up with a test that shows the opposite then please do.

kerimoyle:
Hi all - is it possible to split this PR into something which only addresses the bug in the parents first?  That would make it a whole lot easier to deal with - as is it will overwrite a lot of the validation stuff in other branches, that I don't think this branch needs to touch at all?  Or ... what am I missing here?

hsorby:
> Hi all - is it possible to split this PR into something which only addresses the bug in the parents first? That would make it a whole lot easier to deal with - as is it will overwrite a lot of the validation stuff in other branches, that I don't think this branch needs to touch at all? Or ... what am I missing here?

It is possible but in this instance I don't think it is worth the effort.  The changes to the `validator` are mainly around pulling out bits into smaller chunks so this will be easy to merge with other changes.

nickerso (approved):

src/utilities.h
agarny: I am resolving this conversation since this PR looks good to @nickerso and I, but @kerimoyle feel free to create an issue for this, if needed and still relevant.

agarny (approved):

PR #405, opened 2019-09-27
Address Entity parents.

hsorby:
This is how I think parents of entities should be handled.

Adresses #320.


kerimoyle:
This feels like going backwards in time ... we've only just changed into differentiating between parentModel and parentComponent types, which is actually really useful in validation.  It would be great to be able to use that same functionality (ie - two separate functions) to retrieve both the parent model (even if the component is a long way down an encapsulation chain) as well as the component's immediate parent component.  I don't see that functionality present here any more?  

I would love to have something like:

```
model->addComponent(c1)
c1->addComponent(c2)
c2->addComponent(c3)

c1->addVariable(v1)
c2->addVariable(v2)
c3->addVariable(v3)

v1->parentModel = model
v1->parentComponent = c1

c1->parentComponent = nullptr
c1->parentModel = model

c2->parentComponent = c1
c2->parentModel = model

... etc ...
```

hsorby:
If you look at the `utilities.h` header file I have added a function `parentModel` that retrieves the owning model of the entity if there is one.

kerimoyle:
Missed that, thanks :)  So we assume the type of object returned by `parent()` is always a component?  and parentModel() always a model?  Could we not just keep the parentComponent() name then, just for clarity?

hsorby:
Nope, `parent()` can return a model.

kerimoyle:
Yeah, that's not great ... then every time you call parent() you'd have to test for type before knowing how to handle it... that's what the changes in the earlier PR https://github.com/cellml/libcellml/pull/322 were written to fix?

agarny:
> Yeah, that's not great ... then every time you call parent() you'd have to test for type before knowing how to handle it... that's what the changes in the earlier PR #322 were written to fix?

Indeed, I don't like the idea of having to `static_cast` something to determine whether we are dealing with a parent component or a parent model. It's not clean.

hsorby:
This is kind of how C++ works with inheritance and so on, and anyway you wouldn't be doing a static cast it would have to be .a dynamic one :)

kerimoyle:
It is how C++ works with inheritance, but I don't think that having parents in the highest level entity make sense if they're not actually used anywhere except the components?  Why not just define a parentModel and parentComponent pointer in the components and remove the parent pointer from the entity?

hsorby:
`units` and `variables` have parents?

kerimoyle:
.. but where do we actually use them?

hsorby:
I think they will be used when we look at the `addVariable` and `removeVariable` again and realise that it is not working as desired.

nickerso (approved):
I like this approach. Entities should not need to know anything about derived types.

kerimoyle:
> Entities should not need to know anything about derived types.

True ... but overall I think that our use of inheritance here actually gets in the way of the functionality.  What's wrong with defining parents explicitly in the few cases we need them?  Of course it's possible to use inheritance to define the parent pointer, but as it makes other parts of the code a lot more clumsy, I don't see the advantage?

> I think they will be used when we look at the addVariable and removeVariable again and realise that it is not working as desired.

... could you be a bit more specific about that?  Sorry for all the questions, I just don't understand this PR.

agarny (approved):
I find that it makes the API more cumbersome to use, but... ok, let's see how it goes in the long run...

Issue #404, opened 2019-09-26
Bug in Component->removeVariable()


kerimoyle:
Calls to Component->removeVariable(variable) will remove the variable from the component's variable list, but do not update the parent of the variable to be nullptr.  See test below...

```
TEST(Validator, validateInvalidConnectionsParentlessVariable)
{
    const std::vector<std::string> expectedErrors = {
        "Variable 'variable1' has an equivalent variable 'variable2' which does not reciprocally have 'variable1' set as an equivalent variable.",
    };

    libcellml::Validator v;
    libcellml::ModelPtr m = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr comp1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp2 = std::make_shared<libcellml::Component>();

    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();

    m->setName("modelName");
    comp1->setName("component1");
    comp2->setName("component2");

    v1->setName("variable1");
    v2->setName("variable2");

    v1->setUnits("dimensionless");
    v2->setUnits("dimensionless");

    v1->setInterfaceType("public");
    v2->setInterfaceType("public");

    comp1->addVariable(v1);
    comp2->addVariable(v2);
    m->addComponent(comp1);
    m->addComponent(comp2);

    // Valid connections.
    libcellml::Variable::addEquivalence(v1, v2);

    // Make a variable without a parent component.
    comp2->removeVariable(v2);

    std::cout<<comp2->variableCount()<<"\n";   // <<< prints 0 as expected
    std::cout<<v2->parentComponent()->name()<<"\n";  // << prints component2 ... not expected!
}
```

hsorby:
The last two lines could be better written as:

```
EXPECT_EQ(size_t(0), comp2->variableCount()) // succeeds
EXPECT_EQ("", v2->parentComponent()->name()) // Fails, actual is "component2"
```

hsorby:
I expect this holds for all `remove*` methods (haven't actually checked but fairly confident).

kerimoyle:
Yep, don't worry, this was just me figuring out what was going on, not really a Test test... I'll have to check the other removes too ... thanks for the headsup :)

PR #406, opened 2019-09-27
Update parents of removed items

kerimoyle:
Addresses issue #404 

When an item is removed from a component, update the parent component of that item.


kerimoyle:
Review please @hsorby @nickerso @agarny

kerimoyle:
An update ... 
Because this was holding up the work on #350 I've included the changes here (it was only a few lines) in that branch already.  So don't merge it ... but any thoughts welcome :)

Issue #407, opened 2019-09-30
improve how equivalent variables are handled


kerimoyle:
After understanding better (I hope) how equivalent variables work, I think that the way we interact with them in the code is quite strange.  If they're a set of aliases to a single underlying variable, then we should store them as such.  This means we avoid all the complicated looping/not-looping checks/specs (which don't actually make sense if it's not a network), as well as getting around all the recursive calls to check for unit consistency and reset order validity.   Instead of doing the recursive checking later, I think we should rather update the equivalences as a set when they're created.

Here's my example ...

```
TEST(Validator, fixUpEquivVariables)
{
    libcellml::Validator v;
    libcellml::ModelPtr m = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr comp1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp2 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp3 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr comp4 = std::make_shared<libcellml::Component>();

    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v3 = std::make_shared<libcellml::Variable>();
    libcellml::VariablePtr v4 = std::make_shared<libcellml::Variable>();

    m->setName("modelName");
    comp1->setName("component1");
    comp2->setName("component2");
    comp3->setName("component3");
    comp4->setName("component4");

    v1->setName("variable1");
    v2->setName("variable2");
    v3->setName("variable3");
    v4->setName("variable4");

    v1->setUnits("dimensionless");
    v2->setUnits("dimensionless");
    v3->setUnits("dimensionless");
    v4->setUnits("dimensionless");

    v1->setInterfaceType("public");
    v2->setInterfaceType("public");
    v3->setInterfaceType("public");
    v4->setInterfaceType("public");

    comp1->addVariable(v1);
    comp2->addVariable(v2);
    comp3->addVariable(v3);
    comp4->addVariable(v4);

    // No encapsulation, all components can access each other
    m->addComponent(comp1);
    m->addComponent(comp2);
    m->addComponent(comp3);
    m->addComponent(comp3);

    // Making a set of equivalent variables v1, v2, v3, v4
    libcellml::Variable::addEquivalence(v1, v2);
[...]

PR #408, opened 2019-09-30
Addresses issue407 to improve equiv variables

kerimoyle:
This PR is to look at ways of improving the way we use equivalent variables, in the hope that the implementation can more closely match the spec.  
Addresses #407


kerimoyle:
A little progress, but I need to find out what/when/how the `removeEquivalence` function is used.  For example, this set of connections from the earlier example now passes:
```
// Making a set of equivalent variables v1, v2, v3, v4
    libcellml::Variable::addEquivalence(v3, v4);
    libcellml::Variable::addEquivalence(v4, v1);

    // In reality, all of these variables are equivalent to *three* others
    EXPECT_EQ(size_t(3), v1->equivalentVariableCount()); // passes
    EXPECT_EQ(size_t(3), v2->equivalentVariableCount()); 
    EXPECT_EQ(size_t(3), v3->equivalentVariableCount()); 
    EXPECT_EQ(size_t(3), v4->equivalentVariableCount()); 
```
... and printing the set of equivalentVariables for each of them gives the other three.  So far so good.  The trouble comes when we're no longer storing lists of connected pairs, but collapsing those pairs into sets.  So my question is, when you want to remove an equivalent variable connection, what would you want to end up with:
```
    libcellml::Variable::removeEquivalence(v2, v3);
```
Either:
- the set v1, v4, and then individual v2, and individual v3 (easy enough)
```
 EXPECT_EQ(v4, v1->equivalentVariable(0));
EXPECT_EQ(v1, v4->equivalentVariable(0));
EXPECT_EQ(size_t(0), v2->equivalentVariableCount());
EXPECT_EQ(size_t(0), v3->equivalentVariableCount());
```

- the set v1, v2, v4 and individual v3 (means defining a direction to the removal - first argument is treated differently than the second one), read the function as "remove arg2 from the set of arg1" or something.
```
 EXPECT_EQ(size_t(2), v1->equivalentVariableCount()); // v1 = v2, v4
 EXPECT_EQ(size_t(2), v2->equivalentVariableCount()); // v2 = v1, v4
 EXPECT_EQ(size_t(0), v3->equivalentVariableCount()); // v3 is orphaned
 EXPECT_EQ(size_t(2), v4->equivalentVariableCount()); // v4 = v1, v2
```
Thoughts?

ps - my implementation so far is only for figuring this out - I know the storage is awful, storing n^2 instead of n ...

Issue #386, opened 2019-09-23
Code generation with imports


nickerso:
Now that we have the basic code generation framework in place (#360) we need to extend this to support use with imports.

PR #410, opened 2019-10-02
Work toward supporting code generation for models with imports

nickerso:
Addressing #386.


src/api/libcellml/model.h
```diff
@@ -251,6 +251,16 @@ class LIBCELLML_EXPORT Model: public ComponentEntity, public std::enable_shared_
      */
     bool hasUnresolvedImports();
 
+	/**
+	 * @brief Flatten all imports in this model.
+	 *
+	 * Flatten this model to remove all imports. If any unresolved imports are found
```
hsorby: I think flatten should be changed to actually describe what is going on.  Flatten here has too much of a contextual meaning.

tests/generator/generator.cpp
```diff
@@ -393,6 +393,8 @@ TEST(Generator, algebraicEqnComputedVarOnRhsImport)
 	model->resolveImports(resourcePath("generator/imports/algebraic_eqn_computed_var_on_rhs/"));
 	EXPECT_FALSE(model->hasUnresolvedImports());
 
+	model->flatten();
```
agarny: To have a method to "flatten" a model is fine by me, but... when it comes to code generation, I would expect the generator to realise that the model contains some imports. Therefore, I would expect the generator to create a "flattened" version of that model before generating code for it.

nickerso: my current thinking is that flatten would be changing the current model - I would not expect code generation to change the model. Hence not calling the flatten within the code generation. Code generation should generate code for the model it is given (or fail), not make a new model and generate code for that.

An alternative is to have flatten create a new model which is the flattened version of the source model, but I still don't think that is something that the generator should be doing either. A flattened model can be used for things outside of generation, or at least the same flattened model could be used with multiple profiles...

hsorby: I think that this method should return the flattened model and not flatten the model that it is.

nickerso: Talking it over with @agarny and @hsorby, it is clear that the flatten method should create a new model. I still think the generator should be working on the model I give it, not making its own flattened model internally and generating code for that.

hsorby: I think so too, returning a flattened model from `model->flatten` would work in nicely with this.

agarny: +1 on the flatten method creating a new model. However, if we are to do this, then we might want to rename the method to something like `model->flattenedModel()`? `flatten()` on its own tells me that `model` is going to be flattened, which is not what we want.

As for the code generator working on the model it is given... I guess I could eventually get convinced that it's the way to go...

nickerso:
closing in favour of #483.

Issue #381, opened 2019-09-05
Validation: generate a warning if some units differ by some multiplicative factor


agarny:
Ideally, this would have been done as part of issue #361, but we need to be able to generate warnings for this while we can only generate errors at this stage. In other words, we need to address issue #314 first before addressing this issue.

kerimoyle:
This is already in progress  - along with the other things mentioned in the first statement of PR #314 :) . It did indeed used to be part of the larger validation PR, but as it's not actually required for validation, it was removed...

PR #412, opened 2019-10-03
Return multiplier in equivalent units (i381)

kerimoyle:
Adjusting the validator to return multiplier mismatch as well as base unit comparisons.  This is not a validation issue, but is returned only so that the generator can make use of it.  NB: A multiplier mismatch alone will _not_ trigger a validation error, but where there is a legitimate validation error caused by mismatched base units, the multiplier error is reported as well.  In all cases the multiplier is found by a call to `Validator::unitsAreEquivalent()`.

Addresses #381


agarny:
Is that supposed to be closed in the end?

kerimoyle:
Yes, I gave up here ... see 413?

agarny:
Ok.

Issue #409, opened 2019-10-02
Code generation for connections that use compatible, but not equivalent, units (i.e. need to be scaled)


agarny:
Now that the framework is in place for code generation with algebraic or ODE models (#360) we need to extend this to support models with connections that use equivalent units that need conversion (e.g. `mV` and `V`).

agarny:
Depends on issue #381.

agarny:
Depends on issue #545.

agarny:
Depends on issue #563.

agarny:
Depends on issue #571.

PR #413, opened 2019-10-03
Reporting multiplier from equivalent units

kerimoyle:
This PR addresses #381 in preparation for #409.  It has changed the argument list of `validator::unitsAreEquivalent` to return a multiplier such that:
 `v1->units() = v2->units*10^(multiplier)`

The validator will not report mismatched multipliers as an error if that is all that is wrong with the connection, but where there is a base unit mismatch the multiplier error is reported as well.

--------------------------------

**I HAVE NO CLUE WHAT IS GOING ON ... :( **

This is a really straightforward change: two files changed, about 50 lines total ... I have no clue why all the branches I've made say that there are 44 commits to put in etc etc ... if anyone can help me get my develop branch sorted that would be brilliant :(

In the meantime I've put the changes on a branch here - I know it's not how it's meant to be, but as I think I'll be nuking my repo now I wanted to get it out of the way.


kerimoyle:
Review please @hsorby @agarny @nickerso

kerimoyle:
After my initial commit I've done a little prep towards getting something useful for the `Generator` for @agarny.  

Some things to think about:
- I've moved a couple of the functions out of the `Validator` into the `utilities.cpp` file (`isStandardUnit` and `isStandardPrefix`) too, as they made more sense there I think.
- I'm not really a fan of the public function call signature - at the moment it needs the model as well as two units to be specified.  It would be cleaner to have a function like this:
     `double factor = u1->scalingFactor(u2);`
... but that means that the parent model has to be set correctly before calling it.  Perhaps moving this to be 'owned' by the model makes more sense?  All unit names will be interpreted in the context of the model anyway:
`double factor = model->scalingFactor(u1,u2); `??

- At present it returns the log10 of the difference between them, ie: 
`u1 = (10^factor)*u2`.  I'm not sure which is preferred, this or the actual number?

Thoughts on how to do that better welcome :)

kerimoyle:
Need to recheck the user-created unit multiplier checks ... not convinced they're correct at the moment.

agarny:
> * I've moved a couple of the functions out of the `Validator` into the `utilities.cpp` file (`isStandardUnit` and `isStandardPrefix`) too, as they made more sense there I think.

That seems reasonable to me.

> * I'm not really a fan of the public function call signature - at the moment it needs the model as well as two units to be specified.  It would be cleaner to have a function like this:
>   `double factor = u1->scalingFactor(u2);`
>   ... but that means that the parent model has to be set correctly before calling it.  Perhaps moving this to be 'owned' by the model makes more sense?  All unit names will be interpreted in the context of the model anyway:
>   `double factor = model->scalingFactor(u1,u2); `??

Without thinking about internal implications, I prefer `double factor = u1->scalingFactor(u2);`. In fact, when generating some code for a model, we know that that model is valid. If, for whatever reason, we were to make that call using an invalid model then I would probably expect to get `0.0` back.

> * At present it returns the log10 of the difference between them, ie:
>   `u1 = (10^factor)*u2`.  I'm not sure which is preferred, this or the actual number?

I am not sure I understand what you are saying here. `scalingFactor()` is to return a `double`, so...?

kerimoyle:
ScalingFactor returns factor, where factor is as in the equation?  Do you
want the log10 of the multiplier so its more like a prefix, or the
multiplier itself?  The model will always be needed, esp for user defined
base units.

agarny:
> ScalingFactor returns factor, where factor is as in the equation? Do you want the log10 of the multiplier so its more like a prefix, or the multiplier itself? The model will always be needed, esp for user defined base units.

I want the multiplier itself, so that it can be directly used for code generation.

I appreciate that a model will always be needed, but surely this can be retrieved from a units, so why need to pass it to `scalingFactor()`?

kerimoyle:
> In fact, when generating some code for a model, we know that that model is valid.

It would be nice if this were more generally useful than just for code generation though ...

> I want the multiplier itself, so that it can be directly used for code generation.

Sold :).

kerimoyle:
Does anyone think it would be useful to be able to create units independent of a model?  eg:

```
    libcellml::UnitsPtr apple = std::make_shared<libcellml::Units>();
    apple->setName("apple");

    libcellml::UnitsPtr dozen_apples = std::make_shared<libcellml::Units>();
    dozen_apples->setName("dozen_apples");
    dozen_apples->addUnit(apple, 0, 1, 12.0); // pass by pointer not name or StandardUnit?
```
... instead of the current ...
`    dozen_apples->addUnit("apple", 0, 1, 12.0); // pass by name and implied parentModel?`
    
Fairly simple overload of `addUnit`, I think, but means that units can be more portable between models?

kerimoyle:
Sooooo ... in order to overload the scalingFactor so that we can also have a nice function signature (eg `factor = u1->scalingFactor(u2)` ) we need to:

- have the parent of a units item updated when it's added to a model?  See #414
- which also means that a unit can only be added to one model at a time?
- remove the parent model from a unit when it's removed from the parent model?
- which means we need to `enable_make_shared_from` on the `Units` (perhaps?)

... thoughts?

agarny:
I don't know what it would mean to libCellML's implementation, but as a user of libCellML I would certainly expect to be able to do something like `factor = u1->scalingFactor(u2)`. (Sorry for not being more helpful... :))

Otherwise, I can see that our [master libCellML repository](https://github.com/cellml/libcellml/branches) has a [`i381_multiplier_report_equiv_units` branch](https://github.com/cellml/libcellml/tree/i381_multiplier_report_equiv_units), which clearly has nothing to do there...

src/units.cpp
```diff
@@ -157,24 +157,35 @@ double Units::UnitsImpl::doGetScalingFactor(const ModelPtr &model, const std::st
...
     } else if (isStandardUnitName(u2Name)) {
         updateBaseUnitCount(model, unitMap, multiplier, u2Name, 1, 0, -1);
+        found2 = true;
     }
 
-    return multiplier;
+    if (found1 && found2) {
+        return (std::pow(10, multiplier));
+    }
+    return (0.0);
```
agarny, line 188: Minor: no parentheses around what we return.

kerimoyle:
Otherwise, I can see that our master libCellML repository has a i381_multiplier_report_equiv_units branch, which clearly has nothing to do there...

Yep, I know @agarny ... but I can't reset my current repo until the outstanding branches are merged in.  So as soon as the two branches waiting are in #358 #350 (I think it's beyond a nudge, more like a shove now?? @hsorby @nickerso ), then I can update my base repo and get rid of this way of doing it ... 

Otherwise, this PR will need to wait until the units parents bug is fixed, see #414

kerimoyle:
Deleting this PR so I can move the branch elsewhere.

Issue #414, opened 2019-10-04
parentModel returns null for Units 


kerimoyle:
Now that we have the `parentModel(entity)` function I've been trying to use it in the unit multiplier checking, and I have two questions:
- is there a reason that we don't include the `utilities.cpp` file in the test coverage (I think?) I've wanted to use the `parentModel` function in a test, so I went looking to duplicate whatever manner it had been included in elsewhere to follow it ... but couldn't find it?
- Secondly, would we expect this test to pass or not?  or do we have to manually set the unit parent to model separately from including the unit in the model?

```
TEST(Units, parentsOfUnits)
{
    libcellml::ModelPtr model = std::make_shared<libcellml::Model>();
    model->setName("I_am_a_model");

    libcellml::UnitsPtr u = std::make_shared<libcellml::Units>();
    u->setName("u");

    model->addUnits(u);  // should this set the parent model of the unit??

    libcellml::ModelPtr parent = parentModel(u);
    EXPECT_FALSE(parent == nullptr);
}
```

hsorby:
I think you'll find that `utilities.cpp` does get covered:

`utilities.cpp                      89                 0   100.00%          26                 0   100.00%         322                 0   100.00%`

hsorby:
The utilities file is internal to the library so you won't see any tests directly testing those functions.

hsorby:
I would expect that it would be impossible to write that test.

kerimoyle:
The behaviour it's showing is still wrong, no?  When units are added to a model they don't take the model as their parent?  I should have written it like this, maybe:

```
TEST(Units, parentsOfUnits)
{
    libcellml::ModelPtr model = std::make_shared<libcellml::Model>();
    model->setName("I_am_a_model");

    libcellml::UnitsPtr u = std::make_shared<libcellml::Units>();
    u->setName("u");

    model->addUnits(u);  // should this set the parent model of the unit??

    libcellml::ModelPtr parent = u->parent();
    EXPECT_FALSE(parent == nullptr);
}
```

hsorby:
Yes that test would be the behaviour that I would expect (although again I don't think that test would compile).   If that test doesn't behave like that then yes it's definitely a bug.

kerimoyle:
You're right, had to cast to a `ModelPtr` to get it to compile, but it does indeed return `nullptr`, so it is a bug.

kerimoyle:
I've opened #415 so you can see ...

PR #415, opened 2019-10-07
i414 Bug in unit parents pointer

kerimoyle:
When a `Units` item is added to a `Model`, the parent pointer of the `Units` is not being set to the `Model`.

Addresses #414.


hsorby:
I have some further code to address this issue but was hoping not to add it until #400 had been dealt with as it relies on that work.

hsorby:
I like this test and it is just the sort of thing we need.

agarny:
@kerimoyle: try to avoid creating branches in the main repo... :)

kerimoyle:
Removing this and will replace with a clean branch from elsewhere ...

PR #418, opened 2019-10-08
I381 multiplier report equiv units

kerimoyle:
This branch and PR replaces #413 which branched from prime.


kerimoyle:
Ready for review now please @agarny @hsorby @nickerso

agarny:
Nope, not ready to review. Still a lot of commented out code in there and my [previous comment](https://github.com/cellml/libcellml/pull/413#pullrequestreview-297856637) hasn't been taken into account. Also, `scalingFactor()` requires a model to be passed, which is [not what I want or would even expect](https://github.com/cellml/libcellml/pull/413#issuecomment-538780959).

kerimoyle:
> Nope, not ready to review. 

Whoops, sorry ... I just got too excited that it was all untangled and passing tests that I forgot it wasn't finished.  Currently waiting for the units parents bug to be fixed in #417

src/api/libcellml/component.h
```diff
@@ -37,12 +37,19 @@ class LIBCELLML_EXPORT Component: public ComponentEntity, public ImportedEntity
                                   public std::enable_shared_from_this<Component>
 #endif
 {
-public:
+private:
```
agarny: FWIW, we are becoming increasingly inconsistent. Shouldn't we stick to having the `public`, `protected` and then `private` sections in that order?

src/api/libcellml/component.h
```diff
@@ -37,12 +37,19 @@ class LIBCELLML_EXPORT Component: public ComponentEntity, public ImportedEntity
...
     Component(); /**< Constructor */
+
+public:
     ~Component() override; /**< Destructor */
-    Component(const Component &rhs); /**< Copy constructor */
-    Component(Component &&rhs) noexcept; /**< Move constructor */
-    Component &operator=(Component rhs); /**< Assignment operator */
+    Component(const Component &rhs) = delete; /**< Copy constructor */
+    Component(Component &&rhs) noexcept = delete; /**< Move constructor */
+    Component &operator=(Component rhs) = delete; /**< Assignment operator */
```
agarny, line 44: Similarly (i.e. wrt consistency), it might be nice to have our constructors before the destructor.

src/api/libcellml/component.h
```diff
@@ -37,12 +37,19 @@ class LIBCELLML_EXPORT Component: public ComponentEntity, public ImportedEntity
...
+
+    template<typename... Args>
```
agarny, line 46: I haven't looked at the rationale for now using templates, but... I thought we wanted to avoid having to do that "at all costs"?...

hsorby:
Sooooo ... it turns out that when you start using `enable_shared_from_this` you have to change what you allow users to be able to do.  You can easily get into undefined behaviour using `enable_shared_from_this`.  One of the things I have done here is to make the constructor private thus forcing users to use the `create` static method (which it turns out can be used with variadic templates to enable argument passing).  Once this is done we no longer need the move semantics stuff as we are now solely dealing with shared pointers.  I had a little problem with the Python interfacing but think I have resolved that now.  This is still a work in progress as I look to see if this implementation works similarly across all entities.

hsorby:
This is ready for review, the next hot pull request to be considered.

src/api/libcellml/generator.h
```diff
@@ -49,6 +49,12 @@ class LIBCELLML_EXPORT GeneratorVariable
     GeneratorVariable(GeneratorVariable &&rhs) noexcept; /**< Move constructor */
     GeneratorVariable &operator=(GeneratorVariable rhs); /**< Assignment operator */
 
+    template<typename... Args>
+    static std::shared_ptr<GeneratorVariable> create(Args &&... args) noexcept
+    {
+        return std::shared_ptr<GeneratorVariable> {new GeneratorVariable {std::forward<Args>(args)...}};
+    }
+
```
agarny, line 57: Doesn't our best practice require the body of a function to be in a `.cpp` file and not a `.h` file? (Same elsewhere in this PR.)

src/api/libcellml/units.h
```diff
@@ -367,9 +376,19 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
+     * @brief Return the scaling factor difference between two units.
+     * 
+     * This can be interpreted as factor, where u1 = (factor)*u2.  This method
+     * does not check to see if the units are compatible.
+     * 
+     * @param units1 The first units to compare.
+     * @param units2 The second units to compare.
+     * @return The factor units1/unitsf2.
+     */
+    static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2);
```
agarny, line 386: It might be nice to also have a non-static version of `scalingFactor()`, i.e.

```
double scalingFactor(const UnitsPtr &nits);
```

to return the scaling factor difference between the current and the given units.

src/bindings/interface/importsource.i
```diff
@@ -2,6 +2,8 @@
 
 #define LIBCELLML_EXPORT
 
+%include "createconstructor.i"
+
```
agarny: Why the empty line? Also, why is `entity.i` included before `types.i`? Kind of the same for `model.i` below (and elsewhere too).

src/parser.cpp
```diff
@@ -314,15 +314,16 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
...
     while (childNode) {
         if (childNode->isCellmlElement("component")) {
-            ComponentPtr component = std::make_shared<Component>();
+            const std::string name;
+            ComponentPtr component = libcellml::Component::create(name);
             loadComponent(component, childNode);
             model->addComponent(component);
         } else if (childNode->isCellmlElement("units")) {
-            UnitsPtr units = std::make_shared<Units>();
+            UnitsPtr units = libcellml::Units::create();
```
agarny: We shouldn't need to specify the `libcellml` namespace. IOW, I would expect to see:

```
UnitsPtr units = Units::create();
```

Same is true elsewhere in this PR.

src/units.cpp
```diff
@@ -118,6 +121,17 @@ struct Units::UnitsImpl
...
+     * @brief Returns the log of the scaling factor between two similar units
+     * 
+     *  The factor returned can be interpreted as u1 = (factor)*u2
+     * 
+     * @param model The @c ModelPtr containing the units
+     * @param u1 The first @c UnitsPtr to compare
+     * @param u2 The second @c UnitsPtr to compare
+     * @return factor The scaling factor between u1 and u2
+     */
+    double doGetScalingFactor(const ModelPtr &model, const std::string &u1Name, const std::string &u2Name);
```
agarny: Am I missing something or is there no implementation for this method?...

src/units.cpp
```diff
@@ -126,43 +140,118 @@ std::vector<Unit>::iterator Units::UnitsImpl::findUnit(const std::string &refere
...
+    }
+
+    return updated;
 }
 
-Units::Units(const Units &rhs)
-    : NamedEntity(rhs)
-    , ImportedEntity(rhs)
-    , mPimpl(new UnitsImpl())
+double doGetScalingFactor(const UnitsPtr &units1, const UnitsPtr &units2)
```
agarny: This method is only used in `Units::scalingFactor()`, so I would the body of `doGetScalingFactor()` there.

src/units.cpp
```diff
@@ -118,6 +121,17 @@ struct Units::UnitsImpl
     std::vector<Unit> mUnits; /**< A vector of unit defined for this Units.*/
 
     std::vector<Unit>::iterator findUnit(const std::string &reference);
+    /**
+     * @brief Returns the log of the scaling factor between two similar units
+     * 
+     *  The factor returned can be interpreted as u1 = (factor)*u2
```
agarny: Hmm... I would consider `u1` to be our reference, so I would expect the returned factor to be interpreted as `u2 = factor*u1`.

src/units.cpp
```diff
@@ -126,43 +140,118 @@ std::vector<Unit>::iterator Units::UnitsImpl::findUnit(const std::string &refere
...
+        double exp;
+        double mult;
+        double expMult;
+        for (size_t i = 0; i < units->unitCount(); ++i) {
+            units->unitAttributes(i, ref, pre, exp, expMult, id);
+            mult = std::log10(expMult);
+            if (!isStandardUnitName(ref)) {
+                auto model = owningModel(units);
+                auto refUnits = model->units(ref);
+                updated = updateUnitMultipliers(unitMultiplierMap, multiplier, refUnits, exp * uExp, logMult + mult * uExp + standardPrefixList.at(pre) * uExp, direction);
```
agarny, line 162: I haven't checked the code in details, but couldn't there be a case where `updated` was previously `true` and then `updateUnitMultipliers()` returns `false`, making `updated` `false`? Same question for the other `updateUnitMultipliers()` method.

src/api/libcellml/generator.h
hsorby: It's a templated method the definition has to be there.

src/utilities.cpp
```diff
@@ -426,6 +426,24 @@ ModelPtr owningModel(const EntityPtr &entity)
     return model;
 }
 
+bool isStandardUnitName(const std::string &name)
+{
+    bool result = false;
+    if (standardUnitsList.count(name) != 0) {
+        result = true;
+    }
+    return result;
```
agarny: There are a lot of cases like this, so what is wrong with simply have:
```
return standardUnitsList.count(name) != 0;
```

src/utilities.h
```diff
@@ -375,4 +375,41 @@ ModelPtr owningModel(const EntityPtr &entity);
...
+/**
+ * @brief Check if the provided @p node is a supported MathML element.
+ *
+ * Checks if the provided @p node is one of the supported MathML elements defined in the table
+ * of supported MathML elements from the CellML specification version 2.0 document.
+ *
+ * @param node The @c XmlNode node to check against the list of supported MathML elements.
+ * @return @c true if @node is a supported MathML element and @c false otherwise.
+ */
+//bool isSupportedMathMLElement(const XmlNodePtr &node);
```
agarny: Shouldn't this be removed since we have no implementation for it in the end?

src/variable.cpp
```diff
@@ -216,7 +224,7 @@ void Variable::swap(Variable &rhs)
 {
     std::swap(mPimpl, rhs.mPimpl);
 }
-
+*/
```
agarny: Delete the above commented out code.

tests/bindings/python/test_component.py
```diff
@@ -22,10 +22,10 @@ def test_inheritance(self):
...
 
         x = Component()
-        self.assertIsInstance(x, libcellml.ComponentEntity)
-        self.assertIsInstance(x, libcellml.ImportedEntity)
-        self.assertIsInstance(x, libcellml.NamedEntity)
-        self.assertIsInstance(x, libcellml.Entity)
+        self.assertIsInstance(x, libcellml.componententity.ComponentEntity)
+        self.assertIsInstance(x, libcellml.importedentity.ImportedEntity)
+        self.assertIsInstance(x, libcellml.namedentity.NamedEntity)
+        self.assertIsInstance(x, libcellml.entity.Entity)
```
agarny, line 28: What is the rationale behind using a longer form of classes? I find it less convenient...

hsorby: This is because they are no longer available from the package init file.  This has been done because they are not classes that users should use or interact with in Python.

tests/coverage/coverage.cpp
```diff
@@ -269,6 +115,7 @@ TEST(Coverage, model)
 
     vec.push_back(rm);
     vec.insert(vec.begin(), ao);
+    */
 }
```
agarny: Delete that test since it's not relevant anymore.

src/api/libcellml/generator.h
agarny: Thanks for reminding me why I am not a fan of templates... :)

tests/bindings/python/test_component.py
agarny: Oh, ok, fair enough.

agarny (changes requested):
I understand that this PR contains some unfinished work. Still, I reckon there are some minor issues that could be addressed here. So, will happily approve once whatever can be addressed here has been addressed... here. :)

src/api/libcellml/component.h
hsorby: All classes currently go
* Constructor
* Destructor
* Copy
* Move
* Assignment

If the constructor is not protected or private.  Seems pretty consistent.

hsorby:
I have resolved all the review issues that relate to tidiness and correctness.

src/api/libcellml/component.h
agarny: Ok, my bad, I hadn't realised that. By default, I would have expected all types of constructors to be first and then the destructor. At least, it seems consistent indeed. So, happy with that.

agarny (approved):

src/api/libcellml/units.h
nickerso: see #447.

nickerso (approved):

Issue #175, opened 2017-08-31
Maths validation doesn't seem to pull in parent namespace declarations


nickerso:
I'm getting validation errors such as:
```
Namespace prefix cellml for units on cn is not defined
```
in models where the `cellml` namespace prefix is declared on the `model` element. Might need to rework some existing validation tests which may not be expecting the correct errors.

nickerso:
See also this comment in the `Parser::ParserImpl::loadComponent(...)` method:
```c++
// TODO: copy any namespaces declared in parents into the math element
//       so math is a valid subdocument.
```

nickerso:
This was semi-addressed by #420, but further work to tidy up things is required. Closing this one in favour of #424.

Issue #402, opened 2019-09-26
Code maintenance


agarny:
This issue aims at cleaning up the code, refactoring it whenever needed, etc. Things to be done (feel free to add to the list!):
 - [ ] As much as possible (and definitely in our comments), we should have 120 characters per line. (Yes, 80 characters used to be good when people didn't have wide screens.)
 - [ ] Go through the Clang-Tidy checks that have been disabled, and reenable whatever we can (e.g. `modernize-use-trailing-return-type`) (see issue #397).
 - [ ] Fix up implementation header file includes to follow the coding standard.
 - [ ] Make our lists of items consistent (see comment https://github.com/cellml/libcellml/pull/400#pullrequestreview-296551890), i.e. replace things like
```
const std::vector<std::string> expectedErrors = {
    "Encapsulation in model 'model_name' has an invalid child element 'component_free'."};
```
with
```
const std::vector<std::string> expectedErrors = {
    "Encapsulation in model 'model_name' has an invalid child element 'component_free'.",
};
```
 - [ ] Replace occurrences of "mathml" with "MathML".
 - [ ] Ensure that we use the same convention for file/folder names (e.g. `folder_name`, `my_super_duper_file.txt`).
 - [ ] Replace `std::shared_ptr<Xxx>` with `XxxPtr` wherever possible (e.g. `Component::create()`).
 - [ ] Replace calls to `std::string::at()` with `std::string::operator[]` if we know the index is valid and if, therefore, we don't handle a possible exception raised by `std::string::at()`.
 - [ ] Replace `std::list` with `std::vector` wherever possible (`std::vector` is much faster than `std::list`).
 - [ ] Clean up `while` loops and conditionals (see discussion [here](https://github.com/cellml/libcellml/pull/642#discussion_r445844944)).
 - [ ] Use `XxxPtr` rather than `std::shared_ptr<const Xxx>` (e.g. `getVariableIndexInComponent()`).
 - [ ] Ensure that documentation is consistent. That's a big one, but there is no consistency in our documentation, e.g. sometimes we use "Return" and other times "Returns" when describing the value returned by a function (see [here](https://github.com/cellml/libcellml/pull/631#discussion_r448723275)).
 - [ ] Ensure that we don't create trivial variables for returning.
 - [ ] Ensure that we use British spelling (as opposed to US spelling).
 - [ ] Use the `auto` keyword whenever possible in variable declarations.
 - [ ] Replace bits of code like:
```
Importer::Importer()
[...]

PR #420, opened 2019-10-08
Maths validation bring in cellml xml namespace

hsorby:
Using a text based method to bring in the cellml xml namespace (if required) into math sub documents on validation.

Addresses issue #175.

### DIFF
--- a/src/validator.cpp
+++ b/src/validator.cpp
@@ -30,6 +30,8 @@ limitations under the License.
 #include <cmath>
 #include <stdexcept>
 
+#include <iostream>
+
 #include <libxml/uri.h>
 
 namespace libcellml {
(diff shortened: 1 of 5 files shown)


hsorby:
This pull request is ready for review.

src/validator.cpp
```diff
@@ -16,22 +16,22 @@ limitations under the License.
...
 #include "namespaces.h"
 #include "utilities.h"
 #include "xmldoc.h"
 
-#include <algorithm>
-#include <cmath>
-#include <stdexcept>
-
-#include <libxml/uri.h>
-
```
agarny, line 34: Why reordering our header files using an order that is different from what was done in a recent suggested reordering?! So much inconsistency.

agarny (approved):
This looks good to me.

src/validator.cpp
hsorby: I thought this was the consistent ordering defined by the style guide.  Associated header first, system headers, project headers.

agarny: Am giving up. Will see that as part of issue #402.

src/validator.cpp
```diff
@@ -835,9 +835,20 @@ void Validator::ValidatorImpl::validateWhen(const WhenPtr &when, const ResetPtr
 
 void Validator::ValidatorImpl::validateMath(const std::string &input, const ComponentPtr &component)
 {
+    std::string modifiedInput = input;
+
+    // It may be that we need to copy over the cellml namespace from the enclosing document.
+    auto cellmlNamespaceText = " xmlns:cellml=\"" + std::string(CELLML_2_0_NS);
+    if (input.find("cellml:units") != std::string::npos && input.find(cellmlNamespaceText) == std::string::npos) {
```
nickerso, line 842: This will only work if we assume a `cellml` prefix is used.

nickerso (approved):
I think this works as well as our current handling of MathML and namespaces in parsing. So I think this improves the code, but needs further work to address these issues. See also https://github.com/cellml/libcellml/issues/118#issuecomment-235479335

PR #421, opened 2019-10-08
Test to show removeVariable bug

kerimoyle:
Addresses #404


hsorby:
It's an ok test but why involve `v2` and make the connection?  Why not just add a variable to a component then remove it and test its parent?

kerimoyle:
It's because I butchered it from another one which showed orphaned
connections, but you're right, it could be simpler.

hsorby:
Ready for review.

kerimoyle:
I think the trouble I'm having with this PR is probably better solved with really-super-clear documentation as I don't see a better way to think about it.  If you have two variables which don't have a direct connection and call removeEquivalence, the indirect connection remains.  I don't see a clean way that a function like removeAllConnectionWithThisOneVariable could work as it means changing the connections of other, possibly very distant variables.  So while I think that this implementation is the best we can do, we need to somehow make sure that people understand its workings ... Here's a test which fails:

Perhaps renaming `removeEquivalence` to `removeDirectEquivalence` would do the trick?

```
TEST(Variable, removeIndirectEquivalence)
{
    libcellml::ModelPtr m = libcellml::Model::create();
    libcellml::ComponentPtr comp1 = libcellml::Component::create();

    libcellml::VariablePtr v1 = libcellml::Variable::create();
    libcellml::VariablePtr v2 = libcellml::Variable::create();
    libcellml::VariablePtr v3 = libcellml::Variable::create();

    m->setName("modelName");
    comp1->setName("component1");

    v1->setName("variable1");
    v2->setName("variable2");
    v3->setName("variable2");

    v1->setUnits("dimensionless");
    v2->setUnits("dimensionless");
    v3->setUnits("dimensionless");

    comp1->addVariable(v1);
    comp1->addVariable(v2);
    comp1->addVariable(v3);
    m->addComponent(comp1);

    EXPECT_TRUE(v1->hasParent());
    EXPECT_TRUE(v2->hasParent());
    EXPECT_TRUE(v3->hasParent());

    // Make v1=v2=v3
    libcellml::Variable::addEquivalence(v1, v3);

    // All variables connected ...
    EXPECT_TRUE(v1->hasEquivalentVariable(v2));
    EXPECT_TRUE(v2->hasEquivalentVariable(v3));
    EXPECT_TRUE(v1->hasEquivalentVariable(v3));

    // ... but not directly through v1-v2
    EXPECT_FALSE(v1->hasDirectEquivalentVariable(v2));
    EXPECT_TRUE(v1->hasDirectEquivalentVariable(v3));
    EXPECT_TRUE(v2->hasDirectEquivalentVariable(v3));

    // Remove v1=v2 - remove indirect equivalence
    libcellml::Variable::removeEquivalence(v1, v2);

    EXPECT_TRUE(v1->hasEquivalentVariable(v3)); // nothing should have changed here
    EXPECT_TRUE(v2->hasEquivalentVariable(v3));
    EXPECT_TRUE(v1->hasDirectEquivalentVariable(v3));
    EXPECT_TRUE(v2->hasDirectEquivalentVariable(v3));

    EXPECT_FALSE(v1->hasDirectEquivalentVariable(v2)); // It never was anyway, no change here
    
    // Just removed this (or so we imagine) ... but connection remains.  
    // Need to document this really clearly as it's not really expected - test fails.
    EXPECT_FALSE(v1->hasEquivalentVariable(v2));
}
```

kerimoyle:
... or rename `hasEquivalentVariable` to `hasIndirectEquivalentVariable` (though names are getting a bit long now). Currently we have a bit of a name mismatch:

`addEquivalence` -> adds a **direct** equivalence
`removeEquivalence` -> removes a **direct** equivalence
`hasEquivalence` -> reports on an **indirect** equivalence

`hasDirectEquivalence` -> reports on a direct equivalence

hsorby:
Seems like we should rename `hasDirectEquivalence` to `hasIndirectEquivalence`. And swap the implementation of the `has*` methods.

kerimoyle:
Yep, I think I agree, so:

`addEquivalence` -> adds a **direct** equivalence
`removeEquivalence` -> removes a **direct** equivalence
`hasEquivalence` -> reports on a **direct** equivalence (used to be `hasDirectEquivalentVariable`)

`hasIndirectEquivalence` -> reports on an **indirect** equivalence (used to be 'hasEquivalentVariable`)

NB change of name from *`EquivalentVariable` to *`Equivalence` too?

hsorby:
Probably going to stick with `hasIndirectEquivalentVariable` as this reads better.

hsorby:
As an alternative what about the following options
- `hasEquivalentVariable(searchEquivalenceNetwork = false)`
- `hasEquivalence(searchNetwork = false)`
- `hasEquivalence(onlyDirect = false)`
- `hasEquivalenceWith(direct = false)`

thus combining the two methods into one?

agarny:
> As an alternative what about having `hasEquivalentVariable(searchEquivalenceNetwork = false)` thus combining the two methods into one?

Sounds good to me, but would rename the parameter: `hasEquivalentVariable(bool directEquivalence = true )`.

kerimoyle:
> hasEquivalence(onlyDirect = false)

This one gets my vote - the language is consistent (has/add/remove Equivalence) and I think that the default will be to search everywhere rather than only the direct ones?

hsorby:
I quite like the idea of having
- `bool hasEquivalenceWith(const VariablePtr &variable, bool directEquivalence = true)`

This then keeps using the word *equivalence* throughout which I think makes sense.  I think it also reads nicely which hopefully makes the API easier to use.

kerimoyle:
> I quite like the idea of having
> 
> * `bool hasEquivalenceWith(const VariablePtr &variable, bool directEquivalence = true)`
> 
> This then keeps using the word _equivalence_ throughout which I think makes sense. I think it also reads nicely which hopefully makes the API easier to use.

Except for the icky English ... (should be isEquivalentTo?) Maybe we can just be clearer and say add/remove/hasConnection instead of _equivalence_?  It's truer in a sense too ... 

But if function names are our biggest worry, we're in pretty good shape ...

hsorby:
>     // Remove v1=v2 - remove indirect equivalence
>    libcellml::Variable::removeEquivalence(v1, v2);

One thing that is quite important here is that the `removeEquivalence` method as used here returns false and in doing so informs the user that the equivalence was not removed.

tests/variable/variable.cpp
```diff
@@ -1165,3 +1165,121 @@ TEST(Variable, modelUnitsAttributeBeforeNameAttribute)
...
+    EXPECT_FALSE(libcellml::Variable::removeEquivalence(v1, v2));
+
+    EXPECT_TRUE(v1->hasIndirectEquivalentVariable(v3)); // nothing should have changed here
+    EXPECT_TRUE(v2->hasIndirectEquivalentVariable(v3));
+    EXPECT_TRUE(v1->hasEquivalentVariable(v3));
+    EXPECT_TRUE(v2->hasEquivalentVariable(v3));
+
+    EXPECT_FALSE(v1->hasEquivalentVariable(v2)); // It never was anyway, no change here
+
+    // Just removed this (or so we imagine) ... but connection remains.
```
kerimoyle: Change comments now test has changed?

src/bindings/interface/variable.i
```diff
@@ -16,11 +16,11 @@ objects are updated)";
 %feature("docstring") libcellml::Variable::equivalentVariable
 "Returns the equivalent variable at the given index.";
 
-%feature("docstring") libcellml::Variable::hasDirectEquivalentVariable
+%feature("docstring") libcellml::Variable::hasEquivalentVariable
 "Tests if the given variable is in this variable's set of equivalent
```
kerimoyle, line 20: Need to switch the docstrings here too

kerimoyle:
> One thing that is quite important here is that the removeEquivalence method as used here returns false and in doing so informs the user that the equivalence was not removed.

... wasn't removed or doesn't exist ... either way, it's confusing for the user.  I'm happy with the rename as it makes it a lot more consistent and clearer.  Just a couple of places where the docstrings need to be swapped also.

kerimoyle:
While we're in the area ... 

Writing these tutorials I've found it really useful to be able to work with the returned pointers from the get methods (eg: component->variable("i_exist") ) rather than creating a whole bunch of VariablePtrs.  The only trouble comes when you use it for a variable name which doesn't exist, which causes a segfault instead of failing nicely.  eg:

` libcellml::Variable::addEquivalence(c1->variable("i_exist"), c2->variable("i_dont"));`

... will crash it all.  Could we add a check for `nullptr` to the addEquivalence method, and the others in that group?

hsorby:
Yes, yes we can.

hsorby:
Ready for review.

src/api/libcellml/variable.h
```diff
@@ -251,20 +264,22 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
      * @deprecated This method is currently needed by our Printer and Validator classes, but this is
      * not a method that a libCellML user should ever need. It will therefore be removed at some point.
      */
-    bool hasDirectEquivalentVariable(const VariablePtr &equivalentVariable) const;
+    bool hasEquivalentVariable(const VariablePtr &equivalentVariable) const;
```
nickerso: is this method still deprecated?

hsorby: Probably not, may also look to add a parameter to this method and remove the need for `hasIndirectEquivalentVariable`.

hsorby:
Ready for review.

nickerso (approved):

kerimoyle:
Is this PR still happening?  Since it came from my branch I can't review it, sorry ...

hsorby:
Yes it is, I thought we could double team it.  If you are happy with it I will approve it for 'us'.

kerimoyle:
OK, great.  I wasn't sure with all the other similar work going on what the story was.  I'll have a proper look in the morning and let you know then.

src/api/libcellml/variable.h
```diff
@@ -242,30 +254,21 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
+     * equivalent network of this variable.
      *
-     * @return @c true if the @p equivalentVariable is equivalent to this variable and
-     * @c false otherwise.
+     * @return @c true if the @p equivalentVariable is in this variable's equivalent variables set or
+     * if @p considerIndirectEquivalences is @c true then @c true if the @p equivalentVariable is in
+     * this variable's equivalence network.  In all other cases @c false is returned.
      */
-    bool hasEquivalentVariable(const VariablePtr &equivalentVariable) const;
+    bool hasEquivalentVariable(const VariablePtr &equivalentVariable, bool considerIndirectEquivalences = false) const;
```
kerimoyle, line 272: I thought it was the other way around (so checks for indirect connections by default)?

src/printer.cpp
```diff
@@ -157,7 +157,7 @@ void buildMaps(const ModelPtr &model, ComponentMap &componentMap, VariableMap &v
             if (variable->equivalentVariableCount() > 0) {
                 for (size_t k = 0; k < variable->equivalentVariableCount(); ++k) {
                     VariablePtr equivalentVariable = variable->equivalentVariable(k);
-                    if (equivalentVariable->hasDirectEquivalentVariable(variable)) {
+                    if (equivalentVariable->hasEquivalentVariable(variable)) {
```
kerimoyle, line 160: See earlier comment.  If the default is false then this line's behaviour has changed?

src/validator.cpp
```diff
@@ -987,22 +987,16 @@ void Validator::ValidatorImpl::validateConnections(const ModelPtr &model)
                                 mValidator->addError(err);
                             }
 
-                            if (equivalentVariable->hasDirectEquivalentVariable(variable)) {
+                            if (equivalentVariable->hasEquivalentVariable(variable)) {
```
kerimoyle, line 990: Ditto here

src/variable.cpp
```diff
@@ -94,11 +103,22 @@ struct Variable::VariableImpl
      * @param equivalentVariable The variable to test for equivalence to this one.
      * @return @c true if the variables are equivalent @c false otherwise.
      */
-    bool hasDirectEquivalentVariable(const VariablePtr &equivalentVariable) const;
+    bool hasEquivalentVariable(const VariablePtr &equivalentVariable, bool considerIndirectEquivalences = false) const;
```
kerimoyle, line 107: check default here too

src/variable.cpp
```diff
@@ -246,39 +273,47 @@ size_t Variable::equivalentVariableCount() const
...
+ * Traverse the variable equivalence network to determine if the two given variables
+ * are equivalent.  Returns true if they are equivalent and false otherwise.
+ *
+ * @param variable1 The first variable to test.
+ * @param variable2 The second variable to test.
+ * @param testedVariables Vector of previoulsy tested variables.
+ *
+ * @return True if the two given variables are equivalent, false otherwise.
+ */
+bool haveEquivalentVariables(const Variable *variable1,
```
kerimoyle, line 314: maybe `areEquivalentVariables()` ? the "have" is throwing me a little

tests/bindings/python/test_variable.py
```diff
@@ -163,16 +128,51 @@ def test_has_equivalent_variable(self):
 
         self.assertFalse(v1.hasEquivalentVariable(v1))
         self.assertTrue(v1.hasEquivalentVariable(v2))
-        self.assertTrue(v1.hasEquivalentVariable(v3))
+        self.assertFalse(v1.hasEquivalentVariable(v3))
```
kerimoyle, line 131: check defaults

tests/variable/variable.cpp
```diff
@@ -528,17 +494,51 @@ TEST(Variable, hasEquivalentVariable)
 
     EXPECT_FALSE(v1->hasEquivalentVariable(v1));
     EXPECT_TRUE(v1->hasEquivalentVariable(v2));
-    EXPECT_TRUE(v1->hasEquivalentVariable(v3));
+    EXPECT_FALSE(v1->hasEquivalentVariable(v3));
```
kerimoyle, line 497: check defaults

tests/variable/variable.cpp
```diff
@@ -1165,3 +1165,243 @@ TEST(Variable, modelUnitsAttributeBeforeNameAttribute)
...
+    // Make v1=v2=v3
+    libcellml::Variable::addEquivalence(v2, v3);
+    libcellml::Variable::addEquivalence(v1, v3);
+
+    // All variables connected ...
+    EXPECT_TRUE(v1->hasEquivalentVariable(v2, true));
+    EXPECT_TRUE(v2->hasEquivalentVariable(v3, true));
+    EXPECT_TRUE(v1->hasEquivalentVariable(v3, true));
+
+    // ... but not directly through v1-v2
```
kerimoyle, line 1267: check defaults

tests/variable/variable.cpp
```diff
@@ -1165,3 +1165,243 @@ TEST(Variable, modelUnitsAttributeBeforeNameAttribute)
...
+
+TEST(Variable, leaveExpiredVariable)
+{
+    libcellml::VariablePtr v1 = libcellml::Variable::create();
+    libcellml::VariablePtr v2 = libcellml::Variable::create();
+    libcellml::VariablePtr v3 = libcellml::Variable::create();
+    libcellml::Variable::addEquivalence(v1, v2);
+    v2.reset();
+    libcellml::Variable::addEquivalence(v1, v3);
+    EXPECT_FALSE(v1->hasEquivalentVariable(v2));
```
kerimoyle, line 1294: check defaults

kerimoyle:
A few comments:
- I'd thought that the default for testing equivalence was to test indirect too but this seems to use the other way around.  Reading it again I think I was confused, but also wonder about the rationale for testing direct equivalence only by default?  For example:
   - You have a gate component which you now encapsulate deeper into a new channel component.  
It now has a different relationship to the variables it was previously connected to, so any tests you have will behave differently?  Then again, you wouldn't be allowed to have a direct equivalence then anyway ... gah, I'm so confused ... 
- when we compare one variable to another like this `v1->hasEquivalentVariable(v2)` then the function name makes sense.  When both the variables are in the argument list like this `haveEquivalentVariables(v1,v2)` it's a bit weird?
- (probably a separate PR) are we better using `auto` now that we're using `create() `everywhere?

src/api/libcellml/variable.h
hsorby: No that changed as per your comments above.  The default position is to now consider only directly equivalent variables and then optionally indirect variables.

hsorby:
I don't think our tests should use `auto` exclusively, I think it is fine for the tests to mix it up that way we show different ways of doing things.  I think this is more beneficial than having everything exactly the same.

src/api/libcellml/variable.h
nickerso: agreed, direct equivalence by default (i.e., there is a `connection` in this model between the variables).

src/printer.cpp
nickerso: no, direct equivalence is the same as defaulting to not checking indirect equivalences. So this behaviour has not changed.

src/variable.cpp
nickerso: pretty sure this is correct, please un-resolve if there is still confusion.

nickerso (approved):

src/api/libcellml/variable.h
kerimoyle: OK

kerimoyle:
OK @hsorby, good to go :)

hsorby (approved):
@kerimoyle and I think this PR is good to go.

Issue #422, opened 2019-10-08
Provide python casts for Entity::parent()


kerimoyle:
I've just updated my local python library to the develop branch, and the functionality which relied on component->parentComponent() and ->parentModel() is now gone (as expected from changes in PR #405).  I'm wondering what the better approach is for the python bindings?
- to provide a SWIG interface for a pre-cast pointer to each type, so that the parentComponent() and parentModel() calls (or something similar) in Python work, or
- to require the user to do the casting themselves in Python whenever they need to retrieve a parent of any kind?

Returning an entity pointer (which isn't really used anywhere else?) isn't as useful as a pointer to a concrete class would be ... (at least in my head!)

Thoughts anyone?

hsorby:
I'm not sure you need one in Python.  If it is a `Model` object you use it as a model object if it is a `Component` object you use it as a component.

kerimoyle:
That's what I'd imagined originally, but the SWIG wrapping in Python gives
you an Entity, not either a Component or Model, so you don't have access to
the functions which are tied to those derived classes.  Returning a cast
pointer behaved as I expected though.

kerimoyle:
Here's an example of how I've needed it and what I get.  The program is designed to load a cellml model using libcellml and translates it into a database class structure; this function connects the database objects to each other through the connections/equivalent variables.

```
def connect_equivalent_variables(component, in_entity):
    in_component = in_entity.component(component.cellml_index)

    for v in range(0, in_component.variableCount()):
        in_variable = in_component.variable(v)

        variable = component.variables.filter(name=in_variable.name()).first()

        for ev in range(0, in_variable.equivalentVariableCount()):

            in_equiv = in_variable.equivalentVariable(ev)  # returns a Variable

            # -------------------------------------
            in_equiv_parent = in_equiv.parent()  # returns an Entity
            # ent_name = in_equiv_parent.name()  # cannot be accessed because NamedEntity is a subclass ... ditto all the other functions that are useful in the child classes
            # -------------------------------------
            in_equiv_component = in_equiv.parentAsComponent()  # returns a Component
            in_equiv_grandparent = in_equiv_component.parentAsComponent()  # returns a Component, if it exists or None
            if not in_equiv_grandparent:
                in_equiv_grandparent = in_equiv_component.parentAsModel()  # returns a Model

            out_equiv_component = component.model.all_components.filter(name=in_equiv_component.name()).first()

            # Look for variable in the sibling component set
            if out_equiv_component:
[...]

PR #423, opened 2019-10-08
Expose Entity->parent() as Model or Component for Python

kerimoyle:
Addresses #422 

... and yeah, I know there's no test, but that's because I can't figure out how to write one ... I just need to get the parent pointer in the correct typecast in python.  These changes do what I need, but not sure what others will think?

### DIFF
--- a/src/api/libcellml/entity.h
+++ b/src/api/libcellml/entity.h
@@ -111,6 +111,20 @@ public:
      */
     bool hasAncestor(const EntityPtr &entity) const;
 
+    /**
+     * @brief Return the parent pointer cast to a Model
+     * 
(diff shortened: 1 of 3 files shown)


hsorby:
I'm not a fan of this approach.  Trying alternatives without success so far.

hsorby:
I have created an alternative to this in pull request #435 try it out and see if it works (it does for me).

nickerso:
Should this one be closed in favour of #435 ? or is it still not clear what alternative to go with?

hsorby:
I think this should be closed in favour of #435 waiting for an independent opinion on whether #435 works in the required situation or not.

kerimoyle:
Works for me, so closing.

Issue #424, opened 2019-10-09
Improve handling of namespaces in math strings


nickerso:
We currently do a rather poor job of handling XML namespaces with math strings. As described in the discussion starting here: https://github.com/cellml/libcellml/issues/118#issuecomment-235479335; we should make sure the math strings are fully self contained XML fragments which define all required XML namespaces when parsing models and similarly update the documentation of the `setMath()` methods to clearly define that the strings are expected to be self-contained fragments when set by user code.

See also: https://github.com/cellml/libcellml/pull/420#discussion_r332825594; for the validation code that will need to updated.

This is closely related to #175, which was semi-addressed by #420.

hsorby:
I have done some work on this and I have mapped the namespace to the math when parsing a model.  When validating a model it removes the cellml namespace information and attributes using libxml2 before asking MathML DTD for its opinion.

kerimoyle:
Here's a simple test based on the current develop branch which is really confusing me:

```
TEST(Validator, transformedCellml1To2)
{
    libcellml::Parser parser;
    libcellml::ModelPtr model = parser.parseModel(fileContents("segfault.cellml"));
    printErrors(parser);

    libcellml::Validator validator;
    validator.validateModel(model);
    printErrors(validator);

    EXPECT_EQ(size_t(0),validator.errorCount());
}
```
where the `segfault.cellml` file is this (I know the maths is wonky):

```
<?xml version="1.0" encoding="UTF-8"?>
<model xmlns="http://www.cellml.org/cellml/2.0#" xmlns:cellml="http://www.cellml.org/cellml/2.0#"
	name="segfault" id="segfault_id">
    <component name="component1"></component>
    <component name="component2">
        <variable name="a" units="dimensionless"/>
        <math xmlns="http://www.w3.org/1998/Math/MathML">
            <apply><eq/>
                <ci>a</ci>
                <cn cellml:units="dimensionless">10</cn>
            </apply>
        </math>
    </component>
    <encapsulation>
        <component_ref component="component1">
            <component_ref component="component2"/>
        </component_ref>
    </encapsulation>
</model>
```

which gives validation errors of:
```
Namespace prefix cellml for units on cn is not defined., 
Math cn element has an invalid attribute type 'cellml:units' in the cellml namespace., 
CellML identifiers must contain one or more basic Latin alphabetic characters., 3.1.3
Math cn element with the value '10' does not have a valid cellml:units attribute., 
[...]

PR #334, opened 2019-05-29
Documentation updates

nickerso:
This is a work in progress to address a range of documentation issues: 

- [ ] #315 Clarify parts of the coding guidelines
- [ ] #258 Add information about SWIG and Python
- [ ] #221 Improve development setup documentation
- [ ] #149 Figure numbers not showing in documentation
- [ ] #373 Documentation for clang-tidy

- [ ] #425 Tutorials for libCellML end users

PR #426, opened 2019-10-10
Improve math validation

hsorby:
Here we are shifting to using libXml2 to handle the namespaces of elements.  This should work in a larger variety of cases than what is currently handled.

Also, a change is made from the previous validation where the math strings need to self-sufficient for them to validate with no errors.

Addresses #424.


src/parser.cpp
```diff
@@ -232,6 +232,55 @@ void Parser::ParserImpl::updateModel(const ModelPtr &model, const std::string &i
     loadModel(model, input);
 }
 
+XmlNamespaceMap attributeNamespaces(const XmlNodePtr &node)
```
hsorby: Could move this function to src/xmlnode.cpp.

src/parser.cpp
```diff
@@ -232,6 +232,55 @@ void Parser::ParserImpl::updateModel(const ModelPtr &model, const std::string &i
...
+    XmlNamespaceMap namespaceMap;
+    auto tempAttribute = node->firstAttribute();
+    while (tempAttribute != nullptr) {
+        namespaceMap[tempAttribute->namespacePrefix()] = tempAttribute->namespaceUri();
+        tempAttribute = tempAttribute->next();
+    }
+    return namespaceMap;
+}
+
+XmlNamespaceMap determineMissingNamespaces(const XmlNamespaceMap &namespaceMap1, const XmlNamespaceMap &namespaceMap2)
```
hsorby: Could move this function to src/xmlnode.cpp.

src/parser.cpp
```diff
@@ -232,6 +232,55 @@ void Parser::ParserImpl::updateModel(const ModelPtr &model, const std::string &i
...
+    for (it = namespaceMap1.begin(); it != namespaceMap1.end(); ++it) {
+        auto result = namespaceMap2.find(it->first);
+        if (result == namespaceMap2.end()) {
+            undefinedNamespaces[it->first] = it->second;
+        }
+    }
+    return undefinedNamespaces;
+}
+
+XmlNamespaceMap traverseTreeForUndefinedNamespaces(const XmlNodePtr &node)
```
hsorby: Could move this function to src/xmlnode.cpp.

src/parser.cpp
```diff
@@ -398,8 +447,17 @@ void Parser::ParserImpl::loadComponent(const ComponentPtr &component, const XmlN
...
-            //       so math is a valid subdocument.
+            // Copy any namespaces that do not feature as a namespace definition
+            // of the math node into the math node.
+            auto mathElementDefinedNamespaces = childNode->definedNamespaces();
+            auto possiblyUndefinedNamspaces = traverseTreeForUndefinedNamespaces(childNode);
+            auto undefinedNamespaces = determineMissingNamespaces(possiblyUndefinedNamspaces, mathElementDefinedNamespaces);
+            XmlNamespaceMap::const_iterator it;
+            for (it = undefinedNamespaces.begin(); it != undefinedNamespaces.end(); ++it) {
+                childNode->addNamespaceDefinition(it->second, it->first);
+            }
```
hsorby: This is a big change from the previous state.  Now we look through the math tree and determine which namespaces are used and compares them to what is defined.  The missing namespaces are then added to the root node for the math.

src/validator.cpp
```diff
@@ -902,11 +884,12 @@ void Validator::ValidatorImpl::validateMath(const std::string &input, const Comp
     // Iterate through ci/cn elements and remove cellml units attributes.
     XmlNodePtr mathNode = node;
     validateAndCleanMathCiCnNodes(node, component, variableNames, bvarNames);
-    // Get the MathML string (with cellml:units attributes already removed) and remove the CellML namespace.
-    // While the removeSubstring() approach for removing the cellml namespace before validating with the MathML DTD
-    // is not ideal, libxml does not appear to have a better way to remove a namespace declaration from the tree.
+
+    // Remove the cellml namespace definition.
+    mathNode->removeNamespaceDefinition(CELLML_2_0_NS);
```
hsorby: Here we are doing a proper libXml2 removal of the CellML namespace from the subdocument.

tests/validator/validator.cpp
```diff
@@ -1112,7 +1127,7 @@ TEST(Validator, validMathCnElementsMissingCellMLNamespace)
     m->addComponent(c);
 
     v.validateModel(m);
-    EXPECT_EQ(size_t(0), v.errorCount());
+    EXPECT_EQ_ERRORS(expectedErrors, v);
```
hsorby, line 1422: This is an important change from previous validation.  Where there were no errors now there are a lot.  This is because the CellML namespace has not been bought through from the enclosing model.

hsorby:
This is ready for review.

kerimoyle:
I've just started looking and (selfishly ...) went straight to the model I wanted to use to see if it could be read.  (This is what I pared down to give the earlier test which segfaulted on my machine).  
The model is the Hodgkin Huxley 1952 from PMR, run through your wee script to translate from 1 to 2, and now using this branch I get a validator error (just switching out your final validation test file name): 

```
Math in component 'membrane' contains 'time' as a bvar ci element but it is already a variable name., 
Math in component 'sodium_channel_m_gate' contains 'time' as a bvar ci element but it is already a variable name., 
W3C MathML DTD error: No declaration for attribute cellml of element math., 
Math in component 'sodium_channel_h_gate' contains 'time' as a bvar ci element but it is already a variable name., 
Math in component 'potassium_channel_n_gate' contains 'time' as a bvar ci element but it is already a variable name.,
```

... I'm not sure whether the glitch is in the file, the translation, or the validation?  The middle error is a new one, but the others I think have been found before by @agarny ?

I'll have a look at the actual code now ... :)

src/xmlnode.h
```diff
@@ -57,9 +63,57 @@ class XmlNode
...
     std::string namespaceUri() const;
 
+    /**
+     * @brief Add a namespace definition to this XML element.
+     *
+     * Add a libXml2 namespace definition to this XML element using the given URI
+     * and prefix.
+     *
+     * @param uri The @c std::string representation of the XML namespace URI.
+     * @param prefix The @c std::string representatin of the XML namespace prefix.
```
kerimoyle: @param prefix The @c std::string representatin of the XML namespace prefix
check spelling ...

tests/validator/validator.cpp
```diff
@@ -1764,3 +1771,47 @@ TEST(Validator, unitComplexCycle)
 
     EXPECT_EQ_ERRORS(expectedErrors, v);
 }
+
+TEST(Validator, duplicatedCellMLUnitsOnCiElement)
+{
+    const std::string math =
+        "<math xmlns:cellml=\"http://www.cellml.org/cellml/2.0#\" xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
+        "  <ci cellml:units=\"dimensionless\" cellml:units=\"second\">B</ci>\n"
```
kerimoyle, line 2111: I'd thought that we didn't define units on the `ci` elements, only the `cn` ones?

tests/validator/validator.cpp
```diff
@@ -1764,3 +1771,47 @@ TEST(Validator, unitComplexCycle)
...
+        "</math>\n";
+
+    const std::vector<std::string> expectedErrors = {
+        "Attribute cellml:units redefined.",
+        "Could not get a valid XML root node from the math on component 'componentName'.",
```
kerimoyle, line 2116: Also don't understand how this error message correlates with the problem of duplicated namespace - not sure how that affects the root node?

src/validator.cpp
```diff
@@ -980,34 +983,23 @@ void Validator::ValidatorImpl::validateAndCleanMathCiCnNodes(XmlNodePtr &node, c
...
-        // Check that cellml:units has been set.
-        if (ciType) {
-            if (unitsAttribute != nullptr) {
-                ErrorPtr err = std::make_shared<Error>();
-                err->setDescription("Math ci element with value '" + textNode + "' has a cellml:units attribute with name '" + unitsAttribute->value() + "'.");
-            }
+        bool checkUnitsIsInModel = false;
+        // Check that cellml:units has been set if appropriate for type.
+        if (ciType && unitsAttribute != nullptr) {
+            checkUnitsIsInModel = validateCiCnUnits(component, unitsName, "ci", textNode);
```
kerimoyle: `ci` elements shouldn't contain units?

kerimoyle:
I can't add comments to some parts of the review pane:
- src/validator.cpp lines 1037, 1048: The description is odd, seems to be missing something in the middle or else why the gap after element?  `err->setDescription("Math has a '" + childNode->name() + "' element" + " that is not a supported MathML element.");`

- I'm confused about whether or not units are allowed/required/forbidden on `ci` elements.  A discussion ages ago with (I think, @agarny or @MichaelClerx?) said that we didn't allow units to be specified on `ci` elements as this was then an over-definition, as the variables had to have their own units anyway?

src/xmlnode.h
hsorby: If you see typos in documentation or comments just fix them.

tests/validator/validator.cpp
hsorby: It's not defined by the spec. I think.  The spec. doesn't say that you can't do that and somebody is free to write that sort of thing whether it is a validation error is at this point undefined.  I wouldn't mind someone pointing to the section in the spec. that refutes this though.

tests/validator/validator.cpp
hsorby: I think libXml2 takes exception to this really bad xml and gives you nothing back.

hsorby:
`Math in component 'membrane' contains 'time' as a bvar ci element but it is already a variable name.` messages like this may be a hang over from an older spec.  I think this should not be an error but I don't think the spec. is clear on this.

hsorby:
> * src/validator.cpp lines 1037, 1048: 

Message is ok I think just needs a little tidying.

> * I'm confused about whether or not units are allowed/required/forbidden on `ci` elements.  

I'm going with 'not defined by the spec. ' until proven otherwise.

hsorby:
`W3C MathML DTD error: No declaration for attribute cellml of element math.` . this might indicate that the math is not being cleaned properly, something to look into further.

kerimoyle:
> . I'm going with 'not defined by the spec. ' until proven otherwise.

You're right, it's not spelled out in the spec, except to say that we don't permit other unexpected attributes inside tags, so why would we here ...?  It *is* spelled out as required on the `cn` tags though.  There is a discussion in the specification document around this already?

kerimoyle:
> Math in component 'membrane' contains 'time' as a bvar ci element but it is already a variable name. messages like this may be a hang over from an older spec. I think this should not be an error but I don't think the spec. is clear on this.

I think @agarny said earlier that it shouldn't be an error (have searched for comment but can't find it ...).  I don't think it can be an error, as we need to define the variables outside the mathml, so however the variable is used inside it (as `<ci>` or `<bvar>`) shouldn't matter?

RE units in <ci> elements, see discussion [here](https://docs.google.com/document/d/1MleZJF6DYp4FO4vxslAHInrRFMqrSpARa1s_dPhwYU0/edit?disco=AAAACq5-iuI): 

> so ci elements cannot have units? or just don't have to?

> That's right, they can't have them in CellML 1 or 2.0, and units are just defined in the variable.
> So the ci uses the unit that the variable has in the component that the math is defined in

hsorby:
Yes I have made changes somewhere that I have temporarily misplaced that will only deal with cellml units on cn elements and require all ci elements to reference a variable in the component.  Once I where I put it that is.

hsorby:
All fixed up and ready to review now.

tests/validator/validator.cpp
```diff
@@ -608,20 +605,67 @@ TEST(Validator, invalidMathMLVariables)
...
 
+TEST(Validator, invalidSimpleMathmlCellMLUnits)
+{
+    const std::string math =
+        "<math  xmlns:cellml=\"http://www.cellml.org/cellml/2.0#\" xmlns=\"http://www.w3.org/1998/Math/MathML\"><apply><bvar><ci cellml:units=\"dimensionless\">B</ci></bvar></apply></math>";
+    const std::vector<std::string> expectedErrors = {
+        "CellML identifiers must contain one or more basic Latin alphabetic characters.",
+        "Model does not have a valid name attribute.",
+        "CellML identifiers must contain one or more basic Latin alphabetic characters.",
+        "Component does not have a valid name attribute.",
```
kerimoyle, line 616: Would it be practical to start reporting components with relationship to their parents now that we can do that easily?  It would make it a lot easier to locate the source of the error ... eg: "Component inside parent component "somethingHere" does not have a valid name" etc?

tests/validator/validator.cpp
```diff
@@ -608,20 +605,67 @@ TEST(Validator, invalidMathMLVariables)
...
-        "    <cn cellml:units=\"invalid\">oops</cn>\n"
+        "    <cn cellml:units=\"invalid\" cellml:value=\"zero\">oops</cn>\n"
         "    <apply>\n"
         "      <plus/>\n"
         "      <ci>A</ci>\n"
         "      <apply>\n"
         "        <plus/>\n"
         "        <bvar>\n"
-        "          <ci cellml:units=\"dimensionless\" cellml:value=\"zero\">new_bvar</ci>\n"
+        "          <ci cellml:units=\"dimensionless\">new_bvar</ci>\n"
```
kerimoyle, line 668: Also report on ci elements which have units?

hsorby: No we are not doing that the spec. only mentions `cellml:units` on `cn` elements so letting the MathML DTD complain about units on `ci` elements.

kerimoyle: Now looking below I understand this even less than before ... in line 700 are we saying that W3C needs a declaration of units for ci elements and that it's not there (but it is) or that it's not allowed? I am confused ...

tests/validator/validator.cpp
```diff
@@ -647,16 +691,14 @@ TEST(Validator, invalidMathMLCiAndCnElementsWithCellMLUnits)
...
-        "MathML ci element has no child.",
+        "Math cn element has an invalid attribute type 'value' in the cellml namespace.  Attribute 'units' is the only CellML namespace attribute allowed.",
+        "Math has a cn element with a cellml:units attribute 'invalid' that is not a valid reference to units in the model 'modelName' or a standard unit.",
+        "MathML ci element has the child text 'new_bvar' which does not correspond with any variable names present in component 'componentName'.",
+        "MathML ci element has the child text 'undefined_variable' which does not correspond with any variable names present in component 'componentName'.",
         "CellML identifiers must contain one or more basic Latin alphabetic characters.",
         "Math cn element with the value '2.0' does not have a valid cellml:units attribute.",
-        "W3C MathML DTD error: Namespace prefix cellml for value on ci is not defined.",
-        "No declaration for attribute cellml:value of element ci.",
+        "W3C MathML DTD error: No declaration for attribute units of element ci.",
```
kerimoyle, line 700: Error message is ambiguous ... not sure whether it's saying we must or must not have units on ci

tests/validator/validator.cpp
kerimoyle: In other sections of the validator/parser we do not permit random stuff to be present in the cellml ... not sure why this would be any different?  It's not (currently) explicitly forbidden, but if it is allowed, then we need to read it and validate it against the units defined for that variable elsewhere in the code.  I don't like the idea of having two possibly conflicting definitions permitted, and the user being unaware which is actually being used.

tests/validator/validator.cpp
kerimoyle: Could we change the error to be something more general then?  "This was so badly formed I didn't know where to start parsing" instead of pointing the user to an error which doesn't make sense?

kerimoyle (commented):
Looking good so far, but I'd like to make sure that the error messages returned are useful to the user.  This could perhaps be addressed en masse in a separate PR, but there are a couple that should be done here first.

tests/validator/validator.cpp
hsorby: It's a message from libXml2 when using the MathML DTD we don't have any control over what is says.

hsorby:
Ready for review.

kerimoyle:
Earlier issues from my review are still there ... Have we resolved where the units thing for <ci> tags sits yet?

tests/validator/validator.cpp
hsorby: No, the error itself holds a reference to the component that has the error.

hsorby:
I thought we had, we can only concern ourselves with `cellml:units` as an attribute on `cn` MathML elements.  Leaving others alone and letting the MathML DTD report errors on malformed MathML.

tests/validator/validator.cpp
kerimoyle: .. but the user doesn't get to see it unless they know to go hunting?  If you're interacting with libCellML only though a basic C++ file then this kind of error does nothing to help you.

kerimoyle (dismissed):
Approving these so we can keep working, but would like to create/extend the issue around error messages as many of our current ones (including ones here) aren't as helpful as they could be.

hsorby:
Ready for re-review.

src/debug.h
```diff
@@ -0,0 +1,55 @@
...
+*/
+
+#pragma once
+
+#include <iostream>
+#include <sstream>
+
+namespace libcellml {
+
+struct dbg
```
nickerso: can this be consolidated with the `debug` that has gone into the test utilities? or this is really something different? (Separate issue, not something that I think needs to hold this one up!)

src/parser.cpp
nickerso: I would think it would be better suited to `xmlnode.cpp` but not particularly fussed :)

src/validator.cpp
```diff
@@ -828,27 +799,18 @@ void Validator::ValidatorImpl::validateMath(const std::string &input, const Comp
...
     // Iterate through ci/cn elements and remove cellml units attributes.
     XmlNodePtr mathNode = node;
-    validateAndCleanMathCiCnNodes(node, component, variableNames, bvarNames);
-    // Get the MathML string (with cellml:units attributes already removed) and remove the CellML namespace.
-    // While the removeSubstring() approach for removing the cellml namespace before validating with the MathML DTD
-    // is not ideal, libxml does not appear to have a better way to remove a namespace declaration from the tree.
+    validateAndCleanMathCiCnNodes(node, component, variableNames);
+
+    // Remove the cellml namespace definition.
+    if (mathNode->hasNamespaceDefinition(CELLML_2_0_NS)) {
```
nickerso, line 808: why are we removing the CellML namespace?

tests/validator/validator.cpp
nickerso: The API docs should be sufficient to tell the user where to go looking for this information. Which we know is not the case currently and part of #334, so I think can resolve this in terms of this PR.

tests/validator/validator.cpp
nickerso: Looks like this confusion has been resolved, right? We're only cleaning the MathML strings of what we know should be there, anything else may be picked up by validating against the W3C MathML DTD.

tests/validator/validator.cpp
nickerso: CellML 2.0 explicitly specifies that it is invalid to put random stuff in random places. For MathML, we're only using the W3C MathML 2.0 DTD to validate the math strings, so if it is not picked up by that then its not something we can check for. Having said that, once we get "warning", you'd expect this specific warning would be a potential model debugging check that could be useful to avoid that potential misunderstanding about the duplication of a variable's units.

nickerso (dismissed):
Couple of outstanding comments but nothing that should hold this one up. Assuming the CI comes back with a green tick I am happy with this one.

src/debug.h
hsorby: No it can't be consolidated with the `debug()` in test utilities.  I find it kind of useful but still has the same problem of leaving `#include <iostream>` around.

src/validator.cpp
hsorby: Because this is about to be passed to the MathML DTD and if we have a stand alone math document it is likely to have a `cellml` namespace definition here.

src/debug.h
agarny: At least, it would be nice to have `dbg` renamed to `Debug`...

cmake/test_libxml2.cpp
```diff
@@ -0,0 +1,32 @@
...
+  xmlParserCtxtPtr context = xmlNewParserCtxt();	
+  xmlSetStructuredErrorFunc(context, structuredErrorCallback);	
+  xmlDocPtr xmlDoc = xmlCtxtReadDoc(context, reinterpret_cast<const xmlChar *>(mathmlString.c_str()), "/", nullptr, XML_PARSE_DTDVALID);	
+  xmlFreeParserCtxt(context);	
+  xmlSetStructuredErrorFunc(nullptr, nullptr);	
+  xmlFreeDoc(xmlDoc);
+  xmlCleanupParser();
+  xmlCleanupGlobals();	
+}
+
```
agarny: Am I missing something or is this file not used anywhere?

tests/parser/parser.cpp
```diff
@@ -1858,6 +1858,110 @@ TEST(Parser, xmlComments)
...
+        "  </component>\n"
+        "</model>\n";
+
+    libcellml::Parser parser;
+    parser.parseModel(input);
+
+    EXPECT_EQ(size_t(0), parser.errorCount());
+}
+
+TEST(Parser, mathWithMathmlNamespaceOnModel)
```
agarny, line 1939: Nice test!

tests/parser/parser.cpp
```diff
@@ -1858,6 +1858,110 @@ TEST(Parser, xmlComments)
...
+        "  </component>\n"
+        "</model>\n";
+
+    libcellml::Parser parser;
+    parser.parseModel(input);
+
+    EXPECT_EQ(size_t(0), parser.errorCount());
+}
+
+TEST(Parser, mathWithNamespacesDefinedOnTheMathNodeNonStandardPrefix)
```
agarny: What was the intention here? The test name gives the impression that we wanted to test a non-standard prefix for the CellML namespace, which is the case here, but it's not definied on the math node (as claimed in the name of the test), but on the model (as can be seen in the code). So, which one should it be? (I don't think it matters since the most important, here, is probably to test the non-standard prefix?)

hsorby: I think the test should be renamed to `mathWithNonStandardCellMLPrefix`.

agarny: Sounds good to me. Done in commit d4327175.

tests/validator/validator.cpp
```diff
@@ -608,20 +605,67 @@ TEST(Validator, invalidMathMLVariables)
...
+    m->addComponent(c);
+
+    v.validateModel(m);
+    EXPECT_EQ_ERRORS(expectedErrors, v);
+}
+
+TEST(Validator, invalidMathmlCellMLNsOnNode)
+{
+    const std::string math =
+        "<math  xmlns:cellml=\"http://www.cellml.org/cellml/2.0#\" xmlns=\"http://www.w3.org/1998/Math/MathML\"><apply><cellml:bvar><ci cellml:units=\"dimensionless\">B</ci></cellml:bvar></apply></math>";
```
agarny, line 634: I wish this test was based on a physical CellML file rather than inline code like here, which makes it more difficult to assess/review/understand/etc.

Comment also valid elsewhere.

nickerso: whereas I like to see them inline to make it easy to see everything in one place :)

At least for these small test models.

agarny: Sure, in this particular case, it was more about the lack of indentation and new lines. When it's some (short) CellML code that is indented and has new lines, then it's fine.

agarny (approved):
Looks good to me.

nickerso (approved):

tests/validator/validator.cpp
kerimoyle: I much prefer cellml inline but with indents for code that's not a real model.  This means that it can be commented so we can follow the tests' purposes more easily, and saves hunting around for files when they only contain a couple of lines.  Indents are needed though ... :)

Issue #428, opened 2019-10-14
Export of compiler features.


hsorby:
Currently the compiler features used by the libCellML library do not get exported.  This results in consuming projects not picking up the compiler features required for building against the libCellML library.  This needs to be fixed.

agarny:
Not sure I understand what you are talking about here...?

hsorby:
When someone (like @kerimoyle ) is trying to create an application like a tutorial (the consumer here) the compiler flags that are required to build successfully against the libCellML library are not propagated to the consuming target.  This results in a build failure.

agarny:
Oh, ok, got it now. Thanks.

nickerso:
Should we (can we?) look to add a test for this working? or that is really outside of the scope of our testing?

hsorby:
It is outside the scope of our unit testing but would still be a good test.  Would need to have a think about how this could be added.

kerimoyle:
Yuss! Huzzah!

Issue #425, opened 2019-10-09
Tutorials for libcellml use


kerimoyle:
Before releasing libcellml out into the big wide world we need to have some user tutorials covering:

- [ ] installation/setup for use with a C++ project
   - [ ] Windows
   - [ ] MacOS
   - [ ] Linux
- [ ] installation/setup for use with a Python project
   - [ ] Windows
   - [ ] MacOS
   - [ ] Linux
- [x] range of user tutorials and examples for C++
- [x] range of user tutorials and examples for Python
- [ ] background information and biology copied out of OpenCOR tutorial

@nickerso suggested I follow the use cases addressed by the OpenCOR tutorials - if anyone has any suggestions for what else should be included, please let me know.

nickerso:
For reference, this is the OpenCOR tutorial I'm talking about :)

https://tutorial-on-cellml-opencor-and-pmr.readthedocs.io/en/latest/

kerimoyle:
I've started by trying to make some template CMake files to build simple examples which use the libcellml library, but I'm confused about how we actually want people to access them (assuming I've got the generation stuff right).
I want to get the basics straight in my own head first so am writing the most bare-bones version of everything.  I've turned off all the optional extras (tests, coverage etc), set it to Release, and run `make -j` and then `make install` in a local directory to install it all.

NB: Confusion about simple stuff is my speciality ... apologies for the dumb questions ... 

**Question 1:** Do we want users to use the same` #include <libcellml>` idiom that we use in the tests?  In which case, is that a module or a library in CMake parlance? The only way I could get any joy building my wee program was to mention the *.h files by name and specify the include path in the CMakeLists.txt ... which causes problems later (see below).

CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.0.0)
project(tutorial1 VERSION 0.1.0)

add_executable(tutorial1 main.cpp)
target_include_directories(tutorial1 PUBLIC BEFORE "/Users/kmoy001/libcellml-tutorials/library/include")
```

main.cpp
```
#include <iostream>
//#include <libcellml>        // Does not work as module or include or library?
#include <libcellml/model.h>
#include <libcellml/parser.h>

int main() {
  std::cout << "hello world\n";
  libcellml::Parser parser;
}
```

**Question 2:** Which version of C++ compiler are we supporting?  There was talk recently about upgrades to 19(?) but I'm getting errors like this:
```
[...]

PR #429, opened 2019-10-14
Use target_compile_features function.

hsorby:
target_compile_features function propagates compiler flags onto consuming projects via the INTERFACE_COMPILE_FEATURES flag of exported targets.  Simply setting the required CXX standard via properties does not do this.

Addresses #428.


nickerso (dismissed):
Haven't had a chance to test this, but looks like it should work.

hsorby:
If you are interested in testing the following might help (some of which has been modified from #425.)

 [Assumes hsorby is a label for https://github.com/hsorby/libcellml.git, 'libcellml' is the name of the directory libcellml is cloned too, you are on a Unices based machine.]

```
git fetch hsorby export_compiler_features_i428
git checkout -b temp_test hsorby/export_compiler_features_i428
cd ..
mkdir tmp_build
cd tmp_build
cmake -DINSTALL_PREFIX=../install-libcellml ../libcellml
make -j4
make install
```

Then:

```
cd ..
mkdir tmp_app
mkdir tmp_app_build
cd tmp_app
```

In the `tmp_app` directory create the following two files with the contents as specified:

CMakeLists.txt
```
cmake_minimum_required(VERSION 3.10.2)
project(test_app VERSION 0.1.0)

find_package(libCellML REQUIRED)

add_executable(test_app main.cpp)
target_link_libraries(test_app PUBLIC cellml)
```

main.cpp
```
#include <iostream>
#include <libcellml>

int main()
{
  std::cout << "Hello libCellML!" << std::endl;
  libcellml::Parser parser;
}
```

Then configure and build the application:
```
cd ../tmp_app_build
cmake -DCMAKE_PREFIX_PATH=../install-libcellml ../tmp_app
make
```

The run the application:
```
./test_app
```

You should get the following output:
```
Hello libCellML!
```

kerimoyle:
Thanks @hsorby.  I've tried your test and got it going with a couple of tweaks.   FYI this is what I got when trying to build the new project:
`cmake -DCMAKE_PREFIX_PATH=../install-libcellml ..`

Gives output:
```
-- The C compiler identification is AppleClang 10.0.1.10010046
-- The CXX compiler identification is AppleClang 10.0.1.10010046
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at CMakeLists.txt:4 (find_package):
  By not providing "FindlibCellML.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "libCellML", but CMake did not find one.

  Could not find a package configuration file provided by "libCellML" with
  any of the following names:

    libCellMLConfig.cmake
    libcellml-config.cmake

  Add the installation prefix of "libCellML" to CMAKE_PREFIX_PATH or set
  "libCellML_DIR" to a directory containing one of the above files.  If
  "libCellML" provides a separate development package or SDK, be sure it has
  been installed.
```

I just needed to specify the child directories too, and then it worked fine:

`cmake -DCMAKE_PREFIX_PATH=../install-libcellml/lib/cmake/libCellML .. `

So I have a couple of questions:
- What's the reason for the single "../" at the beginning of the command above?  The actual relative path from the current place is two directories down not just one?
- Are we planning on distributing libCellML so that people need to build their own copy (like this), or are we aiming for precompiled libraries somehow instead?

hsorby:
You or I may have missed a `cd ..` so we could be one level out.  Disclaimer: I didn't actually test the above commands just straight outta my head.

hsorby:
> * What's the reason for the single "../" at the beginning of the command above?  The actual relative path from the current place is two directories down not just one?

Could be a mistake in the instructions.

> * Are we planning on distributing libCellML so that people need to build their own copy (like this), or are we aiming for precompiled libraries somehow instead?

Pre-compiled binaries somehow but still want instructions on using built libraries or installed libraries from own build.

kerimoyle:
> You or I may have missed a `cd ..` so we could be one level out. Disclaimer: I didn't actually test the above commands just straight outta my head.

That's what I thought too, but when I added in the extra `../` to make it make sense (to me..) it didn't work.  Your way was right, but I don't understand why!

hsorby:
I have tested my own commands now and believe I have it right, if you try and follow them again they should work 🤞

hsorby:
> Your way was right, but I don't understand why!

Not sure I understand this, which way was right?

kerimoyle:
This worked:
`cmake -DCMAKE_PREFIX_PATH=../install-libcellml/lib/cmake/libCellML ..`

- single `../` not the double that I would have expected
- exact path to the config file needed; just the install directory path was not enough.

hsorby:
Strange, the base directory of the install root works for me.  You could try the absolute path to the install root and see what happens.  From a fresh build directory of course.

kerimoyle:
The base directory worked the second time, but only after passing the full
path the first time. I guess there are leftovers somewhere ...

nickerso (approved):
Can confirm that @hsorby's directions worked perfectly for me.

hsorby:
Anybody else able to review this pretty simple pull request.

agarny (approved):
Can confirm that @hsorby's directions also work for me.

Issue #432, opened 2019-10-18
Simple first order model causing segfault.


hsorby:
The following model segfaults if you try and create it as an independent executable but if added as a test in the generator test suite passes.

```
void main()
{
    //  1.a   Create the model instance
    libcellml::ModelPtr model = std::make_shared<libcellml::Model>();
    model->setName("Tutorial4_FirstOrderModel");

    //  1.b   Create a component and add it into the model
    libcellml::ComponentPtr component = std::make_shared<libcellml::Component>();
    component->setName("IonChannel");
    model->addComponent(component);

    //  2.a   Define the mathematics.
    std::string mathHeader = "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">";

    // dy/dt = alpha_y*(1-y) - beta_y*y
    std::string equation1 =
        "<apply>\
            <eq/>\
            <apply>\
                <diff/>\
                <bvar>\
                    <ci>t</ci>\
                </bvar>\
                <ci>y</ci>\
            </apply>\
            <apply>\
                <minus/>\
                <apply>\
                    <times/>\
                    <ci>alpha_y</ci>\
                    <apply>\
                        <minus/>\
                        <cn cellml:units=\"dimensionless\">1</cn>\
                        <ci>y</ci>\
                    </apply>\
                </apply>\
                <apply>\
                    <times/>\
                    <ci>beta_y</ci>\
                    <ci>y</ci>\
                </apply>\
            </apply>\
        </apply>";
    // i_y = g_y*power(y,gamma)*(V-E_y)
    std::string equation2 =
        "<apply>\
            <eq/>\
            <ci>i_y</ci>\
            <apply>\
                <times/>\
                <ci>g_y</ci>\
                <apply>\
                    <minus/>\
                    <ci>V</ci>\
                    <ci>E_y</ci>\
                </apply>\
                <apply>\
                    <power/>\
                    <ci>y</ci>\
                    <ci>gamma</ci>\
                </apply>\
            </apply>\
        </apply>";
    std::string mathFooter = "</math>";

    //  2.b   Add the maths to the component.  Note that there is only one maths
    //        string stored, so parts which are appended must create a viable
    //        MathML2 string when concantenated.  To clear any string which is
    //        already stored, simply call setMath("") with an empty string.
    component->setMath(mathHeader);
    component->appendMath(equation1);
    component->appendMath(equation2);
    component->appendMath(mathFooter);

[...]

Issue #431, opened 2019-10-16
Segfault for maths in encapsulated components


kerimoyle:
See test below:

```
TEST(Validator, mathInEncapsulatedComponentSegfault)
{
    libcellml::ModelPtr model = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr c1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr c2 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr c3 = std::make_shared<libcellml::Component>();

    libcellml::Validator validator;

    model->setName("model");
    c1->setName("c1");
    c2->setName("c2");
    c3->setName("c3");

    model->addComponent(c1);
    model->addComponent(c2);
    c2->addComponent(c3);

    const std::string math =
        "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
        "  <apply>\n"
        "    <eq/>\n"
        "    <ci>v</ci>\n"
        "    <cn cellml:units=\"dimensionless\">1</cn>\n"
        "  </apply>\n"
        "</math>\n";

    libcellml::VariablePtr v = std::make_shared<libcellml::Variable>();
    v->setName("v");
    v->setUnits("dimensionless");
    c3->addVariable(v);

    validator.validateModel(model);
    EXPECT_EQ(size_t(0), validator.errorCount());  // passes fine ... 

    c3->setMath(math);

    validator.validateModel(model);  // SEGFAULTS HERE
    EXPECT_EQ(size_t(0), validator.errorCount());
}
```

hsorby:
FYI: Soon the math in a component will need to be a complete sub-document so this
```
"<math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
```
should be this
```
"<math xmlns=\"http://www.w3.org/1998/Math/MathML\" xmlns:cell=\"http://www.cellml.org/cellml/2.0#\">\n"
```
to validate without errors.

agarny:
Wouldn't it be sufficient to have something like:

```
<?xml version='1.0' encoding='UTF-8'?>
<model name="my_model" xmlns="http://www.cellml.org/cellml/1.0#" xmlns:cellml="http://www.cellml.org/cellml/2.0#">
    <component name="my_component">
        ...
        <math xmlns="http://www.w3.org/1998/Math/MathML">
            ...
            <cn cellml:units="dimensionless">1</cn>
            ...
[...]

PR #433, opened 2019-10-18
Segfault investigation

hsorby:
Fixing segfaulting issue with some models when using the libCellML library in a stand-alone application.

Addresses #432. 

This pull request is built on top of other pending pull requests and should be considered after they have been merged.


src/generator.cpp
```diff
@@ -553,8 +553,8 @@ struct Generator::GeneratorImpl
 
     Generator::ModelType mModelType = Generator::ModelType::UNKNOWN;
 
-    std::list<GeneratorInternalVariablePtr> mInternalVariables;
-    std::list<GeneratorEquationPtr> mEquations;
+    std::list<GeneratorInternalVariablePtr> mInternalVariables = {};
+    std::list<GeneratorEquationPtr> mEquations = {};
```
agarny: This shouldn't be needed...?

hsorby: No it's not I thought clang tidy might remind me to remove it.

hsorby:
This pull request is now ready for review.

kerimoyle:
Looks good to me :) The test from #431 is also included and passing, so this PR will close that issue too.

kerimoyle (dismissed):

tests/isolated/generator.cpp
```diff
@@ -0,0 +1,211 @@
+/*
```
kerimoyle, line 1: I don't understand how this is more "isolated" than the other tests?  It's still linking to the library in the same way, so it doesn't really test anything different?

tests/parser/parser.cpp
```diff
@@ -37,11 +37,11 @@ TEST(Parser, invalidXMLElements)
     const std::vector<std::string> expectedErrors = {
         "Specification mandate value for attribute bearded.",
         "Specification mandates value for attribute bearded.",
-        "Opening and ending tag mismatch: Dwarf line 3 and ShortGuy.",
-        "Opening and ending tag mismatch: Hobbit line 4 and EvenShorterGuy.",
-        "Opening and ending tag mismatch: Wizard line 5 and SomeGuyWithAStaff.",
-        "Opening and ending tag mismatch: Elf line 6 and fellows.",
-        "Premature end of data in tag fellowship line 2.",
+        "LibXml2 error: Opening and ending tag mismatch: Dwarf line 3 and ShortGuy.",
```
kerimoyle, line 40: I like the starting tags!  Makes it a bit more contextual when you're trying to interpret the error :)

agarny (approved):

nickerso (approved):

Issue #430, opened 2019-10-16
Validator doesn't find variables with invalid units in encapsulated components


kerimoyle:
A test and a question:
- See test below.  I'd expect the validator to report both sets of unknown units, but it only reports the first one...
- After playing around with the API for the last few days I'm wondering about what "best practice" is for things like the independent variables.  It's technically permitted to use the {} to define limited scope, and seems (segfaults and test errors aside ... ) to work reasonably well (in that it gives readable code without stupidly stupid variable names).  The alternative is to encourage thousands of top-level variables all referencing say time, for example, with different symbols? eg: t_environment, t_component1, t_component2 etc etc.  Since I'm writing these tutorials it makes sense to use the best practice in them!

```
TEST(Validator, unfoundUnitsInEncapsulatedComponents)
{
    // Create a model with three components and one variable in each to be shared between them
    libcellml::ModelPtr model = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr c1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr c2 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr c3 = std::make_shared<libcellml::Component>();

    libcellml::Validator validator;

    model->setName("model");
    c1->setName("c1");
    c2->setName("c2");
    c3->setName("c3");

    model->addComponent(c1);
    model->addComponent(c2);
    c2->addComponent(c3);

    {
        libcellml::VariablePtr v = std::make_shared<libcellml::Variable>();
        v->setName("v");
        v->setUnits("dimensionless");
        c1->addVariable(v);
    }
    {
        libcellml::VariablePtr v = std::make_shared<libcellml::Variable>();
        v->setName("v");
        v->setUnits("i_dont_exist");
        c2->addVariable(v);
    }
    {
        libcellml::VariablePtr v = std::make_shared<libcellml::Variable>();
        v->setName("v");
        v->setUnits("neither_do_I_but_Im_not_reported");
        c3->addVariable(v);
    }

    validator.validateModel(model);
    printErrors(validator);
    EXPECT_EQ(size_t(2), validator.errorCount());
}
```

PR #434, opened 2019-10-22
Unfound units validation error.

hsorby:
Addresses #430.


hsorby:
Ready to review.

src/validator.cpp
```diff
@@ -679,10 +679,10 @@ void Validator::ValidatorImpl::validateVariable(const VariablePtr &variable, con
         mValidator->addError(err);
     } else if (!isStandardUnitName(variable->units())) {
         ComponentPtr component = std::dynamic_pointer_cast<Component>(variable->parent());
-        ModelPtr model = std::dynamic_pointer_cast<Model>(component->parent());
+        ModelPtr model = owningModel(component);
         if ((model != nullptr) && !model->hasUnits(variable->units())) {
             ErrorPtr err = std::make_shared<Error>();
-            err->setDescription("Variable '" + variable->name() + "' has an invalid units reference '" + variable->units() + "' that does not correspond with a standard unit or units in the variable's parent component or model.");
+            err->setDescription("Variable '" + variable->name() + "' has an units reference '" + variable->units() + "' that does not correspond with a standard units and is not a units defined in the variable's model.");
```
agarny: "... a units..." rather than "... an units..."

tests/validator/validator.cpp
```diff
@@ -1764,3 +1764,36 @@ TEST(Validator, unitComplexCycle)
 
     EXPECT_EQ_ERRORS(expectedErrors, v);
 }
+
+TEST(Validator, unfoundUnitsInEncapsulatedComponents)
+{
+    // This test resulted from https://github.com/cellml/libcellml/issues/430
```
agarny: Don't need this. We can always work it out from GitHub blame, if (really) needed.

tests/validator/validator.cpp
```diff
@@ -1764,3 +1764,36 @@ TEST(Validator, unitComplexCycle)
...
+    const std::vector<std::string> expectedErrors = {
+        "Variable 'v' has an units reference 'neither_do_I_but_Im_not_reported' that does not correspond with a standard units and is not a units defined in the variable's model.",
+        "Variable 'v' has an units reference 'i_dont_exist' that does not correspond with a standard units and is not a units defined in the variable's model.",
+    };
```
agarny, line 1773: I kind of wish there was some logical order in the way errors are reported. Based on the model, I would expect the issue with the `i_dont_exist` unit to be reported first and then the one with the `neither_do_I_but_Im_not_reported` unit since `c3` is a child of `c2`. It's not critical and has nothing to do with this issue as such, just that it would be "cleaner" (a bad word here, I know!).

Otherwise, `neither_do_I_but_Im_not_reported` probably ought to be renamed (since it's now reported...!)?

agarny (dismissed):
Looks good pending some minor comments.

tests/validator/validator.cpp
kerimoyle: After being on the "other side" while creating the tutorials, I can second that the error messages leave a lot to be desired in terms of usefulness-in-debugging-your-model!  I suggest a separate PR that deals with them all ...

src/validator.cpp
```diff
@@ -679,10 +679,10 @@ void Validator::ValidatorImpl::validateVariable(const VariablePtr &variable, con
         mValidator->addError(err);
     } else if (!isStandardUnitName(variable->units())) {
         ComponentPtr component = std::dynamic_pointer_cast<Component>(variable->parent());
-        ModelPtr model = std::dynamic_pointer_cast<Model>(component->parent());
+        ModelPtr model = owningModel(component);
```
kerimoyle, line 682: What's the rationale for having a separate function `owningModel` outside the Entity class? It's not really consistent with other behaviour?

src/validator.cpp
kerimoyle: I find the use of "units" as a singular term really clumsy ... maybe we can make it clearer using "unit" and "compound unit" to make the distinction between the `unit` and `units` classes?

kerimoyle (dismissed):
Looks fine to me.  I think that we need to look at error messages separately as they're not currently very useful to an end user (eg: try not naming one item in a large model and see how long it takes you to figure out where it is ...).  I'd also like to see whether others feel as I do about the unit vs units thing - I think it makes all the language around it clumsy and hard to understand, so would prefer another term for the current Units items if possible (though probably not in this PR).

src/validator.cpp
hsorby: The `owningModel` function has nothing to do with the `Entity` class.  It is merely a utility function that saves a little bit of work.

tests/validator/validator.cpp
hsorby: It is possible to tidy this up a little it is the order the tree is searched in, currently we are doing a depth first search thus the deeper error gets found first.

src/validator.cpp
hsorby: I think it should be " ...a units... " because the "u" in units makes a consonant sound and not a vowel sound.

kerimoyle: it's not the "a" vs "an" that I'm talking about - I agree that "a unit" is better than "an unit".  It's the plural vs singular of "unit" meaning different things: we use the plural of "unit" to mean something other than more than one unit, we use it to mean the `libCellML::Units` items.  I think we should call our `libCellML::Units` class something like `CompoundUnit` so that the distinction is clearer and easier to write/talk about.

tests/validator/validator.cpp
kerimoyle: Yep, exactly.  Also reporting parents of components in their name would be helpful too.

src/validator.cpp
kerimoyle: ... and it shouldn't it be included in the Entity class because ...?

src/validator.cpp
hsorby: That was really a comment to @agarny on the English not a comment on the use of the term "units" in a singular manner.

src/validator.cpp
hsorby: The `Entity` class should know nothing of  the `Model` class.

src/validator.cpp
agarny: > I think it should be " ...a units... " because the "u" in units makes a consonant sound and not a vowel sound.

Yes, that's what I meant and my point since right now the message reads "... an units..."

tests/validator/validator.cpp
agarny: > It is possible to tidy this up a little it is the order the tree is searched in, currently we are doing a depth first search thus the deeper error gets found first.

I thought it might have been the case, and I clearly don't think this is the way it should be done.

> Yep, exactly. Also reporting parents of components in their name would be helpful too.

That's indeed what the code generator does when raising issues.

agarny:
> Looks fine to me. I think that we need to look at error messages separately as they're not currently very useful to an end user (eg: try not naming one item in a large model and see how long it takes you to figure out where it is ...). I'd also like to see whether others feel as I do about the unit vs units thing - I think it makes all the language around it clumsy and hard to understand, so would prefer another term for the current Units items if possible (though probably not in this PR).

I have always found the "unit" vs. "units" business confusing. I appreciate that is "needed" in CellML, but that's exactly where it should stay... in CellML. When reporting errors, we shouldn't be using "units" unless we are talking about severals units.

src/validator.cpp
agarny: > I think it should be " ...a units... " because the "u" in units makes a consonant sound and not a vowel sound.

This still needs to be corrected.

agarny (changes requested):
Still that "a units" rather than "an units" issue to be fxed.

src/validator.cpp
hsorby: Sorry missed the first use was only looking at the second.

agarny (dismissed):

agarny (approved):

tests/validator/validator.cpp
kerimoyle: I suggest we do a complete overhaul of all the errors, but in a separate PR.

kerimoyle (approved):

PR #435, opened 2019-10-22
Get SWIG to cast to derived type.

hsorby:
SWIG blasts over any derived type information when making bindings, we can however help it out with a `typemap`.

Addresses #422.


hsorby:
This is ready for review, in particular it should be checked to see if it resolves the issue raised.  It does for me :).

kerimoyle:
Thanks Hugh - I'll have a proper look at whether this solves my problem tomorrow.  I've been fiddling with the python installation so want to make sure that's all good before testing this out.

kerimoyle:
Yep, all good, thanks.

kerimoyle (approved):

nickerso (approved):

Issue #448, opened 2019-11-11
Rename clear* methods to remove*


hsorby:
Most of our methods are paired with add/remove but some have add/clear this issue looks to make all such methods add/remove.

Issue #443, opened 2019-11-05
consistency in clearing math strings


nickerso:
as per https://github.com/cellml/libcellml/pull/416#pullrequestreview-306336108 we need to make sure we are consistent in the way we clear math strings from our objects.

hsorby:
I think this method should be called `removeMath` as that is the pairing we have used elsewhere in the code (add/remove).

hsorby:
While we are at it we should probably add a `removeUnits` method to the `Variable` class.

kerimoyle:
> While we are at it we should probably add a `removeUnits` method to the `Variable` class.

... also see test in https://github.com/cellml/libcellml/pull/417 for bug with units

PR #449, opened 2019-11-11
Rename clear* methods to remove* methods.

hsorby:
Addresses #448


hsorby:
This one is now ready for review.

src/bindings/interface/logger.i
```diff
@@ -10,7 +10,7 @@
 %feature("docstring") libcellml::Logger::addError
 "Adds an error to this logger.";
 
-%feature("docstring") libcellml::Logger::clearErrors
+%feature("docstring") libcellml::Logger::removeAllErrors
```
kerimoyle, line 13: What's the rationale behind `removeAllErrors` - are we able to remove them selectively eg with a `removeError(index)` function?  That could be useful if/when we get to having different levels or types of error.  If not, I would prefer it just be `removeErrors`

kerimoyle (commented):
Can we please have a removeMath method for components too (as in #443) or is that coming later?

src/bindings/interface/logger.i
hsorby: The rationale is that when we are removing all types of thing from an object we have used `removeAll*` to do that job.

- `removeAllUnits`
- `removeAllComponents`
- `removeAllResets`
- `removeAllEquivalences`
- `removeAllVariables`

hsorby: There is currently no functionality to remove/take selective errors.

hsorby:
> Can we please have a removeMath method for components too (as in #443) or is that coming later?

It is coming in #450 which is built upon this pull request so will come in later.  This pull request is only dealing with the renaming of methods.

src/bindings/interface/logger.i
kerimoyle: Ok, thanks.

kerimoyle (approved):
All good with me.

agarny (approved):

Issue #458, opened 2019-11-15
Factory `create` method for all classes.


hsorby:
Now that we are using `enable_shared_from_this` some classes are allowing only the use of the factory create method while others have both the factory create method and the usual object construction method.  This issue is to make all libCellML classes use the factory `create` method for construction of an object and privatise the constructors to force the use of said method.

This issue follows from discussion on #453.

PR #460, opened 2019-11-16
Use `create` method for object construction only

hsorby:
This work forces the use of the `create` method for constructing new objects for all classes in the library.

Addresses #458.


hsorby:
Ready for review.

src/api/libcellml/component.h
```diff
@@ -43,11 +43,26 @@ class LIBCELLML_EXPORT Component: public ComponentEntity, public ImportedEntity
...
+     * @brief Create a @c Component object.
+     *
+     * Factory method to create a @c Component.  Create a
+     * blank component with::
+     *
+     *   ComponentPtr component = libcellml::Component::create();
+     *
+     * or a named component with name "Component" with::
+     *
+     *   ComponentPtr component = libcellml::Component::create(std::string("Component"));
```
agarny: Might want to simplify this a bit:

```
ComponentPtr component = libcellml::Component::create("Component");
```

src/api/libcellml/model.h
```diff
@@ -44,11 +44,26 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
+     * @brief Create a @c Model object.
+     *
+     * Factory method to create a @c Model.  Create a
+     * blank model with::
+     *
+     *   ModelPtr model = libcellml::Model::create();
+     *
+     * or a named model with name "Model" with::
+     *
+     *   ModelPtr model = libcellml::Model::create(std::string("Model"));
```
agarny: Simplify with:

```
ModelPtr model = libcellml::Model::create("Model");
```

src/api/libcellml/reset.h
```diff
@@ -37,11 +37,26 @@ class LIBCELLML_EXPORT Reset: public OrderedEntity
...
-    }
+    /**
+     * @brief Create a @c Reset object.
+     *
+     * Factory method to create a @c Reset.  Create a
+     * blank reset with::
+     *
+     *   ResetPtr reset = libcellml::Reset::create();
+     *
+     * or an ordered reset with order "10" with::
```
agarny: Drop the double quotes, it's not a string.

src/api/libcellml/variable.h
```diff
@@ -37,11 +37,26 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
+     * @brief Create a @c Variable object.
+     *
+     * Factory method to create a @c Variable.  Create a
+     * blank variable with::
+     *
+     *   VariablePtr variable = libcellml::Variable::create();
+     *
+     * or a named variable with name "Variable" with::
+     *
+     *   VariablePtr variable = libcellml::Variable::create("Variable");
```
agarny, line 50: Simplify with:

```
VariablePtr variable = libcellml::Variable::create("Variable");
```

agarny (changes requested):
Looks good, pending some (very) minor issues.

agarny (approved):

nickerso:
Its not clear to me why different objects have different `create` methods. It maybe makes sense that classes derived from `NamedEntity` have a create method with a `name` argument, but why a `Reset` create with an `order` attribute? Or perhaps why no `Variable` create method with name and units? (or did I miss it?)

hsorby:
> Its not clear to me why different objects have different `create` methods. It maybe makes sense that classes derived from `NamedEntity` have a create method with a `name` argument, but why a `Reset` create with an `order` attribute? Or perhaps why no `Variable` create method with name and units? (or did I miss it?)

I have added some convenience methods to see how they would work.  A `Reset` must have an order set so it seemed useful to have a method that allowed the creation of a `Reset` with an order.

I haven't done all possible convenience methods just the ones that I thought about as I modified the `create` method.  'Missing' convenience methods can be added at any time for those that have been missed.

nickerso (approved):

kerimoyle:
I'd have thought that having a reset with the test_variable and the reset_variable made more sense as a constructor than the order alone ... but as you say, there are a whole load of combinations ...

hsorby:
These are all good ideas we should probably put them into an issue on convenience constructors :)

Issue #467, opened 2019-11-18
Internal variables in environmentchecks.cmake are not declared properly.


hsorby:
Some CMake variables are declared `... INTERNAL BOOL ... ` this should read ` ... CACHE INTERNAL ... `.

PR #468, opened 2019-11-18
Correct declaration of internal CMake variables.

hsorby:
Addresses #467.

### DIFF
--- a/cmake/environmentchecks.cmake
+++ b/cmake/environmentchecks.cmake
@@ -84,25 +84,25 @@ else()
 endif()
 
 if(CLANG_FORMAT_EXE AND GIT_EXE)
-  set(CLANG_FORMAT_TESTING_AVAILABLE TRUE INTERNAL BOOL "Executables required to run the ClangFormat test are available.")
+  set(CLANG_FORMAT_TESTING_AVAILABLE TRUE CACHE INTERNAL "Executables required to run the ClangFormat test are available.")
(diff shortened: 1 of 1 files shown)


hsorby:
Ready for review.

agarny (approved):

cmake/environmentchecks.cmake
```diff
@@ -84,25 +84,25 @@ else()
 endif()
 
 if(CLANG_FORMAT_EXE AND GIT_EXE)
-  set(CLANG_FORMAT_TESTING_AVAILABLE TRUE INTERNAL BOOL "Executables required to run the ClangFormat test are available.")
+  set(CLANG_FORMAT_TESTING_AVAILABLE TRUE CACHE INTERNAL "Executables required to run the ClangFormat test are available.")
```
nickerso, line 87: just to check, `INTERNAL BOOL` is just wrong, right? It doesn't actually define a boolean type variable? And the implication of `INTERNAL` being a string-valued variable has no affect on the way we're using this? (or am I reading the docs wrong?)

hsorby: Yes `INTERNAL BOOL` is just wrong.  Being string valued has no affect on the way we're using this.  The string still evaluates to true, it's just not a constant true as it would be if it were a boolean.

nickerso (approved):

Issue #465, opened 2019-11-18
New API comparison methods for comparable and non-comparable units in units class


awickens:
Potentially adding in a new method called `areComparable`, which would check if units were of the same form (e.g. kilograms and grams). and add another method `areDimensionallyComparable` which checks if the units are exactly the same (e.g. kilograms and kilograms).

hsorby:
hmm, probably better to suggest the new API method for the units class.

PR #469, opened 2019-11-18
Proposed unit tests and function implementations for comparable and non-comparable units

awickens:
Proposed changes for #465.


src/units.cpp
```diff
@@ -477,7 +477,7 @@ void createUnitMap(const UnitsPtr &units, std::map<std::string, double> &unitMap
 
     } else if (isStandardUnitName(units->name())) {
         auto unit = standardUnitsList.find(units->name());
-        for (auto u : unit->second) {
+        for (auto const &u : unit->second) {
```
agarny: Also `auto const` and `const auto` are the same, we use the latter in libCellML. So, I would change `auto const &u` to `const auto &u` (same below).

src/api/libcellml/units.h
```diff
@@ -401,30 +401,30 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
+	 * @brief Return a bool which indicates if the units are loosely comparable.
+	 * 
+	 * Two units are loosely comparable if they share the same units ONLY (kg,s,m,...),
+	 * this means we are uninterested in unit multipliers. 
+	 *
+	 * @param1 units1 The first units to compare.
+	 * @param2 units2 The second units to compare.
+	 *
+	 * @return bool indicating whether units are comparable (true), false otherwise.
+	 */
```
agarny: Wrong indentation. Configure your editor to use spaces rather than tabs.

src/units.cpp
```diff
@@ -464,4 +464,99 @@ double Units::scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2)
     return 0.0;
 }
 
+void createUnitMap(const UnitsPtr &units, std::map<std::string, double> &unitMap, double exp = 1.0)
```
hsorby: Define a typemap for this `unitMap`; Something like `using UnitsMap = std::map<std::string, double>;`

src/units.cpp
```diff
@@ -464,4 +464,99 @@ double Units::scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2)
...
+        return false;
+    }
+    if ((units1->isImport()) || (units2->isImport())) {
+        return false;
+    }
+
+    std::map<std::string, double> units1Map;
+    createUnitMap(units1, units1Map);
+    std::map<std::string, double> units2Map;
+    createUnitMap(units2, units2Map); // creating the maps to compare units over
```
hsorby: If you are creating something like you are implying with the name `createUnitMap` you should just return the unit map from the function.  So this would look like `UnitsMap units1Map = createUnitMap(units1);` (using the typemap defined above.

src/units.cpp
```diff
@@ -464,4 +464,99 @@ double Units::scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2)
...
+            it++;
+        }
+        return true;
+    }
+    return false;
+}
+
+bool Units::isDimensionallyEquivalentTo(const UnitsPtr &units1, const UnitsPtr &units2)
+{
+    return ((Units::isEquivalentTo(units1, units2)) && (Units::scalingFactor(units1, units2) == 1.0));
```
hsorby: I feel we have gone a bit heavy on the braces here.

hsorby (changes requested):
From an intial review I have a few changes but looks promising.

src/units.cpp
agarny: Indeed, I would go with:
```
return Units::isEquivalentTo(units1, units2) && (Units::scalingFactor(units1, units2) == 1.0); 
```

src/api/libcellml/units.h
```diff
@@ -402,11 +402,11 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
-	 * 
-	 * Two units are loosely comparable if they share the same units ONLY (kg,s,m,...),
-	 * this means we are uninterested in unit multipliers. 
-	 *
-	 * @param1 units1 The first units to compare.
+	   * 
+	   * Two units are loosely comparable if they share the same units ONLY (kg,s,m,...),
+	   * this means we are uninterested in unit multipliers. 
+	   *
+	   * @param1 units1 The first units to compare.
```
agarny: Please, configure your editor properly. No tabs, only spaces!

awickens: `Tools -> Options -> Text Editor -> [Language] -> Tabs` then change to insert spaces? I did that prior but I'll re-check to make sure it hasn't magically decided to revert back.

agarny: I don't know which editor you are using (Visual Studio?), but personally I am using [Qt Creator](https://www.qt.io/development-tools) and it's all working fine for me there.

awickens: Yeah, I'm using Visual Studio. To be extra safe I just went through and re-indented the `units.h` docstrings using spaces only, so hopefully that resolves the issue.

awickens:
Code ready for re-review.

src/units.cpp
```diff
@@ -464,6 +464,91 @@ double Units::scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2)
...
+
+void updateUnitMap(const UnitsPtr &units, unitsMap &unitMap, double exp = 1.0)
+{
+    if (units->isBaseUnit()) {
+        auto found = unitMap.find(units->name());
+        if (found == unitMap.end()) {
+            unitMap.emplace(units->name(), 1.0);
+        } else {
+            auto unit = unitMap.find(units->name());
+            unit->second += exp;
```
nickerso: `found->second += exp;` ?

src/units.cpp
```diff
@@ -464,6 +464,91 @@ double Units::scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2)
...
+            auto found = units2Map.find(unit);
+
+            if (found == units2Map.end()) {
+                return false;
+            }
+            if ((found->second < it->second) || (found->second > it->second)) {
+                return false;
+            }
+
+            it++;
```
nickerso: `++it;` is generally preferred (I think).

src/api/libcellml/units.h
```diff
@@ -400,6 +400,33 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
      */
     static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2);
 
+    /**
+     * @brief Return a bool which indicates if the units are loosely comparable.
```
nickerso: not a fan of "loosely comparable" - much prefer the brief description used in the python bindings!

src/api/libcellml/units.h
```diff
@@ -400,6 +400,33 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
+     * 
+     * Two units are loosely comparable if they share the same units ONLY (kg,s,m,...),
+     * this means we are uninterested in unit multipliers. 
+     *
+     * @param1 units1 The first units to compare.
+     * @param2 units2 The second units to compare.
+     *
+     * @return bool indicating whether units are comparable (true), false otherwise.
+     */
+    static bool isEquivalentTo(const UnitsPtr &units1, const UnitsPtr &units2);
```
nickerso: `isEquivalentTo` doesn't seem appropriate for a static method. I'd go with `areEquivalent` or maybe just `equivalent`.

src/api/libcellml/units.h
```diff
@@ -400,6 +400,33 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
+     * Two units are dimensionally comparable if they are the same unit, e.g. volts and
+     * volts. This means we can check for the same base units across all parameters (kg,s,m), 
+     * using the exponent, multiplier, and prefix of the unit.
+     * 
+     * @param1 units1 The first units to compare.
+     * @param2 units2 The second units to compare.
+     *
+     * @return bool indicating whether units are dimensionally comparable (true), false otherwise.
+     */
+    static bool isDimensionallyEquivalentTo(const UnitsPtr &units1, const UnitsPtr &units2);
```
nickerso: as above, I'd go with `areDimensionallyEquivalent` or `dimensionallyEquivalent`.

nickerso (dismissed):
Couple of minor issues to change, but otherwise looking good. Great range of tests!

src/api/libcellml/units.h
agarny: I would go with `equivalent`.

src/api/libcellml/units.h
agarny: I would go with `dimensionallyEquivalent`.

src/units.cpp
awickens: Ah yes, that's a bit less redundant than what I had before.

src/units.cpp
agarny: Indeed, it's preferred, I removed it in the end. :)

src/api/libcellml/units.h
awickens: @agarny did you make those changes in the commits? (Just checking as I don't want to re-commit changes which have been already been made).

awickens: Otherwise I'll sort those requested changes now.

agarny: Commits e30945b and a16e090 are indeed your branch. Just do a `git pull`.

src/api/libcellml/units.h
```diff
@@ -389,17 +389,44 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
     static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2);
 
+    /**
+     * @brief Return whether the two Units are equivalent.
+     *
+     * Two Units are considered to be equivalent if they share the same units,
+     * independently of their dimension (e.g. volt and volt, volt and millivolt).
+     *
+     * @param1 units1 The first Units to compare.
+     * @param2 units2 The second Units to compare.
```
hsorby: I would use either `The first @c Units to compare.` or `The first units to compare.`

src/api/libcellml/units.h
```diff
@@ -389,17 +389,44 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
+    static bool equivalent(const UnitsPtr &units1, const UnitsPtr &units2);
+
+    /**
+     * @brief Return whether the two Units are dimensionally equivalent.
+     *
+     * Two Units are considered to be dimensionally equivalent if they share the
+     * exact same units (e.g. volt and volt, but not volt and millivolt).
+     *
+     * @param1 units1 The first Units to compare.
+     * @param2 units2 The second Units to compare.
```
hsorby: Repeat last comment.

src/units.cpp
```diff
@@ -464,6 +464,86 @@ double Units::scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2)
     return 0.0;
 }
 
+using unitsMap = std::map<std::string, double>;
```
hsorby: `unitsMap` is a name of something therefore `UnitsMap`.

src/units.cpp
```diff
@@ -464,6 +464,86 @@ double Units::scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2)
...
+    // Initial checks.
+    if ((units1 == nullptr) || (units2 == nullptr)) {
+        return false;
+    }
+    if ((units1->isImport()) || (units2->isImport())) {
+        return false;
+    }
+
+    unitsMap units1Map;
+    updateUnitsMap(units1, units1Map);
```
hsorby: Urgh, this I do detest.  If I was doing this I would probably create a small shim function to create my units map and inside that function have these two lines of code.

hsorby (changes requested):
Some more minors to fix.  Otherwise looking good.

src/units.cpp
agarny: Well spotted!

src/units.cpp
```diff
@@ -508,55 +497,78 @@ void createUnitMap(const UnitsPtr &units, std::map<std::string, double> &unitMap
...
-    createUnitMap(units2, units2Map); // creating the maps to compare units over
+    UnitsMap units1Map = createUnitsMap(units1);
+    UnitsMap units2Map = createUnitsMap(units2);
 
     if (units1Map.size() == units2Map.size()) {
-        auto it = units1Map.begin();
-
-        while (it != units1Map.end()) {
-            std::string unit = it->first;
+        for (auto &units : units1Map) {
```
hsorby: I'm surprised that there isn't a warning that this should be `const auto &units`.

hsorby (approved):
Looks good to me now.  Hopefully CI will return a green tick.

agarny (approved):

Issue #472, opened 2019-11-19
Seg fault when passing in nonstandard units without parent model to scalingFactor


awickens:
When passing in non-standard units without a parent model to `scalingFactor` and hence `updateUnitMultipliers`, the lines 

`auto model = owningModel(units);
 auto refUnits = model->units(ref)`

Return an error (clearly, as the units have no `owningModel` and hence `model` is null).

awickens:
Another area of interest is what do we expect to return if our nonstandard units have no `owningModel`? Would we expect `updateUnitMultipliers` to return the correct multiple or will we simply return 0?

PR #473, opened 2019-11-19
Tests which show seg fault when calling units without parent model

awickens:
Shows seg fault problems related to issue #472.


src/api/libcellml/units.h
```diff
@@ -385,6 +385,23 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
+	*
+	* @param multiplier The multiplier to find.
+	* @param units The units to find the multiplier for.
+	* @param uExp The exponential of the units.
+	* @param logMult The log multiplier of the units.
+	* @param direction The direction to update multiplier. Either 1 or -1.
+	*
+	* @return Either @c true or @c false, depending if the units were successfully updated.
+	*/
+    static bool Units::updateUnitMultiplier(double &multiplier, const UnitsPtr &units, double uExp, double logMult, int direction);
```
hsorby: This doesn't belong in the public API.

hsorby (changes requested):

src/api/libcellml/units.h
awickens: I'll make it private now.

hsorby: Doesn't even need to be private it was just fine being a function in the `units.cpp` implementation unit.

src/api/libcellml/units.h
```diff
@@ -415,14 +422,7 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
 	*/
-    static bool Units::updateUnitMultiplier(double &multiplier, const UnitsPtr &units, double uExp, double logMult, int direction);
-
-private:
-    Units(); /**< Constructor */
-    explicit Units(const std::string &name); /**< Constructor with std::string parameter*/
-
-    struct UnitsImpl; /**< Forward declaration for pImpl idiom. */
-    UnitsImpl *mPimpl; /**< Private member to implementation pointer */
+    static bool updateUnitMultiplier(double &multiplier, const UnitsPtr &units, double uExp, double logMult, int direction);
```
hsorby: There are really no private methods this is because we use the Pimpl idiom.  Any private methods should be added as public methods of the private implementation.

hsorby (changes requested):

src/api/libcellml/units.h
awickens: So we would just declare it inside of the `Units::UnitsImpl` struct? (In that case, would I keep the docstring as well or just declare the function inside?) 

Or would I just leave the function as it was before (no declaration, just in units.cpp)?

hsorby: If it doesn't require any access to the details of the units then just leave it as a function, documenting it is for other developers to understand the intentions of the function that should be done no matter where the function lives.

awickens:
Ready for re-review.

tests/units/units.cpp
```diff
@@ -758,6 +758,33 @@ TEST(Units, compareScalingFactorWithNullptrAsBothParameters)
...
+{
+    libcellml::UnitsPtr u1 = libcellml::Units::create();
+    u1->setName("u1");
+    u1->addUnit("apple", 0, 1.0, 1.0);
+
+    libcellml::UnitsPtr u2 = libcellml::Units::create();
+    u2->setName("u2");
+    u2->addUnit(libcellml::Units::StandardUnit::LUX, 0, 1.0, 1.0);
+
+    EXPECT_EQ(0.0, libcellml::Units::scalingFactor(u1, u2)); // Currently function not evaluating
```
agarny: What do you mean by "Currently function not evaluating"?

awickens: Ah, that's commenting left over from when I originally pushed the unit tests to show issue #472. That can be gotten rid of (if it hasn't already been done).

agarny: Ok, will remove it (together with some other minor cleaning up).

src/units.cpp
```diff
@@ -206,10 +221,11 @@ bool updateUnitMultipliers(double &multiplier,
             mult = std::log10(expMult);
             if (isStandardUnitName(ref)) {
                 multiplier += direction * (logMult + (standardMultiplierList.at(ref) + mult + standardPrefixList.at(pre)) * exp);
-            } else {
+                updated = true;
+            } else if (owningModel(units)) {
                 auto model = owningModel(units);
```
hsorby: I'm not a fan of this sort of thing, calling the same function twice in succession.  The test should be on the `model` object where we can clearly state that we don't want a `nullptr`.

hsorby (dismissed):

agarny (approved):

src/units.cpp
```diff
@@ -206,10 +221,13 @@ bool updateUnitMultipliers(double &multiplier,
...
             if (isStandardUnitName(ref)) {
                 multiplier += direction * (logMult + (standardMultiplierList.at(ref) + mult + standardPrefixList.at(pre)) * exp);
+                updated = true;
             } else {
                 auto model = owningModel(units);
-                auto refUnits = model->units(ref);
-                updated = updateUnitMultipliers(multiplier, refUnits, exp * uExp, logMult + mult * uExp + standardPrefixList.at(pre) * uExp, direction);
+                if (model != nullptr) {
+                    auto refUnits = model->units(ref);
+                    updated = updateUnitMultiplier(multiplier, refUnits, exp * uExp, logMult + mult * uExp + standardPrefixList.at(pre) * uExp, direction);
```
nickerso, line 229: #476 addresses the issue of not checking a units is actually found.

nickerso: imported units will also break this, I think.

nickerso (approved):
some issues to address in future changes, but improves the current state of the code.

nickerso:
Off topic question, but how come the changes from @awickens didn't resolve/dismiss @hsorby's requested changes review?

Issue #476, opened 2019-11-21
Units part of a model but without base unit in the model break updateUnitMultiplier


awickens:
If the `Units` created are *both* non-standard and part of a `model`, we would expect the units making up that `Unit` to also appear in the `model`. However, currently the only checks are in place for either `Units` which are a `standardUnit` or `Units` which have an `owningModel`, not if the `Units` has an `owningModel` but no base `Unit` in the model. Running tests cause a seg fault.

PR #477, opened 2019-11-21
Failure of scalingFactor when no base unit in defined model

awickens:
Causes a seg fault, as per #476.


src/units.cpp
```diff
@@ -226,6 +226,10 @@ bool updateUnitMultiplier(double &multiplier,
                 auto model = owningModel(units);
                 if (model != nullptr) {
                     auto refUnits = model->units(ref);
+                    if ((refUnits == nullptr) || (refUnits->isImport() == true)) {
```
agarny: Replace `(refUnits->isImport() == true)` with `refUnits->isImport()`.

src/units.cpp
```diff
@@ -226,6 +226,10 @@ bool updateUnitMultiplier(double &multiplier,
...
+                        updated = false;
+                        return updated;
```
agarny: Replace those two lines with `return false;`.

tests/units/units.cpp
```diff
@@ -915,7 +915,7 @@ TEST(Units, checkingOwningModelBothUnits)
 
     model->addUnits(u1);
     model->addUnits(u2);
-    model->addUnits(u3); // Should fail. No base units defined in model
+    model->addUnits(u3); // u not defined in model
```
agarny: No need for such comments. (Same for the one below).

awickens: Will make those changes now.

src/units.cpp
```diff
@@ -226,9 +226,8 @@ bool updateUnitMultiplier(double &multiplier,
                 auto model = owningModel(units);
                 if (model != nullptr) {
                     auto refUnits = model->units(ref);
-                    if ((refUnits == nullptr) || (refUnits->isImport() == true)) {
-                        updated = false;
-                        return updated;
+                    if ((refUnits == nullptr) || (refUnits->isImport())) {
```
agarny: I am being very pedantic here, but: no need for parentheses around `refUnits->isImport()`.

awickens: Okay, has been changed.

awickens:
Ready for re-review.

hsorby (changes requested):
None of those tests are using an imported `Units`.  A test with an imported `Units` should be added.

awickens:
Ready for re-review.

tests/units/units.cpp
```diff
@@ -891,3 +891,164 @@ TEST(Units, complicatedMultiplicationFactorUnits)
...
+    libcellml::UnitsPtr u2 = libcellml::Units::create();
+    u2->setName("u2");
+    u2->addUnit(libcellml::Units::StandardUnit::RADIAN, 0, 1.0, 1.0);
+
+    libcellml::ImportSourcePtr import = libcellml::ImportSource::create();
+    import->setUrl("I_am_a_url");
+
+    u1->setImportSource(import);
+
+    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u2));
```
nickerso, line 1002: I can see why this is giving `1.0` rather than `0.0` but I would expect it to fail. I suspect this is actually a bug in `Units::isBaseUnit()` which should probably be checking for imports....but need to think about this some more.

awickens: Yeah, that would likely be it. We check if the the unit `isBaseUnit()` first, so in that test case we never actually reach the point where we would check `refUnits->isImport()`, since both `Units` are only comprised of base ones.

awickens: I wonder if it would be even necessary to have such a condition, since the only way an imported `Unit` would fail is if there were units defined in there which were not in the original `model` and not `base`, and hence unable to be evaluated. I guess the question is if we want to ban all `imports` or just the ones which have units which are not `base`.....

awickens: @nickerso: I've just changed `Units::updateUnitMultipliers` to return 0 for that case. I guess it would be easier to not allow any imports rather than creating a whole ton of edge cases.

awickens:
Ready for re-review.

src/units.cpp
```diff
@@ -206,6 +206,11 @@ bool updateUnitMultiplier(double &multiplier,
 {
     bool updated = false;
     auto unitsName = units->name();
+
+    if (units->isImport()) {
```
nickerso: I don't think we should do this. The error is in `isBaseUnit()` which should be covered in #478, so I'd suggest removing this check here.

nickerso (changes requested):
I think we should defer fixing this in regard to imported units to #478, so these changes look good to me. I suspect removing the check on imports will need some changes to the tests for now.

src/units.cpp
awickens: So for now, simply remove the check and address it in a further pull request which focuses on #478? It will only change the one unit test.

nickerso (dismissed):

tests/units/units.cpp
```diff
@@ -891,3 +891,164 @@ TEST(Units, complicatedMultiplicationFactorUnits)
     // Incompatible units but return a scaling factor.
     EXPECT_EQ(1e-07, libcellml::Units::scalingFactor(incredible_pile_of_square_apples, bunch_of_bananas));
 }
+
+TEST(Units, checkingOwningModelBothUnits)
+{
+    libcellml::ModelPtr model = libcellml::Model::create();
+
+    libcellml::UnitsPtr u = libcellml::Units::create();
+    u->setName("u");
```
hsorby: This does nothing

tests/units/units.cpp
```diff
@@ -891,3 +891,164 @@ TEST(Units, complicatedMultiplicationFactorUnits)
...
+    u2->addUnit("u", "kilo", 2.0, 0.001); // standard, exponent.
+    u2->addUnit("dimensionless");
+
+    libcellml::UnitsPtr u3 = libcellml::Units::create();
+    u3->setName("u3");
+    u3->addUnit("u", "kilo", 4.0, 0.001); // standard, exponent.
+
+    model->addUnits(u1);
+    model->addUnits(u2);
+    model->addUnits(u3);
```
hsorby: Do we have a test where `u3` here is not added to the model but has a reference to an units that is in the model?

tests/units/units.cpp
```diff
@@ -891,3 +891,164 @@ TEST(Units, complicatedMultiplicationFactorUnits)
...
+
+    EXPECT_EQ(0.0, libcellml::Units::scalingFactor(u1, u2));
+    EXPECT_EQ(0.0, libcellml::Units::scalingFactor(u2, u3));
```
hsorby: I don't see what the `u2, u3` test shows us here that is different from the `u1, u2` test?

hsorby (changes requested):

tests/units/units.cpp
awickens: Ah, that might be a hang up from an earlier version of the unit test. I'll get rid of the redundancies in there now.

tests/units/units.cpp
awickens: We don't currently, but I can see how perhaps that would give something we wouldn't expect - I think at the moment the `owningModel` call means that such a combination would work, but considering that the units should all be part of the same model it might not make sense for it to work.

tests/units/units.cpp
awickens: Changes now fixed, the extra test passes.

tests/units/units.cpp
```diff
@@ -909,16 +906,34 @@ TEST(Units, checkingOwningModelBothUnits)
...
+TEST(Units, checkingOwningModelWithUnitBaseOnlyInModel)
+{
+    libcellml::ModelPtr model = libcellml::Model::create();
+
+    libcellml::UnitsPtr u = libcellml::Units::create();
+    u->setName("apple");
+
+    libcellml::UnitsPtr u1 = libcellml::Units::create();
+    u1->setName("u1");
+    u1->addUnit("u", "kilo", 2.0, 0.001); // standard, exponent.
```
hsorby: This unit should be `apple` right. Everything required for the scaling factor comparison is present except that `u2` has not been added to the model.

hsorby (changes requested):

tests/units/units.cpp
awickens: Yep, that's correct. It's been fixed and re-pushed.

hsorby (approved):

nickerso (approved):

nickerso:
Closes #476.

Issue #439, opened 2019-11-02
model copy constructor is not making a new copy of the model


nickerso:
When making a copy of a model (or any ``ComponentEntity`` really) I would expect it to make a new independent copy of the model. This doesn't seem to be the case currently. The vector of child components is copied, but this will simply have a list of the same component pointers, not new component objects.

nickerso:
As per #440, it seems that the copy constructor is simply making  a new pointer to the same "data". Changing this will likely have a large impact? So maybe the way to go for me to get what I need for now is to implement a clone (shallow and deep) method and take it from there.

kerimoyle:
When I was writing the online version I made a deep and shallow copy function, but found pretty quickly that the shallow one didn't really make much sense.  I think it would be cleaner to have a recursive duplicate (where the child items, through as many generations as needed are copied too).  I can't think of a use case for the shallow copy in our current situation except perhaps for units (but even then, not really since they have to be within the model anyway) ... what was it you had in mind for the shallow copy?

nickerso:
yep. Shallow isn't likely to be useful for what I need now, but should be a subset of the deep. I think.

agarny:
From http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0201r0.pdf, a "`shared-ptr`’s copy constructor is shallow". So, it looks like having a deep clone would be the way forward and... the aforementioned link provides a class that does just that.

hsorby:
I would like to see a function `duplicateModel`, with the work I am doing #418 the copy constructor will not be available due to the use of `enable_shared_from_this`.

agarny:
I am not sure we want a `duplicateModel()` method. Instead, I would expect something like `Entity::clone(bool deep = true)`?

hsorby:
I don't mind that but I would probably drop the `deep` parameter and just make the method do a deep copy job.

hsorby:
I can't see how a shallow copy could be useful and so far no use case exists to show that we need one.

agarny:
Fine by me (I merely put it in there because it had been mentioned before).

hsorby:
First tier clone-able objects (or objects that need to be cloned to create a cloned model):

- [ ] Model
- [ ] Component
- [ ] Variable
- [ ] Units
- [ ] ImportSource
- [ ] Reset

Second tier clone-able objects (or objects that may be helpful to be clone-able):

- [ ] GeneratorProfile

hsorby:
[...]

PR #479, opened 2019-11-24
Add clone method.

hsorby:
Adding clone method to all libCellML entities that are used by the `Model` class.

- [x] Model
- [x] Component
- [x] Variable
- [x] Units
- [x] ImportSource
- [x] Reset

Addresses #439.

### DIFF
--- a/src/api/libcellml/units.h
+++ b/src/api/libcellml/units.h
@@ -400,6 +400,8 @@ public:
      */
     static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2);
 
+    UnitsPtr clone() const;
+
 private:
     Units(); /**< Constructor */
     explicit Units(const std::string &name); /**< Constructor with std::string parameter*/
--- a/src/api/libcellml/units.h
+++ b/src/api/libcellml/units.h
@@ -400,6 +400,15 @@ public:
      */
     static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2);
 
+    /**
+     * @brief Create a clone of this units.
+     *
+     * Creates a full separate copy of this units without copying
+     * the parent. Thus the cloned (returned) version of this units
+     * will not have a parent set even if this units does.
+     *
+     * @return a new @c UnitsPtr to the cloned units.
+     */
     UnitsPtr clone() const;
 
 private:
--- a/src/component.cpp
+++ b/src/component.cpp
@@ -282,4 +282,9 @@ bool Component::hasReset(const ResetPtr &reset) const
     return mPimpl->findReset(reset) != mPimpl->mResets.end();
 }
 
+ComponentPtr Component::clone() const
+{
+    return create();
+}
+
 } // namespace libcellml
--- a/src/component.cpp
+++ b/src/component.cpp
@@ -282,6 +282,21 @@ bool Component::hasReset(const ResetPtr &reset) const
     return mPimpl->findReset(reset) != mPimpl->mResets.end();
 }
 
+size_t getVariableIndexInComponent(const std::shared_ptr<const Component> &component, const VariablePtr &variable)
+{
+    size_t index = 0;
+    bool found = false;
+    while (index < component->variableCount() && !found) {
+        if (component->variable(index) == variable) {
+            found = true;
+        } else {
+            ++index;
+        }
+    }
+
+    return index;
+}
+
 ComponentPtr Component::clone() const
 {
     auto c = create();
@@ -300,7 +315,18 @@ ComponentPtr Component::clone() const
 
     for (size_t index = 0; index < resetCount(); ++index) {
         auto r = reset(index);
-        c->addReset(r->clone());
+        auto rClone = r->clone();
+        c->addReset(rClone);
+        size_t variableIndex = getVariableIndexInComponent(shared_from_this(), r->variable());
+        if (variableIndex < variableCount()) {
+            auto v = c->variable(variableIndex);
+            rClone->setVariable(v);
+        }
+        size_t testVariableIndex = getVariableIndexInComponent(shared_from_this(), r->testVariable());
+        if (testVariableIndex < variableCount()) {
+            auto v = c->variable(testVariableIndex);
+            rClone->setTestVariable(v);
+        }
     }
 
     for (size_t index = 0; index < componentCount(); ++index) {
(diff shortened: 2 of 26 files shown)


kerimoyle:
Not sure if it helps or hinders, but the python code I wrote to do this kind of thing is here:  https://github.com/kerimoyle/cellml_as_database/blob/development/main/copy.py
I think all the functions should work independently as well as in a cascade for a "deep" copy.

hsorby:
Ready for review.

src/api/libcellml/units.h
```diff
@@ -400,6 +400,8 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
      */
     static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2);
 
+    UnitsPtr clone() const;
```
agarny, line 412: Need documentation.

src/model.cpp
```diff
@@ -339,4 +339,159 @@ bool Model::hasUnresolvedImports()
...
+    for (size_t index = 0; index < model->componentCount(); ++index) {
+        indexStack.push_back(index);
+        auto c = model->component(index);
+        recordVariableEquivalences(c, map, indexStack);
+        generateEquivalenceMap(c, map, indexStack);
+        indexStack.pop_back();
+    }
+
+    return map;
+}
```
agarny: This method is not needed as such. Its body could be directly in `Model::clone()`.

agarny: In fact, `Model::clone()` could declare `map` and `indexStack`, and then call a generic `generateEquivalenceMap()` (since both `Model` and `Component` have components)? This would avoid duplicated code.

tests/clone/clone.cpp
```diff
@@ -0,0 +1,550 @@
...
+    v->setUnits("lightning");
+    v->setInitialValue(-3.0);
+    v->setInterfaceType(libcellml::Variable::InterfaceType::PRIVATE);
+    v->setId("id_unique");
+    v->setName("variable");
+
+    auto vClone = v->clone();
+
+    EXPECT_EQ(nullptr, vClone->parent());
+    EXPECT_NE(v->units(), vClone->units());
```
agarny, line 163: I take it that the units are not equal because of their parent being different? If so, it would be nice to explicitely check that? (Same for the other similar test below.)

agarny: In fact, it would be nice to confirm in all our tests (e.g. `variableWithUnitsPtrSet `) that it is indeed the parent that is making the units different.

agarny: Oops, I have just realised that I might have got it wrong. I was thinking of `compareUnits()` while here we are "simply" using `EXPECT_NE`, so it's probably checking the units address or something?

tests/clone/clone.cpp
```diff
@@ -0,0 +1,550 @@
...
+TEST(Clone, modelWithUnits)
+{
+    auto m = libcellml::Model::create();
+    m->setId("unique_model");
+    m->setName("model");
+
+    auto mClone = m->clone();
+
+    compareModel(m, mClone);
+}
```
agarny, line 514: `modelWithUnits`? I can't see any...

tests/clone/clone.cpp
```diff
@@ -0,0 +1,550 @@
...
+    c1->setName("child_1");
+    c2->setName("child_2");
+
+    c->addComponent(c1);
+    c->addComponent(c2);
+
+    auto mClone = m->clone();
+
+    compareModel(m, mClone);
+}
```
agarny, line 539: Shouldn't `c` be added as a component to `m`?

agarny (changes requested):
Looks good overall, just a few things that need clarifications/fixing/etc.

tests/clone/clone.cpp
hsorby: This is comparing the address of the pointers to show that they are distinct and not pointing to the same resource.

agarny: Ok, what I realised then.

src/model.cpp
hsorby: The problem is the `shared_from_this` it is a subtly different type of shared pointer which can't be converted to the shared pointers that we typically use.

tests/clone/clone.cpp
hsorby: O for awesome on that one.

tests/clone/clone.cpp
hsorby: Double O for awesome, I'm on fire.

src/model.cpp
agarny: Ok, I thought it might have been a problem. Fair enough then re avoiding duplication of code. However, we might still want to move the contents of the method to `Model::clone()`. Not critical though.

tests/clone/clone.cpp
agarny: Makes me wonder whether I haven't missed other similar awesomeness from you... :)

hsorby: That has to be the limit of my awesomeness! surely 😆

agarny: I would hope! Hmm... :)

hsorby:
All review issues have been resolved with the latest changes.

agarny (approved):

src/component.cpp
```diff
@@ -282,4 +282,59 @@ bool Component::hasReset(const ResetPtr &reset) const
     return mPimpl->findReset(reset) != mPimpl->mResets.end();
 }
 
+size_t getVariableIndexInComponent(const std::shared_ptr<const Component> &component, const VariablePtr &variable)
```
nickerso, line 285: Is this different to the `getVariableIndexInComponent` in `utilities.cpp` ? (is this the `shared_from_this` issues again?)

hsorby: Yes,

hsorby: because the method is `const`, `shared_from_this` returns a `std::shared_ptr<const *>` object.  As it should, unfortunately.  We probably could do with adding a `const` form of our shared pointers I guess?

hsorby: Maybe worth trying as a separate issue.

nickerso: thanks for the explanation and yes to trying in a separate issue!

nickerso (approved):

Issue #480, opened 2019-11-28
Validation for units/components in models


awickens:
Currently, since the addition of the `equivalent()` and `dimensionallyEquivalent()` methods into `libcellml`, there is no associated validation checks in `validator.cpp`. This would mean adding some unit tests and model checks which use `equivalent()` and `dimensionallyEquivalent()`, done in #465.

agarny:
@awickens: it would help if you could reference where that work on `equivalent()` and `dimensionallyEquivalent()` was done. Right now, we know because it was completed yesterday, but... in the future it would be good if we could easily go back to it without having to waste time searching for it.

kerimoyle:
They weren't included in the validator because we're allowed (apparently) to have non-matching multipliers between equivalent variables as long as the base units are the same.  It's waiting for the Issue class (or similar) to be written so that we can return different levels of warning (see https://github.com/cellml/libcellml/pull/368 for discussions)

PR #482, opened 2019-12-01
Fixes for equivalent() method

awickens:
Implementing code which works towards addressing #480 by checking that units are correctly mapped to one another.


awickens:
Still currently working through issues in checking for duplicate mappings, but any feedback on whether I am heading in the right direction for mapping checks (especially in regard to the new function `checkUnitHomogeneity()`) would be appreciated.

src/validator.cpp
```diff
@@ -1249,4 +1259,47 @@ void Validator::ValidatorImpl::checkUnitForCycles(const ModelPtr &model, const U
...
+                            && (std::find(checkedPairs.begin(), checkedPairs.end(), secondCheckPairing) == checkedPairs.end())) {
+                            checkedPairs.push_back(firstCheckPairing);
+
+                            // Check equivalence
+                            bool equivalent = variable->units()->equivalent(variable->units(), equivalentVariable->units());
+                            if (!equivalent) {
+                                auto unitsName = variable->units() == nullptr ? "" : variable->units()->name();
+                                auto equivalentUnitsName = equivalentVariable->units() == nullptr ? "" : equivalentVariable->units()->name();
+                                ErrorPtr err = Error::create();
+                                err->setDescription("Error: Variables '" + variable->name() + "' and '" + equivalentVariable->name() + "' do not have the same unit reduction.");
```
hsorby: We don't need the `Error: ` prefix in the error description, one of those that is what it says on the tin things.

src/units.cpp
```diff
@@ -426,15 +426,16 @@ void updateUnitsMap(const UnitsPtr &units, UnitsMap &unitsMap, double exp = 1.0)
             std::string pre;
             std::string id;
             double expMult;
-            units->unitAttributes(i, ref, pre, exp, expMult, id);
+            double e;
+            units->unitAttributes(i, ref, pre, e, expMult, id);
             if (isStandardUnitName(ref)) {
                 auto unit = standardUnitsList.find(ref);
```
hsorby: This is a poor choice of variable name.

src/units.cpp
```diff
@@ -456,6 +457,32 @@ UnitsMap createUnitsMap(const UnitsPtr &units)
...
+        if (it->second == 0.0) {
+            auto found = unitsMap.find("dimensionless");
+            if (found == unitsMap.end()) {
+                unitsMap.emplace(std::make_pair("dimensionless", 0.0));
+                it = unitsMap.erase(it);
+                --it;
+            } else {
+                it = unitsMap.erase(it);
+                --it;
+            }
```
hsorby: I don't understand your thinking here what is it that you are trying to achieve?

src/validator.cpp
```diff
@@ -258,6 +258,13 @@ struct Validator::ValidatorImpl
     void checkUnitForCycles(const ModelPtr &model, const UnitsPtr &parent,
                             std::vector<std::string> &history,
                             std::vector<std::vector<std::string>> &errorList);
+
+    /**
+    * @brief checks model mapping to ensure all variables mapped to other variables have the same units.
+    *
+    * @param model The model containing the variables and mappings to be tested.
+    */
+    void checkUnitHomogeneity(const ModelPtr &model);
```
hsorby: Brief should start with a capital letter, like any sentence.  There should also be a full explanation given in the lines below.

src/validator.cpp
```diff
@@ -1249,4 +1259,47 @@ void Validator::ValidatorImpl::checkUnitForCycles(const ModelPtr &model, const U
     }
 }
 
+void Validator::ValidatorImpl::checkUnitHomogeneity(const ModelPtr &model)
+{
+    std::vector<std::pair<VariablePtr, VariablePtr>> checkedPairs;
+    std::vector<std::pair<VariablePtr, VariablePtr>> errorPairs;
```
hsorby: Declare this using `using`.

hsorby (commented):
Some preliminary comments and one quiestion.

src/units.cpp
awickens: Yep, most of this is preliminary until I actually get code running (just OSX compiler mucking things up at the moment). I'll probably rename it to something like `uExp` for better clarity.

src/units.cpp
awickens: The problem this part of the code was trying to address was issues when things become `dimensionless`. One problem I found when doing validation was that if we had units which reduced from some complicated expression to "meters^0.0, amperes^1.0, kg^2.0" and another unit which reduced to "seconds^0.0, amperes^1.0, kg^2.0", both those units should return `true` when passed into `equivalent()`. However, the `map` in its original form would keep `meters` and `seconds` in it, so the idea here is to add a `dimensionless` placeholder if anything has an `exponent` of 0.0, to eliminate such cases of unit cancellation returning the incorrect output.

awickens: Although thinking about it now it's probably not necessary to do this; would be much cleaner to instead get rid of both any units with an `exponent` of 0.0, and any units with `dimensionless` in them.

src/validator.cpp
```diff
@@ -260,8 +260,13 @@ struct Validator::ValidatorImpl
                             std::vector<std::vector<std::string>> &errorList);
 
     /**
-    * @brief checks model mapping to ensure all variables mapped to other variables have the same units.
+    * @brief Checks @p model mappings to ensure all variables mapped to other variables have the same units.
+    * Any errors are logged in the @c Validator.
```
hsorby: Brief statements can only have one sentence just delete the second sentence you mention it in the full description.

hsorby (commented):
Minor documentation correction.

awickens:
Ready for review.

tests/validator/validator.cpp
```diff
@@ -2242,3 +2251,218 @@ TEST(Validator, unfoundUnitsInEncapsulatedComponents)
 
     EXPECT_EQ_ERRORS(expectedErrors, v);
 }
+
+TEST(Validator, validateUnitsOfCompenentsSimple)
```
kerimoyle: Spelling of test name?

tests/validator/validator.cpp
```diff
@@ -2242,3 +2251,218 @@ TEST(Validator, unfoundUnitsInEncapsulatedComponents)
...
+
+    libcellml::Variable::addEquivalence(v1, v2);
+
+    v->validateModel(model);
+
+    EXPECT_EQ(size_t(0), v->errorCount());
+}
+
+TEST(Validator, validateUnitsOfCompenentsComplex)
+{
```
kerimoyle: validateUnitsOfComponentsComplex? spelling of "compenents" is not correct

tests/validator/validator.cpp
```diff
@@ -1878,6 +1882,7 @@ TEST(Validator, unitIllDefinedEquivalentUnits)
         "CellML identifiers must contain one or more basic Latin alphabetic characters.",
         "Variable 'v2' does not have a valid units attribute.",
         "Variable 'v1' has units of '' and an equivalent variable 'v2' with non-matching units of ''. The mismatch is: ",
+        "Variables 'v1' and 'v2' do not have the same unit reduction.",
```
kerimoyle: I think that if we're going to add a statement like this to the error message, it should ideally come before the detailed statement telling the same thing (which is above it).  It's probably also part of a wider improvement of our error messages, but returning a check for matching units when no units are specified doesn't really make sense either?  I'd rather have the validator stop checking at that point so that the user isn't given loads of messages when the only real problem is that they haven't specified any units?

kerimoyle:
Hi Andrew! Actually looking at this now I'm unsure why we need to test units twice?  There's already code (which you'll see from the error messages already in the tests) which tests for unit reduction between equivalent variables?  I would have thought that one or the other was needed, but I don't know what the difference is? I am coming into this a little late in the day so have probably missed something ...

awickens:
@kerimoyle Hi Keri, looking at it it does look like I've pretty much done the same thing twice (unlucky!). It was supposed to be related to unit equivalences in the May 2017 CellML 2.0 specification, in particular section 19.10.6 (stuff about map_variables reducing to identical units for all tuples). Worst comes to worst I can get rid of the stuff which I've unnecessarily redone, but I would still like to keep the extra unit tests which fixed the issues with `equivalence()` in the `Units` class.

awickens:
@hsorby @agarny @nickerso I'm pretty sure what I've done here is very similar (albeit slightly different and not as well done) to what has already been done before, especially looking at similarities between `checkUnitHomogeneity()` and `unitsAreEquivalent()` along with `validateConnections()`. I'm going to make the call to get rid of the validation functions I've written, but keep the `equivalence()` fixes which solve issue #484, if that's reasonable?

agarny:
I am not fully up-to-date (as neither is your branch for that matter :)), but I would say that it sounds reasonable. So, I imagine you are going to use Keri's implementation then?

awickens:
Yes, Keri's is much better as it actually tells you what the unit and multiplier mismatch is. Once I feel like I've made more progress on #497 I'll update the pull request and re-push.

awickens:
Ready for re-re-review.

nickerso (dismissed):
looks good assuming the CI passes everything when that eventually finishes...

agarny (approved):
Looks good to me.

src/units.cpp
```diff
@@ -431,7 +431,7 @@ void updateUnitsMap(const UnitsPtr &units, UnitsMap &unitsMap, double exp = 1.0)
     if (units->isBaseUnit()) {
         auto found = unitsMap.find(units->name());
         if (found == unitsMap.end()) {
-            unitsMap.emplace(units->name(), 1.0);
+            unitsMap.emplace(units->name(), exp);
```
kerimoyle, line 434: If we're dealing with a base unit then shouldn't the exponent be 1.0 all the time?  I'm not sure which way around is better ... ho hum.

kerimoyle (approved):
Looks fine to me.  If the tests pass them I'm happy :)

src/units.cpp
agarny: I guess it should be `1.0`, but to have `exp` is not only the same in this particular case, but it also makes things look "correct" (when looking at the code). So, I would leave it as-is.

PR #483, opened 2019-12-02
Model flattening for code generation.

hsorby:
Addresses #386.


hsorby:
Ready for review.

src/model.cpp
```diff
@@ -221,6 +222,78 @@ size_t Model::unitsCount() const
...
+                    model->addUnits(u);
+                    modelUnits = u;
+                }
+                v->setUnits(modelUnits);
+            }
+        }
+    }
+}
+
+void traversComponentTreeLinkingUnits(const ComponentPtr &component)
```
nickerso: I want to add an 'e' to the `travers`...

src/model.cpp
```diff
@@ -221,6 +222,78 @@ size_t Model::unitsCount() const
...
+        if (u != nullptr) {
+            auto model = owningModel(u);
+            unlinked = model == nullptr && !isStandardUnitName(u->name());
+        }
+    }
+
+    return unlinked;
+}
+
+bool traveseComponentTreeForUnlinkedUnits(const ComponentPtr &component)
```
nickerso: ok, now I'm starting to think these are typos that were not picked up in code-completion tools :)

tests/model/model_flattening.cpp
```diff
@@ -0,0 +1,623 @@
...
+
+    auto printer = libcellml::Printer::create();
+
+    auto a = printer->printModel(modelWithUnitsImports);
+    EXPECT_EQ(modelImportingUnits, a);
+}
+
+TEST(ModelFlattening, importedUnits)
+{
+    const std::string unitsDefiniitonModel =
```
nickerso: these typos should probably be fixed, but will be easier to do them all at once in a proper editor rather than online via Github.com :)

nickerso (dismissed):

agarny:
Not sure whether @hsorby is back to work, but before I can review this, a conflict should be resolved. I am happy to do it, if needed. @nickerso?

nickerso:
go for it @agarny - thanks!

src/model.cpp
agarny: Done.

src/model.cpp
agarny: Fixed too.

tests/model/model_flattening.cpp
agarny: Done.

tests/model/model_flattening.cpp
```diff
@@ -0,0 +1,623 @@
...
+
+    model->flatten();
+
+    auto printer = libcellml::Printer::create();
+
+    auto a = printer->printModel(model);
+    auto modelNonImportVersion = parser->parseModel(fileContents("generator/hodgkin_huxley_squid_axon_model_1952/model.cellml"));
+    auto e = printer->printModel(modelNonImportVersion);
+    // Sadly cannot get the order of the connections to match each other, everything else does.
+    // EXPECT_EQ(e, a);
```
agarny: I am going to reenable that check since we clearly want it. I am just going to double check the "original" HH52 model and modify accordingly, if it is acceptable.

agarny: Done.

tests/model/model_flattening.cpp
```diff
@@ -0,0 +1,623 @@
...
+        "    <variable name=\"x\" units=\"dimensionless\" interface=\"public_and_private\"/>\n"
+        "    <variable id=\"sin\" units=\"dimensionless\" name=\"sin\" interface=\"public_and_private\"/>\n"
+        "    <math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
+        "  <apply id=\"actual_sin\"><eq/>\n"
+        "    <ci>sin</ci>\n"
+        "    <apply><sin/>\n"
+        "      <ci>x</ci>\n"
+        "    </apply>\n"
+        "  </apply>\n"
+        "</math>\n"
```
agarny: This is not "nice". I would expect the indentation to be right. Will try to "fix" it.

tests/units/units.cpp
```diff
@@ -345,6 +345,173 @@ TEST(Units, replaceUnits)
...
+    c1->addVariable(v1);
+    m->addComponent(c1);
+
+    // Adding units to model after adding to a variable
+    // means the units are not linked.
+    v1->setUnits("a_unit");
+    m->addUnits(u1);
+
+    EXPECT_TRUE(m->hasUnlinkedUnits());
+}
```
agarny, line 445: Not sure about the need of this test. It's the same as `hasUnlinkedUnitsWhenBaseUnitsAddedToModelAfterNameUsedForVariableUnits` except that we add `second` to `a_unit`, but why should that make things different when it comes to `hasUnlinkedUnits()`? What am I missing?...

nickerso: Isn't it to test that there is no special treatment of base units that is affecting this behaviour? (as we've found out in other aspects of units)

tests/units/units.cpp
```diff
@@ -345,6 +345,173 @@ TEST(Units, replaceUnits)
...
+
+    v4->setUnits("a_unit");
+
+    m->addUnits(u1);
+    v1->setUnits("a_unit");
+    v2->setUnits(u1);
+    v3->setUnits(u1);
+
+    EXPECT_TRUE(m->hasUnlinkedUnits());
+}
```
agarny, line 513: Again, I am not quite sure why the `hasUnlinkedUnitsLayeredComponentsWithVariables` and `hasUnlinkedUnitsSiblingComponentsWithVariables` are needed. I feel like what they are testing is already tested in previous tests...?

tests/units/units.cpp
agarny: Hmm... not convinced. We have `v1->setUnits("a_unit")`, which in itself results in an unlinked unit since `u1` is only added to the model after setting the units to `v1`. Anyway, I am not going to have this hold back the PR. I just thought I would mention my "concerns" so that @hsorby can look into it when he's back.

Right now, the only thing that I want before accepting this PR is to fix the indentation of the math, i.e. https://github.com/cellml/libcellml/pull/483#discussion_r363150645.

agarny:
At this stage, the only thing that is left for me (besides my "concerns" about unnecessary tests, which I am not going to hold against this PR) is to fix the math indentation issue, if possible (see https://github.com/cellml/libcellml/pull/483#discussion_r363150645).

src/model.cpp
```diff
@@ -472,4 +577,208 @@ ModelPtr Model::clone() const
...
+            if (!rebasedVector.empty()) {
+                rebasedMap[rebasedKey] = rebasedVector;
+            }
+        }
+    }
+
+    return rebasedMap;
+}
+
+void getComponentNames(const ComponentPtr &component, NameList &names)
```
kerimoyle: Since this returns the component names, should it be consistent with function names elsewhere and ditch the "get"?

src/model.cpp
```diff
@@ -472,4 +577,208 @@ ModelPtr Model::clone() const
...
+    flattenComponent(parent, component, componentIndex);
+
+    for (size_t index = 0; index < component->componentCount(); ++index) {
+        auto c = component->component(index);
+        flattenComponentTree(component, c, index);
+    }
+}
+
+void Model::flatten()
+{
```
kerimoyle, line 759: Again for consistency ... should this be flattenModel() (like validateModel and parseModel ?)

kerimoyle:
Do we expect users to be allowed to import two separate (and individually invalid) models and have them combine to something sensible?  For example, one model has all the units, another model has all the variables to which the units are linked by name.  Would we expect the combined model to report unlinked units or not?

src/model.cpp
agarny: Although it's an internal method, I agree. Done.

src/model.cpp
agarny: Hmm, indeed. This means that we also need to rename `Model::clone()` to `Model::cloneModel()`. Done.

nickerso: this is a method on `Model`, so I don't think it needs to have the model, that is by definition what is being flattened or cloned.

agarny: Oops, I agree with @nickerso. For the `Validator` and `Parser` classes, it's a different story since we are indeed validating/parsin a model, not a validator/parser.

agarny:
> Do we expect users to be allowed to import two separate (and individually invalid) models and have them combine to something sensible? For example, one model has all the units, another model has all the variables to which the units are linked by name. Would we expect the combined model to report unlinked units or not?

I would say that there is nothing that prevents us from importing two separate (and individually invalid) models. Then, sure, this may result in unlinked units, but it shouldn't if you flatten your model (unless your combined model is also invalid).

tests/model/model_flattening.cpp
agarny: Done.

agarny (approved):
Looks good to me apart from a few tests that I am not sure are needed. I guess I must be missing something, so not holding it against this PR.

nickerso (approved):

Issue #436, opened 2019-10-22
Equivalent variables not validated properly


kerimoyle:
Here's a test showing that we don't check for invalid connections between equivalent variables or for interfaces:

```
TEST(Validator, variableEquivalenceErrorsNotRaised)
{
    // Create a model with three components and one variable in each to be shared between them
    libcellml::ModelPtr model = std::make_shared<libcellml::Model>();
    libcellml::ComponentPtr c1 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr c2 = std::make_shared<libcellml::Component>();
    libcellml::ComponentPtr c3 = std::make_shared<libcellml::Component>();

    libcellml::Validator validator;

    model->setName("model");
    c1->setName("c1");
    c2->setName("c2");
    c3->setName("c3");

    // Two siblings in the model, one child in component2
    model->addComponent(c1);
    model->addComponent(c2);
    c2->addComponent(c3);

    libcellml::VariablePtr v1 = std::make_shared<libcellml::Variable>();
    v1->setName("v1");
    v1->setUnits("dimensionless");

    libcellml::VariablePtr v2 = std::make_shared<libcellml::Variable>();
    v2->setName("v2");
    v2->setUnits("dimensionless");

    libcellml::VariablePtr v3 = std::make_shared<libcellml::Variable>();
    v3->setName("v3");
    v3->setUnits("dimensionless");

    c1->addVariable(v1);
    c2->addVariable(v2);
    c3->addVariable(v3);

    libcellml::Variable::addEquivalence(v1, v3); // invalid equivalence, too far away
    validator.validateModel(model);
    // expect errors saying that components are too distant for variable equivalence, and no specification of interface type
    EXPECT_FALSE(validator.errorCount() == size_t(0));

    // Change to a valid configuration
    libcellml::Variable::removeEquivalence(v1, v3);
    libcellml::Variable::addEquivalence(v1, v2);
    libcellml::Variable::addEquivalence(v2, v3);

    // Now expect errors only related to the interface type missing
    validator.validateModel(model);
    EXPECT_FALSE(validator.errorCount() == size_t(0));

    // Add the interface types and expect no errors
    v1->setInterfaceType("public_and_private");
    v2->setInterfaceType("public_and_private");
    v3->setInterfaceType("public_and_private");

    validator.validateModel(model);
    EXPECT_EQ(size_t(0), validator.errorCount());
}
```

hsorby:
[...]

PR #485, opened 2019-12-04
Start validating variable interfaces derived from equivalences.

hsorby:
Addresses #436.


hsorby:
Ready for review.

src/utilities.cpp
```diff
@@ -457,4 +457,37 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
...
+}
+
+bool areEntitiesSiblings(const EntityPtr &entity1, const EntityPtr &entity2)
+{
+    auto entity1Parent = entity1->parent();
+    return entity1Parent != nullptr && entity1Parent == entity2->parent();
+}
+
+InterfaceTypePair determineInterfaceType(const VariablePtr &variable1, const VariablePtr &variable2)
+{
```
kerimoyle, line 539: should this function name just be interfaceType() to follow our other get/set format?

hsorby: It is not getting an interface type it is working out what the interface type for both variables should be.

src/validator.cpp
```diff
@@ -949,6 +964,50 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+            component2Name = component2->name();
+        }
+        ErrorPtr err = Error::create();
+        err->setDescription("Invalid equivalence between '" + variable1->name() + "' and '" + variable2->name() + "', the separation of components '" + component1Name + "' and '" + component2Name + "' is too great.");
+        err->setVariable(variable1);
+        err->setKind(Error::Kind::CONNECTION);
+        mValidator->addError(err);
+    } else {
+        if (!validateVariableInterface(variable1, pair.first)) {
+            ErrorPtr err = Error::create();
```
kerimoyle: Can we add a "I expected (something) or (something) but I found (something else)" end to these statements? Makes it more useful to the user

tests/variable/variable.cpp
```diff
@@ -1165,3 +1165,213 @@ TEST(Variable, modelUnitsAttributeBeforeNameAttribute)
...
+    v3->setUnits("dimensionless");
+
+    c1->addVariable(v1);
+    c2->addVariable(v2);
+    c3->addVariable(v3);
+
+    libcellml::Variable::addEquivalence(v1, v2);
+    libcellml::Variable::addEquivalence(v2, v3);
+
+    EXPECT_EQ("public", v1->interfaceType());
```
kerimoyle, line 1470: See comment on #436 - I'd expect this to fail validation as there is no interface type set, so should default to None?

tests/variable/variable.cpp
```diff
@@ -1165,3 +1165,213 @@ TEST(Variable, modelUnitsAttributeBeforeNameAttribute)
...
+
+    c1->addVariable(v1);
+    c2->addVariable(v2);
+
+    v2->setInterfaceType("private");
+
+    libcellml::Variable::addEquivalence(v1, v2);
+
+    EXPECT_EQ("private", v1->interfaceType());
+    EXPECT_EQ("public_and_private", v2->interfaceType());
```
kerimoyle: Ditto here - we're changing what the user has specified by assuming we know better?  I think that's dangerous ... ?

hsorby:
Addressed review comments and is ready for re-review.

hsorby:
Ready for re-review.

hsorby:
This could do with another review.

src/utilities.h
```diff
@@ -197,6 +198,21 @@ const std::vector<std::string> supportedMathMLElements = {
...
+/**
+ * @brief Map to convert an interface type into its string form.
+ *
+ * An internal map used to convert a Variable InterfaceType enum class member into its string form.
+ */
+static const std::map<Variable::InterfaceType, const std::string> interfaceTypeToString = {
+    {Variable::InterfaceType::NONE, "none"},
+    {Variable::InterfaceType::PRIVATE, "private"},
+    {Variable::InterfaceType::PUBLIC, "public"},
+    {Variable::InterfaceType::PUBLIC_AND_PRIVATE, "public_and_private"}};
```
kerimoyle, line 212: Is there a reason that the strings shouldn't be camel case, as elsewhere?

src/validator.cpp
```diff
@@ -952,63 +962,131 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
-                }
+void Validator::ValidatorImpl::validateEquivalenceStructure(const VariablePtr &variable)
+{
+    for (size_t index = 0; index < variable->equivalentVariableCount(); ++index) {
+        auto equivalentVariable = variable->equivalentVariable(index);
+        if (equivalentVariable->hasEquivalentVariable(variable)) {
+            auto component = std::dynamic_pointer_cast<Component>(equivalentVariable->parent());
+            if (component == nullptr) {
+                ErrorPtr err = Error::create();
+                err->setDescription("Variable '" + equivalentVariable->name() + "' is an equivalent variable to '" + variable->name() + "' but '" + equivalentVariable->name() + "' has no parent component.");
```
kerimoyle, line 1093: Need to add component names to the variable (where poss) so that it's easier for the user to find later

src/validator.cpp
```diff
@@ -1047,11 +1125,11 @@ bool Validator::ValidatorImpl::isCellmlIdentifier(const std::string &name)
...
-                                                  const VariablePtr &v1,
-                                                  const VariablePtr &v2,
-                                                  std::string &hints,
-                                                  double &multiplier)
+bool unitsAreEquivalent(const ModelPtr &model,
+                        const VariablePtr &v1,
+                        const VariablePtr &v2,
+                        std::string &hints,
+                        double &multiplier)
 {
```
kerimoyle, line 1163: Probably not relevant to this PR, but what do you think about (later) removing these functions from the validator in favour of the equivalent() and compatible() ones in #554 ?

src/variable.cpp
```diff
@@ -478,6 +467,14 @@ void Variable::removeInterfaceType()
     mPimpl->mInterfaceType.clear();
 }
 
+bool Variable::hasInterfaceType(InterfaceType interfaceType) const
+{
+    if (interfaceType == Variable::InterfaceType::NONE && mPimpl->mInterfaceType.empty()) {
+        return true;
+    }
+    return mPimpl->mInterfaceType == interfaceTypeToString.find(interfaceType)->second;
```
kerimoyle, line 475: I don't understand why we have an `enum` for the `interfaceType` if we store it as a string in the variable?  Why not go one way or the other?

tests/validator/validator.cpp
```diff
@@ -836,62 +836,6 @@ TEST(Validator, validateInvalidConnectionsVariableWithoutParentComponent)
     EXPECT_EQ_ERRORS(expectedErrors, v);
```
kerimoyle, line 880: Should all of these `expectedError` strings be renamed to `e` as in earlier PR?

kerimoyle (changes requested):

kerimoyle:
In general I'm really happy to see this go through (especially that the user has control over what they fix and when :) ), but I'm curious about why we have an enum for interface types when all the work (setting, getting, comparison etc) is done through strings?  We don't allow this approach elsewhere (types of issues, generator profiles, etc) so I'm not sure why this case is different?

src/validator.cpp
hsorby: Sounds like a great idea.

src/utilities.h
hsorby: Because they match what the spec. says.

src/validator.cpp
hsorby: I think we have an issue for going over all descriptions don't we?

tests/validator/validator.cpp
hsorby: The tests don't need to be identical in their coding style for me.

hsorby:
Allowing strings opens up a whole new ball game of things to watch out for whereas enums reduce the field of what is allowed.

Enums can also be quite convenient for application developers.

src/variable.cpp
hsorby: I don't have too much of a problem in offering both here.

tests/variable/variable.cpp
hsorby: I think this has now been fixed so it doesn't happen anymore.

tests/validator/validator.cpp
agarny: And that's why I hate it: it's a mess.

hsorby: It needs to be different so you can try doing things differently, if everything is done the same you are not testing over a very wide base.

agarny: We are talking about variable names here...

hsorby: I thought we were talking about all things to do with test coding style.  I don't have a problem with changing variable names.

agarny: Re coding style in general, it's also a mess for me and I don't see how how it helps with testing over a wider base...? Anyway, those are "only" tests, I guess...

src/utilities.h
kerimoyle: Makes sense :)

kerimoyle:
> Allowing strings opens up a whole new ball game of things to watch out for whereas enums reduce the field of what is allowed.
Enums can also be quite convenient for application developers.

Yes, that's kind of my point ... why are we passing strings around the place here instead of using only enums (as we do in other parts, code generation, issues etc?)

src/api/libcellml/model.h
```diff
@@ -347,13 +347,14 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
+     * set correctly then false is returned.
      *
-     * If the interface type for a variable cannot be determined it is left unchanged.
+     * If the interface type for a variable cannot be set correctly it is left unchanged.
      *
-     * @return True if the @c Model has had a variable interface changed,
-     * false otherwise.
+     * @return @c true if all the variable interfaces in the @c Model are correct and
+     * @c false otherwise.
      */
```
kerimoyle, line 358: I like this behaviour a lot more, makes more sense to the end user.

src/variable.cpp
kerimoyle: Why forbid it everywhere else then?  eg: generator profiles, issue kinds etc.  Should we add string overloads to those functions too?

kerimoyle:
Random thought:  Since we now have a function `fixVariableInterfaces` (based on equivalent variables), do we need the other direction too?  eg: `fixEquivalentVariables` (based on interfaces)?

hsorby:
> Random thought: Since we now have a function `fixVariableInterfaces` (based on equivalent variables), do we need the other direction too? eg: `fixEquivalentVariables` (based on interfaces)?

I don't think you can create equivalent variables based on the interface type also I haven't seen a need for this so I would say no we don't need this.

src/variable.cpp
hsorby: No, we shouldn't.  If anything we should remove the string from here.  This was a very early piece of code used for the very start where we were printing models hence why it is this way

src/validator.cpp
hsorby: Leaving this for the whole error message sweep.

nickerso (dismissed):

src/api/libcellml/variable.h
```diff
@@ -418,6 +418,17 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
      */
     void removeInterfaceType();
 
+    /**
+     * @brief Test if this variable has the given interface type.
+     *
+     * Test if this variable has the smae interface as @p interfaceType.
```
kerimoyle: *same

src/api/libcellml/variable.h
```diff
@@ -418,6 +418,17 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
+     *
+     * @param interfaceType The interface type to test for.
+     *
+     * @return True if the interface type is the same as the given interface type, false otherwise.
```
kerimoyle: should this be @c true and @c false?

src/bindings/interface/variable.i
```diff
@@ -52,6 +52,9 @@ reference.";
 %feature("docstring") libcellml::Variable::interfaceType
 "Returns this variable's interface type as string.";
```
kerimoyle: Return (not returns)

src/utilities.cpp
```diff
@@ -463,4 +463,99 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
     return index;
 }
 
+bool isEntityChildOf(const EntityPtr &entity1, const EntityPtr &entity2)
+{
```
kerimoyle, line 467: Is there ever a situation when entity1 is nullptr? If so this will segfault?

src/utilities.cpp
```diff
@@ -463,4 +463,99 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
...
+    return entity1->parent() == entity2;
+}
+
+bool areEntitiesSiblings(const EntityPtr &entity1, const EntityPtr &entity2)
+{
+    auto entity1Parent = entity1->parent();
```
kerimoyle, line 473: ditto here

src/utilities.cpp
```diff
@@ -463,4 +463,99 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
...
+    return entity1Parent != nullptr && entity1Parent == entity2->parent();
+}
+
+using PublicPrivateRequiredPair = std::pair<bool, bool>;
+
+/**
+ * @brief Determine whether a public and/or private interface is required for the givn @p variable.
+ *
```
kerimoyle, line 481: *given not givn

src/utilities.cpp
```diff
@@ -463,4 +463,99 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
...
+ * Determine whether a public and/or private interface is required for the givn @p variable.  Returns
+ * a pair of booleans where the first item in the pair indicates that a public interface is required,
```
kerimoyle: indicates *whether* a public interface is required?  (ditto below)

src/utilities.cpp
```diff
@@ -463,4 +463,99 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
...
+ * the second item in the pair indicates that a private interface is required, and if both are items
+ * in the pair are false then this indicates an error has occured and the interface type cannot be
```
kerimoyle: @c false

src/api/libcellml/model.h
```diff
@@ -343,6 +343,21 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
     void flatten();
 
+    /**
+     * @brief Fix @c Variable interfaces throughout the model.
+     *
+     * Traverses the model investigating variable equivalences to set the appropriate
+     * interface type.  If all equivalent variables have the correct interface set then
+     * the method returns true.  If one or more equivalent variable interface cannot be
```
kerimoyle: @c true and @c false?

src/utilities.cpp
```diff
@@ -463,4 +463,99 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
...
+            return std::make_pair(false, false);
+        }
+    }
+    return pair;
+}
+
+/**
+ * @brief Get the interface type for the given public private pair.
+ *
+ * Get the interface type for the @p pair.  The default return type is
```
kerimoyle, line 517: libcellml::Variable::InterfaceType::NONE

src/utilities.cpp
```diff
@@ -463,4 +463,99 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
...
+ * NONE.
+ *
+ * @param pair The pair to determine the interface type for.
```
kerimoyle: The pair for which the interface type is determined

src/utilities.h
```diff
@@ -432,6 +448,57 @@ bool isStandardPrefixName(const std::string &name);
...
+ */
+bool isEntityChildOf(const EntityPtr &entity1, const EntityPtr &entity2);
+
+/**
+ * @brief Test to determine if @p entity1 and @p entity2 are siblings.
+ *
+ * Test to determine if @p entity1 and @p entity2 are siblings.  Returns
+ * @c true if @p entity1 and @p entity2 are siblings, @c false otherwise.
+ *
+ * @param entity1 An @c Entity to test if it is a sibling to @p entity2.
```
kerimoyle, line 468: An @c Entity to test for relationship to @p entity2 ?

src/utilities.h
```diff
@@ -432,6 +448,57 @@ bool isStandardPrefixName(const std::string &name);
...
+ * @param entity2 An @c Entity to test if it is a sibling to @p entity1.
+ *
+ * @return @c true if @p entity1 and @p entity2 are siblings, @c false otherwise.
+ */
+bool areEntitiesSiblings(const EntityPtr &entity1, const EntityPtr &entity2);
+
+/**
+ * @brief Determine the interface type of the @p variable.
+ *
+ * Determines the interface type of the given @p variable.  If the variable
```
kerimoyle: Determine (not determines)

src/utilities.h
```diff
@@ -432,6 +448,57 @@ bool isStandardPrefixName(const std::string &name);
...
+ * type is determined to be NONE this indicates an error (if the variable has
+ * at least one equivalent variable).
+ *
```
kerimoyle, line 481: For variables with at least one equivalent variable, returning an **interface** type of NONE indicates an error.

src/utilities.h
```diff
@@ -432,6 +448,57 @@ bool isStandardPrefixName(const std::string &name);
...
+ * @param variable The variable to determine the interface type for.
+ *
+ * @return The @p variable's interface type.
+ */
+Variable::InterfaceType determineInterfaceType(const VariablePtr &variable);
+
+/**
+ * @brief Traverses the component tree looking for variables with equivalences.
```
kerimoyle: Traverse (not traverses) and Search (below) cf searches

src/utilities.h
```diff
@@ -432,6 +448,57 @@ bool isStandardPrefixName(const std::string &name);
...
+ *
+ * Searches through the component tree starting at @p component looking for variables
+ * with equivalences.  Variables found in the component tree with equivalences are added
+ * to the @p variables list.
+ *
+ * @param component The @c Component to search.
+ * @param variables The list of equivalent variables found.
```
kerimoyle, line 496: should "list" say something about the type?  or it's generic enough? (or maybe I'm just getting confused because list means something specific?)

src/validator.cpp
```diff
@@ -35,6 +35,38 @@ limitations under the License.
 
 namespace libcellml {
 
+/**
+* @brief Validate that equivalent variable pairs in the @p model
```
kerimoyle, line 41: Model cf model? (not sure ...)

src/validator.cpp
```diff
@@ -35,6 +35,38 @@ limitations under the License.
...
+* have equivalent units.
+* Any errors will be logged in the @c Validator.
+*
+* Any difference in base units is reported as an error in the @c Validator, but the multiplier difference does not trigger a validator error.
+* Where the base units are equivalent, the multiplier may be interpreted as units_of_v1 = (10^multiplier)*units_of_v2
```
kerimoyle, line 46: I think we should change this to be consistent with other multipliers - ie: return the actual multiplier, not the log10 of it?

kerimoyle: And also the order - in the Units file it's units_of_v2 = units_of_v1*multiplier ?

src/validator.cpp
```diff
@@ -35,6 +35,38 @@ limitations under the License.
...
+*
+* @param model The model containing the variables
```
kerimoyle: fullstop

src/validator.cpp
kerimoyle: On second thoughts, if this isn't shown to users then I'm happy enough

src/validator.cpp
```diff
@@ -35,6 +35,38 @@ limitations under the License.
...
+*/
+bool unitsAreEquivalent(const ModelPtr &model, const VariablePtr &v1, const VariablePtr &v2, std::string &hints, double &multiplier);
+
+/**
+* @brief Utility function used by unitsAreEquivalent to compare base units of two variables.
+*
+* @param model The model containing the variables.
+* @param unitMap A list of the exponents of base variables.
+* @param uName String name of the current variable being investigated.
+* @param standardList Nested map of the conversion between built-in units and the base units they contain
```
kerimoyle: fullstop

src/validator.cpp
```diff
@@ -952,63 +962,131 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
     }
 }
 
-void Validator::ValidatorImpl::validateConnections(const ModelPtr &model)
+/**
+ * @brief Test to see if the given variables are reachable from their parent components.
+ *
+ * Determine if the variables parents are reachable from one another.  That is they either
+ * have a parent/child relationship or a sibling relationship in the models component
```
kerimoyle: model's

src/validator.cpp
```diff
@@ -952,63 +962,131 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+ * hierarchy.
+ *
+ * Both variables must have a valid parent.
```
kerimoyle, line 995: Is it enough to just say it here? you don't need to test it?

src/validator.cpp
```diff
@@ -952,63 +962,131 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+ *
+ * @param variable1 The first variable.
+ * @param variable2 The second variable.
+ *
+ * @return True if the parents of the given variables are reachable in the model component
```
kerimoyle: @c true

tests/connection/connection.cpp
```diff
@@ -18,6 +18,31 @@ limitations under the License.
...
+    libcellml::Variable::addEquivalence(v1, v2);
+    EXPECT_EQ(size_t(0), v1->equivalentVariableCount());
+    EXPECT_FALSE(v1->hasEquivalentVariable(v2));
+}
+
+TEST(Variable, addEquivalenceNullptrBothParameters)
+{
+    libcellml::VariablePtr v1 = nullptr;
+    libcellml::VariablePtr v2 = nullptr;
+    libcellml::Variable::addEquivalence(v1, v2);
```
kerimoyle, line 43: there's nothing being tested here?

tests/validator/validator.cpp
```diff
@@ -1522,7 +1466,7 @@ TEST(Validator, validMathCnElementsMissingCellMLNamespace)
         "  </apply>\n"
         "</math>\n";
 
-    const std::vector<std::string> expectedErrors {
+    const std::vector<std::string> expectedErrors = {
```
kerimoyle, line 1469: is there a convention around initialising through the constructor cf through assignment?  in an earlier PR the equals were all removed ...?

tests/validator/validator.cpp
```diff
@@ -2395,6 +2357,333 @@ TEST(Validator, unfoundUnitsInEncapsulatedComponents)
...
+
+    validator->validateModel(model);
+
+    EXPECT_EQ_ERRORS(expectedErrors, validator);
+}
+
+TEST(Validator, variableInterfaceShouldBePublic)
+{
+    const std::vector<std::string> expectedErrors = {
+        "Variable 'v1' has no interface type set. The interface type required is 'public'.",
```
kerimoyle: Can we add components to the error message? there might be a whole load of 'v1's in the model?

kerimoyle (changes requested):
Mostly typos in comments, and it would be good to get the component names into the error strings here too.

src/api/libcellml/model.h
```diff
@@ -343,6 +343,21 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
+     * set correctly then false is returned.
+     *
+     * If the interface type for a variable cannot be set correctly it is left unchanged.
```
agarny: `... correctly[,] it...`

src/utilities.cpp
```diff
@@ -463,4 +463,99 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
...
+    auto entity1Parent = entity1->parent();
+    return entity1Parent != nullptr && entity1Parent == entity2->parent();
+}
+
+using PublicPrivateRequiredPair = std::pair<bool, bool>;
+
+/**
+ * @brief Determine whether a public and/or private interface is required for the givn @p variable.
+ *
+ * Determine whether a public and/or private interface is required for the givn @p variable.  Returns
```
agarny: *given not givn

src/utilities.cpp
```diff
@@ -463,4 +463,99 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
...
+ * a pair of booleans where the first item in the pair indicates that a public interface is required,
+ * the second item in the pair indicates that a private interface is required, and if both are items
```
agarny: `... if both items...`, i.e. delete `are`.

src/utilities.cpp
```diff
@@ -463,4 +463,99 @@ size_t getVariableIndexInComponent(const ComponentPtr &component, const Variable
...
+        auto equivalentVariable = variable->equivalentVariable(index);
+        auto componentOfVariable = variable->parent();
+        auto componentOfEquivalentVariable = equivalentVariable->parent();
+        if (componentOfVariable == nullptr || componentOfEquivalentVariable == nullptr) {
+            return std::make_pair(false, false);
+        }
+        if (areEntitiesSiblings(componentOfVariable, componentOfEquivalentVariable)) {
+            pair.first = true;
+        } else if (isEntityChildOf(componentOfVariable, componentOfEquivalentVariable)) {
+            pair.first = true;
```
agarny: Merge these two tests since they both set `pair.first` to `true`...?

tests/connection/connection.cpp
hsorby: Yes there is, it is testing that it doesn't blow up.

tests/validator/validator.cpp
hsorby: I don't remember which PR that was.  I noticed there was a difference in how they we declared and made them all the same, went with the majority format.

tests/validator/validator.cpp
hsorby: Leaving this for the error message sweep.

src/utilities.cpp
hsorby: Nope.

src/utilities.cpp
hsorby: Nope.

src/bindings/interface/variable.i
hsorby: I really think it should be 'Returns' this is what the majority of Python doc strings use.

src/validator.cpp
hsorby: That is why it is said because it is not being tested.

src/utilities.cpp
hsorby: I considered doing this but thought it was clearer and cleaner to separate it out.

agarny: Not sure it is, but it's not critical, so I won't insist. :)

tests/validator/validator.cpp
```diff
@@ -2395,6 +2357,333 @@ TEST(Validator, unfoundUnitsInEncapsulatedComponents)
...
+    libcellml::Variable::addEquivalence(v1, v3);
+    validator->validateModel(model);
+
+    EXPECT_EQ_ERRORS(expectedErrors, validator);
+}
+
+TEST(Validator, variableEquivalenceUnreachableAndReachableTogether)
+{
+    const std::vector<std::string> expectedErrors = {
+        "Variable 'v1' has an interface type set to 'public' which is not the correct interface type for this variable. The interface type required is 'private'.",
```
agarny: As an end-user, I have no idea what I could do with such an error message. I imagine that it's indirectly related to trying to create an equivalence between `v1` and `v2`, and then between `v2` and `v3`? Indeed, for those equivalences to work, I would expect `v1` to have a private interface.

tests/validator/validator.cpp
```diff
@@ -2395,6 +2357,333 @@ TEST(Validator, unfoundUnitsInEncapsulatedComponents)
...
+        "The equivalence between 'v2' in component 'c2'  and 'v4' in component 'c4' is invalid. Component 'c2' and 'c4' are neither siblings nor in a parent/child relationship.",
+        "The equivalence between 'v4' in component 'c4'  and 'v2' in component 'c2' is invalid. Component 'c4' and 'c2' are neither siblings nor in a parent/child relationship.",
```
agarny: Those two messages are effectively the same, so I really wish we would only generate one of them. To have two will only confuse people.

tests/validator/validator.cpp
hsorby: Seems pretty clear to me. `v1` has interface type `public` but it should be `private`.  I can only take away that the variable `v1` interface type should be set to `private`.

agarny: Ok, sure, the message tells me what needs to be done to fix the problem, but that's the thing: I can't tell from the message what the problem is? Maybe some people don't care, but as an end-user I do or how can I avoid that problem in the future?

As I said earlier, and as far as I can tell, the problem is because of trying to create an equivalence between `v1` and `v2` and then between `v2` and `v3`. So, I would expect two error messages that reference those equivalences and then the fact that they are possible because of `v1` having a public interface rather than a private interface.

tests/validator/validator.cpp
```diff
@@ -2395,6 +2357,333 @@ TEST(Validator, unfoundUnitsInEncapsulatedComponents)
...
+    validator->validateModel(model);
+
+    EXPECT_EQ_ERRORS(expectedErrors, validator);
+}
+
+TEST(Validator, variableInterfaceShouldBePublic)
+{
+    const std::vector<std::string> expectedErrors = {
+        "Variable 'v1' has no interface type set. The interface type required is 'public'.",
+        "Variable 'v2' has an interface type set to 'private' which is not the correct interface type for this variable. The interface type required is 'public'.",
```
agarny: Other cases of where I wish we would also be told about the problem, not only its solution.

tests/validator/validator.cpp
hsorby: This error has nothing to do with the equivalence between `v2` and `v3` and only to do with `v1` and `v2`.

I can only say that you would have to read the spec. for understanding.  I don't think it is desirable to say go and read the spec. my compiler doesn't tell me to go and read the standard every time I get something wrong.  I either get told that that is an error or that is an  error *did you mean ...*.

agarny: > This error has nothing to do with the equivalence between `v2` and `v3` and only to do with `v1` and `v2`.

It indirectly has something to do with `v2` and `v3` since we need to go through `v1` and we can only do that if the interface is `private`.

> I can only say that you would have to read the spec. for understanding. I don't think it is desirable to say go and read the spec. my compiler doesn't tell me to go and read the standard every time I get something wrong. I either get told that that is an error or that is an error _did you mean ..._.

I did say what I would expect to see as an error, but we clearly seem to disagree. So be it, I will keep reviewing, but will leave it to others to approve. :)

tests/variable/variable.cpp
```diff
@@ -429,6 +429,30 @@ TEST(Variable, setGetInterfaceType)
     EXPECT_EQ(interfaceTypeString4, v4->interfaceType());
 }
 
+TEST(Variable, hasInterfaceType)
```
agarny, line 432: Might want to test against the `none`, `public`, `private` and `public_and_private` interfaces after each call to `setInterfaceType()`/`removeInterfaceType()`?

tests/variable/variable.cpp
```diff
@@ -1405,3 +1429,271 @@ TEST(Variable, hasEquivalentVariableWithNullptr)
...
+    c1->addVariable(v1);
+    c2->addVariable(v2);
+    c3->addVariable(v3);
+
+    libcellml::Variable::addEquivalence(v1, v3);
+
+    EXPECT_FALSE(model->fixVariableInterfaces());
+
+    EXPECT_EQ("", v1->interfaceType());
+    EXPECT_EQ("", v3->interfaceType());
```
agarny, line 1620: I imagine that `fixVariableInterfaces()` couldn't set the variable interfaces because `v1` and `v3` are too far away? If so, we might just want to add a comment to that effect?

src/validator.cpp
```diff
@@ -952,63 +962,131 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+    if (isEntityChildOf(parent1, parent2)) {
+        reachable = true;
+    } else if (isEntityChildOf(parent2, parent1)) {
+        reachable = true;
+    } else if (areEntitiesSiblings(parent1, parent2)) {
+        reachable = true;
+    }
+
+    return reachable;
+}
```
agarny: I find this kind of code unnecessarily complicated to read. I would simply have:

```
bool reachableEquivalence(const VariablePtr &variable1, const VariablePtr &variable2)
{
    auto parent1 = variable1->parent();
    auto parent2 = variable2->parent();

    return isEntityChildOf(parent1, parent2)
           || isEntityChildOf(parent2, parent1)
           || areEntitiesSiblings(parent1, parent2);
}
```

src/utilities.h
hsorby: Seems right.

src/utilities.h
hsorby: I think it is describing the parameter sufficiently well.

tests/variable/variable.cpp
hsorby: I think making sure the previous is gone and the new is in is sufficient here.

nickerso (dismissed):

src/validator.cpp
```diff
@@ -1035,20 +1057,26 @@ void Validator::ValidatorImpl::validateVariableInterface(const VariablePtr &vari
...
+            auto it = std::find(alreadyReported.begin(), alreadyReported.end(), reversePair);
+            if (it == alreadyReported.end()) {
+                VariablePair pair = std::make_pair(variable, equivalentVariable);
+                alreadyReported.push_back(pair);
+                auto unitsName = variable->units() == nullptr ? "" : variable->units()->name();
+                auto equivalentUnitsName = equivalentVariable->units() == nullptr ? "" : equivalentVariable->units()->name();
+                ErrorPtr err = Error::create();
+                err->setDescription("Variable '" + variable->name() + "' has units of '" + unitsName + "' and an equivalent variable '" + equivalentVariable->name() + "' with non-matching units of '" + equivalentUnitsName + "'. The mismatch is: " + hints);
+                err->setModel(model);
+                err->setKind(Error::Kind::UNITS);
```
kerimoyle, line 1077: Do we need a better error code that just UNITS here?  The stuff under that spec heading has no relevance to this situation ...?

tests/validator/validator.cpp
kerimoyle: I think that (soon!) having a better place to point people towards will help here - that way they can read all about the problem and its solutions without having to have enormously long error messages returned.

src/validator.cpp
hsorby: This one comes under "interpretation of map_variables" I think.

kerimoyle:
I'm happy with this now, thanks @hsorby.

kerimoyle (approved):

tests/validator/validator.cpp
kerimoyle: We can add more information to the informative spec around the causes and reasons - it will save having enormously long error messages returned from the API.

nickerso (approved):

Issue #478, opened 2019-11-24
isBaseUnits doesn't consider imports


nickerso:
Currently `Units::isBaseUnits()` only checks for no child units to determine if a units is a base units. This fails when the units is an imported units. The check needs to traverse imports to determine if the units is a base units, or return `false` if unable to make that determination (e.g., imports are not resolved).

PR #488, opened 2019-12-05
Re-implementation of isBaseUnit to consider imports

awickens:
Changes which work towards addressing #478.

### DIFF
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -217,8 +217,8 @@ bool Units::isBaseUnit() const
         ImportSourcePtr importedSource = importSource();
         if (importedSource != nullptr) {
             ModelPtr model = importedSource->model();
-            if (model != nullptr && model->hasUnits(name())) {
-                auto unit = model->units(name());
+            if (model != nullptr && model->hasUnits(importReference())) {
+                auto unit = model->units(importReference());
                 return unit->unitCount() == 0;
             }
         }
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -213,13 +213,12 @@ UnitsPtr Units::create(const std::string &name) noexcept
 bool Units::isBaseUnit() const
 {
     if (isImport()) {
-        // If our model is unresolved or our import source is unresolved then return false
         ImportSourcePtr importedSource = importSource();
         if (importedSource != nullptr) {
             ModelPtr model = importedSource->model();
             if (model != nullptr && model->hasUnits(importReference())) {
                 auto unit = model->units(importReference());
-                return unit->unitCount() == 0;
+                return unit->isBaseUnit(); // Call isBaseUnit recursively until unit is no longer an import
             }
         }
         return false;
--- a/src/units.cpp
+++ b/src/units.cpp
@@ -212,6 +212,19 @@ UnitsPtr Units::create(const std::string &name) noexcept
 
 bool Units::isBaseUnit() const
 {
+    if (isImport()) {
+        // If our model is unresolved or our import source is unresolved then return false
+        ImportSourcePtr importedSource = importSource();
+        if (importedSource != nullptr) {
+            ModelPtr model = importedSource->model();
+            if (model != nullptr && model->hasUnits(name())) {
+                auto unit = model->units(name());
+                return unit->unitCount() == 0;
+            }
+        }
+        return false;
+    }
+
     return unitCount() == 0;
 }
 
--- a/tests/units/units.cpp
+++ b/tests/units/units.cpp
@@ -1486,8 +1486,8 @@ TEST(Units, isBaseUnitsImported)
 
     model->addUnits(u);
 
-    u->setImportSource(import);
     import->setModel(model);
+    u->setImportSource(import);
 
     EXPECT_TRUE(u->isImport());
     EXPECT_TRUE(u->isBaseUnit());
--- a/tests/units/units.cpp
+++ b/tests/units/units.cpp
@@ -1527,7 +1527,7 @@ TEST(Units, isNotBaseUnitsImported)
     libcellml::ImportSourcePtr import = libcellml::ImportSource::create();
     import->setUrl("I_am_a_url");
     import->setModel(model);
-    
+
     u4->setImportSource(import); // u4 part of a separate model
     u4->setImportReference("u3");
 
@@ -1639,7 +1639,7 @@ TEST(Units, isNotBaseUnitButUnitIsBaseImport)
     u2->addUnit(libcellml::Units::StandardUnit::AMPERE, 0, 1.0, 1.0);
     u3->setName("u3");
     u3->addUnit("u2", 0, 1.0, 1.0);
-    u4->setName("some_other_u"); 
+    u4->setName("some_other_u");
 
     model->addUnits(u1);
     model->addUnits(u2);
--- a/tests/units/units.cpp
+++ b/tests/units/units.cpp
@@ -1484,7 +1484,10 @@ TEST(Units, isBaseUnitsImported)
     libcellml::ImportSourcePtr import = libcellml::ImportSource::create();
     import->setUrl("I_am_a_url");
 
+    model->addUnits(u);
+
     u->setImportSource(import);
+    import->setModel(model);
 
     EXPECT_TRUE(u->isImport());
     EXPECT_TRUE(u->isBaseUnit());
@@ -1498,8 +1501,8 @@ TEST(Units, isNotBaseUnitsImported)
     libcellml::UnitsPtr u1 = libcellml::Units::create();
     libcellml::UnitsPtr u2 = libcellml::Units::create();
     libcellml::UnitsPtr u3 = libcellml::Units::create();
-    
-    u1->setName("u1"); // Base unit 
+
+    u1->setName("u1"); // Base unit
     u2->setName("u2");
     u3->setName("u3");
     u2->addUnit("u1", 0, 1.0);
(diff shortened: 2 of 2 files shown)


tests/units/units.cpp
```diff
@@ -1471,3 +1471,198 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+    // Making a base unit
+    libcellml::UnitsPtr u = libcellml::Units::create();
+    u->setName("u");
+
+    libcellml::ImportSourcePtr import = libcellml::ImportSource::create();
+    import->setUrl("I_am_a_url");
+
+    model->addUnits(u);
+
+    import->setModel(model);
```
hsorby: You have made a circular reference here, no way a shared pointer can release the memory thus the memory leak.

hsorby (changes requested):

tests/units/units.cpp
hsorby: The model that gets set to the import is the model that has the units you are referencing.

awickens: So instead, to avoid the circular reference, we would create two units with the same name, one in the `model` and one outside the `model`, then checking?

hsorby: You have to create a whole separate model that has the units defined that another model is going to import from see [tests/model/model_flattening.cpp](https://github.com/hsorby/libcellml/blob/b83d125346d04323e9482754ac937a556da860f8/tests/model/model_flattening.cpp#L69) for an example.

awickens:
Ready for review.

awickens:
Ready for re-review.

src/units.cpp
```diff
@@ -212,6 +212,19 @@ UnitsPtr Units::create(const std::string &name) noexcept
...
 bool Units::isBaseUnit() const
 {
+    if (isImport()) {
+        // If our model is unresolved or our import source is unresolved then return false
+        ImportSourcePtr importedSource = importSource();
+        if (importedSource != nullptr) {
+            ModelPtr model = importedSource->model();
+            if (model != nullptr && model->hasUnits(importReference())) {
+                auto unit = model->units(importReference());
+                return unit->unitCount() == 0;
```
nickerso: I think here we should be doing `return unit->isBaseUnit();` to handle the case when you're importing an imported units...if I have my types correct ;) Then we only have the check on `unitCount()` defined in one place and handle an arbitrarily deep import hierarchy.

nickerso (changes requested):
Tests looking good, but would be useful to test a multi-level import hierarchy.

nickerso (dismissed):

tests/units/units.cpp
```diff
@@ -1471,3 +1471,327 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+{
+    libcellml::ModelPtr model = libcellml::Model::create();
+    model->setName("model");
+
+    libcellml::UnitsPtr u1 = libcellml::Units::create();
+    libcellml::UnitsPtr u2 = libcellml::Units::create();
+    libcellml::UnitsPtr u3 = libcellml::Units::create();
+    libcellml::UnitsPtr u4 = libcellml::Units::create();
+
+    u1->setName("u1"); // Base unit
```
hsorby: `u1` is redundant here it really takes no part in this test.

tests/units/units.cpp
```diff
@@ -1471,3 +1471,327 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+    import->setModel(model);
+
+    u4->setImportSource(import); // u4 part of a separate model
+    u4->setImportReference("u3");
+
+    EXPECT_TRUE(u4->isImport());
+    EXPECT_FALSE(u4->isBaseUnit());
+}
+
+TEST(Units, isNotBaseUnitsCheckUsingImport)
```
hsorby: Overly complicated test for running `isImport` false check when no imports are used.  Most of this test can be deleted.

tests/units/units.cpp
```diff
@@ -1471,3 +1471,327 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+    u3->setName("u3");
+    u2->addUnit("u1", 0, 1.0);
+    u2->addUnit(libcellml::Units::StandardUnit::AMPERE, 0, 1.0, 1.0);
+    u3->setName("u3");
+    u3->addUnit("u2", 0, 1.0, 1.0);
+    u4->setName("some_u");
+
+    model->addUnits(u1);
+    model->addUnits(u2);
+    model->addUnits(u3);
```
hsorby: Again most of this is redundant as it is not used.

tests/units/units.cpp
```diff
@@ -1471,3 +1471,327 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+
+    libcellml::ImportSourcePtr import = libcellml::ImportSource::create();
+    import->setUrl("I_am_a_url");
+
+    u4->setImportSource(import);
+    u4->setImportReference("u3");
+
+    EXPECT_TRUE(u4->isImport());
+    EXPECT_FALSE(u4->isBaseUnit()); // expect false as unit import is unresolved
```
hsorby: Rather than have the comment here it would be better to have the test:
`u4->importSource()->model() == nullptr`.

tests/units/units.cpp
```diff
@@ -1471,3 +1471,327 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+    model->addUnits(u1);
+
+    u2->setImportSource(import);
+    import->setModel(model);
+
+    EXPECT_TRUE(u2->isImport());
+    EXPECT_FALSE(u2->isBaseUnit()); // expect false as import ref name unresolved
+}
+
+TEST(Units, isNotBaseUnitButUnitIsBaseImport)
```
hsorby: This looks like the same test as `isNotBaseUnitsImported`.

tests/units/units.cpp
```diff
@@ -1471,3 +1471,327 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+}
+
+TEST(Units, isBaseUnitAndUnitIsBaseImport)
+{
+    libcellml::ModelPtr model = libcellml::Model::create();
+    model->setName("model");
+
+    libcellml::UnitsPtr u1 = libcellml::Units::create();
+    libcellml::UnitsPtr u2 = libcellml::Units::create();
+    libcellml::UnitsPtr u3 = libcellml::Units::create();
```
hsorby: These units are redundant in this test.

tests/units/units.cpp
```diff
@@ -1471,3 +1471,327 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+    import->setModel(model);
+
+    u4->setImportSource(import);
+    u4->setImportReference("u1");
+
+    EXPECT_TRUE(u4->isImport());
+    EXPECT_TRUE(u4->isBaseUnit()); // expect true as unit import is resolved and is base
+}
+
+TEST(Units, isBaseUnitImportedTwice)
```
hsorby, line 1575: Remove units that are not really part of the test.  `u2`, `u3`, `u5`, and `u6` don't actually play a part of this *unit* test.

tests/units/units.cpp
```diff
@@ -1471,3 +1471,327 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+    i2->setModel(m1);
+    u4->setImportSource(i2);
+    u4->setImportReference("u1");
+
+    EXPECT_TRUE(u7->isImport());
+    EXPECT_TRUE(u4->isImport());
+    EXPECT_TRUE(u7->isBaseUnit()); // expect true as unit import is resolved and is base
+}
+
+TEST(Units, isNotBaseUnitImportedTwice)
```
hsorby, line 1611: There are also some redundant variable names declared here also.

hsorby (changes requested):
I think we need to remove a lot of the redundant code that really has no effect on the *unit* test.  Also I think the names of variables should be made clearer e.g. a units of an imported model `i1` would be called `u_i1` this way we can make natural connections without having to trace through the code.

awickens:
Changes addressed - ready for another review.

tests/units/units.cpp
```diff
@@ -1508,103 +1502,58 @@ TEST(Units, isNotBaseUnitsImported)
...
 
-    u4->setImportSource(import);
-    u4->setImportReference("u3");
+    u2->setImportSource(import);
+    u2->setImportReference("u1");
 
-    EXPECT_TRUE(u4->isImport());
-    EXPECT_FALSE(u4->isBaseUnit()); // expect false as unit import is unresolved
+    EXPECT_TRUE(u2->isImport());
+    EXPECT_TRUE(u2->importSource()->model() == nullptr);
```
hsorby: or `EXPECT_EQ(nullptr, u2->importSource()->model());`

tests/units/units.cpp
```diff
@@ -1619,179 +1568,77 @@ TEST(Units, isBaseUnitImportReferenceUnresolved)
...
-    libcellml::ImportSourcePtr import = libcellml::ImportSource::create();
-    import->setUrl("I_am_a_url");
-    import->setModel(model);
-
-    u4->setImportSource(import);
-    u4->setImportReference("u1");
-
-    EXPECT_TRUE(u4->isImport());
-    EXPECT_TRUE(u4->isBaseUnit()); // expect true as unit import is resolved and is base
+    EXPECT_TRUE(u2->importReference() == "");
```
hsorby: Or `EXPECT_TRUE(u2->importReference().empty());`

tests/units/units.cpp
```diff
@@ -1471,3 +1471,174 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+TEST(Units, isBaseUnitImportedTwice)
+{
+    libcellml::ModelPtr m_i1 = libcellml::Model::create();
+    m_i1->setName("m_i1");
+    libcellml::ModelPtr m_i2 = libcellml::Model::create();
+    m_i2->setName("m_i2");
+
+    libcellml::UnitsPtr u_i1 = libcellml::Units::create();
+    libcellml::UnitsPtr u_i2 = libcellml::Units::create();
+    libcellml::UnitsPtr u_import = libcellml::Units::create();
```
hsorby: I would rename this guy to `u_i0`.

tests/units/units.cpp
```diff
@@ -1471,3 +1471,174 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+    u_import->setName("some_other_u");
+
+    m_i1->addUnits(u_i1);
+    m_i2->addUnits(u_i2);
+
+    libcellml::ImportSourcePtr i2 = libcellml::ImportSource::create();
+    i2->setUrl("I_am_a_url");
+    i2->setModel(m_i2);
+    u_import->setImportSource(i2);
+    u_import->setImportReference("u_i2");
```
hsorby: I would have `u_i1->setImportSource(i2);` etc.

tests/units/units.cpp
```diff
@@ -1471,3 +1471,174 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+
+    libcellml::ImportSourcePtr i1 = libcellml::ImportSource::create();
+    i1->setUrl("I_am_another_url");
+    i1->setModel(m_i1);
+    u_i2->setImportSource(i1);
+    u_i2->setImportReference("u_i1");
```
hsorby: I would have `u_i0->setImportSource(i1)`.

tests/units/units.cpp
```diff
@@ -1471,3 +1471,174 @@ TEST(Units, compareEquivalentSameSizeButDifferentExponent)
...
+
+    EXPECT_TRUE(u_import->isImport());
+    EXPECT_TRUE(u_i2->isImport());
```
hsorby: So that this would end up being:
```
EXPECT_TRUE(u_i0->isImport());
EXPECT_TRUE(u_i1->isImport());
EXPECT_FALSE(u_i2->isImport());
```

hsorby (changes requested):
Looking much better some minor renaming I think to the last two tests and a small amount of test answer rejigging and this will be good to go.

tests/units/units.cpp
awickens: That's what threw the error when I recommitted, it's `u2->importReference().empty()` in the most recent commit.

hsorby (approved):

nickerso (approved):
liking the refactored tests :)

Issue #496, opened 2019-12-11
Validating math in components


awickens:
If we have math set in our `component` class there is no way of verifying dimensional mismatch in units (i.e. between `mA` and `kA`). Although CellML only needs to account for unit mismatch at the connection level (after discussing with @agarny), it would be nice to have some way to notify the user of any unit multiplier mismatch within the mathematics of a component - not being interested in the SI units themselves, but more the scaling factor between them (if any).

Issue #493, opened 2019-12-10
Code generation: ensure that a model is runnable


agarny:
For code generation to work, a model must be valid, i.e. a call to `Validator::validateModel()` must return without any errors.

However, a valid model is no guarantee that code can be generated from it. Indeed, `Validator::validateModel()` makes sure that a model is valid with regards to the CellML specifications, but **not** whether a model is runnable.

So, as part of processing a model, our generator should check that the model is indeed runnable, and return errors if it is not.

An example of a valid model, but not runnable model, can be found in issue #492 (second example) where we have a model that has a variable that is both initialised and computed. Indeed, the only case where this is allowed is if the variable is computed through an ODE. However, in that example, the variable is computed through an algebraic equation.

In summary, this issue is about going through our `Generator` class and removing any "assumption" we make with regards to a model being runnable.

kerimoyle:
Just to note that in addition to the two in #492 there is also the malformed MathML (as in #459) which segfaults too...

agarny:
Closing this issue in favour of issue #499.

PR #497, opened 2019-12-11
Checks for Units in Component math in code generation

awickens:
Work towards creating verification for units in MathML strings as part of a `component`. 

Addresses #496.


nickerso:
Just an initial observation, the `Validator` should be validating against the rules defined in the CellML specification. Checking dimensional consistency within equations is not part of validation (against the spec). These are not errors in the CellML model, they are errors in the mathematical model that is being encoded in CellML.

So I would expect this to be done by a something other than the `Validator` class or as a different method in the current `Validator`. And to probably really start using the extended `Issue` ideas from #368.

agarny:
> So I would expect this to be done by a something other than the `Validator` class or as a different method in the current `Validator`. And to probably really start using the extended `Issue` ideas from #368.

Agreed. Probably as part of checking whether a model is runnable (see issue #493), although I feel like that code should be extracted from our `Generator` class (see issue #499). Those issues are somewhat all related. So, we will need to be careful in which order they are tackled. I feel like issue #499 should be worked on first. Then this issue is really a subset of issue #493.

In the meantime, and as a prerequisite to those issues, there is indeed the `Issue` class that really ought to be implemented. So, that would probably by my priority at this stage.

src/generator.cpp
```diff
@@ -1682,8 +1682,8 @@ UnitsMap processEquationUnitsAst(const GeneratorEquationAstPtr &ast, UnitsMap un
                     VariablePtr variable = getVariable(ast);
                     ComponentPtr component = (variable != nullptr) ? std::dynamic_pointer_cast<Component>(variable->parent()) : nullptr;
                     ModelPtr model = (component != nullptr) ? owningModel(component) : nullptr;
-                    std::string compName = (component != nullptr) ? component->name() : nullptr;
-                    std::string modelName = (model != nullptr) ? model->name() : nullptr;
+                    std::string compName = (component != nullptr) ? component->name() : " ";
```
agarny: Can't help but wonder why `" "` rather than `""`? (Keeping in mind that I have no idea what the code is supposed to be doing... :))

awickens: No reason at all, actually. I'm just committing to make sure unit tests pass but coverage certainly won't - I'll change it once I'm satisfied with the test I have (which is a work in progress, I got rid of the old ones).

awickens:
Currently at the stage where the units checking works - still need to check for multipliers, though -  but it would be nice if anyone shared their thoughts on the error messages. I was thinking of putting the actual mathematical expression into the error message, rather than just the operation and the component it is in (although it was sufficient when debugging some of the larger models in `test_generator`).

kerimoyle:
Hey @awickens It might be worth holding off on the error messages for a bit - we're trying to figure out some "rules" for them at the moment (see #514), so any feedback on them now would be probably subject to change!

awickens:
Right, makes sense. I'll hold off on making any changes to the messages for the moment.

agarny:
Agreed with @kerimoyle, don't worry too much about the error message for now.

src/generator.cpp
```diff
@@ -1515,10 +1515,14 @@ void updateBaseUnitCount(const ModelPtr &model,
                     multiplier += direction * (logMult + (standardMultiplierList.at(ref) + mult + standardPrefixList.at(pre)) * exp);
                 }
             }
-        } else if (unitMap.find(uName) == unitMap.end()) {
+        } 
+        
+        // Leaving this as a comment for the moment: This would only be necessary if we had a base unit which was *not* in the standard units list - uncertain if this will ever occur within a formal model.
```
kerimoyle: Yes, it will still be needed.  Users should be able to define their own custom base units and build others from them.  Most obvious units that don't use our current base variables are for cost/dollars/francs/whatever or pH.

src/generator.cpp
```diff
@@ -1868,7 +1868,7 @@ double processEquationMultiplierAst(const GeneratorEquationAstPtr &ast, std::vec
 
             // The only time we check multiplier mismatch is in a comparision operation.
             if (isDirectComparisonOperator(ast)) {
-                if (leftMult != rightMult && ast->mLeft != nullptr && ast->mRight != nullptr) {
+                if ((leftMult > rightMult || leftMult < rightMult) && ast->mLeft != nullptr && ast->mRight != nullptr) {
```
agarny: Wow, this is *ugly*! Please use [`areEqual()`](https://github.com/cellml/libcellml/blob/4f19a5c917c60f7f345ca11d9f7649d33d8b3bac/src/utilities.cpp#L163-L166) instead.

awickens:
I know this can't really be merged in until #521 has been approved (so the errors can be changed to warnings instead), but any review of the code would be appreciated.

src/utilities.h
```diff
@@ -44,9 +44,9 @@ const std::map<std::string, std::map<std::string, double>> standardUnitsList = {
     {"ampere", {{"ampere", 1.0}}},
     {"becquerel", {{"second", -1.0}}},
     {"candela", {{"candela", 1.0}}},
-    {"coulomb", {{"ampere", -1.0}, {"second", 1.0}}},
+    {"coulomb", {{"ampere", 1.0}, {"second", 1.0}}},
     {"dimensionless", {{"dimensionless", 1.0}}},
-    {"farad", {{"ampere", 2.0}, {"kilogram", -1.0}, {"metre", -2.0}, {"second", -4.0}}},
+    {"farad", {{"ampere", 2.0}, {"kilogram", -1.0}, {"metre", -2.0}, {"second", 4.0}}},
```
kerimoyle: Well caught! This was my bad, taken from a typo in the spec.

tests/resources/generator/validate_units_in_model.cellml
```diff
@@ -0,0 +1,232 @@
...
+                <eq/>
+                <ci>j</ci>
+                <apply>
+                    <power/>
+                    <ci>h</ci>
+                    <cn cellml:units="second">2</cn>
+                </apply>
+            </apply>
+        </math>
+        <!-- Not allowed - 2 is not dimensionless-->
```
kerimoyle: Do you also have a test when the power is a variable (ie: <ci> vs <cn>), to test that it's dimensionless?

src/generator.cpp
```diff
@@ -1285,6 +1313,649 @@ void Generator::GeneratorImpl::processEquationAst(const GeneratorEquationAstPtr
...
+           || (type == libcellml::GeneratorEquationAst::Type::COTH);
+}
+
+bool isDerivativeOperator(const GeneratorEquationAstPtr &ast)
+{
+    const GeneratorEquationAst::Type type = ast->mType;
+    return type == libcellml::GeneratorEquationAst::Type::DIFF;
+}
+
+bool isBottomVariableOperator(const GeneratorEquationAstPtr &ast)
```
kerimoyle: I think that the "b" in "bvar" stands for "bound"? Pretty sure it's not bottom :)

src/generator.cpp
```diff
@@ -1285,6 +1313,649 @@ void Generator::GeneratorImpl::processEquationAst(const GeneratorEquationAstPtr
...
+        else if (unitMap.find(uName) == unitMap.end()) {
+            unitMap.emplace(std::pair<std::string, double>(uName, 1.0 * uExp));
+        } else {
+            const auto &iter = unitMap.find(uName);
+            unitMap.at(iter->first) += (iter->second * uExp);
+        }
+    }
+}
+
+void updateBaseMultiplier(const ModelPtr &model,
```
kerimoyle: See PR #526 ... there was a bug somewhere around this area, just need to check that yours is ok?

src/generator.cpp
```diff
@@ -1285,6 +1313,649 @@ void Generator::GeneratorImpl::processEquationAst(const GeneratorEquationAstPtr
...
+                model = (ast->mVariable != nullptr) ? owningModel(ast->mVariable) : nullptr;
+                uName = (ast->mUnits != nullptr) ? ast->mUnits->name() : "dimensionless";
+                if (!(uName == "dimensionless")) {
+                    updateBaseUnitCount(model, unitMap, uName, 1, 0);
+                }
+            }
+            return unitMap;
+        }
+
+        // We know if we have reached an internal vertex that we have a mathematical operation as it's type.
```
kerimoyle: its not it's

src/generator.cpp
```diff
@@ -1285,6 +1313,649 @@ void Generator::GeneratorImpl::processEquationAst(const GeneratorEquationAstPtr
...
+                    correctUnits = isDimensionless(leftMap) && isDimensionless(rightMap); // If we have a variable as our power both the power and the quantity it is being applied to must be dimensionless
+                } else if (ast->mLeft->mType == GeneratorEquationAst::Type::DEGREE) {
+                    correctUnits = isDimensionless(leftMap);
+                    leftMap = rightMap;
+                } else {
+                    correctUnits = isDimensionless(rightMap); // Otherwise we just check the power for dimensionlessness
+                }
+
+                // If we have a square root operation
+                if (power == 0.0 && ast->mRight == nullptr) {
```
kerimoyle: I don't quite follow this bit?

src/utilities.h
```diff
@@ -44,9 +44,9 @@ const std::map<std::string, std::map<std::string, double>> standardUnitsList = {
     {"ampere", {{"ampere", 1.0}}},
     {"becquerel", {{"second", -1.0}}},
     {"candela", {{"candela", 1.0}}},
-    {"coulomb", {{"ampere", -1.0}, {"second", 1.0}}},
+    {"coulomb", {{"ampere", 1.0}, {"second", 1.0}}},
```
kerimoyle: And this one was just my mistake

kerimoyle:
HI @awickens!  I've had a quick peek and it's looking good (and I'm so glad you caught my units typos!)  I wouldn't worry about waiting for creating `Error`s or waiting for the `Issue` class - just make the errors here as best you can, but realise it could change once we're certain about how to finesse the messages.  It is useful to understand what kinds of problems will be reported though, so the error messages are a bit chicken and egg.  

I did wonder about whether the generator checks of maths should be done independently of the `Generator` class though?  So that people can check for errors in their maths/model formulation without necessarily needing to generate code to do it?  @agarny this is what you and I talked about a few weeks back? Thoughts?

agarny:
> I did wonder about whether the generator checks of maths should be done independently of the `Generator` class though? So that people can check for errors in their maths/model formulation without necessarily needing to generate code to do it? @agarny this is what you and I talked about a few weeks back? Thoughts?

Indeed, this should be done as part of issues #493/#499.

awickens:
Ready for review.

tests/resources/generator/validate_units_in_model.cellml
awickens: I will have to double check, but I think in some of the other models used as unit tests which @agarny put in (perhaps in the coverage test as well) there are instances of variables used as powers.

src/generator.cpp
awickens: Yep, will change tomorrow.

src/generator.cpp
awickens: There are instances in the AST where a root number is not specified as a child (i.e, the root node will only have a variable child and no second child giving a root number - it happens a couple of times in some of the bigger models in the unit tests for the `validator`). So, if there is no child (a power cannot be found, so 0.0 is returned), the assumption is that the intended operation is a square root operation.

nickerso: MathML defines a default square-root: https://www.w3.org/TR/REC-MathML/chap4_4.html#sec4.4.3.11

tests/resources/generator/validate_units_in_model.cellml
agarny: Indeed, the coverage test (at least) has instances of variables used as powers ([here](https://github.com/cellml/libcellml/blob/7e991e0f69d2512576659d1c41ef068ae95f7c27/tests/resources/generator/coverage/model.cellml#L396-L412) for example).

src/generator.cpp
kerimoyle: I'm not sure we should be making any assumptions anywhere?  Isn't that the point of an unambiguous model description? Rather return an error ...

tests/resources/generator/validate_units_in_model.cellml
kerimoyle: Are we even allowed to have dimensional stuff as an exponent?  How on earth do the units handle that?

agarny: As far as I can remember, only a `dimensionless` exponent makes sense. This is the reason `n` in `m^n` is `dimensionless`.

src/generator.cpp
awickens: I think a few of the models used for testing (not the ones I added in) would spontaneously combust if I was to just return an error. Perhaps when the AST gets moved out/exposed then some discussion can be made as to whether default roots like ```sqrt``` can be made into their own type of internal node.

tests/resources/generator/validate_units_in_model.cellml
awickens: Yep, @kerimoyle it doesn't make sense to have dimensional stuff as exponents (obviously), but to the best of my recollection there is nothing preventing someone from writing a CellML file intentionally with exponents which are not dimensionless, hence why it is still good to check if it is in case someone's messed up the CellML or incorrectly defined a variable with units.

src/generator.cpp
nickerso: the definition of the root operator is very clear in this regard. It is not an assumption. A root with two arguments has the degree (number) and the thing it is applied to; a root with a single argument has degree 2. (sorry if I'm missing something, trying to be helpful without reading everything ;-)

src/generator.cpp
awickens: Yes, double checked now and I have the correct format (with the ```standardMultiplierList.at(uName)```) which was missed out in ```units.cpp```.

agarny:
Just wondering... is this PR ready? If so, I might take over it and update it with our current `develop` branch. @awickens, @hsorby, @kerimoyle and @nickerso?

awickens:
@agarny coming out of the woodwork after a couple months, this was probably ready in March but I wouldn't be certain now, especially since there seems to be some code reshuffling going on where units checking is done elsewhere (looking at the Analyser class being implemented).

agarny:
> @agarny coming out of the woodwork after a couple months, this was probably ready in March but I wouldn't be certain now, especially since there seems to be some code reshuffling going on where units checking is done elsewhere (looking at the Analyser class being implemented).

Thanks @awickens, it's true that a lot of reshuffling of code has happened (and some more is on its way). So, just wondering, is that something on which you will be keen to work (once PRs #631 and #650) have been merged in, or would you rather leave it to me? (No pressure whatsoever! 🙂)

awickens:
> 
> Thanks @awickens, it's true that a lot of reshuffling of code has happened (and some more is on its way). So, just wondering, is that something on which you will be keen to work (once PRs #631 and #650) have been merged in, or would you rather leave it to me? (No pressure whatsoever! 🙂)

It depends on when you'd like it to be done (although I'd like to finish what I started!). I'm out through the South Island until the 27th, so I wouldn't be able to touch it for a couple of weeks until I get back up to Auckland. If PRs #631 and #650 get merged in within the next couple of days I would leave it up to you to get it done :)

agarny:
> It depends on when you'd like it to be done (although I'd like to finish what I started!). I'm out through the South Island until the 27th, so I wouldn't be able to touch it for a couple of weeks until I get back up to Auckland. If PRs #631 and #650 get merged in within the next couple of days I would leave it up to you to get it done :)

There is no rush, I was merely interested to know whether you were still keen to work on this issue since I agree that you should finish what you started. PR #631 is ready to go, but not PR #650 (it will, now, likely be ready sometime next week), so we can definitely wait until you are back.

awickens:
Okay @agarny, I've finally gotten around to this (about time, too). I just thought I'd run my plan by you to make sure I don't end up moving code over to the incorrect thing and having to revert it all back.

In my head, I envisioned moving this stuff from the ```generator``` class over into the ```analyser``` class, since all the AST exposure stuff has been moved there. This also includes moving the tests over (from memory I made a couple CellML files to check behavior). I suspect the error messages will have changed significantly from when I implemented them as well, so those will likely have to be changed as well. I was wondering if there was anything else in particular which needed to be addressed as well?

agarny:
> Okay @agarny, I've finally gotten around to this (about time, too).

Very nice to see some progress on this. Welcome back! :)

> I just thought I'd run my plan by you to make sure I don't end up moving code over to the incorrect thing and having to revert it all back.
>
> In my head, I envisioned moving this stuff from the `generator` class over into the `analyser` class, since all the AST exposure stuff has been moved there. This also includes moving the tests over (from memory I made a couple CellML files to check behavior). I suspect the error messages will have changed significantly from when I implemented them as well, so those will likely have to be changed as well. I was wondering if there was anything else in particular which needed to be addressed as well?

Yes, to move your work to the analyser is definitely the way forward. About something else that might need to be addressed, I will have to get back to you on this. Like you, it has been a long time since I have looked into this issue. Still, it should be safe to start moving your work to the analyser. (PRs #650 and #763 should be merged on Monday, but hopefully it shouldn't interfere too much with your work.)

hsorby:
Is anyone else seeing just the one file changed here?

awickens:
@hsorby Yeah, I had taken out all the stuff from the ```validator``` (with associated tests) awhile back, in preparation to move it over - unfortunately commitments to what I am doing over this summer have gotten in the way (plus Visual Studio refusing to build properly). I can start moving stuff back in once I sort out what's wrong with Visual Studio.

hsorby:
Hi @awickens, we are looking to move this forward because we want to use it for the building blocks of handling DAEs.  We will start working on this from 02/02/21.  Please don't stress if you cannot do what you were planning by then, we are happy to move this forward from the work you have done.  We are of course still happy to have your contributions at any time.  This is just a note to let you know so we don't double up on work.

awickens:
Hi @hsorby, I've put all the changes I originally made in ```validator.cpp``` as comments in the ```analyser.cpp``` file (as well as the tests I made all that time ago in ```test_analyser.cpp```) - hopefully this helps as this gets moved forwards from today.

agarny:
Thanks a lot @awickens.

agarny:
@hsorby, I will have a look at it tomorrow, if you don't mind.

src/analyser.cpp
```diff
@@ -1197,9 +1197,13 @@ void Analyser::AnalyserImpl::updateUnitsMap(const ModelPtr &model,
         UnitsPtr units = model->units(unitsName);
 
         if (units->isBaseUnit()) {
-            auto iter = unitsMap.find(unitsName);
+            if (unitsMap.find(unitsName) == unitsMap.end()) {
+                unitsMap[unitsName] = unitsExponent;
+            } else {
+                auto iter = unitsMap.find(unitsName);
```
hsorby: Not sure we need to do the find twice here.

agarny: Correct, I was just testing something. This will be corrected in a minute, although I will likely push it later with more upcoming changes.

agarny: Still needed to push things, so that I can test locally on Windows/Linux. So, yes, addressed in commit 12b600a65.

hsorby (commented):
You can't start identifiers with underscore in C++ that is reserved.

agarny:
> You can't start identifiers with underscore in C++ that is reserved.

Argh, why does Clang accept it?! Anyway, fixed in commit b591d08c5.

agarny:
This is replaced with PR #818, following the renaming of our `develop` branch to `main` (and `master` to `release`).

Issue #346, opened 2019-06-12
Switch "pretty printing" to use libXml2


nickerso:
As per the discussion on #312, with the change in #339 we can now switch to using libXml2 to handle the "pretty printing" of generated XML strings.

kerimoyle:
Closed by #511

PR #511, opened 2020-01-07
Printer: pretty-print a model using libxml2

agarny:
This addresses issue #346.


nickerso (approved):

kerimoyle (commented):
Just wondered whether there were parameters that can be set on the prettyprint stuff, or whether it was set in stone to have so many newlines?  I find it a huge amount easier to read/write/debug MathML when the `<apply>` and `<thingtodo>` are on the same line, so that they become the "brackets" of their indented operations?  eg:
```
<apply><eq/>
   <ci>a</ci>
   <apply><plus/>
       <ci>b</ci>
       <ci>c</ci>
   </apply> // closes plus - easy to see
</apply> // closes eq - easy to see
```

cf

```
<apply>
   <eq/>
   <ci>a</ci>
   <apply>
       <plus/>
       <ci>b</ci>
       <ci>c</ci>
   </apply> // closes plus - hard to see
</apply> // closes eq - hard to see
```
If that's not possible or against some kind of standard, then I'll live with it ... :)

agarny:
The only thing you can tell libxml2 (through its `xmlDocDumpFormatMemoryEnc()` method) is whether you want it to format (i.e. pretty-print) the output or not. So, no, you can't get what you are after. At least, not without manually alterating the output, which is clearly not worth it.

kerimoyle:
Okeydokes, I'll live with it then :)

kerimoyle (approved):

Issue #512, opened 2020-01-08
Default and manual file names for header files included in generated code


kerimoyle:
See discussion in #487 and #509.

Here's a test to show behaviour I'd expect from the generator in situations where an interface file is needed.  Since we'll only generate code for a validated model, and since the validator checks the model name for emptiness/weird characters/etc, I think it's reasonable to go with the verbatim model name as the default rather than trying to change it to camel case or anything fancy.  K.I.S.S :)

The test depends on implementation (are filenames stored/passed to the `Generator` or the `GeneratorProfile`?), so I wasn't sure how best to write it, but you get the general idea ...

```
TEST(Generator, setInterfaceFilenameAndDefault)
{
    auto model = libcellml::Model::create("myModelName");
    auto generator = libcellml::Generator::create();
    auto profileC = libcellml::GeneratorProfile::create(libcellml::GeneratorProfile::Profile::C);

    generator->setProfile(profileC);
    generator->processModel(model);

    std::string interfaceCode = generator->interfaceCode();
    EXPECT_TRUE(interfaceCode.find("#include \"myModelName.h\"") != std::string::npos)

    profileC->setInterfaceFileName("customHeaderFile.h"); // ?? or perhaps generator->setInterfaceFileName()??

    EXPECT_TRUE(interfaceCode.find("#include \"customHeaderFile.h\"") != std::string::npos)
    EXPECT_FALSE(interfaceCode.find("#include \"myModelName.h\"") != std::string::npos)
}
```

agarny:
> Here's a test to show behaviour I'd expect from the generator in situations where an interface file is needed. Since we'll only generate code for a validated model, and since the validator checks the model name for emptiness/weird characters/etc, I think it's reasonable to go with the verbatim model name as the default rather than trying to change it to camel case or anything fancy. K.I.S.S :)

Fine with the general idea (which indeed has to be done through `GeneratorProfile`), but not with using the name of a model as the default header file name. To have something like `myModelName.h` just looks wrong to me. For me, it should be `mymodelname.h`. So, because we are going to have the `GeneratorProfile::setInterfaceFileNameString()` method (and `GeneratorProfile::interfaceFileNameString()` one), I don't think we should have a default interface file name based on the name of the model.

kerimoyle:
[...]

PR #513, opened 2020-01-08
Default and set of interface filename in Generator

kerimoyle:
Initial test to show the behaviour I'd like in the `Generator`:
- default `#include` filename is the model's name, not "model.h"
- user can set their own filename easily without having to parse/replace contents of the whole header string

Addresses #512


agarny:
Closing in favour of PR #515.

PR #515, opened 2020-01-08
Generator: customise the name of an interface file

agarny:
Addresses issue #512.


nickerso (approved):

kerimoyle:
Looks good so far - but interested to hear you thoughts on [my latest comment](https://github.com/cellml/libcellml/issues/512#issuecomment-572430294) (esp the last bit):

Instead of a static default name, can the generator to return an error if the filename had not been specified? This way everyone knows what's going on, and there are no surprises, no dodgy filename constructions, and no need to edit a generated file?

agarny:
> Instead of a static default name, can the generator to return an error if the filename had not been specified? This way everyone knows what's going on, and there are no surprises, no dodgy filename constructions, and no need to edit a generated file?

The generator shouldn't return an error if the file name has not been specified. For one, there is no need for such a file name in the Python profile. So, this means that the `GeneratorProfile` class should have a method to let the user specify that no file name is needed, which would become ridiculous.

Also, if we were to return that kind of error then we would have to do something similar for *everything* that can be customised through the `GeneratorProfile` class, which would be a tedious and very time consuming task. As far as I am concerned, it's not worth it, assuming it could be done for *everything*.

kerimoyle (approved):
Thanks @agarny :)

Issue #514, opened 2020-01-08
Discussion of error construction and usage


kerimoyle:
Something new and hinky is happening in the validator ... here's a test to show you what I mean:
The test fails because the units defined inside the parsed model aren't being found when checking the variables inside the manually-created component, giving the message:
```
Description: Variable 't' has a units reference 'millisecond' that does not correspond with a standard units and is not a units defined in the variable's model.
```
There are a couple of things here:
- the error message needs to also report the name of its parent component so we can find it more easily
- if we're going to stick with "units" as a singular noun, how about this for the message:
"Variable 't' in component 'whatever' has units of 'millisecond' which are neither standard nor defined within its model." (just to avoid the awkward "a units" sentences!)
- but the main thing is ... the error shouldn't be happening as the units *are* defined in the model ... or am I missing something dumb?

```
TEST(Validator, unfoundUnitsInParsedModel)
{
    auto parser = libcellml::Parser::create();
    auto validator = libcellml::Validator::create();

    std::string in = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
                     "<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"error_in_units\">"
                     "  <units name=\"millisecond\">"
                     "    <unit prefix=\"milli\" units=\"second\"/>"
                     "  </units>"
                     "  <component name=\"IonChannel\">"
                     "    <variable name=\"t\" units=\"millisecond\"/>"
                     "  </component>"
                     "</model>";

    auto model = parser->parseModel(in);

    validator->validateModel(model);
    EXPECT_EQ(size_t(0), validator->errorCount()); // <<< this is fine!

    // Add a component to represent the voltage dependency of the n-gate
    auto nGate = libcellml::Component::create("nGate");
    model->addComponent(nGate);

    auto t2 = libcellml::Variable::create("t2");
    t2->setUnits("millisecond");
    nGate->addVariable(t2);

    validator->validateModel(model);
    EXPECT_EQ(size_t(0), validator->errorCount()); // <<< test fails!
}
```

agarny:
> * the error message needs to also report the name of its parent component so we can find it more easily

Definitely agree. We might also want to report the name of the model? This is what I have done when adding errors from the code generator (since we can import units/components).

> * if we're going to stick with "units" as a singular noun, how about this for the message:
[...]

PR #516, opened 2020-01-09
Fix inconsistency of hasUnits behaviour

kerimoyle:
Addresses #514


kerimoyle:
@agarny Confirming error in validator :( Not sure why you're not seeing it too?  Where am I going wrong?

kerimoyle:
I've found the source of the error and can fix it so the unit tests pass, but it's showing up a much bigger problem!  There's something really odd about the way the `hasUnits` function behaves, and depending on which overload you use, the final result is quite different.  This was the source of the error I was getting earlier (see the `Validator::validateVariable()` function):

If I call it by reference, then it fails to find units:
`        if ((model != nullptr) && !model->hasUnits(variable->units())) // << test fails`
If I call it by units name, tests behave as expected ...:
`        if ((model != nullptr) && !model->hasUnits(variable->units()->name())) // <<< tests pass`

**... BUT ... this has surfaced some bigger issues:** 

- Of course we need to be sure about how the `hasUnits()` function behaviour is defined, because at present having an overloaded function which gives very different results is rather less-than-sub-optimal (!)

- The difference is coming because the pass-by-reference option is expecting that the model contains linked units, and in my test the units which are added by name are **not** linked before validation.  I tried calling `model->linkUnits()` at the top of the call to `validateModel()`  (to mimic calls to it elsewhere) but this breaks other tests where a unit truly *isn't* defined in the model as it fails to catch them.  Is this expected behaviour?  It seems very strange to me!  The tests which are affected are:
```
Validator.unnamedAndDuplicateNamedVariablesWithAndWithoutValidUnits
Validator.unfoundUnitsInEncapsulatedComponents
```
... both of which fail because the validator now fails to spot genuinely missing units.

- Finally, the current commit with my 8-character-fix is failing the coverage testing because it's not finding the case of non-standard prefix values.  I am so confused as to how changing this one line of code (as above) can alter the structure of the whole validation process that this line is now excluded?  (It's not an issue with the coverage test as I've run it with a print out and it genuinely doesn't access that line any more).  Previously coverage was provided by the `Units.unitsWithPrefixOutOfRange` test.  (an aside: I've not been able to figure out why that line existed in the first place but thought it best to leave it there to show the difference caused ...)

Conclusion ... I am bamboozled, both by behaviour and expected behaviour :( Thoughts, anyone? @agarny @nickerso @awickens

agarny:
> @agarny Confirming error in validator :( Not sure why you're not seeing it too? Where am I going wrong?

I just gave it another try (on my up-to-date `develop` branch) and, as far as I can tell I did the same thing as yesterday. Yet, today, I am seeing the error while I didn't yesterday!? Boh!

> I've found the source of the error and can fix it so the unit tests pass, but it's showing up a much bigger problem! There's something really odd about the way the `hasUnits` function behaves, and depending on which overload you use, the final result is quite different. This was the source of the error I was getting earlier (see the `Validator::validateVariable()` function):
> 
> If I call it by reference, then it fails to find units:
> ` if ((model != nullptr) && !model->hasUnits(variable->units())) // << test fails`
> If I call it by units name, tests behave as expected ...:
> ` if ((model != nullptr) && !model->hasUnits(variable->units()->name())) // <<< tests pass`

I can indeed confirm the above and, yes, this is bad. :(

> * Finally, the current commit with my 8-character-fix is failing the coverage testing because it's not finding the case of non-standard prefix values.  I am so confused as to how changing this one line of code (as above) can alter the structure of the whole validation process that this line is now excluded?

Is that a rhetorical question? I mean, if you look [`Model::ModelImpl::findUnits(const std::string &name)`](https://github.com/cellml/libcellml/blob/912bb4d289feeda94d0c93039eee6c5d40f9ad7a/src/model.cpp#L53-L57) and [`Model::ModelImpl::findUnits(const UnitsPtr &units)`](https://github.com/cellml/libcellml/blob/912bb4d289feeda94d0c93039eee6c5d40f9ad7a/src/model.cpp#L59-L63), you can see that the latter does quite a bit more checks and must, at some point, go through that now-non-covered line.

> Conclusion ... I am bamboozled, both by behaviour and expected behaviour :( Thoughts, anyone? @agarny @nickerso @awickens

I am not super familiar with the units side of libCellML, but what is clear to me is that both `Model::hasUnits(const std::string &name)` and `Model::hasUnits(const UnitsPtr &units)` should behave in the same way, whatever we decide that way should be.

awickens:
@agarny @kerimoyle I've done some digging in the debugger and something weird is definitely happening. The reason why `hasUnits()` appears to have different behavior when using:

`if ((model  != nullptr)  &&  !model->hasUnits(variable->units()))`
and 
`if ((model  != nullptr)  &&  !model->hasUnits(variable->units()->name()))`

Is that `variable()->units()` returns `nullptr` when you call it, but in your added test case, `variable->units()->name()` doesn't, which is strange and probably why the test spontaneously combusts when using the first call. I had a look at the `setUnits` function as well to see if we were somehow not assigning the variable units and only a name when we make the call

`t2->setUnits("millisecond")`,

 but that just brings you back to `hasUnits()` again (which itself is called in `setUnits()`....). I'll do some more digging and see if I can find out why it's not liking `variable->units()`.

**EDIT**: `setUnits()` is indeed the problem. It can't find a model because you set the variable units before you add it to the `model`, so you can't find any units called "millisecond". So the lines 

```   
t2->setUnits("millisecond");
nGate->addVariable(t2);
```

should be

```   
nGate->addVariable(t2);
t2->setUnits("millisecond");
```

And indeed it passes everything if you make that change!

agarny:
> And indeed it passes everything if you make that change!

I can confirm that calling `nGate->addVariable(t2)` before `t2->setUnits("millisecond")` is indeed what should have been done in the first place (I reviewed some code about that stuff earlier this week). Well spotted @awickens!

Now, regarding `Model::hasUnits(const std::string &name)` and `Model::hasUnits(const UnitsPtr &units)`, I can understand why it would be "ok" to have them work in a slightly different way, but in that case the documentation has to be crystal clear about it, or people are bound to come across the same problem as @kerimoyle.

kerimoyle:
> what should have been done in the first place 

I'm glad we've found the source of this problem, but for me the behaviour itself remains a problem.  I'd expect a user to be able to add anything, in any order, at any time, including removing and editing as need be, and still have predictable and consistent behaviour.  Anything else is a recipe for user grumpiness and a debugging nightmare.  I'd have thought that calling the linkUnits function at the start of the validation would be the best way forward, but that has its own problems of behaviour too?  

For the `hasUnits` stuff I really don't think that documentation is the answer here, especially as every other part of libcellml allows flexibility.  We either need to rename the functions to show that they're distinct (overloads should allow flexibility in the argument list, *not* the functionality!).  

Sorry to be the wet blanket, but I don't think this is "fixed" yet :(

awickens:
@kerimoyle To the best of my recollection in CellML the idea is that if two units have the same name, they *must* be the same unit (or something along those lines, I haven't checked the documentation in a while). I agree with what you said about `hasUnits()`, though. I guess we could get rid of the `dimensionallyEquivalent()` unit comparison in the `findUnits(const UnitsPtr &units)` function, and that would make behaviour consistent with the `findUnits(const std::string name)` function - as possibly another option to fix the discrepancy.

kerimoyle:
Hey again @awickens.  Yes, you're right, there certainly can't be two units in the same model with the same name.  That's not the bit I have a problem with.  I think that the order in which the model is constructed should not be allowed to make a difference to the final model.  In other words, the "problem" earlier (where the units were added to the variable before they were present in the model) shouldn't actually be a problem at all - whatever we do behind the scenes must accept this as a user behaviour, and have exactly the same model structure/validity/interpretation/etc as if they were added in a different order.  Does that make more sense? Or am I missing something again? (entirely possible!)

awickens:
Yep, that makes sense - so we would want there to be no difference in whatever order we chose to construct the model. 

It's taken me awhile to wrap my head around how we would do that though - using an old example, in the case of that problem with the lines 

```
t2->setUnits("millisecond");
nGate->addVariable(t2);
```

If we called it in the above order, even though it works when you use variable->name() (where ordering doesn't matter), the actual units of millisecond in t2 (which is empty) is not the same as the units of millisecond in the model - so for me, although I understand why we would want it to not matter in practice, I'm not sure how we could do it without some lines necessarily coming before others (like defining something to be part of a model if you want it to have the same units as something in the model). 

Just my 2 cents though, it's probable @agarny would have a more substantial yes/no answer for re-implementing model construction so the order in which you define things doesn't matter.

agarny:
I agree, I really don't like the fact that:

```
t2->setUnits("millisecond");
nGate->addVariable(t2);
```

currently gives different results than:

```
nGate->addVariable(t2);
t2->setUnits("millisecond");
```

I can understand why this is the case (because of the way things are currently done underneath), but this is just wrong and bond to frustrate people in the end (it would certainly frustrate me). I guess this is one for @hsorby... :)

kerimoyle:
It must be something new though?  Because *all* the tutorials (which up until a few days ago) were totally fine in the order in my test?  So can't we just reset to how it was earlier and figure out a better way to make whatever new functionality has caused the issue?

kerimoyle:
Another thought - would it be possible to call the general `linkUnits` (? can't remember name exactly) which is called before printing and generation, at the top of validation?  This means that we can add whatever in any order, but that it *should* all only get linked at validate-time?  I tried this in that same test earlier, but it had some equally strange behaviour.  I'd imagine that fixing this would be the best approach though ... what do others think?

awickens:
I would guess that it's (probably) the presence of `dimensionallyEquivalent()` in one of the `findUnits()` functions which caused the issue (ordering does matter for that one, and I guess it's fair considering you would want the units to actually be the same when checking if it was in the model).

`linkUnits()` looks like a reasonable way of getting around it, although It does seem like you could still get unexpected behavior depending on the order you added the units in (it adds the first units instance found when iterating through the variables, and thereafter sets model units to a variable if it can find them). So you could get weird things happening there if you didn't ensure the first thing added to the model was the full definition of the units, and not just a name. We could change it to ensure it adds the proper units definition first, I guess?

kerimoyle (commented):
The whole point of this test was to show that we need the flexibility to add these in any order.  Changing the test doesn't help fix that ...

kerimoyle:
>  and I guess it's fair considering you would want the units to actually be the same when checking if it was in the model

... really?  I wouldn't have thought that.  I'd have thought that `hasUnits` would look only for the name, or name of reference passed; there's nothing in the function name or description to indicate that it's looking for dimensional consistency.   

After a bit more digging I think I understand it better now, and have made these changes:

- I've changed the `hasUnits()` function so that both overloads *only* search by name, so that it does what it says on the tin; 
- I added a `hasExactUnits()` function which includes the dimensionality checking that previously existed in the hasUnits(reference) function;
- I've changed the `flattenComponent` function to call the `hasExactUnits` rather than `hasUnits` when trying to resolve name clashes in imported units with different dimensionality.  

Are people are happy with those names/behaviours?  Could also use `hasUnitsNamed()` as the inexact match if that was preferred, or have another argument.  Either way, the vastly different behaviour of the earlier hasUnits(ref or name) has been addressed.

The tests pass (with one exception) on my machine, including the adding-in-any-order flexibility that raised this issue in the first place.  The test which is currently failing is in python and I'm currently trying to debug where it's actually happening (vscode is not behaving for me today!) ...

agarny:
Not happy with that "fix". I would rather address the original issue rather than come up with new API, which we shouldn't need.

Otherwise, @hsorby, it looks like the tests are not being triggered anymore... once again...

kerimoyle:
> I would rather address the original issue

Agreed!  But isn't this the underlying cause of the original issue?

hsorby:
Changes here fix the inconsistency of `hasUnits` behaviour and thus ready for review :).

kerimoyle (commented):
Two thoughts:
- not entirely happy with functions (eg: `unitsCount`) which do more than their name suggests?  In that one they also link units to the model?  I can see how it's easier for us, but it doesn't quite sit right ...
- Are there reasons we can't/shouldn't call `linkUnits` at the beginning of the `validateModel` function (similarly to how it's called before printing/generation?)

hsorby:
Not sure which `unitsCount` you are looking at but that does nothing but return the number of units in the model?

hsorby:
> * Are there reasons we can't/shouldn't call `linkUnits` at the beginning of the `validateModel` function (similarly to how it's called before printing/generation?)

Yes, `linkUnits` modifies the model so we cannot use it in `validateModel`.

kerimoyle:
> Not sure which unitsCount you are looking at but that does nothing but return the number of units in the model? 

@ hsorby Whoops, my bad, I read unexpanded github and matched the wrong function description with the name :)

kerimoyle:
The original reason for this PR still exists ... but somewhere along the way the test demonstrating the bug disappeared.  I'm pretty sure that doesn't count as "fixing" it, so I'm putting it back ... :(

src/validator.cpp
```diff
@@ -591,7 +591,7 @@ void Validator::ValidatorImpl::validateVariable(const VariablePtr &variable, con
         ModelPtr model = owningModel(component);
         if ((model != nullptr) && !model->hasUnits(variable->units())) {
             ErrorPtr err = Error::create();
-            err->setDescription("Variable '" + variable->name() + "' has a units reference '" + unitsName + "' that does not correspond with a standard units and is not a units defined in the variable's model.");
+            err->setDescription("Variable '" + variable->name() + "' in component '" + component->name() + "' has a units reference '" + unitsName + "' which is neither standard nor defined in the parent model.");
```
kerimoyle, line 594: Should we give users a headsup here that they *might* have defined the units but just need to call `linkUnits()` to get them recognised?

kerimoyle:
Tests _including_ original issue are passing so I'm happy now :)  I think it would be good to rework the error message for missing units to tell users that they just need to call `linkUnits`, but will leave that for the other broader error issue.  I can't "approve" this as is on my own branch though ...

agarny:
> Tests _including_ original issue are passing so I'm happy now :) I think it would be good to rework the error message for missing units to tell users that they just need to call `linkUnits`, but will leave that for the other broader error issue. I can't "approve" this as is on my own branch though ...

Hmm, not sure I would mention the need to call `linkUnits()`.

kerimoyle:
> Hmm, not sure I would mention the need to call linkUnits().

The trouble is that at the moment it's possible to create a model which includes the units, and have the validator tell you it can't find them (hence the test that I re-included).  We need to change the way that we tell users what the problem is to reflect this somehow, because the current error message only causes more confusion!

agarny:
Sure, happy for the error message to be reworded, but I wouldn't mention the need to call this or that API. I don't think it has anything to do in such an error message.

hsorby:
Ready for review again.

src/api/libcellml/model.h
```diff
@@ -256,18 +256,19 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
-     * the name given to the variables setUnits() method then the units for
-     * that variable will be unlinked.  This method will link variable units
-     * specified by name to units in the model (if they are found). Any variable
-     * units that cannot be linked to units in the model are left untouched.
-     *
-     * Any @c Variables found that are not linked to model @c Units will be
-     * linked. If a Variable has units that are not attached to the model
-     * then the units will be added to the model as well.
+     * set by name.  If the model that the variable belongs to has
+     * units defined with the same name then that variables units will
```
kerimoyle: Minor grammar/formatting stuff:
"If the @c Model to which the @c Variable belongs has @c Units defined with the same name, then that @c Variable's @c Units will not be linked to the @c Model's @c Units."

kerimoyle:
Is this still active?  Alas, I can't review it but am happy with the code ...

hsorby (dismissed):
@kerimoyle and I are happy with this code.

tests/model/model_flattening.cpp
```diff
@@ -591,393 +591,125 @@ TEST(ModelFlattening, importedComponentWithNameClashes)
...
 
-    auto a = printer->printModel(model);
-    auto modelNonImportVersion = parser->parseModel(fileContents("generator/hodgkin_huxley_squid_axon_model_1952/model.cellml"));
-    auto e = printer->printModel(modelNonImportVersion);
-    EXPECT_EQ(e, a);
+    membrane->setImportSource(membraneImporter);
+    membrane->setImportReference("membrane_for_importing");
 
-    auto generator = libcellml::Generator::create();
+    // Import the sodium channel from a file, and set as encapsulated child of the membrane
```
agarny: Minor: full stop at the end.

tests/model/model_flattening.cpp
```diff
@@ -591,393 +591,125 @@ TEST(ModelFlattening, importedComponentWithNameClashes)
...
+    auto sodiumChannel = libcellml::Component::create("sodium_channel");
+    membrane->addComponent(sodiumChannel);
 
-    generator->processModel(model);
+    auto sodiumImporter = libcellml::ImportSource::create();
+    sodiumImporter->setUrl("basicsodiumchannel.cellml");
```
agarny: We are really inconsistent when it comes to our file name convention. Yet, in most cases we use `_` to separate words, so here I would expect to see `basic_sodium_channel.cellml` (something similar for other file names in this PR).

tests/model/model_flattening.cpp
```diff
@@ -591,393 +591,125 @@ TEST(ModelFlattening, importedComponentWithNameClashes)
...
-        "          <apply>\n"
-        "            <minus/>\n"
-        "            <ci>V</ci>\n"
-        "            <ci>E_Na</ci>\n"
-        "          </apply>\n"
-        "        </apply>\n"
-        "      </apply>\n"
-        "    </math>\n"
-        "  </component>\n"
+        "  <component name=\"imported_heirarchy\"/>\n"
```
agarny: Spelling mistake (and the same spelling mistake in other places in this PR).

agarny (changes requested):
Fine with this PR pending some minor issues.

kerimoyle:
Thanks @agarny, I'll do those now.

tests/model/model_flattening.cpp
kerimoyle: Done

tests/model/model_flattening.cpp
kerimoyle: Done

tests/model/model_flattening.cpp
kerimoyle: Looking at the `modelflattening` folder we're about 50/50 with the naming.  I've changed these new tests by adding the underscore, but wanted to check whether this was a big deal before going through the rest of them, as it affects pretty much every test involving a file and its contents, as well as their names ...?

agarny: That's the other thing, the `modelflattening` folder should be named `model_flattening` to be consistent with what we do elsewhere. As for the files in that folder being 50/50 with the naming, well... the same person was involved. (Not giving names... :p :))

Anyway, only things related to this PR should be "fixed". Anything else should be done as part of issue #402.

kerimoyle:
Changes made ... :)

agarny (approved):

tests/model/model_flattening.cpp
kerimoyle: It's not just this folder though, it's throughout all the tests.  We have .xml as well as .cellml extensions, camelCase, underscores, etc.  In some ways it's good to have variety of names as it "tests" them all out, but it looks inconsistent.  The reason I stopped renaming all the files here is because a lot of the missing underscores seemed to make sense: eg, the "import_of_reallylongfilename" stuff is easier to read when the "import_of_" looks different from the "reallylongfilename" bit, esp when there are several generations of importing.  So I think that those should stay as they are ...

kerimoyle:
@hsorby Please can you re-approve as I can't?

nickerso (approved):
given changes are minor fixes, happy to approve on @hsorby's behalf :)

Issue #519, opened 2020-01-14
Imports are skipping units used only by <cn> terms


kerimoyle:
When you import a component containing units which are *only* referenced by the <cn> terms in the maths, these units are not being imported and the validator returns a "units not found" error. 
 I'll make a PR with the failing test.

eg:  
```
TEST(Model, missingUnitsFromImportOfCnTerms)
{
    std::string a =
        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
        "<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"myModel\">\n"
        "   <units name=\"myUnitsThatReallyExist\">\n"
        "       <unit exponent=\"-1\" units=\"second\"/>\n"
        "   </units>\n"
        "   <component name=\"myComponent\">\n"
        "       <variable name=\"a\" units=\"second\"/>\n"
        "       <math xmlns=\"http://www.w3.org/1998/Math/MathML\" xmlns:cellml=\"http://www.cellml.org/cellml/2.0#\">\n"
        "           <apply><eq/>\n"
        "               <ci>a</ci>\n"
        "               <cn cellml:units=\"myUnitsThatReallyExist\">1</cn>\n"
        "           </apply>\n"
        "       </math>\n"
        "   </component>\n"
        "</model>";
    auto parser = libcellml::Parser::create();
    auto modelFromString = parser->parseModel(a);
    auto validator = libcellml::Validator::create();
    validator->validateModel(modelFromString);
    EXPECT_EQ(size_t(0), validator->errorCount());

    // Import the membrane component from a file
    auto model = libcellml::Model::create("model_from_imports");
    auto c = libcellml::Component::create("c");

    auto imp = libcellml::ImportSource::create();
    imp->setUrl("units_in_cn.cellml");

    c->setImportReference("myComponent");
    c->setImportSource(imp);
    model->addComponent(c);


    EXPECT_TRUE(model->hasUnresolvedImports());
    model->resolveImports(resourcePath());
    EXPECT_FALSE(model->hasUnresolvedImports());
    model->flatten();

    validator->validateModel(model);
    for(size_t i = 0; i < validator->errorCount(); ++i){
        std::cout << validator->error(i)->description()<<std::endl;
    }
    EXPECT_EQ(size_t(0), validator->errorCount());
}
```

hsorby:
[...]

PR #520, opened 2020-01-14
Missing import units

kerimoyle:
Test to demonstrate missing units from imports, as in #519


kerimoyle:
This was a bit deeper than I'd imagined ... I had to change a test: `TEST(ModelFlattening, importingComponentThatAlsoHasAnImportedComponentAsAChild)` which was originally missing the import of units `per_mV_ms` from the `sodiumchannel.cellml `component.  This took ages to find because the test was originally passing, even though the code wasn't actually working properly.  What do you guys think about calling the `Validator` more frequently in tests, not necessarily to assure validity (we should be able to read/write etc invalid models), but just to make sure we know what's actually happening?  

In any case, review please :) @agarny @hsorby @nickerso

kerimoyle:
Review please @agarny @hsorby @nickerso :)

src/api/libcellml/model.h
```diff
@@ -271,6 +272,13 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
      */
     void linkUnits();
 
+    /**
+     * @brief Returns an unordered set of @c Units names which are used within the
+     * MathML in the model.
+     * @return unordered set of names of @c Units
+     */
+    std::unordered_set<std::string> cnUnits();
```
agarny: As far as I can tell, `cnUnits()` is only used internally, so it has nothing to do in the API.

src/model.cpp
```diff
@@ -242,6 +244,55 @@ void linkComponentVariableUnits(const ComponentPtr &component)
     }
 }
 
+void findCnUnits(const XmlNodePtr &node, std::unordered_set<std::string> &cnUnits)
+{
+    XmlNodePtr childNode = node->firstChild();
+    while (childNode) {
```
agarny: Replace `childNode` with `childNode != nullptr`.

src/model.cpp
```diff
@@ -242,6 +244,55 @@ void linkComponentVariableUnits(const ComponentPtr &component)
...
+            }
+        }
+        findCnUnits(childNode, cnUnits);
+        childNode = childNode->next();
+    }
+}
+
+void findComponentCnUnits(const ComponentPtr &component, std::unordered_set<std::string> &cnUnits)
+{
+    // Inspect the mathml in this component for any specified constant <cn> units,
```
agarny: "MathML" not "mathml".

src/model.cpp
```diff
@@ -242,6 +244,55 @@ void linkComponentVariableUnits(const ComponentPtr &component)
...
+    XmlDocPtr doc = std::make_shared<XmlDoc>();
+    doc->parse(math);
+    XmlNodePtr root = doc->rootNode();
+    findCnUnits(root, cnUnits);
+}
+
+void doFindCnUnits(const ComponentPtr &component, std::unordered_set<std::string> &cnUnits)
+{
+    findComponentCnUnits(component, cnUnits);
+    for (size_t c = 0; c < component->componentCount(); c++) {
```
agarny: `++c` not `c++`.

src/model.cpp
```diff
@@ -691,16 +742,29 @@ void flattenComponent(const ComponentEntityPtr &parent, const ComponentPtr &comp
             importedComponentCopy->addComponent(component->component(i));
         }
 
-        // Temporarily add component to new model to find units used.
+        // Temporarily add component to new model to find units used by the variables
```
agarny: Missing full stop at the end of the sentence.

src/model.cpp
```diff
@@ -691,16 +742,29 @@ void flattenComponent(const ComponentEntityPtr &parent, const ComponentPtr &comp
...
         tempModel->linkUnits();
+
         std::vector<UnitsPtr> requiredUnits;
         for (size_t i = 0; i < tempModel->unitsCount(); ++i) {
             auto u = tempModel->units(i);
             requiredUnits.push_back(u);
         }
 
+        // Get the names of the units used by the <cn> elements in all the components' mathml and add them
+        // to the required units
```
agarny: Missing full stop at the end of the sentence. Also, "MathML" not "mathml".

tests/model/model.cpp
```diff
@@ -552,3 +554,47 @@ TEST(Model, setAndCheckIdsAllEntities)
...
+        "       </math>\n"
+        "   </component>\n"
+        "</model>";
+    auto parser = libcellml::Parser::create();
+    auto modelFromString = parser->parseModel(a);
+    auto validator = libcellml::Validator::create();
+    validator->validateModel(modelFromString);
+    EXPECT_EQ(size_t(0), validator->errorCount());
+
+    // Import the membrane component from a file
```
agarny: Not "critical" for tests, but would still add a full stop at the end of the sentence.

tests/model/model_flattening.cpp
```diff
@@ -654,6 +654,10 @@ TEST(ModelFlattening, importingComponentThatAlsoHasAnImportedComponentAsAChild)
         "  <units name=\"per_ms\">\n"
         "    <unit exponent=\"-1\" prefix=\"milli\" units=\"second\"/>\n"
         "  </units>\n"
+        "  <units name=\"per_mV_ms\">\n" // KRM added this in as was not being imported or tested for
```
agarny: Remove comment.

src/model.cpp
```diff
@@ -242,6 +244,55 @@ void linkComponentVariableUnits(const ComponentPtr &component)
...
+    findComponentCnUnits(component, cnUnits);
+    for (size_t c = 0; c < component->componentCount(); c++) {
+        doFindCnUnits(component->component(c), cnUnits);
+    }
+}
+
+std::unordered_set<std::string> Model::cnUnits()
+{
+    std::unordered_set<std::string> cnUnits = {};
+    for (size_t c = 0; c < componentCount(); c++) {
```
agarny: `++c` not `c++`.

src/model.cpp
```diff
@@ -691,16 +742,29 @@ void flattenComponent(const ComponentEntityPtr &parent, const ComponentPtr &comp
...
 
+        // Get the names of the units used by the <cn> elements in all the components' mathml and add them
+        // to the required units
+        auto cnList = tempModel->cnUnits();
+        for (auto &name : cnList) {
+            if (!tempModel->hasUnits(name)) {
+                auto u = importModel->units(name);
+                if (u != nullptr) {
+                    requiredUnits.push_back(u);
+                } // KRM else warning
```
agarny: Remove the comment.

src/model.cpp
kerimoyle: Done!

src/model.cpp
kerimoyle: Done

src/model.cpp
kerimoyle: Happy to change it if need be, but we use "mathml" throughout the rest of the code too?  This should also change?

src/model.cpp
kerimoyle: Done

src/model.cpp
kerimoyle: Done

tests/model/model_flattening.cpp
kerimoyle: Done

src/model.cpp
kerimoyle: Changed to TODO so can be found later

tests/model/model.cpp
kerimoyle: Done

src/model.cpp
agarny: Change the ones in this PR. The others ought to be changed as part of issue #402 (i.e. a PR should remain on target).

src/model.cpp
agarny: Fair enough.

src/model.cpp
kerimoyle: Added fullstop.  We use "mathml" throughout the code too...  should fix all of them separately?

src/api/libcellml/model.h
kerimoyle: Not sure about this one.  We don't use it currently (but I could add a test if you like ;) ) I thought it might be useful to the user to be able to retrieve units referenced from within the MathML rather than added through the API?  Happy to delete if that's not the case.  What do others think?  @hsorby @nickerso ?

hsorby: Definitely doesn't belong in the API.

src/model.cpp
kerimoyle: Done, also changed other occurrences.

src/model.cpp
kerimoyle: Done, ditto other places.

src/api/libcellml/model.h
kerimoyle: Done.

kerimoyle:
All changes made now.  Should be good to go? @agarny @hsorby @nickerso ?

src/api/libcellml/model.h
```diff
@@ -20,6 +20,7 @@ limitations under the License.
 #include "libcellml/exportdefinitions.h"
 
 #include <string>
+#include <unordered_set>
```
agarny: Not needed anymore.

src/bindings/interface/model.i
```diff
@@ -73,6 +73,9 @@ determine the full path to the source model relative to this one.";
 %feature("docstring") libcellml::Model::flatten
 "Instantiate all imported Components and Units to make this model self-contained.";
 
+%feature("docstring") libcellml::Model::cnUnits
+"Returns an unordered set of Units names which are used within the MathML in the model."
+
```
agarny: Not needed anymore.

src/api/libcellml/model.h
kerimoyle: Done

src/bindings/interface/model.i
kerimoyle: Done

kerimoyle:
Is there a flag/check/something I can set somewhere that shows when includes aren't required?  Or is @agarny just really good at spotting them? :)

agarny:
There is https://include-what-you-use.org/. Probably something that we should add to our CI system, @hsorby?

agarny (dismissed):

hsorby (changes requested):
`std::unordered_set<std::string>` should be typedef'd using `using`.

kerimoyle:
> std::unordered_set<std::string> should be typedef'd using using.

Do you mean inside the `internaltypes.h`?  And (just for my info) what's the rationale?

hsorby:
It is unwieldy to run around with something like that better to make it into a type.

kerimoyle:
Even though it's used only once? I figured it was cleaner to be able to see what it was without having to look in a separate file for the definition.  Ah well.  Will move it.

hsorby:
It isn't used only once, it is used multiple times.  With a good typedef we won't need to look at anything as it will be self-explanatory.

hsorby:
It is also not required to go into `internaltypes.h` as (currently) this will only be used in the one file.

kerimoyle:
Should be ok now @hsorby ?

src/model.cpp
```diff
@@ -710,6 +761,21 @@ void flattenComponent(const ComponentEntityPtr &parent, const ComponentPtr &comp
...
+        auto tempModel = Model::create();
+        tempModel->addComponent(importedComponentCopy);
+        tempModel->linkUnits();
+        auto cnList = cnUnits(tempModel);
+        for (auto &name : cnList) {
+            if (!tempModel->hasUnits(name)) {
+                auto u = importModel->units(name);
+                if (u != nullptr) {
+                    requiredUnits.push_back(u);
+                } // TODO else warning
```
hsorby: Comment can go.

tests/model/model.cpp
```diff
@@ -552,3 +554,48 @@ TEST(Model, setAndCheckIdsAllEntities)
...
+    model->addComponent(c);
+
+    EXPECT_TRUE(model->hasUnresolvedImports());
+    model->resolveImports(resourcePath());
+    EXPECT_FALSE(model->hasUnresolvedImports());
+    model->flatten();
+
+    validator->validateModel(model);
+    EXPECT_EQ(size_t(0), validator->errorCount());
+    printErrors(validator);
```
hsorby: This should be deleted.

tests/model/model.cpp
```diff
@@ -552,3 +554,48 @@ TEST(Model, setAndCheckIdsAllEntities)
     const std::string a = printer->printModel(m);
     EXPECT_EQ(a, e);
 }
+
+TEST(Model, missingUnitsFromImportOfCnTerms)
```
hsorby, line 558: I feel like I'm missing the point of this test.  The `cn` term has a units defined in the model so why is it missing?  The missing here refers to something but I know it is not the clearly not missing units but it obviously is for something else which I can't figure out.  Anyone able to clarify what this is actually doing.  As far as I can tell the import is the same model?

hsorby (changes requested):
Minor changes and a question to enlighten me.

tests/model/model.cpp
kerimoyle: It might have been lost in the merge ... but the point of the test is to show that when we currently import a model which uses <cn> to define units, we haven't previously been importing those units too, just the ones associated with variables.

src/model.cpp
kerimoyle: As soon as we have an Issue class we can add the warnings ... this is just a reminder of where they'll be needed.

hsorby: So that should be in an issue then, no need for it to be in the codebase.

tests/model/model.cpp
hsorby: So the test shouldn't already have the units defined then?  If they units we need for the `cn` are already defined then we don't need to import them?

kerimoyle: The test should import a model which uses <cn> units that are not also used by a variable.  That was the point of the whole PR - we were only importing units used by variables, and ignoring those used exclusively in the <cn> tags.

hsorby: Yes I agree but this test doesn't show me that if you import a component from another model that makes use of a `cn` element in some math then the units for that `cn` are imported as well.

kerimoyle: OK, I've found the confusion and updated the comments for others too.  Basically this uses two methods to create a model - one via the `Parser` of the string that's specified in the test, and the other by importing the component from a file which contains exactly the same string.   The `Parser` was creating a valid model, but the import was missing out on importing the units.  The test just demonstrates that this bug is now fixed :)

kerimoyle:
... and once more for luck (crosses fingers ... )

tests/model/model.cpp
```diff
@@ -552,3 +554,57 @@ TEST(Model, setAndCheckIdsAllEntities)
...
```
hsorby, line 558: For me half of this test should be deleted, we only want/need to see the case where the bug was found i.e. the import case.

tests/model/model_flattening.cpp
```diff
@@ -657,6 +657,12 @@ TEST(ModelFlattening, unitsUsedByVariableNotInDirectlyImportedComponent)
         "  </encapsulation>\n"
         "</model>\n";
 
+    auto parser = libcellml::Parser::create();
+    auto e_model = parser->parseModel(e);
+    auto validator = libcellml::Validator::create();
+    validator->validateModel(e_model);
+    EXPECT_EQ(size_t(0), validator->errorCount());
```
hsorby: I don't see the purpose of this addition for this test.

hsorby (dismissed):

tests/model/model.cpp
kerimoyle: It was to show the difference between the parsing and the importing as their behaviour should be consistent.  I guess now it's fixed the parse half can go.

kerimoyle: (oops, replied to wrong thing).

tests/model/model_flattening.cpp
kerimoyle: Done.

hsorby:
I have added some more tests that mean we might need to do a little more work when considering units attached to cn elements.

kerimoyle:
Thanks @hsorby.  I'm confused though (what's new!):
In the test `missingUnitsFromImportOfCnTermsNotDefinedInImportedModel` why do you expect there to be no errors at the end?  I still get a message from the validitor (that I'd expect ...) saying "Error   1: Math has a cn element with a cellml:units attribute 'myUnitsThatIUse' that is not a valid reference to units in the model 'model' or a standard unit."  What am I missing ... ?

hsorby:
Sorry I left that test a little short.  I have added in further bits so that the issue is properly visible.

kerimoyle:
It's not that test I meant - it's the one above it.  I get this output, but I don't get what the comments suggest and miraculously find the units?  

```
[ RUN      ] Model.missingUnitsFromImportOfCnTermsNotDefinedInImportedModel
/Users/kmoy001/libcellml/libcellml-master/src/tests/model/model.cpp:630: Failure
Expected equality of these values:
  size_t(0)
    Which is: 0
  validator->errorCount()
    Which is: 1.         <<<<<<<<<<<< I think this is the correct output as the units don't exist???
Error   1: Math has a cn element with a cellml:units attribute 'myUnitsThatIUse' that is not a valid reference to units in the model 'model' or a standard unit.
[  FAILED  ] Model.missingUnitsFromImportOfCnTermsNotDefinedInImportedModel (33 ms)
```

kerimoyle:
Hi again.  
- I've added in a wee function that will swap out the duplicated units name from the mathml string for the numbered one (eg: from myUnits to myUnits_1 etc).  I know the thoughts on avoiding doing anything with the mathml as a string, but I can't see another way around it?  
- I still don't see the behaviour you're seeing in `missingUnitsFromImportOfCnTermsNotDefinedInImportedModel` ... but I've changed the test to be what I expect, and it passes.
- Re the final test, I've fixed the segfault which it was throwing, but I don't actually understand what you want to demonstrate with the test?  The expected values in your error counts aren't what I would expect in any case (when there are duplicated maths blocks the validator can't find the xml root ...?) which isn't really an import issue at all?  We don't seem to have any tests for this elsewhere, so I think that the multiple mathml blocks should be moved to the validator, **and** specified in the spec if it's actually not allowed ...

kerimoyle:
The tests should all pass now.  I've added a couple to satisfy myself that we don't duplicate the units on import if they're identical.  The final test is changed to reflect what I think is the expected behaviour (depending on whether multiple mathml blocks are allowed anywhere, not just imports).  If this isn't the case then I think it should be a separate issue and PR to this one as it's not really about cn unit imports anymore!
Ready for review again ... @hsorby @agarny @nickerso

kerimoyle:
I've (reluctantly) found a use case which doesn't do what I'm expecting.  This test fails because it creates a duplicate `per_millisecond_1` unit for the `beta_n` variable.  Have I done something wrong with the construction of the test?  I'm still not quite clear on when to use `linkUnits()`.

```
TEST(Model, importUnitsDuplicated)
{
    const std::string in =
        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
        "<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"myImportedModel\">\n"
        "  <units name=\"millisecond\">\n"
        "    <unit prefix=\"milli\" units=\"second\"/>\n"
        "  </units>\n"
        "  <units name=\"per_millisecond\">\n"
        "    <unit exponent=\"-1\" units=\"millisecond\"/>\n"
        "  </units>\n"
        "  <component name=\"myComponent\">\n"
        "    <variable name=\"t\" units=\"millisecond\"/>\n"
        "    <variable name=\"alpha_n\" units=\"per_millisecond\" initial_value=\"1\"/>\n"
        "    <variable name=\"beta_n\" units=\"per_millisecond\" initial_value=\"2\"/>\n"
        "  </component>\n"
        "</model>";

    const std::string expectedModelString =
        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
        "<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"model\">\n"
        "  <units name=\"millisecond\">\n"
        "    <unit prefix=\"milli\" units=\"second\"/>\n"
        "  </units>\n"
        "  <units name=\"per_millisecond\">\n"
        "    <unit exponent=\"-1\" units=\"millisecond\"/>\n"
        "  </units>\n"
        "  <component name=\"c\">\n"
        "    <variable name=\"t\" units=\"millisecond\"/>\n"
        "    <variable name=\"alpha_n\" units=\"per_millisecond\" initial_value=\"1\"/>\n"
        "    <variable name=\"beta_n\" units=\"per_millisecond\" initial_value=\"2\"/>\n"
        "  </component>\n"
        "</model>\n";

    // Create the model by parsing the string above.
    auto importedModel = parser->parseModel(in);

    auto validator = libcellml::Validator::create();
    validator->validateModel(importedModel);
    EXPECT_EQ(size_t(0), validator->errorCount());

    auto c = libcellml::Component::create("c");

    auto importSource = libcellml::ImportSource::create();
    importSource->setUrl("not_required_resolving_import_manually");
    importSource->setModel(importedModel);

    c->setImportReference("myComponent");
    c->setImportSource(importSource);
    model->addComponent(c);

    EXPECT_FALSE(model->hasUnresolvedImports());
    model->flatten();

    validator->validateModel(model);
    printErrors(validator);
    EXPECT_EQ(expectedModelString, printer->printModel(model));
}
```

Fails with:
```
[ RUN      ] Model.importUnitsDuplicated
/Users/kmoy001/libcellml/libcellml-src/tests/model/model.cpp:926: Failure
Expected equality of these values:
  expectedModelString
    Which is: "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"model\">\n  <units name=\"millisecond\">\n    <unit prefix=\"milli\" units=\"second\"/>\n  </units>\n  <units name=\"per_millisecond\">\n    <unit exponent=\"-1\" units=\"millisecond\"/>\n  </units>\n  <component name=\"c\">\n    <variable name=\"t\" units=\"millisecond\"/>\n    <variable name=\"alpha_n\" units=\"per_millisecond\" initial_value=\"1\"/>\n    <variable name=\"beta_n\" units=\"per_millisecond\" initial_value=\"2\"/>\n  </component>\n</model>\n"
  printer->printModel(model)
    Which is: "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"model\">\n  <units name=\"millisecond\">\n    <unit prefix=\"milli\" units=\"second\"/>\n  </units>\n  <units name=\"per_millisecond\">\n    <unit exponent=\"-1\" units=\"millisecond\"/>\n  </units>\n  <units name=\"per_millisecond_1\">\n    <unit exponent=\"-1\" units=\"millisecond\"/>\n  </units>\n  <component name=\"c\">\n    <variable name=\"t\" units=\"millisecond\"/>\n    <variable name=\"alpha_n\" units=\"per_millisecond\" initial_value=\"1\"/>\n    <variable name=\"beta_n\" units=\"per_millisecond_1\" initial_value=\"2\"/>\n  </component>\n</model>\n"
With diff:
@@ +5,7 @@
   </units>
   <units name=\"per_millisecond\">
+    <unit exponent=\"-1\" units=\"millisecond\"/>
+  </units>
+  <units name=\"per_millisecond_1\">
     <unit exponent=\"-1\" units=\"millisecond\"/>
   </units>
@@ -10,5 +13,5 @@
     <variable name=\"t\" units=\"millisecond\"/>
     <variable name=\"alpha_n\" units=\"per_millisecond\" initial_value=\"1\"/>
-    <variable name=\"beta_n\" units=\"per_millisecond\" initial_value=\"2\"/>
+    <variable name=\"beta_n\" units=\"per_millisecond_1\" initial_value=\"2\"/>
   </component>
 </model>\n

[  FAILED  ] Model.importUnitsDuplicated (1 ms)
```

kerimoyle:
After some more digging, this latest test failure comes from one of the bugs which (I think) has already been addressed in #554.  I'm not sure what to do now ...

kerimoyle:
I was wrong, it won't be fixed by the other PR.  Somewhere during cloning the units of the variables are being given different pointers, even though in the original model they're using the same units.  This means that  the usedUnits gets two copies, and hasUnits thinks they're distinct (but named identically):

```
VariablePtr Variable::clone() const
{
    auto v = create();

    // KRM This should search through the parent model to see whether the units already exist
    // Currently it's duplicating the units no matter what ... which causes the problems later.

    if (mPimpl->mUnits != nullptr) {
        v->setUnits(mPimpl->mUnits->clone());
    }
    v->setInitialValue(initialValue());
    v->setInterfaceType(interfaceType());
    v->setId(id());
    v->setName(name());

    return v;
}
```

hsorby:
If people could look at the tests I think this is working a lot better now.

kerimoyle:
Thanks @hsorby, it's way better :)  I had a question about function docstrings though - are these only needed for exposed functions?  I couldn't see the rationale for what was/wasn't documented just looking through the file.  If they're ok (or to be addressed separately when we do all the user-facing stuff), then I'm happy with this to be merged.  I can't review it though ... @agarny @nickerso ?

hsorby:
We are trying to document all functions if you do see one it is something to say in a review comment.  This sort of documentation is important for further down the track so we know what a function/method is meant to be doing.

The official documentation only shows the API documentation and that needs to be really good as that is what users will have to use.

hsorby (dismissed):
@kerimoyle and I approve these changes.

src/xmlnode.h
```diff
@@ -130,6 +130,16 @@ class XmlNode
      */
     bool isElement(const char *name, const char *ns) const;
 
+    /**
+     * @brief Test if this @c XmlNode is an Xml element node.
+     *
+     * Test if this @c XmlNode is an Xml element node.
+     *
+     * @return @c true if this @c XmlNode is an Xml element node and @c false
```
agarny: Replace `Xml` with `XML`.

src/xmlutils.h
```diff
@@ -48,4 +50,15 @@ XmlNamespaceMap determineMissingNamespaces(const XmlNamespaceMap &namespaceMap1,
  */
 XmlNamespaceMap traverseTreeForUndefinedNamespaces(const XmlNodePtr &node);
 
+/**
+ * @brief Turn xml content with potentially multiple root elements in a vector of documents.
+ *
+ * Parse the @p content and try to find Xml documents returning a vector of any Xml documents
+ * found.
+ *
+ * @param content The string to parse for Xml roots.
```
agarny: Replace `xml`/`Xml` with `XML`.

src/xmlnode.cpp
```diff
@@ -204,6 +209,12 @@ bool XmlNode::hasAttribute(const char *attributeName) const
     return found;
 }
 
+xmlNsPtr getAttributeNamespace(const xmlNodePtr &node, const char *attributeName)
+{
+    xmlAttrPtr attribute = xmlHasProp(node, reinterpret_cast<const xmlChar *>(attributeName));
+    return attribute->ns;
```
agarny: Not a big deal (I guess), but I would simplify this with:

```
return xmlAttrPtr attribute = xmlHasProp(node, reinterpret_cast<const xmlChar *>(attributeName))->ns;
```

src/utilities.h
```diff
@@ -430,4 +432,63 @@ bool isStandardPrefixName(const std::string &name);
...
+ * @brief Trim whitespace from the beginning and end of a string.
+ *
+ * Remove whitespace from the beginning and end of a string
+ * returning the result.
+ *
+ * @param s The @c std::string to trim.
+ *
+ * @return The trimmed string.
+ */
+static inline std::string trim_copy(std::string s)
```
agarny: This should be renamed to `trimCopy()`.

src/utilities.h
```diff
@@ -430,4 +432,63 @@ bool isStandardPrefixName(const std::string &name);
...
 size_t getVariableIndexInComponent(const ComponentPtr &component, const VariablePtr &variable);
 
+/**
+ * @brief Trim whitespace from the front of a string (in place).
+ *
+ * Remove whitespace from the front of a string, modifying the passed string.
+ *
+ * @param s The @c std::string to trim.
+ */
+static inline void ltrim(std::string &s)
```
agarny: This probably ought to be renamed `lTrim()` or `leftTrim()`?

src/utilities.h
```diff
@@ -430,4 +432,63 @@ bool isStandardPrefixName(const std::string &name);
...
+}
+
+/**
+ * @brief Trim whitespace from the end of a string (in place).
+ *
+ * Remove whitespace from the end of a string, modifying the passed string.
+ *
+ * @param s The @c std::string to trim.
+ */
+static inline void rtrim(std::string &s)
```
agarny: This probably ought to be renamed `rTrim()` or `rightTrim()`?

src/model.cpp
```diff
@@ -656,19 +749,56 @@ ComponentNameMap createComponentNamesMap(const ComponentPtr &component)
...
+        units->unitAttributes(index, ref, pre, uExp, expMult, id);
+        if (!isStandardUnitName(ref)) {
+            auto u = model->units(ref);
+            if (u != nullptr) {
+                auto requiredUnitsUnits = referencedUnits(model, u);
+                requiredUnits.insert(requiredUnits.end(), requiredUnitsUnits.begin(), requiredUnitsUnits.end());
+                requiredUnits.push_back(u);
+            }
+        }
+    }
```
agarny: We really ought to use more meaningful variable names (e.g. `u` should be replaced with `units`). This is API code, not test code.

tests/model/model.cpp
```diff
@@ -552,3 +554,171 @@ TEST(Model, setAndCheckIdsAllEntities)
...
+        "  <units name=\"per_millisecond\">\n"
+        "    <unit exponent=\"-1\" units=\"millisecond\"/>\n"
+        "  </units>\n"
+        "  <component name=\"myComponent\">\n"
+        "    <variable name=\"alpha_n\" units=\"per_millisecond\" initial_value=\"1\"/>\n"
+        "    <variable name=\"beta_n\" units=\"per_millisecond\" initial_value=\"2\"/>\n"
+        "  </component>\n"
+        "</model>";
+
+    const std::string expectedModelString =
```
agarny: Consistency: replace `expectedModelString` with `e`? Also, move it at the beginning of the test?

agarny (changes requested):
Looks good to me pending some minor issues.

src/model.cpp
kerimoyle: I don't get it ... I've just changed what I thought was a good variable name into "e" at your suggestion? Which is actually better?

tests/model/model.cpp
kerimoyle: Done (though see comment above)

src/utilities.h
kerimoyle: Done

src/utilities.h
kerimoyle: Done

src/utilities.h
kerimoyle: Done

tests/model/model.cpp
agarny: This is test code, so we have been using short names. Not my personal preference, but we ought to be consistent.

src/model.cpp
agarny: This is API code, so we ought to use meaningful names while for test code we tend to use short variable names (see below).

kerimoyle: Okeydokes, done.

src/xmlutils.h
kerimoyle: Done

src/xmlnode.h
kerimoyle: Done

src/xmlnode.cpp
kerimoyle: Done

agarny (approved):

hsorby (approved):

Issue #525, opened 2020-01-22
Bug in how standardUnitMultipliers is used in validator


kerimoyle:
There's been a bug in how we test for multiplier mismatch where there are `standardUnitMultipliers` involved.  This test below should return a warning because there are 1000L per m^3, but it was returning no mismatch.  I'll put the super-quick fix in a new PR.

```
TEST(Validator, unitMultiplierFactorDifference)
{
    libcellml::ValidatorPtr validator = libcellml::Validator::create();
    libcellml::ModelPtr m = createModelTwoComponentsWithOneVariableEach("m", "c1", "c2", "v1", "v2");
    auto c1 = m->component(0);
    auto c2 = m->component(1);
    auto v1 = c1->variable(0);
    auto v2 = c2->variable(0);

    v1->setUnits("litre");

    libcellml::UnitsPtr u = libcellml::Units::create();
    u->setName("big_barrel");
    u->addUnit("metre", 3.0);
    m->addUnits(u);

    v2->setUnits(u);

    libcellml::Variable::addEquivalence(v1, v2); // litre ~= metre^3 // Nope: there are 1000L = 1m^3

    validator->validateModel(m);

    printIssues(validator);

    std::cout << "Issues: " << validator->issueCount() << std::endl;
    std::cout << "Errors: " << validator->errorCount() << std::endl;
    std::cout << "Warnings: " << validator->warningCount() << std::endl;
    std::cout << "Hints: " << validator->hintCount() << std::endl;

    EXPECT_EQ(size_t(1), validator->issueCount());  // << no issues reported, but there should have been
    EXPECT_EQ(size_t(0), validator->errorCount());
    EXPECT_EQ(size_t(1), validator->warningCount());

    EXPECT_EQ_ISSUES(expectedErrors, validator);
}
```

PR #521, opened 2020-01-15
Creating Issue class to replace Error class

kerimoyle:
As discussed in various threads:
#314 
#368 
#493 
#514 

... we'd like to have an Issue class to replace the current Error class, in order to allow the return of non-fatal error messages and more helpful warnings and information.  This would include situations in which the cellml models does not violate the specifications but is still nonsense (would not run as a simulation, contains mismatched units in the mathematics, or mismatched multipliers in the units, etc).

It's also a good place to review the construction of our descriptions to ensure that they're all consistent and helpful.

Here's the current list of things which need a non-fatal message to be added:
-  multiplier mismatch in equivalent units
-  multiplier mismatch in maths
-  units mismatch in maths
-  parsed model parts which will later cause validation errors (?) as in #302
-  cyclic or recursive imports (?) as in #280
-  exponents with units (should be dimensionless) as in #509
-  maths statements which are not assignments (ie: valid MathML/CellML but not runable)
-  anything which will cause issues in the Generator, as in #492, #493, #499,

... anything else ... ?

PR #526, opened 2020-01-22
Fix bug in how standardMultipliers are used in the validator

kerimoyle:
Addresses #525


kerimoyle:
Review please @hsorby @awickens @agarny @nickerso

agarny:
I can see that you fixed something, but... without having had a proper look at the underlying problem, why no test has been modified?

kerimoyle:
Because no test can exist at this stage - the bug is in something which is currently a warning, not an error.  Perhaps see the equivalent fix in action on the other branch (where there is a test?) #521

agarny:
Ok, we probably ought to ignore this PR and focus on PR #521 instead.

kerimoyle:
Closing as this fix is now in #554 .

PR #528, opened 2020-01-23
I314 import warnings

kerimoyle:
For demonstration/discussion only!  This branch comes from #521 so is here only to show what I meant about the `Logger/Issue` thing in action.  See: https://github.com/cellml/libcellml/pull/521#issuecomment-577617784

This implements option 3 and uses it to test for cyclical imports in the `resolveImports` function.


kerimoyle:
See #530 - Instead of attaching a `Logger` to the `Model` class, discussions with @hsorby suggest separating the problematic functions `resolveImports()` and `flatten()` to a separate `Importer` class instead, which will have a `Logger` attached.

kerimoyle:
Closing in favour of #530

Issue #527, opened 2020-01-23
Confused about resolving imports ...


kerimoyle:
I can see the rationale of the recursive call to `resolveImports` to create a model against which to test later, but instead of checking/resolving **only** the item which is being imported, we're resolving the entire thing.  Is this intended?  And if so, how do tests like `TEST(ResolveImports, resolveImportsFromFileLevel0)` manage to "resolve" a model which contains so many broken/nonexistent components?

The reason I'm asking is because I'm working on the warnings for cyclical imports - which works as intended - but make the above test fail, and I can't figure out which test isn't doing what it should be!

```
void resolveImport(const ImportedEntityPtr &importedEntity,
                   const std::string &baseFile)
{
    if (importedEntity->isImport()) {
        ImportSourcePtr importSource = importedEntity->importSource();
        if (!importSource->hasModel()) {
            std::string url = resolvePath(importSource->url(), baseFile);
            std::ifstream file(url);
            if (file.good()) {
                std::stringstream buffer;
                buffer << file.rdbuf();
                ParserPtr parser = Parser::create();
                ModelPtr model = parser->parseModel(buffer.str());
                importSource->setModel(model);
                model->resolveImports(url);  // << this bit will then iterate through *all* the components in the imported model, rather than only getting the one of interest?
            }
        }
    }
}
```

hsorby:
Because `resolveImports` is resolving the whole model and not just the required parts of it then yes it is intended.

`hasUnresolvedImports` only checks down the line of imported components, not the whole model.

kerimoyle:
So why the difference in approach between the two?  I would have expected them to do the same thing, whether that was resolving the whole lot or just the component.  One use case/test I can see giving odd results is if we have a setup like the one below.  In this case, resolving all the components (including c1) will create a circular reference ... but really we shouldn't care as we only wanted to import c4 anyway?

```
separatemodel
- c imports c4 from model 1

model1
 - c1 imports c2 from model2
 - c4 is concrete here

model2 
- c2 imports c3 from model3

model3
- c3 imports c1 from model1   // << cycle created here with c1, 2, 3, 1
```
In my circular checking modifications I remove unrelated components from the imported model.  (I haven't touched units yet).  Do you see this bit (commented below) causing conceptual problems somewhere? 

```
    bool resolveImport(const ImportedEntityPtr &importedEntity,
                   const std::string &destination,
                   const std::string &baseFile,
[...]

PR #529, opened 2020-01-24
Test showing segfault in model->flatten()

kerimoyle:
Demonstrating segfault caused by resolve imports for #527


kerimoyle:
Closing as no longer needed now that the `Importer` class is on the way.

Issue #530, opened 2020-01-28
Create an Importer class to enable error logging when resolving imports


kerimoyle:
After discussions yesterday with @hsorby I'm going to have a go at writing a `Debugger/Helper/Checker/Validator` class which will include the current `Validator` behaviour but also allow for wider checking of the model.  As part of this tweak, there will be a new `Flattener/Importer/??` class into which the current `Model::flatten() `and `Model::resolveImports() `functions and their dependents will be moved.  This `Importer` class will have its own logger, so we'll be able to catch things like the current segfaults/recursive depth errors caused by circular imports (see #527, #528, #529) without needing to attach a logger to the `Model` itself (see https://github.com/cellml/libcellml/pull/521#issuecomment-577617784 )

agarny:
Hmm, indeed, it looks like a reasonable way forward.

PR #531, opened 2020-01-28
Creating Importer class to handle resolving imports and flattening models

kerimoyle:
Addresses #530 

NB this branch sits behind the existing PR #521 so should be merged **after** that one. **Edit** #521 is merged now.


kerimoyle:
@hsorby Here's my first go following our conversation yesterday (NB it's on top of the earlier PR which is waiting for review, so should wait for that one).  I've created an `Importer` class with its own `Logger` and can use it to warn of circular imports when resolving a model.  This has meant moving some local functions out of the `Model` and into the `Importer` ... I played for a bit, but couldn't come up with any kind of overall rationale other than what's here.  Happy to rejig if you can see something I've missed.  Any thoughts/suggestions at this stage?

kerimoyle:
Closing in favour of #608

Issue #534, opened 2020-01-29
removeComponent causes segfault in equivalentVariables


kerimoyle:
I'm looking for some clarity around the "remove" functions.  
1) As far as I can tell, they're one sided (ie: they remove the model's record of the component, but not the component's connection to its parent model), and they don't address any of the other connections that could exist (the equivalent variables, for example).  Is this intentional?  Or should we be also detaching parent/ownership stuff and removing variable equivalences when a component is removed too?  

2) This has exposed something I don't understand in the parser/printer.  I can construct a model using the API, remove a component, and everything is fine.  I can take exactly the same model created by the `Parser`, remove a component and get a segfault (see test below, also in #535).

```
TEST(Model, removeComponentsAPIModel)
{
    // Create two identical models - one through the API, one through the parser
    auto model = libcellml::Model::create("model");
    auto c1 = libcellml::Component::create("c1");
    auto c2 = libcellml::Component::create("c2");

    auto v1 = libcellml::Variable::create("v1");
    auto v2 = libcellml::Variable::create("v2");
    v1->setUnits("dimensionless");
    v2->setUnits("dimensionless");
    c1->addVariable(v1);
    c2->addVariable(v2);
    model->addComponent(c1);
    model->addComponent(c2);

    libcellml::Variable::addEquivalence(v1, v2);

    // Printing this model in order to create a second identical one, printedModel
    auto printer = libcellml::Printer::create();
    auto parser = libcellml::Parser::create();
    auto printedModelString = printer->printModel(model);
    auto printedModel = parser->parseModel(printedModelString);
```

Printing to the terminal gives:
-------------------BEFORE--------------------
Component 0 0x7fa0a6000760 c1
  Variable 0 v1 has 1 equiv variables:
      - address 0 0x7fa0a6000f80
      - name 0 v2
      - parent 0 0x7fa0a60000d0
Component 1 0x7fa0a60000d0 c2
  Variable 0 v2 has 1 equiv variables:
      - address 0 0x7fa0a6000500
      - name 0 v1
      - parent 0 0x7fa0a6000760

Now remove component 1 and print to the terminal again:
```
    model->removeComponent("c1");
```
-------------------AFTER--------------------
Component 0 0x7fa0a60000d0 c2
  Variable 0 v2 has 1 equiv variables:
      - address 0 0x7fa0a6000500
      - name 0 v1
      - parent 0 0x7fa0a6000760
  
[...]

PR #535, opened 2020-01-29
Test to demonstrate segfault in remove component

kerimoyle:
Added test to demonstrate the behaviour described in #534


kerimoyle:
Thanks @hsorby.  Should we be removing the equivalent variable connections when removing a component?  This is what was causing the segfault, not just the equivalent variable count.  Changing the test as in 3d84826 takes away the demonstration of that behaviour ...

hsorby:
I only had the one segfault from the for loop on equivalent variable count.

hsorby:
> Should we be removing the equivalent variable connections when removing a component?
No we shouldn't.

kerimoyle:
> Should we be removing the equivalent variable connections when removing a component?
> No we shouldn't.

I don't understand.  Why does the equivalent variable count change when we remove a component if we're not changing the connections at all?  Isn't that what your test is showing? (I added the first four lines)

```
EXPECT_EQ(size_t(1), model->component(0)->variable(0)->equivalentVariableCount());
EXPECT_NE(nullptr, model->component(0)->variable(0)->equivalentVariable(0));

EXPECT_EQ(size_t(1), model->component(1)->variable(0)->equivalentVariableCount());
EXPECT_NE(nullptr, model->component(1)->variable(0)->equivalentVariable(0));

model->removeComponent("c1");

// I don't understand this line if the connection is supposed to remain? Shouldn't this stay at 1?
EXPECT_EQ(size_t(0), model->component(0)->variable(0)->equivalentVariableCount());
EXPECT_EQ(nullptr, model->component(0)->variable(0)->equivalentVariable(0));
```

kerimoyle:
I've gone through the source code some more and I think that my confusion comes from the fact that we use "removeSomething" to mean "remove it AND delete what it pointed to", when I'd imagined that it simply removed the pointer from the list.  Since we don't create the object when calling the "addSomething" function, this is confusing!

The function names and documentation don't mention that the original object will be deleted, and it *really* limits the way that people can create and reuse objects through the API.  You couldn't have two models which used the same component or units, for example, because removing a set of units from one model would delete it, and cause problems in any other model which referenced them.  I'll try and make a test to show you what I mean.  

Would a clearer way be to have the `add/removeSomething` functions only adding/removing from the list of pointers in the object: **not** touching the contents of the pointers themselves.   We can then have a set of `deleteSomething` functions which could do both at once.  At least then the add/remove behaviour is more predictable, and allows the user to move things around as they want to without deleting them without realising it ...?

hsorby:
I don't agree, if you don't retain a handle to an object that you are removing then it will be deleted.  This is pretty standard behaviour.

We also have the `take*` methods which gives the above behaviour.  The user must decide whether they want to remove something or take it.  Remove to me is doing what I would expect.

kerimoyle:
Somewhere along the line the behaviour that was buggy has been fixed.  Not here, but somewhere ... so this PR doesn't really need to go in anymore unless we just want an extra test.

hsorby (approved):

nickerso (approved):

Issue #532, opened 2020-01-28
Superfluous warnings from ranlib about libcellml.a having no symbols on macOS


agarny:
It doesn't seem to be possible (anymore?) to build a static version of libCellML. Here is what I tried and got on macOS:

```
cd <libCellML>
mkdir build && cd build && cmake -G Ninja -DLIBCELLML_BINDINGS_PYTHON=OFF -DLIBCELLML_BUILD_SHARED=OFF -DLIBCELLML_BUILD_TYPE=Release -DLIBCELLML_COVERAGE=OFF -DLIBCELLML_LLVM_COVERAGE=OFF -DLIBCELLML_UNIT_TESTS=OFF .. && ninja
-- The CXX compiler identification is AppleClang 11.0.0.11000033
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Python: /usr/local/Frameworks/Python.framework/Versions/3.7/bin/python3.7 (found version "3.7.6") found components: Interpreter Development
-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)
-- Found Sphinx: /usr/local/opt/sphinx-doc/bin/sphinx-build
-- Found SWIG: /usr/local/bin/swig (found suitable version "4.0.1", minimum required is "3")
-- Performing Test LLVM_COVERAGE_COMPILER_FLAGS_OK
-- Performing Test LLVM_COVERAGE_COMPILER_FLAGS_OK - Success
-- Performing Test GCC_COVERAGE_COMPILER_FLAGS_OK
-- Performing Test GCC_COVERAGE_COMPILER_FLAGS_OK - Success
-- Found LibXml2: /usr/lib/libxml2.dylib (found version "2.9.4")
-- Performing Test COMPILER_HAS_HIDDEN_VISIBILITY
-- Performing Test COMPILER_HAS_HIDDEN_VISIBILITY - Success
-- Performing Test COMPILER_HAS_HIDDEN_INLINE_VISIBILITY
[...]

PR #536, opened 2020-01-29
Alternative: Prevent superfluous warnings from ranlib about libcellml.a having no symbols on macOS

hsorby:
Addresses #532. 

Rather than playing around with compiler flags it may be better to exclude the debug files from the build unless explicitly enabled.  These functions are not used except when debugging.


agarny (approved):
Looks good and works well for me.

nickerso (approved):

Issue #270, opened 2019-01-10
Add some status information to README


MichaelClerx:
We should let people know the library isn't ready for use yet on the main page I think

@nickerso ?

nickerso:
I keep telling people that for the things it is capable of, that it is definitely preferable to use libCellML than try to learn to use the CellML API (for example, @dewancse has been able to get some model composition code up and running in Python with minimal assistance and much quicker than if he had to start from scratch with the CellML API). And it is good to have non-core developers testing things out, I think.

But I agree that adding some status information prominently to the README would be useful. Perhaps a comment to make sure it is clear that libCellML should be considered alpha until we make a 1.0 release? And some kind of summary about what we know has not been implemented (links to tracker items?)?

agarny:
> But I agree that adding some status information prominently to the README would be useful. Perhaps a comment to make sure it is clear that libCellML should be considered alpha until we make a 1.0 release? And some kind of summary about what we know has not been implemented (links to tracker items?)?

Will look into it.

agarny:
FWIW, before we can deal with this issue, I think we will need to handle PR #267 first.

PR #540, opened 2020-02-16
Added some status information to README.rst

agarny:
Addresses issue #270.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,6 +7,8 @@ The prime libCellML repository is located at https://github.com/cellml/libcellml
 
 The documentation for the libCellML project is available at https://libcellml.readthedocs.io/.
 
+ALTHOUGH libCellML IS ALREADY BEING USED SUCCESSIVELY, IT SHOULD BE CONSIDERED ALPHA UNTIL VERSION 1.0 IS RELEASED.
+
 Scope
 -----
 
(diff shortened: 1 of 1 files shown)


hsorby (commented):
I probably wouldn't go with the all-caps text and I do feel this should be evident through the use of the version numbers and a reference to semantic versioning.

agarny:
That's the thing, based solely on the version number, I wouldn't normally use libCellML (I mean, version 0.1/0.2...!). Yet, it can already be useful to people. It all depends on what you want to do with it. However, one should indeed keep in mind that the API may change at any stage while we haven't reached version 1.0 (as it has on several occasions already).

Anyway, let's see whether @nickerso and @MichaelClerx's view has changed on this.

nickerso:
maybe worth looking at adding some badges to the repo/readme instead? Must be some kind of version info badge?

Otherwise agree with @hsorby's suggestion to not use ALL-CAPS :)

agarny:
The caps were, obviously, used on purpose: to make sure that people don't overlook it (how many people have read the README.rst?).

Not sure about badges: too much hassle for something so trivial.

MichaelClerx (dismissed):
I don't even mind the all-caps :-)

Never seen a badge for development status, wouldn't look there myself - Would much prefer writing at the very start that this is not ready for use, instead of writing as if it's ready but with some hidden disclaimer among the badges :D

nickerso (approved):

nickerso:
whoops, forgot that making changes would dismiss reviews - I just changed one word and happy to accept @MichaelClerx's earlier approval as still valid...

agarny:
I think it's fair enough, so FWIW happy for you to force the merge.

Issue #542, opened 2020-02-18
Prefix (for units): the enumeration should be moved to `units.h`


agarny:
Right now, the `Prefix` enum class is defined in `[libCellML]/src/api/libcellml/enumerations.h` while I would expect it to be defined in `[libCellML]/src/api/libcellml/units.h`, as is the case for `StandardUnit` enum class. This means that, in the end, there is no need for `[libCellML]/src/api/libcellml/enumerations.h`.

kerimoyle:
I can't see a use for `Prefix` outside of `Units` ... but which way makes more sense?  To move the `StandardUnit` to `enumerations.h` or vv?

agarny:
`StandardUnit` and `Prefix` are related to `Units`, so for me they should be in `units.h`. Simple as that. (I "hate" having to look in different places for things that are related.)

agarny:
Closing following the merging of PR #543.

PR #543, opened 2020-02-18
Prefix (for units): the enumeration should be moved to `units.h`

agarny:
Addresses issue #542.


hsorby (approved):
I'll approve these changes.  

With the proviso that I was under the impression that the prefix was put into enumerations as it was (at some point) seen as standalone definitions that aren't actually tied to libCellML units.  Up to this point we have only used them in Units.

agarny:
Could a prefix be used somewhere else than in `Units`? Independent of that, right now, we only use it in `Units`, so this is where I would expect it to be. Then, if at some point, we were to use it somewhere else then, sure, to have it in `enumerations.h` would make sense.

kerimoyle (approved):
Yep, happy with this :)

Issue #545, opened 2020-02-18
Variable: wrongly return a Units object with no Unit children


agarny:
While working on issue #409, I have a case where I have a variable expressed in seconds and another expressed in milliseconds.

For the first variable, libCellML returns a `Units` object with `second` as a name and no `Unit` children. For the second variable, libCellML returns a `Units` object with `millisecond` as a name and one `Unit` child based on the standard unit `second` and `milli` as a prefix.

Everything is fine regarding the second variable's unit, but not the unit of the first variable. Indeed, a `Units` object should only be used to define a new unit, not to return an existing (standard) one.

@hsorby, I think this one might be for you... :)

PR #547, opened 2020-02-19
Dealing with standard units.

hsorby:
This PR looks at defining the way units are handled that are standard units.  The two tests show the new behaviour.

Addresses issue #545.


kerimoyle:
See also #525 and pr #526?

src/printer.cpp
```diff
@@ -200,6 +200,8 @@ std::string Printer::PrinterImpl::printUnits(const UnitsPtr &units) const
             repr += " id=\"" + units->id() + "\"";
         }
         repr += "/></import>";
+    } else if (isStandardUnit(units)) {
+        // Do nothing
```
agarny: Missing full stop at the end of the comment.

src/utilities.cpp
```diff
@@ -431,6 +432,15 @@ bool isStandardUnitName(const std::string &name)
     return standardUnitsList.count(name) != 0;
 }
 
+bool isStandardUnit(const UnitsPtr &units)
+{
+    bool is = false;
+    if (units != nullptr) {
+        is = units->isBaseUnit() && isStandardUnitName(units->name());
+    }
+    return is; //standardUnitsList.count(name) != 0;
```
agarny: Why that comment? I would delete it.

Having said that, I would rewrite the whole method to simply read:

```
bool isStandardUnit(const UnitsPtr &units)
{
    if ((units != nullptr) && units->isBaseUnit() && isStandardUnitName(units->name())) {
        return true;
    }

    return false;
}
```

src/utilities.h
```diff
@@ -390,6 +390,8 @@ void removeComponentFromEntity(const EntityPtr &entity, const ComponentPtr &comp
  */
 bool isStandardUnitName(const std::string &name);
 
+bool isStandardUnit(const UnitsPtr &units);
```
agarny, line 404: Need some documentation for this new method.

tests/printer/printer.cpp
```diff
@@ -235,3 +235,20 @@ TEST(Printer, printModelWithTabs)
...
 }
+
+TEST(Printer, printModelWithStandardUnitsAdded)
+{
+    const std::string e =
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"my_model\"/>\n";
+
+    libcellml::UnitsPtr u1 = libcellml::Units::create("second");
+    libcellml::ModelPtr model = libcellml::Model::create("my_model");
```
agarny, line 245: Minor: might want to swap those two lines? (Just feels "weird" to create a units before a model.)

agarny (changes requested):
Looks fine to me pensing some minor issues.

agarny (approved):

nickerso (approved):

Issue #552, opened 2020-02-21
Bug in equivalent units with dimensionless 


kerimoyle:
When the difference between two units is a `dimensionless` unit, the units are not equivalent, so this test fails.  I think that's not what we want, but not sure if others think otherwise?

```
TEST(Units, customUnitsScalingFactorDimensionless)
{
    auto m = libcellml::Model::create("model");
    auto u1 = libcellml::Units::create("u1");

    m->addUnits(u1);

    auto u2 = libcellml::Units::create("u2");
    u2->addUnit("u1");
    u2->addUnit("dimensionless");

    m->addUnits(u2);

    m->linkUnits();

    EXPECT_TRUE(libcellml::Units::baseEquivalent(u1, u2)); // <<< base units are not the same
    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u2));
    EXPECT_TRUE(libcellml::Units::equivalent(u1, u2));
}
```

hsorby:
I think this should be equivalent.

hsorby:
I have also been wondering how a user might be able to set their own units as dimensionless so that comparisons to `radians` would work.

kerimoyle:
Since `dimensionless` is (or was?) one of the 7+1 base units, it should be able to be used in the same way as others (eg `metre` etc)?  The only difference is that the exponent shouldn't be counted.  Something like radians could be specified as `m/m`, or steradians m^2/m^2 and I think that that should return true in a comparison to dimensionless?

kerimoyle:
Try it out on #554 where this issue was fixed and see ...

Issue #550, opened 2020-02-20
Bug in scalingFactor with custom base units


kerimoyle:
The test below fails:

```
TEST(Units, customUnitsScalingFactorSimple)
{
    auto m = libcellml::Model::create("model");
    auto u1 = libcellml::Units::create("u1");
    m->addUnits(u1);
    auto u2 = libcellml::Units::create("u2");
    u2->addUnit("u1");
    m->addUnits(u2);

    m->linkUnits();

    EXPECT_TRUE(libcellml::Units::baseEquivalent(u1, u2)); // passes
    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u2));  // fails, returns 0.0
    EXPECT_TRUE(libcellml::Units::equivalent(u1, u2)); // fails
}
```
... because of the the line below:
```
double Units::scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2)
{
    if (!Units::baseEquivalent(units1, units2)) {
        return 0.0;
    }
    bool updateUnits1 = false;
    bool updateUnits2 = false;
    if ((units1 != nullptr) && (units2 != nullptr)) {
        if ((units1->unitCount() != 0) && (units2->unitCount() != 0)) { 
// <<<<< the unitCount for the base unit is 0, so this is skipped and the scaling factor is returned as 0
            double multiplier = 0.0;

            updateUnits1 = updateUnitMultiplier(multiplier, units2, 1, 0, 1);
            updateUnits2 = updateUnitMultiplier(multiplier, units1, 1, 0, -1);

            if (updateUnits1 && updateUnits2) {
                return std::pow(10, multiplier);
            }
        }
        if (units1->name() == units2->name()) {
            return 1.0;
        }
    }
    return 0.0; // <<<<< returns 0.0 when a base unit is compared
}
```

I'll include a fix for this at the same time as the other scalingFactor/dimensionality stuff.

hsorby:
This has already been addressed in #547.

kerimoyle:
> This has already been addressed in #547 

Alas, no, the test still fails.  It's not about the standard unit stuff, it's the custom units that are causing it.

kerimoyle:
The fix for this is now in PR #554 in the interests of getting it in without depending on the `Issue` class existing.

kerimoyle:
Closing this PR in favour of #554.

Issue #541, opened 2020-02-17
Units: scalingFactor to check dimensionality?


agarny:
One of our units-based tests reads:

```
TEST(Units, compareMultiplierSimple)
{
    // u1 = 1000*u2
    libcellml::UnitsPtr u1 = libcellml::Units::create();
    u1->setName("u1");
    u1->addUnit("metre", 0, 1.0, 1000.0);
    libcellml::UnitsPtr u2 = libcellml::Units::create();
    u2->setName("u2");
    u2->addUnit("metre", 0, 1.0, 1.0);

    EXPECT_EQ(0.001, libcellml::Units::scalingFactor(u1, u2)); // 10^-3.0 factor difference
    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(u2, u1)); // 10^3.0 factor difference
}
```

It all works fine, but now if we modify the test to read (i.e. copied/pasted `u2->addUnit("metre", 0, 1.0, 1.0);`):

```
TEST(Units, compareMultiplierSimple)
{
    // u1 = 1000*u2
    libcellml::UnitsPtr u1 = libcellml::Units::create();
    u1->setName("u1");
    u1->addUnit("metre", 0, 1.0, 1000.0);
    libcellml::UnitsPtr u2 = libcellml::Units::create();
    u2->setName("u2");
    u2->addUnit("metre", 0, 1.0, 1.0);
    u2->addUnit("metre", 0, 1.0, 1.0);
    u2->addUnit("metre", 0, 1.0, 1.0);

    EXPECT_EQ(0.001, libcellml::Units::scalingFactor(u1, u2)); // 10^-3.0 factor difference
    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(u2, u1)); // 10^3.0 factor difference
}
```

then we can see that the test still passes while it clearly shouldn't. @kerimoyle, any views on this?

kerimoyle:
Hi @agarny.  I agree - I think that the test should fail with a scaling factor of 0 as the base units are no longer the same (m vs m^3).  Not sure where the difference is coming from ... @awickens do you have any thoughts?

kerimoyle:
Neither the `dimensionallyEquivalent` nor the `equivalent` functions return true, so I think that this should be the first check inside the `scalingFactor` function: if the base units are not the same then the scaling factor returns 0.

awickens:
[...]

Issue #539, opened 2020-01-30
Need to clarify documentation/names of units and scaling factors?


kerimoyle:
1) This is misleading:
```
    /**
     * @brief Return the scaling factor difference between two @c Units.
     *
     * This can be interpreted as factor, where units2 = (factor)*units1.  This method
     * does not check to see if the units are compatible.
     *
     * @param units1 The first units to compare.
     * @param units2 The second units to compare.
     *
     * @return The factor units1/units2. <<< This bit is wrong
     */
    static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2);
```
- the return and the equation given say different things ...
- the method returns 0 when the units do not have the same base?  What does "does not check to see if the units are compatible" mean?

2) I'm not sure about the naming of the `equivalent` and `dimensionallyEquivalent` functions.  I'm happy with them, just not the names.  The "dimensional" bit of "dimensional analysis" refers to the base units, not the scaling factor ... so I would have thought that these names should be around the other way?

hsorby:
> * @return The factor units1/units2. <<< This bit is wrong

Should be an easy fix.

hsorby:
> * the method returns 0 when the units do not have the same base?  What does "does not check to see if the units are compatible" mean?

The documentation could be changed or the implementation.  If it is working this way then I guess the documentation should be changed to match.

hsorby:
> so I would have thought that these names should be around the other way?

I agree, the names should be swapped.

kerimoyle:
Related to #541

PR #554, opened 2020-02-21
i539, i541, i550, i552 dimensional checks and scaling factor bugs

kerimoyle:
This PR addresses a few small inconsistencies found in the equivalent variables testing and scaling factors.  The changes are:
- the two function names `Units::equivalent()` and `Units::dimensionallyEquivalent()` are now replaced with `Units::compatible()` and `Units::equivalent()` addressing #539 
- two `Units` must be dimensionally equivalent (ie: baseEquivalent = True) before a scaling factor is returned, otherwise it returns 0, addressing #541 
- the bug where a "dimensionless" mismatch between two units made them not-equivalent is fixed, addressing #552 
- the bug where custom base units were wrongly returning a scaling factor of 0 is fixed, addressing  #550


hsorby:
I have been thinking about this and I think we should change the names of the functions to be
 - `equivalent` for deciding if units are exactly equivalent (this one doesn't change obviously).
 - `compatible` for deciding if units only differ by scale.

kerimoyle:
`compatible` is much nicer, thanks.

src/api/libcellml/units.h
```diff
@@ -442,7 +450,7 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
      *
      * @return @c true if the two @c Units are equivalent, @c false otherwise.
      */
-    static bool equivalent(const UnitsPtr &units1, const UnitsPtr &units2);
+    static bool baseEquivalent(const UnitsPtr &units1, const UnitsPtr &units2);
```
hsorby: I think this method should be called `compatible`.

src/units.cpp
```diff
@@ -457,43 +459,99 @@ void updateUnitsMap(const UnitsPtr &units, UnitsMap &unitsMap, double exp = 1.0)
...
+//     }
+
+//     if (requireDimensionless) {
+//         unitsMap.emplace(std::make_pair("dimensionless", 0.0));
+//     }
+
+//     return unitsMap;
+// }
+
+bool Units::usesImportedUnits() const
```
hsorby: Does this function come across as an extension to `isImport`?  It does to me making me think that this belongs as part of the checks to determine if a units is imported or not.

src/units.cpp
```diff
@@ -457,43 +459,99 @@ void updateUnitsMap(const UnitsPtr &units, UnitsMap &unitsMap, double exp = 1.0)
...
-    bool requireDimensionless = false;
     auto it = unitsMap.begin();
     while (it != unitsMap.end()) {
         if (it->second == 0.0) {
             it = unitsMap.erase(it);
-            requireDimensionless = true;
-        } else if (it->first == "dimensionless") {
-            it->second = 0.0;
-            ++it;
+        } else if ((it->first == "dimensionless") || (it->first == "radian") || (it->first == "steradian")) {
```
hsorby: Does this mean we have to know the names of all `dimensionless` units?

hsorby (changes requested):

src/units.cpp
kerimoyle: It's currently not, no.  But I do think it's better that a unit is called "imported" if it or any of its dependencies are.  Would all our imported things need that kind of check? If so, can move it.

src/units.cpp
kerimoyle: Unless you're happy to say that `radians != dimensionless`, then yes, we need to specify them here.  I wasn't sure about this (see #552)...

kerimoyle: Scratch that, I think you're right, thanks :)

src/units.cpp
hsorby: We might need to let this one percolate to see if it makes sense tomorrow.

kerimoyle:
Thoughts needed on two points:  
**1.** I'm unsure about the multiplier behaviour when we have exponents of multiplied units (related to what I think is a typo in the equation in the spec).  This PR contains a test (currently commented out) that doesn't behave as I'd expected?
```
TEST(Units, scalingFactorBetweenMultipliedUnits)
{
    libcellml::UnitsPtr u1 = libcellml::Units::create();
    u1->setName("u1");
    u1->addUnit("metre", 0, 3.0, 1000.0);
    libcellml::UnitsPtr u2 = libcellml::Units::create();
    u2->setName("u2");
    u2->addUnit("metre", 0, 1.0, 1.0);
    u2->addUnit("metre", 0, 1.0, 1.0);
    u2->addUnit("metre", 0, 1.0, 1.0);

     // Returns 1e+9 instead?? ie: (1000m)^3 not (1000)m^3 ??
    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(u2, u1));
}
```

**2.** I've updated the tests so that we have full coverage, but in doing so have found some behaviour I'm not sure about.  At present, if a `Units` or any of its children is imported then the comparison functions (equivalent, compatible, scalingFactor) will return false/zero.  But ... in the test I wrote for coverage, both units were in fact identical as both imported the same thing.  This gives a false negative?  I'm pretty sure that we shouldn't be going down the rabbithole of checking contents of imports before they're resolved, but wanted to check what others thought.  The test is below:

```
TEST(Units, compatibleGrandchildUnitsImported)
{
    auto model = libcellml::Model::create("nurseryrhymes");
    auto imported_units = libcellml::Units::create("imported_units");
    auto imp = libcellml::ImportSource::create();
    imp->setUrl("i_am_a_url");
    imported_units->setImportSource(imp);

    auto u0 = libcellml::Units::create("u0");
    u0->addUnit("imported_units");
    auto u1 = libcellml::Units::create("u1");
    u1->addUnit("u0");
    auto u2 = libcellml::Units::create("u2");
    u2->addUnit("u1");
    auto u3 = libcellml::Units::create("u3");
    u3->addUnit("imported_units");  // So u3 and u2 are identical, but use imports ...

    model->addUnits(u0);
    model->addUnits(u1);
    model->addUnits(u2);
    model->addUnits(u3);
    model->addUnits(imported_units);

    // Expect false only because the first units "imported_units" is imported... otherwise it would be true.
    EXPECT_FALSE(libcellml::Units::compatible(u2, u3));
}
```

Thoughts needed ... @agarny @hsorby  @nickerso :) Thanks!

src/units.cpp
kerimoyle: ... nudge nudge ... @hsorby any more thoughts on this?

hsorby: I have had more thoughts but haven't decided on a course of action.

hsorby: I'm thinking that `isImport` needs to stay as it is but that we might need an additional method like `requiresImports` for components and units that have imported entities as part of their definition.

hsorby: Or maybe look at it from the other side and ask the question `isDefined`.

kerimoyle: I think that `isDefined` is a little less clear because it lumps together import of self and import of children.  If `isImport` is going to stay as is, then I reckon a `requiresImport` or `usesImport` or `hasChildImport` function would make sense.

nickerso:
see cellml/cellml-specification#19

nickerso:
Perhaps something for @awickens to check if the corrected equation from cellml/cellml-specification#19 is being used...

hsorby:
Some of this may have been addressed in #520 (not the formula changes though).

kerimoyle:
> Some of this may have been addressed in #520 (not the formula changes though).

Alas, no ... I included the test in 520 and it still fails. I'm not sure whether I'm misinterpreting the equation or not, or whether this is the expected outcome?

kerimoyle:
I'm coming to the conclusion that it's not a problem with the (fixed) equation - it's a problem with the `scalingFactor` function.  This test fails too:

```
TEST(Units, scalingFactorBetweenMultipliedUnits)
{
    libcellml::UnitsPtr u1 = libcellml::Units::create("u1");
    u1->addUnit("metre", 0, 3.0, 1000.0);
    libcellml::UnitsPtr u2 = libcellml::Units::create("u2");
    u2->addUnit("metre", 0, 3.0, 1.0);

     // Returns 1e+9 instead?? ie: (1000m)^3 not (1000)m^3 ??
    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(u2, u1));
}
```

kerimoyle:
There is indeed a bug in the `scalingFactor` routine: 
`mult` and `logMult` were switched.
```
multiplier += direction * (mult + (standardMultiplierList.at(ref) + logMult + standardPrefixList.at(pre)) * exp);
```

kerimoyle:
The final test is one I don't understand:  I would have expected unresolvable units to return false but the test is looking for true?  Any thoughts @hsorby @awickens @agarny @nickerso ?
```
TEST(Units, compareEquivalentNonStandardWhichCannotBeResolvedUnits)
{
    libcellml::UnitsPtr u1 = libcellml::Units::create();
    u1->setName("a");
    u1->addUnit("millisecond", -1);

    libcellml::UnitsPtr u2 = libcellml::Units::create();
    u2->setName("b");
    u2->addUnit("millisecond", -1);

    EXPECT_TRUE(libcellml::Units::equivalent(u1, u2));
}
```

hsorby:
The units are unresolvable but they match so the units are equivalent.

awickens:
I would expect both the mappings to have nothing in them as well - I ran the test in the debugger and it seems like the second mapping has something in it, when there shouldn't be anything present. 

Edit: No, the mapping bits are fine (had `u1` somewhere where it shouldn't be). The mapping comparison part of the checks in `equivalent()` is fine, it appears to be a problem in `scalingFactor()` - maybe set the multiplier to zero if no units can be resolved?

kerimoyle:
> The units are unresolvable but they match so the units are equivalent.

I'd be worried about that ...  Elsewhere when units are simply imported we say that we are unable to compare them and return false - I don't see why two units which simply have the same name specified should be thought equivalent?  As far as we know they could be referencing two separate, but identically named units?  

Edit:  @hsorby  See second test point in this comment: https://github.com/cellml/libcellml/pull/554#issuecomment-589901277

kerimoyle:
> maybe set the multiplier to zero if no units can be resolved?

That's what should be happening if they're incompatible, so yes, I think it's reasonable to return the same thing if they can't be resolved.

awickens:
> > The units are unresolvable but they match so the units are equivalent.
> 
> I'd be worried about that ... Elsewhere when units are simply imported we say that we are unable to compare them and return false - I don't see why two units which simply have the same name specified should be thought equivalent? As far as we know they could be referencing two separate, but identically named units?

I agree that this would be a problem if we imported from two different `models`. If the model they are imported from is the same, then we would expect them to be the same unit, but there is a high likelihood of mismatch if they come from two different models (which wouldn't be constrained by the specification in that case). Then again, we would expect all units in the model we are importing *into* to also obey specification rules.....

Perhaps a check when importing units from different models into the model would be the way to go as a failsafe?

kerimoyle:
With this test the problem is also that there is no model at all - so even defining millisecond units doesn't change anything!  Do we want to be able to compare units without having to define a model?  If so, we will never be able to go beyond the first layer of unit nesting as there's nowhere to search for the linked units, so it seems a little pointless?

hsorby:
It is not a pointless test, it is showing the behaviour of the code in that situation and what you can expect to happen.

kerimoyle:
I agree that it's not pointless ... but I disagree that it's testing behaviour that I'd expect.  I thought that the rationale for returning `true` from equivalence testing was when we could actually test something.  There are two issues:
- Without a model, we don't have any source of comparison for a test,  
- If the `millisecond` units were defined as an imported units item, the test should still fail because we're not going down the imports route at that point ... 
... so I don't see how missing even more information gives us confidence to return true to this test? I would have expected false.  

If we are going to allow testing of units without models, I think we need to do it via another function which makes clear that only the *names* of the *first generation* of unit children can be compared ... because they're definitely not doing the same thing.

kerimoyle:
Hi @agarny @nickerso 
Trying to sort out this one last test and I need a bit more clarity on what we expect to happen when we can't resolve units.  I know @hsorby feels that [this test](https://github.com/cellml/libcellml/pull/554#issuecomment-591730096) should return `true` but in the interest of not making changes and then undoing them I wanted to get everyone's thoughts before I make bigger changes. At present we have a test which expects two units having:
- the same name
- no parent model(s)
- unresolvable first-generation child units
to return `true` I think it should be `false` to be in keeping with how we handle other situations where we don't have enough information (imports, other unresolvable/parentless units).  This is the only thing holding up this PR so I'm keen to get it figured out ... any thoughts? Thanks :)

agarny:
I would expect that test to return `false` since we can't resolve those units. Simple as that.

src/api/libcellml/units.h
```diff
@@ -457,7 +465,7 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
      * @return @c true if the two @c Units are dimensionally equivalent, @c false
      * otherwise.
      */
-    static bool dimensionallyEquivalent(const UnitsPtr &units1, const UnitsPtr &units2);
+    static bool equivalent(const UnitsPtr &units1, const UnitsPtr &units2);
```
agarny, line 471: If we are to change the name of the methods (i.e. `equivalent()` -> `compatible()` and `dimensionallyEquivalent()` -> `equivalent()`) then we also ought to update their documentation.

kerimoyle: Done :)

src/api/libcellml/units.h
```diff
@@ -428,41 +428,41 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
     /**
      * @brief Return the scaling factor difference between two @c Units.
      *
-     * This can be interpreted as factor, where units2 = (factor)*units1.  This method
-     * does not check to see if the units are compatible.
+     * This can be interpreted as factor, where units2 = (factor)*units1.  If the units
+     * are not compatible the factor returned is 0.
```
agarny: `0.0` rather than `0` since we are returning a `double`?

kerimoyle: Gah! I just changed it back because "0.0." looked weird ...

agarny: It may look weird, but personally I like it when it's clear, from the documentation, what type we are returning.

src/units.cpp
kerimoyle: I've done a bit more thinking ... and I reckon that it doesn't make sense to have a generic function between both `components` and `units` looking for import child dependencies (ie: beyond the `isImport`() function).  Since "children" look different to components and units, there's not really any sense in forcing them to use the same `importedentity` function.  I'm going to change the name of this function to be `requiresImports` and make a separate PR with something similar for the component class.  This PR is all about units and scaling ...

src/api/libcellml/units.h
```diff
@@ -428,41 +428,41 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
      */
     static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2);
 
     /**
-     * @brief Test to determine whether two @c Units are equivalent or not.
+     * @brief Test to determine whether two @c Units are compatible or not.
      *
-     * Two @c Units are considered to be equivalent if they share the same units,
+     * Two @c Units are considered to be compatible if they share the same units,
      * independently of their dimension (e.g. volt and volt are equivalent as are
```
hsorby: I think this should change to independently of their scaling.  Dimension to me means we would consider volt and volt squared as being compatible.

hsorby (changes requested):

.gitignore
```diff
@@ -18,3 +18,4 @@ CMakeLists.txt.user
 
 # VS Code system files
 .vscode/
+*.profraw
```
agarny: Remove this. (I believe @hsorby has already discussed this before.)

src/api/libcellml/units.h
```diff
@@ -417,47 +417,55 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
      */
     size_t unitCount() const;
 
+    /**
+     * @brief Check whether there are any imported child @c Units.
+     *
+     * @return @c true when these @c Units rely on @c Units which are imported,
+     * or @c false otherwise.
+     */
+    bool requiresImports() const;
```
agarny, line 426: I am not sure I would make this method part of the API, not least because it's only being used internally.

src/api/libcellml/units.h
```diff
@@ -417,47 +417,55 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
+
     /**
      * @brief Return the scaling factor difference between two @c Units.
      *
-     * This can be interpreted as factor, where units2 = (factor)*units1.  This method
-     * does not check to see if the units are compatible.
+     * This can be interpreted as factor, where units2 = (factor)*units1.  If the units
+     * are not compatible the factor returned is 0.0.
```
agarny: Replace with:

```
* This can be interpreted as `factor`, where units2 = factor*units1.  If the units
* are not compatible the factor returned is 0.0.
```

? I.e. use `` ` `` around factor and remove the parentheses around its second occurrence.

src/api/libcellml/units.h
```diff
@@ -417,47 +417,55 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
      */
     static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2);
 
     /**
-     * @brief Test to determine whether two @c Units are equivalent or not.
+     * @brief Test to determine whether two @c Units are compatible or not.
      *
-     * Two @c Units are considered to be equivalent if they share the same units,
+     * Two @c Units are considered to be compatible if they share the same units,
      * independently of their dimension (e.g. volt and volt are equivalent as are
```
agarny: Replace "equivalent" with "compatible".

src/api/libcellml/units.h
```diff
@@ -417,47 +417,55 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
      *
-     * @return @c true if the two @c Units are equivalent, @c false otherwise.
+     * @return @c true if the two @c Units are compatible, @c false otherwise.
      */
-    static bool equivalent(const UnitsPtr &units1, const UnitsPtr &units2);
+    static bool compatible(const UnitsPtr &units1, const UnitsPtr &units2);
 
     /**
-     * @brief Test to determine whether two @c Units are dimensionally equivalent or not.
+     * @brief Test to determine whether two @c Units are exactly equivalent or not.
```
agarny: Remove "exactly" or we should then rename `equivalent()` to `exactlyEquivalent()`.

src/bindings/interface/units.i
```diff
@@ -65,6 +65,10 @@ Only the first matching unit is removed.
 "Returns the attributes of the unit specified by index, reference, or
 StandardUnit.";
 
+%feature("docstring") libcellml::Units::requiresImports
+"Returns 'True' when this Units object depends on an imported child
+Units object, or 'False' otherwise.";
```
agarny: Minor: replace `'`' with `` ` ``. This is what (most) other interfaces use.

src/bindings/interface/units.i
```diff
@@ -76,10 +80,10 @@ extract the units with the given `name`.";
 "Returns the scaling factor between two Units objects.";
 
 %feature("docstring") libcellml::Units::equivalent
-"Tests if two Units are equivalent.";
+"Tests if two Units are equivalent in base units as well as scaling factor.";
 
-%feature("docstring") libcellml::Units::dimensionallyEquivalent
-"Tests if two Units are dimensionally equivalent.";
+%feature("docstring") libcellml::Units::compatible
+"Tests if two Units have the same dimensional basis.";
```
agarny, line 83: Minor: might want to swap `compatible` and `equivalent`, so we have the same order as in our C++ API.

src/api/libcellml/units.h
kerimoyle: But ... that defeats the point of dimensional analysis? if you remove the dimension checking? I'm confused ...

src/units.cpp
```diff
@@ -401,19 +398,21 @@ size_t Units::unitCount() const
 
 double Units::scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2)
 {
+    if (!Units::compatible(units1, units2)) {
+        return 0.0;
+    }
+
```
hsorby: I would like to see a default argument added to this method so that I can actively ignore the compatibility check.

hsorby (changes requested):

src/api/libcellml/units.h
kerimoyle: Ahhh ... I get it now, thanks.

src/units.cpp
```diff
@@ -527,9 +562,10 @@ bool Units::equivalent(const UnitsPtr &units1, const UnitsPtr &units2)
     return false;
 }
 
-bool Units::dimensionallyEquivalent(const UnitsPtr &units1, const UnitsPtr &units2)
+bool Units::equivalent(const UnitsPtr &units1, const UnitsPtr &units2)
 {
-    return Units::equivalent(units1, units2) && (Units::scalingFactor(units1, units2) == 1.0);
+    // Units must be compatible to return a non-zero scaling factor
+    return (Units::scalingFactor(units1, units2) == 1.0);
```
agarny: Remove the surrounding parentheses. Also, full stop at the end of the comment.

src/units.cpp
```diff
@@ -506,9 +532,18 @@ bool Units::equivalent(const UnitsPtr &units1, const UnitsPtr &units2)
     if ((units1->isImport()) || (units2->isImport())) {
         return false;
     }
-
-    UnitsMap units1Map = createUnitsMap(units1);
-    UnitsMap units2Map = createUnitsMap(units2);
+    if ((units1->requiresImports()) || (units2->requiresImports())) {
+        return false;
+    }
+    bool isValid = true;
```
agarny: No need to initialise `isValid`.

src/api/libcellml/units.h
hsorby: One would think that `exactly` in `exactlyEquivalent` would be redundant.

src/units.cpp
agarny: ... hmm, unless the compiler is not clever enough to realise that `createUnitsMap()` initialises it...

src/api/libcellml/units.h
kerimoyle: I put it in to make it (I thought) clearer ... will reword the docs not the function name.

src/bindings/interface/units.i
kerimoyle: We are super inconsistent throughout ... there are lots of back ticks and single string quotes everywhere!

src/units.cpp
kerimoyle: What do you expect it to return in that situation?  I'd wondered about getting a scaling factor that included the mismatch of units ... is that what you're talking about, or something else?

src/api/libcellml/units.h
hsorby: then dimensionality is a better term.

src/units.cpp
```diff
@@ -461,43 +460,70 @@ void updateUnitsMap(const UnitsPtr &units, UnitsMap &unitsMap, double exp = 1.0)
...
 
-bool Units::equivalent(const UnitsPtr &units1, const UnitsPtr &units2)
+bool Units::requiresImports() const
+{
+    // Function to check child unit dependencies for imports.
+    std::string ref;
+    std::string prefix;
+    double exponent;
+    double multiplier;
+    std::string id;
```
agarny: Declare them only if we have a parent (i.e. no need to declare variables if they are not going to be used).

src/units.cpp
```diff
@@ -461,43 +460,70 @@ void updateUnitsMap(const UnitsPtr &units, UnitsMap &unitsMap, double exp = 1.0)
...
+            auto child = model->units(ref);
+            if (child == nullptr) {
+                continue;
+            }
+            if (child->isImport()) {
+                return true;
+            }
+            if (child->requiresImports()) {
+                return true;
+            }
```
agarny: Replace with:
```
if (child->isImport() || child->requiresImports()) {
    return true;
}
```

tests/units/units.cpp
```diff
@@ -1060,19 +1060,132 @@ TEST(Units, scalingFactorWithTwoEmptyUnits)
...
+//     u1->addUnit("metre", 0, 3.0, 1000.0);
+//     libcellml::UnitsPtr u2 = libcellml::Units::create();
+//     u2->setName("u2");
+//     u2->addUnit("metre", 0, 1.0, 1.0);
+//     u2->addUnit("metre", 0, 1.0, 1.0);
+//     u2->addUnit("metre", 0, 1.0, 1.0);
+
+//     EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(u2, u1));
+//     EXPECT_EQ(0.001, libcellml::Units::scalingFactor(u1, u2));
+// }
```
agarny: If that test is not needed anymore then remove it.

tests/units/units.cpp
```diff
@@ -1060,19 +1060,132 @@ TEST(Units, scalingFactorWithTwoEmptyUnits)
...
     u1->setName("u1");
     u1->addUnit("metre", 0, 1.0, 1.0);
     libcellml::UnitsPtr u2 = libcellml::Units::create();
     u2->setName("u2");
     u2->addUnit("second", 0, 1.0, 1.0);
 
-    // Incompatible units but return a scaling factor of 1.0.
+    // Incompatible units return a scaling factor of 0.0.
+    EXPECT_EQ(0.0, libcellml::Units::scalingFactor(u1, u2));
+    EXPECT_EQ(0.0, libcellml::Units::scalingFactor(u2, u1));
```
agarny, line 1088: Do we really need that second test?

src/bindings/interface/units.i
hsorby: So we should add it in to the style guide doc that we want back ticks so we know which style we should use.

tests/units/units.cpp
```diff
@@ -1060,19 +1060,132 @@ TEST(Units, scalingFactorWithTwoEmptyUnits)
...
+    m->addUnits(u2);
+    m->addUnits(u3);
+    m->addUnits(u4);
+    m->addUnits(u5);
+    m->addUnits(u6);
+
     EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u2));
     EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u2, u1));
+    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u3, u3));
+    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u4, u4));
```
agarny: Do we really want to test the scaling factor between `u3` and `u3`, and then between `u4` and `u4`? Shouldn't it rather be between `u3` and `u4`, and then between `u4` and `u3`?

src/units.cpp
hsorby: Yes to just getting the scaling factor without being concerned about the mismatch in units.

tests/units/units.cpp
```diff
@@ -1060,19 +1060,132 @@ TEST(Units, scalingFactorWithTwoEmptyUnits)
...
+    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u3));
+    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u2, u4));
+    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u4));
+    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u2, u3));
+    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u5, u6));
+    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u5));
```
agarny: The tests between `u2` and `u4`, `u2` and `u3` are not needed. Also, we might want to add a test between `u6` and `u5`. In the end, I would simply have:

```
EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u2));
EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u2, u1));
EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u3, u4));
EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u4, u3));
EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u5, u6));
EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u6, u5));
EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u3));
EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u5));
```

tests/units/units.cpp
```diff
@@ -1138,8 +1240,8 @@ TEST(Units, complicatedMultiplicationFactorUnits)
     EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u2));
     EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u3, u4));
     EXPECT_EQ(1e-08, libcellml::Units::scalingFactor(incredible_pile_of_square_apples, square_apple));
-    // Incompatible units but return a scaling factor.
-    EXPECT_EQ(1e-07, libcellml::Units::scalingFactor(incredible_pile_of_square_apples, bunch_of_bananas));
+    // Incompatible units but return a scaling factor of 0
```
hsorby: // Incompatible units so we return a scaling factor of 0.0.

tests/units/units.cpp
```diff
@@ -1249,8 +1351,9 @@ TEST(Units, checkScalingFactorOneUnitImported)
 
     u1->setImportSource(import);
 
-    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u1, u2));
-    EXPECT_EQ(1.0, libcellml::Units::scalingFactor(u2, u1));
+    // KRM Changed to be 0.0 because the units u1 contain an import?
```
hsorby: delete

tests/units/units.cpp
agarny: Replace `of 0` with `of 0.0.`

hsorby (changes requested):

tests/units/units.cpp
agarny: Remove comment.

tests/units/units.cpp
```diff
@@ -1249,8 +1351,9 @@ TEST(Units, checkScalingFactorOneUnitImported)
...
+    EXPECT_EQ(0.0, libcellml::Units::scalingFactor(u1, u2));
+    EXPECT_EQ(0.0, libcellml::Units::scalingFactor(u2, u1));
```
agarny, line 1341: Is that second test really needed (since we are returning `0.0`)?

src/units.cpp
kerimoyle: But what "factor"?  a number independent of base units? or the mismatch? 
Option 1:  
`   1e6 = scalingFactor(millisecond, kilogram)`
Option 2:
`   1e6(kg)/(s) = scalingFactor(millisecond, kilogram)`

tests/units/units.cpp
```diff
@@ -1704,8 +1807,42 @@ TEST(Units, isEquivalentBaseUnitImported)
...
+    u1->addUnit(libcellml::Units::StandardUnit::RADIAN, 0, 1.0, 1.0);
+
+    libcellml::UnitsPtr u2 = libcellml::Units::create();
+    u2->setName("u2");
+    u2->addUnit(libcellml::Units::StandardUnit::RADIAN, 0, 1.0, 1.0);
+    u2->addUnit("u3", 0, 1.0, 1.0);
+
+    model->addUnits(u);
+    model->addUnits(u1);
+    model->addUnits(u2);
```
agarny, line 1821: Shouldn't `u3` be added to `model`?

.gitignore
kerimoyle: Done.

tests/units/units.cpp
```diff
@@ -1725,27 +1862,31 @@ TEST(Units, isEquivalentBaseUnitNotInModel)
     u2->setName("u2");
     u2->addUnit(libcellml::Units::StandardUnit::RADIAN, 0, 1.0, 1.0);
     u2->addUnit("u", 0, 1.0, 1.0);
+    // test should fail because base unit "u" is not present in u1, regardless of whether "apples" exists or not
```
agarny: Remove comment. There is no test around that comment, so no idea what this refers to.

tests/units/units.cpp
```diff
@@ -1725,27 +1862,31 @@ TEST(Units, isEquivalentBaseUnitNotInModel)
...
 
     model->addUnits(u1);
     model->addUnits(u2);
+    // model->linkUnits();
```
agarny: Remove.

tests/units/units.cpp
```diff
@@ -1725,27 +1862,31 @@ TEST(Units, isEquivalentBaseUnitNotInModel)
...
+    // Units "u" not in model.  Units "apples" doesn't exist.  Units are not linked.
+    // Previously returned false because of mismatched number of child units.
```
agarny: Previously? What does this mean?

tests/units/units.cpp
```diff
@@ -2004,3 +2145,111 @@ TEST(Units, scalingFactorAcrossStandardUnits)
...
+    auto model = libcellml::Model::create("nurseryrhymes");
+    auto u0 = libcellml::Units::create("sheep");
+    u0->addUnit("wool");
+
+    auto u1 = libcellml::Units::create("bo_peep");
+    u1->addUnit("sheep");
+
+    auto u2 = libcellml::Units::create("flock");
+
+    model->addUnits(u0);
```
agarny, line 2160: `u0` is not at all used in this test, so why have it?

tests/units/units.cpp
```diff
@@ -2004,3 +2145,111 @@ TEST(Units, scalingFactorAcrossStandardUnits)
...
+    model->addUnits(u1);
+    model->addUnits(u2);
+    auto s = libcellml::Units::scalingFactor(u1, u2);
+
+    EXPECT_EQ(0.0, s);
```
agarny: Replace with:

```
EXPECT_EQ(0.0, libcellml::Units::scalingFactor(u1, u2));
```

tests/units/units.cpp
```diff
@@ -2004,3 +2145,111 @@ TEST(Units, scalingFactorAcrossStandardUnits)
...
+    auto u1 = libcellml::Units::create("bo_peep");
+    u1->addUnit("sheep");
+
+    auto u2 = libcellml::Units::create("flock");
+    auto u3 = libcellml::Units::create("sheep");
+    u2->addUnit("sheep");
+
+    m1->addUnits(u1);
+    m2->addUnits(u2);
+    m2->addUnits(u3);
```
agarny, line 2181: `u3` is not used, so remove it.

src/units.cpp
hsorby: I don't see how you can return option 2.  Option 1 was what I was expecting.

tests/units/units.cpp
```diff
@@ -2004,3 +2145,111 @@ TEST(Units, scalingFactorAcrossStandardUnits)
...
+    u3->addUnit("imported_units");
+
+    model->addUnits(u0);
+    model->addUnits(u1);
+    model->addUnits(u2);
+    model->addUnits(u3);
+    model->addUnits(imported_units);
+
+    // Expect false only because the first units "imported_units" is imported... otherwise it would be true.
+    EXPECT_FALSE(libcellml::Units::compatible(u2, u3));
```
agarny, line 2218: The name of test implies that the units should be compatible and yet the test itself expect them not to be. So, update the name of the test?

Otherwise, am I to understand that the `u2` and `u3` are not compatible because don't, in this test, resolve the import and therefore cannot determine what `imported_units` is about exactly? If so, we might want to make it clear in the comment.

tests/units/units.cpp
```diff
@@ -2004,3 +2145,111 @@ TEST(Units, scalingFactorAcrossStandardUnits)
...
+    libcellml::UnitsPtr u1 = libcellml::Units::create();
+    u1->setName("u1");
+    u1->addUnit("metre", 0, 3.0, 1000.0);
+    libcellml::UnitsPtr u2 = libcellml::Units::create();
+    u2->setName("u2");
+    u2->addUnit("metre", 0, 1.0, 1.0);
+    u2->addUnit("metre", 0, 1.0, 1.0);
+    u2->addUnit("metre", 0, 1.0, 1.0);
+
+    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(u2, u1));
```
agarny, line 2232: Might want to test the scaling factor between `u1` and `u2` too.

tests/units/units.cpp
```diff
@@ -2004,3 +2145,111 @@ TEST(Units, scalingFactorAcrossStandardUnits)
...
+TEST(Units, scalingFactorBetweenMultipliedUnits)
+{
+    libcellml::UnitsPtr u1 = libcellml::Units::create();
+    u1->setName("u1");
+    u1->addUnit("metre", 0, 3.0, 1000.0);
+    libcellml::UnitsPtr u2 = libcellml::Units::create();
+    u2->setName("u2");
+    u2->addUnit("metre", 0, 3.0, 1.0);
+
+    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(u2, u1));
```
agarny, line 2245: Might want to test the scaling factor between `u1` and `u2` too.

agarny (changes requested):
A lot of changes! (Far too many, in fact, for a PR.)

src/api/libcellml/units.h
kerimoyle: Done.

src/api/libcellml/units.h
kerimoyle: Done.

src/api/libcellml/units.h
kerimoyle: Done

src/api/libcellml/units.h
kerimoyle: Done.

src/bindings/interface/units.i
kerimoyle: Done (for this instance).

src/units.cpp
kerimoyle: Done.

.gitignore
hsorby: Doesn't look done.

src/units.cpp
kerimoyle: Done

src/units.cpp
kerimoyle: Done.

src/bindings/interface/units.i
kerimoyle: Done.

src/units.cpp
kerimoyle: We know the base unit mismatch - it would be easy enough to create a `Units` item that filled the gap.

tests/units/units.cpp
kerimoyle: Since you've asked for reverse tests to be added below, I'm going to go with "yes" ;)

agarny: The difference, though, is that here we return `0.0` while in the other case I would expect to see `s` and `1.0/s`. However, I am happy with testing both ways in all cases.

tests/units/units.cpp
kerimoyle: u2, u4 -> comparison needed because previously a mismatch of "dimensionless" caused this test to fail, same with u2, u3.  I'll add the test from u6 to u5 as you suggest.

tests/units/units.cpp
kerimoyle: Done.

tests/units/units.cpp
kerimoyle: Done

tests/units/units.cpp
kerimoyle: For completeness, the reverse arrangement is tested too.

tests/units/units.cpp
kerimoyle: I've updated it, thanks.

tests/units/units.cpp
kerimoyle: Previously = earlier behaviour of the equivalence testing functions.  I'll remove the comment, no longer relevant.

src/units.cpp
hsorby: If I wanted that I would expect to call a function `unitsMismatch`.

tests/units/units.cpp
kerimoyle: Updated to better show behaviour.

tests/units/units.cpp
kerimoyle: Updated to show behaviour better.

tests/units/units.cpp
kerimoyle: Done.

tests/units/units.cpp
kerimoyle: Done.

tests/units/units.cpp
kerimoyle: Done.

tests/units/units.cpp
kerimoyle: Done.

tests/units/units.cpp
kerimoyle: Whoops, thanks!

tests/units/units.cpp
kerimoyle: Done.

tests/units/units.cpp
kerimoyle: Done.

tests/units/units.cpp
kerimoyle: Done.

src/units.cpp
kerimoyle: Done.

src/api/libcellml/units.h
kerimoyle: I can imagine it being useful when you parse a model and want to check whether it depends on any imported instances beyond what you've already got? I'll add a test to show it.

kerimoyle (commented):
... how come I can review my own PR?

agarny:
> ... how come I can review my own PR?

Can you really?

kerimoyle:
image
Ah, never mind!  I can "review" but not "approve" :)

src/units.cpp
```diff
@@ -461,43 +460,68 @@ void updateUnitsMap(const UnitsPtr &units, UnitsMap &unitsMap, double exp = 1.0)
...
+    // Function to check child unit dependencies for imports.
+
+    if (hasParent()) {
+        std::string ref;
+        std::string prefix;
+        double exponent;
+        double multiplier;
+        std::string id;
+
+        auto model = std::dynamic_pointer_cast<Model>(parent());
```
hsorby:
```
auto model = std::dynamic_pointer_cast<Model>(parent());
if (model != nullptr) {
...
}
```

I think this is a better approach here.

tests/units/units.cpp
agarny: Still not clear where `u0` is being used in this test...

tests/units/units.cpp
agarny: Can't see the change...?

tests/resources/requires_imports.cellml
```diff
@@ -0,0 +1,12 @@
...
+    <units name="grandpa">
+        <unit units="dada"/>
+    </units>
+    <units name="dada">
+        <unit units="son"/>
+    </units>
+    <import xlink:href="units_definitions.cellml" xmlns:xlink="http://www.w3.org/1999/xlink">
+        <units units_ref="ms" name="son"/>
+    </import>
+</model>
```
agarny: Need an empty line at the end of the file.

tests/units/units.cpp
kerimoyle: See full context below.  The test is intended to show that where a child units item (`sheep` here) cannot be found in the same model as the units being tested (`flock` vs `bo_peep`) then it returns 0.0.  This is why `u3` is added to only one model.
```
TEST(Units, compatibleUnitsInDifferentModelsChildNotFound)
{
    auto m1 = libcellml::Model::create("nurseryrhymes");
    auto m2 = libcellml::Model::create("farmwork");
    auto u1 = libcellml::Units::create("bo_peep");
    u1->addUnit("sheep");

    auto u2 = libcellml::Units::create("flock");
    auto u3 = libcellml::Units::create("sheep");
    u2->addUnit("sheep");

    m1->addUnits(u1);
    m2->addUnits(u2);
    m2->addUnits(u3);

    // Child unit of "sheep" exists only in m2, return 0.0.
    EXPECT_EQ(0.0, libcellml::Units::scalingFactor(u1, u2));
}
```

tests/units/units.cpp
kerimoyle: See full context below.  `u0` is needed to provide a lower generation against which the parent units need to be resolved.

```
TEST(Units, scalingFactorGrandchildUnitsNotFound)
{
    auto model = libcellml::Model::create("nurseryrhymes");

    auto u0 = libcellml::Units::create("sheep");
    u0->addUnit("wool");

    auto u1 = libcellml::Units::create("bo_peep");
    u1->addUnit("sheep");

    auto u2 = libcellml::Units::create("flock");
    u2->addUnit("sheep");

    model->addUnits(u0);
    model->addUnits(u1);
    model->addUnits(u2);

    // Child units of "wool" are not found within "sheep" units.  Comparison should return 0.0.
    EXPECT_EQ(0.0, libcellml::Units::scalingFactor(u1, u2));
}
```

hsorby (changes requested):
I would still like to see a default argument added to `Units::scalingFactor`.

tests/units/units.cpp
agarny: Oops, my bad, I hadn't seen that `u2` uses `u3`. All good.

kerimoyle:
> I would still like to see a default argument added to Units::scalingFactor.

I'm doing it now ... :)

tests/units/units.cpp
agarny: 👍

src/units.cpp
kerimoyle: Done.

src/api/libcellml/units.h
kerimoyle:
```
TEST(Units, parsedUnitsRequiringImports)
{
    auto model = parser->parseModel(fileContents("requires_imports.cellml"));

    // When parsing a model it's hard to know whether the units it uses are imported
    // or not if the import is in a child units item.  Using the Units::requiresImports()
    // function should tell us.

    EXPECT_EQ(size_t(3), model->unitsCount());
    EXPECT_FALSE(model->units("grandpa")->isImport());
    EXPECT_TRUE(model->units("grandpa")->requiresImports());
}
```

src/api/libcellml/units.h
```diff
@@ -426,18 +426,35 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
...
+     * compatible the factor returned is 0.0.
      *
      * @param units1 The first units to compare.
-     * @param units2 The second units to compare.
+     * @param units2 The second units to compare. 
      *
-     * @return The factor units2/units1, where the units are compatible, or 0.0 otherwise.
+     * @return The factor units2/units1.  Where the units are incompatible the factor returned is 0.0.
      */
     static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2);
```
hsorby: I think a default argument would be sufficient, I don't see a need for a separate function.

```
static double scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2, bool checkCompatibility = true);
```

I don't think there is any point passing a bool by reference either.

hsorby (changes requested):

src/api/libcellml/units.h
agarny: Agreed.

kerimoyle: Done.

src/api/libcellml/units.h
```diff
@@ -448,12 +433,12 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
      *
      * @param units1 The first units to compare.
      * @param units2 The second units to compare.
-     * @param checkCompatibility Set @c true for compatiblily checking, or @false to ignore base units. 
+     * @param checkCompatibility Set @c true for compatiblily checking, or @c false to ignore base units.
```
hsorby: ... the default value is @c true.

src/api/libcellml/units.h
```diff
@@ -448,12 +433,12 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
      *
      * @param units1 The first units to compare.
      * @param units2 The second units to compare.
-     * @param checkCompatibility Set @c true for compatiblily checking, or @false to ignore base units. 
+     * @param checkCompatibility Set @c true for compatiblily checking, or @c false to ignore base units. 
      *
      * @return The factor units2/units1.  Where the units are incompatible and @c checkCompatibility
```
hsorby: `@p checkCompatibility` and not `@c checkCompatibility`

kerimoyle:
All changes made now.  @hsorby @agarny

agarny (dismissed):

hsorby (changes requested):

src/api/libcellml/units.h
kerimoyle: Done.

src/api/libcellml/units.h
kerimoyle: Done.

hsorby (approved):

agarny (approved):

Issue #556, opened 2020-02-24
Multiple MathML blocks allowed?


kerimoyle:
Related to https://github.com/cellml/cellml-specification/issues/27

If (and I'm not sure they are) multiple `<math>` blocks within a single component are valid CellML, then we are not currently reading them in libcellml.  If they're not, then we need to say so in the spec.  This test fails:

```
TEST(Validator, validMathInMultipleMathMlBlocks)
{
    const std::string math =
        "<math xmlns:cellml=\"http://www.cellml.org/cellml/2.0#\" xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
        "  <apply>\n"
        "    <eq/>\n"
        "    <ci>A</ci>\n"
        "    <apply>\n"
        "      <plus/>\n"
        "      <cn cellml:units=\"dimensionless\">1</cn>\n"
        "    </apply>\n"
        "  </apply>\n"
        "</math>\n"
        "<math xmlns:cellml=\"http://www.cellml.org/cellml/2.0#\" xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
        "  <apply>\n"
        "    <eq/>\n"
        "    <ci>B</ci>\n"
        "    <apply>\n"
        "      <plus/>\n"
        "      <cn cellml:units=\"dimensionless\">2</cn>\n"
        "    </apply>\n"
        "  </apply>\n"
        "</math>\n";

    libcellml::ValidatorPtr v = libcellml::Validator::create();
    libcellml::ModelPtr m = libcellml::Model::create();
    libcellml::ComponentPtr c = libcellml::Component::create();
    libcellml::VariablePtr v1 = libcellml::Variable::create();
    libcellml::VariablePtr v2 = libcellml::Variable::create();

    m->setName("modelName");
    c->setName("componentName");
    v1->setName("A");
    v2->setName("B");
    v1->setUnits("dimensionless");
    v2->setUnits("dimensionless");

    c->addVariable(v1);
    c->addVariable(v2);
    c->setMath(math);
    m->addComponent(c);

    v->validateModel(m);
    EXPECT_EQ(size_t(0), v->errorCount());
    printErrors(v);
}
```

Reporting errors:
```
[ RUN      ] Validator.validMathInMultipleMathMlBlocks
/Users/kmoy001/libcellml/libcellml-src/tests/validator/validator.cpp:472: Failure
Expected equality of these values:
  size_t(0)
    Which is: 0
  v->errorCount()
    Which is: 2
[...]

PR #557, opened 2020-02-25
Multiple MathML blocks

kerimoyle:
Tests to demonstrate #556.  There are three tests here which will fail at the moment, but should be allowed: more than one MathML block in:
- a component
- a reset test_value
- a reset reset_value

Currently all return libXML2 errors about being unable to find a root node.


kerimoyle:
This is now unneeded because the bug has been fixed in #520 :) ... but I guess it's worth keeping the tests as they're the only ones which involve multiple MathML blocks.

tests/validator/validator.cpp
```diff
@@ -426,6 +426,50 @@ TEST(Validator, validMath)
     EXPECT_EQ(size_t(0), v->errorCount());
 }
 
+TEST(Validator, validMathInMultipleMathMlBlocks)
```
agarny: Casing, it should be: `validMathInMultipleMathMLBlocks`.

tests/validator/validator.cpp
```diff
@@ -1162,6 +1206,40 @@ TEST(Validator, resetNoResetValue)
     EXPECT_EQ(expectedError, validator->error(0)->description());
 }
 
+TEST(Validator, resetMultipleMathMlInResetValue)
```
agarny: Casing, it should be: `resetMultipleMathMLInResetValue `.

tests/validator/validator.cpp
```diff
@@ -1200,6 +1278,40 @@ TEST(Validator, resetNoTestValue)
     EXPECT_EQ_ERRORS(expectedErrors, validator);
 }
 
+TEST(Validator, resetMultipleMathMlInTestValue)
```
agarny: Casing, it should be: `resetMultipleMathMLInTestValue `.

agarny (changes requested):
I would merge those three tests into one. Especially the last two, they are 99% the same.

tests/validator/validator.cpp
kerimoyle: Gah ... in earlier PRs I was told it went the other way  ... :|

hsorby:
> I would merge those three tests into one. Especially the last two, they are 99% the same.

While they are 99% the same I wouldn't merge these into one as they are testing different API calls.

tests/validator/validator.cpp
agarny: TBH, I would personally use `resetMultipleMathmlInTestValue`, but I have seen `xxxYyyyMathMLZzzz` in quite a few places.

kerimoyle: I've changed it once ... and you have edit permission ... so you can feel free to change it again if the mood takes you ;)

nickerso (approved):
I agree that it would be useful to keep these tests as they are not covered elsewhere. Might be useful to start making some functions to do the common setup tasks to address the comment by @agarny , but that can be done as a separate clean up later.

agarny (approved):

PR #560, opened 2020-02-27
Code generator: list the different variables under the "correct" component (#546)

### LINKED ISSUE #546: Code generator: list the different variables under the "correct" component
If we look at the [C code](https://github.com/cellml/libcellml/blob/b1865f87d3f6264d4b69aec901f635918e4c496e/tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.c#L14-L42) generated for the Hodgking-Huxley 1952 model, we can see that `time` is listed under the `membrane` component, as are `i_Na`, `i_K` and `i_L` for example.

However, "ideally", `time` would be listed under the `environment` component while `i_Na`, `i_K` and `i_L` would be listed under the `sodium_channel`, `potassium_channel` and `leakage_current` component, respectively.

The problem is that CellML 2.0 doesn't have the notion of directionality at the connection level. So, we can't tell what the reference variable is. However, when it comes to computed variables, we could assume that the "reference" component is the one that contains the corresponding equation. This leaves the VOI, for which things are trickier. Maybe we could assume that its "reference" component is the first component that defines it...?

FWIW, the reason for wanting the above is that tools like OpenCOR need to make model parameters available to end-users, and to list them under their "reference" component makes sense...

To be done:
 - [x] Have the VOI listed under the first component in which it is defined.
 - [x] Have state variables listed under the component in which they are computed, independent of where they are initialised.
 - [x] Have variables listed under the component in which they are computed (e.g. algebraic variable) or initialised (i.e. constants).

agarny:
Addresses issue #546.


src/generator.cpp
```diff
@@ -1222,18 +1247,42 @@ void Generator::GeneratorImpl::processEquationAst(const GeneratorEquationAstPtr
...
                 mGenerator->addError(err);
             } else {
-                mVoi = variable;
+                // We have found our variable of integration, but this may not
+                // be the one defined in our first component (i.e. the component
+                // under which we are likely to expect to see the variable of
+                // integration to be defined), so go through our components and
+                // look for the first occurrence of our variable of integration.
+
+                ModelPtr model = owningModel(std::dynamic_pointer_cast<Component>(variable->parent()));
```
hsorby: Why are you dynamic casting the returned result?

src/generator.cpp
```diff
@@ -1222,18 +1247,42 @@ void Generator::GeneratorImpl::processEquationAst(const GeneratorEquationAstPtr
 
                 mGenerator->addError(err);
             } else {
-                mVoi = variable;
+                // We have found our variable of integration, but this may not
+                // be the one defined in our first component (i.e. the component
+                // under which we are likely to expect to see the variable of
+                // integration to be defined), so go through our components and
+                // look for the first occurrence of our variable of integration.
```
hsorby, line 1282: What constitutes our first occurrence of the VOI? What metric do you use to determine the first occurrence?

agarny: Directionality has been removed in CellML 2.0, so we can't tell where a variable was originally defined. However, when it comes to the VOI, it has historically been defined in the first component of a CellML document. So, we go through the different components and look for the first one that defines the VOI, and we consider that component as being the component in which the VOI was "originally" defined. So, yes, very much a heuristic approach.

src/generator.cpp
agarny: Clearly an oversight at my end. So, yes, well spotted. Consider it fixed.

agarny: Fixed in commit df08fd8.

src/generator.cpp
```diff
@@ -752,6 +773,48 @@ GeneratorInternalVariablePtr Generator::GeneratorImpl::generatorVariable(const V
...
+
+GeneratorVariablePtr Generator::GeneratorImpl::variableFirstOccurrence(const VariablePtr &variable,
+                                                                       const ComponentPtr &component)
+{
+    // Recursively look for the first occurrence of the given variable in the
+    // given component.
+
+    GeneratorVariablePtr voi = nullptr;
+
+    for (size_t j = 0; j < component->variableCount(); ++j) {
```
hsorby: I think the break condition should be put into the continuation decision:

`j < component->variableCount() && voi == nullptr`

src/generator.cpp
```diff
@@ -752,6 +773,48 @@ GeneratorInternalVariablePtr Generator::GeneratorImpl::generatorVariable(const V
...
+            voi = GeneratorVariable::create();
+
+            voi->mPimpl->populate(testVariable, component, GeneratorVariable::Type::VARIABLE_OF_INTEGRATION);
+
+            break;
+        }
+    }
+
+    if (voi == nullptr) {
+        for (size_t i = 0; i < component->componentCount(); ++i) {
```
hsorby: same here:
`i < component->variableCount() && voi == nullptr`

hsorby (changes requested):

src/generator.cpp
agarny: Good point! Done in 4754c2b.

src/generator.cpp
agarny: Good point! Done in 4754c2b.

hsorby (dismissed):

agarny:
Ping @nickerso and @kerimoyle. Would be nice to have this one in so that I can carry on with my work on issue #409...

kerimoyle:
Hey @agarny.  Overall I'm happy (one tweak to comment wording), but I had a couple of questions:
- What (if any) are the implications of us trying to work out the home component for a variable, and somehow getting it wrong?  Are there any downstream effects that you can imagine?  It's not really a libcellml thing, but it would be good to know for the documentation and tutorials.
- Following the first point, should we make it possible for the user to specify which variable in a connected set is the "real" one?  This could help with the DAE stuff?

src/api/libcellml/generator.h
```diff
@@ -69,6 +71,20 @@ class LIBCELLML_EXPORT GeneratorVariable
...
     VariablePtr variable() const;
 
+    /**
+     * @brief Get the @c Component for this @c GeneratorVariable.
+     *
+     * Return the @c Component of this @c GeneratorVariable. It is the
+     * @c Component in which the @c Variable is first defined (in the case of
+     * the variable of integration), initialised (in the case of a constant) or
+     * actually computed (in the case of a state, computed constant or algebraic
+     * variable). It may, or not, be the same @c Component as the parent
```
kerimoyle: Either "It may or may not be the same ..." or "It is not necessarily the same ..."

agarny: Thanks, fixed in 4a0fad52.

agarny:
> * What (if any) are the implications of us trying to work out the home component for a variable, and somehow getting it wrong?  Are there any downstream effects that you can imagine?  It's not really a libcellml thing, but it would be good to know for the documentation and tutorials.

As you know, CellML 2.0 doesn't support directionality at the connection level anymore. So, because of that, we may indeed report the wrong home component for a variable. This is particularly true for the VOI since we can't rely on its initial condition (since it doesn't have one!) or on where it is being computed (since it's not!). The only thing we can do is rely on some heuristic based on exitsing CellML 1.0/1.1 models, i.e. the VOI tends to be defined in the very first component.

For a DAE-based variable, we might very well get it wrong if such a variable is determined using two equations that are in two different components. In that case, we can't decide which component should be the home one. So, most likely, we might have to go for the one we come across first.

Now, the only implication I can see for getting the home component "wrong" is that a variable will be, in a tool like OpenCOR, be listed in the "wrong" component, and I am not sure there is much we can do about it.

> * Following the first point, should we make it possible for the user to specify which variable in a connected set is the "real" one?  This could help with the DAE stuff?

How would you expect to do this? I am not sure libCellML should have anything to do with this since an end-user may not know anything about a given model. If anything, I would expect the CellML standard to provide a means to specify a variable's home component. However, it's not needed to describe a model, so I am not sure @nickerso would agree with such a feature. :)

kerimoyle:
> If anything, I would expect the CellML standard to provide a means to specify a variable's home component. However, it's not needed to describe a model, so I am not sure @nickerso would agree with such a feature. :)

... I don't follow the logic ... if we wrongly assume which is the home component does the order in which equations are applied change? so the effective model changes? so a unique home component is needed to describe the model unambiguously? so it should be in the standard?  It feels weird to talk about heuristics at the same time as talking about unique and explicit definitions?

agarny:
> if we wrongly assume which is the home component does the order in which equations are applied change?

No, it doesn't change. To know a variable's home component is only useful when you want to list variables in a tree-like structure (i.e. as in OpenCOR). It cannot affect the order in which equations are to be executed since we are relying on variables from different components being equivalent or not.

> so the effective model changes?

No, it doesn't.

> so a unique home component is needed to describe the model unambiguously?

No, it's not.

> so it should be in the standard?

It would solve our "problem" of home component, ut it's not needed to describe a model. So, strictly speaking, it shouldn't be in the standard.

> It feels weird to talk about heuristics at the same time as talking about unique and explicit definitions?

See above. :)

kerimoyle:
OK, good, thanks.  Final question - do you think it would be useful for users outside of code generation to be able to identify the home component like this?  ie - have a `Variable::definedIn()` or `Variable::setByVariable()` function that returns the name/pointer/something to the dominant variable of the equivalent set?  Where people are importing and combining different models this could be useful for debugging?

agarny:
It might be useful but it has nothing to do with CellML as such. So, for me, such a method has nothing to do in libCellML.

kerimoyle:
... but the whole point of this PR is nothing to do with CellML either? It's all about what's useful to OpenCOR?

agarny:
This PR is about code generation, not OpenCOR. As part of code generation, we generate some information about our different variables (see `VOI_INFO`, `STATE_INFO` and `VARIABLE_INFO`). As part of that information, we mention the component associated with a given variable. So, what this PR does is to mention a component that is as meaningful as possible, not just the first one we come across.

kerimoyle (approved):

src/generator.cpp
```diff
@@ -752,6 +773,48 @@ GeneratorInternalVariablePtr Generator::GeneratorImpl::generatorVariable(const V
...
+bool Generator::GeneratorImpl::sameOrEquivalentVariable(const VariablePtr &variable1,
+                                                        const VariablePtr &variable2)
+{
+    // Return whether the given variables are the same or are equivalent (be it
+    // directly or indirectly).
+
+    return (variable1 == variable2) || variable1->hasEquivalentVariable(variable2, true);
+}
+
+GeneratorVariablePtr Generator::GeneratorImpl::variableFirstOccurrence(const VariablePtr &variable,
```
hsorby, line 785: The items I raised in this method have been reverted?

hsorby (changes requested):

src/generator.cpp
agarny: ?? I don't believe I have reverted anything, so not quite sure what you are talking about here.

src/generator.cpp
hsorby: This diff is not showing the change in [4754c2b](https://github.com/cellml/libcellml/commit/4754c2bb071fe4b0f66445d6a5daba2afb51e9e5)

src/generator.cpp
hsorby: Change is not showing in this diff.

hsorby (approved):

src/generator.cpp
agarny: I guess you hadn't refreshed your page or something... :)

src/generator.cpp
agarny: Problem of refresh, I guess... :)

PR #564, opened 2020-03-04
I563 scaling factors standard units

### LINKED ISSUE #563: Cannot "properly" get the scaling factor between volt and millivolt
PR #547 addresses that issue for base units, but not for built-in units that are not base units. Thus, if we modify [this code](https://github.com/cellml/libcellml/blob/7e991e0f69d2512576659d1c41ef068ae95f7c27/tests/units/units.cpp#L2123-L2131) and replace `second` with `volt` then the test fails:

```
TEST(Units, scalingFactorAcrossStandardUnits)
{
    libcellml::UnitsPtr u1 = libcellml::Units::create("volt");
    libcellml::UnitsPtr u2 = libcellml::Units::create("millivolt");
    u2->addUnit(libcellml::Units::StandardUnit::VOLT, libcellml::Units::Prefix::MILLI);

    EXPECT_EQ(0.001, libcellml::Units::scalingFactor(u1, u2)); // Returns 0 and not 0.001
    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(u2, u1)); // Returns 0 and not 1000.0
}
```

I can get what I want by passing `false` as the third argument to `scalingFactor()`, but this is not right since I want to be sure that `volt` and `millivolt` are compatible.

kerimoyle:
Addresses #563


kerimoyle:
Review please @agarny @hsorby @nickerso

kerimoyle:
Ignore this for now ... need to add a test before it's ready.

hsorby:
> Ignore this for now ... need to add a test before it's ready.

So ... the idea behind test driven development is to write the test first?

kerimoyle:
Was the usage of isBaseUnit changed at some stage to mean "is not imported" instead of "is one of the seven fundamental units or custom"?  I'm wondering if it's clearer to rename it so we don't get confused?

kerimoyle:
> Ignore this for now ... need to add a test before it's ready.
> 
> So ... the idea behind test driven development is to write the test first?

Yep, I did that :P. I just need to reinstate the old one too ...

kerimoyle:
Righteho, ready now @agarny @hsorby

src/units.cpp
```diff
@@ -427,7 +427,16 @@ using UnitsMap = std::map<std::string, double>;
...
+    if (isStandardUnitName(units->name())) {
+        auto unitsListIter = standardUnitsList.find(units->name());
+        for (const auto &baseUnitsComponent : unitsListIter->second) {
+            auto unitsMapIter = unitsMap.find(baseUnitsComponent.first);
+            if (unitsMapIter == unitsMap.end()) {
+                unitsMap[baseUnitsComponent.first] = 0.0;
+            }
+            unitsMap[baseUnitsComponent.first] += baseUnitsComponent.second * exp;
+        }
+    } else if (units->isBaseUnit()) {
```
agarny: I would swap the `isBaseUnit()` and `isStandardUnitName()` tests. It shouldn't make much difference in terms of speed, just that it would be more logical to me. :)

tests/units/units.cpp
```diff
@@ -2121,6 +2121,16 @@ TEST(Units, isNotBaseUnitImportedTwice)
...
+{
+    libcellml::UnitsPtr u1 = libcellml::Units::create("volt");
+    libcellml::UnitsPtr u2 = libcellml::Units::create("millivolt");
+    u2->addUnit(libcellml::Units::StandardUnit::VOLT, libcellml::Units::Prefix::MILLI);
+
+    EXPECT_EQ(0.001, libcellml::Units::scalingFactor(u1, u2));
+    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(u2, u1));
+}
+
+TEST(Units, scalingFactorAcrossBaseUnits)
```
agarny: Same here, I would swap the two tests.

agarny (changes requested):
Looks good to me pending some (very) minor changes.

tests/units/units.cpp
kerimoyle: Done!

src/units.cpp
kerimoyle: Alas, that's what caused the bug in the first place, so the order has to stay :)

agarny: Hmm... indeed, I can see that now. Bad data structure, but nevermind...

kerimoyle: It's partly because of the isBaseUnit function I think ... I thought that was simply a test for the SI units, or custom ones, but it's looking at imports instead.  This is where the bug *really* came from - because the `volt` had no children and was not imported, it was wrongly being called a baseUnit.  The effects of it do need to be outside that function though, which is why I've put the base unit update for standard units as the fix, rather than changing that function.

agarny (dismissed):

agarny:
A very simple one that would be nice to have merged in today... :) @hsorby @nickerso

hsorby:
> Was the usage of isBaseUnit changed at some stage to mean "is not imported" instead of "is one of the seven fundamental units or custom"? I'm wondering if it's clearer to rename it so we don't get confused?

Yes the definition of `isBaseUnit` was changed so that it only returned true if it knew the units was a base unit or not.  With an unresolved import the method will always return `false` as you cannot know if it is a base unit or not.  The method documentation details what it means to be a base unit.

kerimoyle:
Does that mean that the `isBaseUnit` function needs to be changed to test for built in units too (so doesn't then return the misleading `true` to standard (but not base) units like `volt`)?  If that's the case then we will need to do some other reworking of the unit comparisons ...

agarny (approved):

nickerso (approved):

agarny:
Am waiting for the tests to go through before merging it in...

Issue #571, opened 2020-03-11
Units: wrong scaling factor


agarny:
The following test fails while I wouldn't expect it to.

```
TEST(Units, scalingFactorWrong)
{
    auto model = libcellml::Model::create();
    auto X = libcellml::Units::create("X");
    X->addUnit("mole");
    auto mX = libcellml::Units::create("mX");
    mX->addUnit("X", "milli");
    model->addUnits(X);
    model->addUnits(mX);

    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(mX, X));

    X->addUnit("litre", -1.0);

    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(mX, X)); // <--- scalingFactor() wrongly returns 1!
}
```

agarny:
To further illustrate the problem:

```
TEST(Units, scalingFactorWrong3)
{
    auto model = libcellml::Model::create();
    auto X = libcellml::Units::create("X");
    X->addUnit("mole");
    auto mX = libcellml::Units::create("mX");
    mX->addUnit("X", "milli");
    model->addUnits(X);
    model->addUnits(mX);

    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(mX, X));

    X->addUnit("litre", -3.0);

    EXPECT_EQ(1000.0, libcellml::Units::scalingFactor(mX, X)); // <--- scalingFactor() wrongly returns 1e-06!
}
```

agarny:
@kerimoyle, I guess this one might be for you? Worst case scenario, I will look into it tomorrow since it's blocking issue #409 / PR #572.

PR #570, opened 2020-03-09
Debug functions automatically available in Debug configurations

hsorby:
Addresses #567.

Issue #567, opened 2020-03-07
Debug functions: make them automatically available when building a debug version


agarny:
Right now, if we want to take advantage of debug functions, we must manually define `ENABLE_DEBUG_FUNCTIONS` as a boolean equal to `ON`. I understand from @hsorby that the reason is that we then don't have to worry about coverage (and maybe some other things?). However, this is a bit annoying. When I build in debug mode, I would expect to have access to the debug functions. Simple as that. (I keep having to search what I need to define to get access to those functions.)

hsorby:
I can look into this, it is a little awkward for MSVC but should still be doable.

agarny:
Great, that would be nice.

PR #572, opened 2020-03-11
Code generation for connections that use compatible, but not equivalent, units (i.e. need to be scaled)

### LINKED ISSUE #409: Code generation for connections that use compatible, but not equivalent, units (i.e. need to be scaled)

agarny:
Addresses issue #409.

Things to do:
 - [x] Wait for PR #521 to be merged in
 - [x] Wait for issue #567 (PR #570) to be addressed;
 - [x] Wait for issue #571 (PR #573) to be addressed; and
 - [x] Remove debug code.


agarny:
FWIW, I am basically waiting for PR #570 to be merged in, but apart from that, this should be ready for review. Ping @hsorby, @nickerso, @kerimoyle.

agarny:
Ok, PR #570 has now been merged in, so we are all good for review. Have a proper look at the test itself (i.e. the contents of the `tests/resources/generator/cellml_unit_scaling` folder). Ping @hsorby, @nickerso, @kerimoyle.

src/debug.cpp
```diff
@@ -54,4 +58,375 @@ void printStringStringMap(const StringStringMap &map)
...
+
+    if ((trunk->mPrev != nullptr) && (trunk->mPrev->mStr == SPACES)
+        && ((trunk->mStr == SPACES) || (trunk->mStr == TRUNK))) {
+        std::cout << " ";
+    }
+
+    std::cout << trunk->mStr;
+}
+
+std::string doPrintAst(const GeneratorEquationAstPtr &ast)
```
hsorby, line 96: I don't think we need all the spacing in this function.

src/internaltypes.h
```diff
@@ -22,6 +22,13 @@ limitations under the License.
 #include "libcellml/types.h"
 #include "libcellml/variable.h"
 
+#undef NAN
+
+#ifdef __linux__
+#    undef TRUE
+#    undef FALSE
+#endif
```
hsorby: Why are you wrapping `TRUE` and `FALSE` but not `NAN`?

src/internaltypes.h
```diff
@@ -32,6 +39,128 @@ using EquivalenceMap = std::map<IndexStack, std::vector<IndexStack>>; /**< Type
 using NameList = std::vector<std::string>; /**< Type definition for list of names. */
 using StringStringMap = std::map<std::string, std::string>; /**< Type definition for map of string to string. */
 
+struct GeneratorEquationAst;
+using GeneratorEquationAstPtr = std::shared_ptr<GeneratorEquationAst>;
+using GeneratorEquationAstWeakPtr = std::weak_ptr<GeneratorEquationAst>;
+
+struct GeneratorEquationAst
```
hsorby: This isn't an internal type it is a full on declaration.

tests/generator/generator.cpp
```diff
@@ -936,6 +936,41 @@ TEST(Generator, cellmlMappingsAndEncapsulations)
     EXPECT_EQ(fileContents("generator/cellml_mappings_and_encapsulations/model.py"), generator->implementationCode());
 }
 
+TEST(Generator, cellmlUnitScaling)
```
hsorby: I think we could start with a simpler test to show working units scaling.

hsorby (changes requested):
Initial review.

src/generator.cpp
```diff
@@ -30,11 +30,11 @@ limitations under the License.
...
+
+#include "internaltypes.h"
 #include "utilities.h"
 #include "xmldoc.h"
 
-#undef NAN
-
 #ifdef __linux__
 #    undef TRUE
 #    undef FALSE
```
hsorby, line 40: Defined elsewhere.

src/debug.cpp
agarny: Yes, I guess that this has never really been discussed as such. It's true that I like my code to be "spacious", but (most of?) the libCellML code is not so. Happy for you to edit as you see fit.

src/internaltypes.h
agarny: Simply being `NAN` is originally defined on all our supported platforms while `TRUE` and `FALSE` are only originally defined on Linux...? I will confirm this later today.

src/internaltypes.h
agarny: I agree. The background to this is that I wanted/needed the `printAst()` method, so I added it to our `debug.[cpp|h]` file, which in turn implied that I had to add the `GeneratorEquationAst` structure to `internaltypes.h`. From there, I must confess that I haven't looked back, i.e. I haven't tried to "improve" that aspect of the PR. Will see later whether there is something I can do about it, however feel free to make suggestions, if you have any.

tests/generator/generator.cpp
agarny: Fair point, not least since this is how I actually came up with this test in the end, i.e. step by step. Ok, I will go back to my commit log and split this test in several sub ones, following my thinking process back then.

src/internaltypes.h
agarny: I have just checked on Windows, Linux and macOS and, indeed, `NAN` needs to be undefined on all three platforms while `TRUE` and `FALSE` only need to be undefined on Linux.

nickerso:
Suggest a test that looks at the `initial_value` being a variable that is scaled across a connection, with a model like the randomly created without checking one given below:

```
<?xml version='1.0' encoding='UTF-8'?>
<model name="my_model" xmlns="http://www.cellml.org/cellml/2.0#" xmlns:cellml="http://www.cellml.org/cellml/2.0#">
    <units name="mV">
        <unit prefix="milli" units="volt"/>
    </units>
    <units name="V">
        <unit units="volt"/>
    </units>
    <component name="bob">
        <variable interface="public" name="variable_in_mV" units="mV" initial_value="10"/>
    </component>
    <component name="fred">
        <variable initial_value="bob" name="variable_in_V" units="V"/>
        <variable name="bob" units="V" interface="public"/>
    </component>
    <connection component_1="bob" component_2="fred">
        <map_variables variable_1="variable_in_mV" variable_2="bob"/>
    </connection>
</model>
```

src/internaltypes.h
agarny: This is fixed in commit 1075d7c6. @hsorby, is that fine with you? If so then feel free to resolve this conversation.

src/debug.cpp
agarny: I shall consider this conversation resolved as I am not sure there is a need to remove spaces as such to have this PR approved. However, I will definitely try to mimic more closely the style used in the rest of the library (yuck! :p :)).

hsorby (changes requested):

src/generator.cpp
agarny: I guess you mean undefined elsewhere. If so, then yes indeed, but... we have:

```
#include "generatorequationast.h"
#include "utilities.h"
#include "xmldoc.h"

#ifdef __linux__
#    undef TRUE
#    undef FALSE
#endif
```

`TRUE` and `FALSE` get undefined in `generatorequationast.h`, but then we are including `utilities.h` and `xmldoc.h`, which I imagine must indirectly redefine `TRUE` and `FALSE`, hence we need to undefined them again...

agarny: FWIW, I confirmed that on Linux.

tests/generator/generator.cpp
agarny: Our "big" cellmlUnitScaling test has now been replaced with smaller tests that address unit scaling, see our "new" cellmlUnitScalingVoiIndirect, cellmlUnitScalingVoiDirect, cellmlUnitScalingConstant, cellmlUnitScalingState, cellmlUnitScalingStateInitialisation and cellmlUnitScalingRate tests.

src/internaltypes.h
agarny: I am resolving this since it has been done. If it is not suitable then another comment will anyway have to be made since this one is now considered outdated by GitHub.

agarny:
> Suggest a test that looks at the `initial_value` being a variable that is scaled across a connection, with a model like the randomly created without checking one given below:
> 
> ```
> <?xml version='1.0' encoding='UTF-8'?>
> <model name="my_model" xmlns="http://www.cellml.org/cellml/2.0#" xmlns:cellml="http://www.cellml.org/cellml/2.0#">
>     <units name="mV">
>         <unit prefix="milli" units="volt"/>
>     </units>
>     <units name="V">
>         <unit units="volt"/>
>     </units>
>     <component name="bob">
>         <variable interface="public" name="variable_in_mV" units="mV" initial_value="10"/>
>     </component>
>     <component name="fred">
>         <variable initial_value="bob" name="variable_in_V" units="V"/>
>         <variable name="bob" units="V" interface="public"/>
>     </component>
>     <connection component_1="bob" component_2="fred">
>         <map_variables variable_1="variable_in_mV" variable_2="bob"/>
>     </connection>
> </model>
> ```

I can indeed confirm that our code generator doesn't currently handle variables that are initialised using another variable. So now working on that and your test above.

agarny:
@nickerso: regarding the variable reference that is used to initialise another variable, I imagine that it has to be a constant? In other words, if `x` is initialised to, say, `3` then `y` can be initialised to `x`. However, if `x` is a computed variable (e.g. `x = 3+5` or `x = a+b`) then `y` cannot be initialised to `x`?

agarny:
> @nickerso: regarding the variable reference that is used to initialise another variable, I imagine that it has to be a constant? In other words, if `x` is initialised to, say, `3` then `y` can be initialised to `x`. However, if `x` is a computed variable (e.g. `x = 3+5` or `x = a+b`) then `y` cannot be initialised to `x`?

It's just that I cannot seem to find anything about this in the interpretation section of the `initial_value` attribute in the sepcs...?

nickerso:
> It's just that I cannot seem to find anything about this in the interpretation section of the `initial_value` attribute in the sepcs...?

As with everything, its not in the spec :) Models can always be constructed that are not simulate-able, and this is nothing different here. Personally I do use computed constants to set initial values (e.g., temperature corrections), and I know the CellML API supports that, can check in OpenCOR. If your initial value was a state variable reference, then an implementation may choose to use the initial value of that state variable, but pretty sure that is not in the spec and doesn't need to be.

agarny:
> As with everything, its not in the spec :) Models can always be constructed that are not simulate-able, and this is nothing different here. Personally I do use computed constants to set initial values (e.g., temperature corrections), and I know the CellML API supports that, can check in OpenCOR. If your initial value was a state variable reference, then an implementation may choose to use the initial value of that state variable, but pretty sure that is not in the spec and doesn't need to be.

Ok, when it comes to the code generator, I am only going to support constants and computed constants. At least, to start with. Then, we can always (try to) support other things in the future, if we want.

agarny:
@nickerso: ok, I have given this a bit more thoughts and, for now at least, I am only going to support constant variables, i.e.
```
x = 1;
y = x;
```

Originally, an `initial_value` was a `double`. So, with CellML 1.1 and the ability to specify a variable reference, I would expect that reference variable to point to a (true) constant.

To support computed constants would make code generation tricky, I think. Indeed, we have `initializeStatesAndConstants()` to initialise our states and constants, and then `computeComputedConstants()` to initialise our computed constants. `initializeStatesAndConstants()` should be called only once while `computeComputedConstants()` may be called more than once (e.g. after a user decides to modify the value of a constant in OpenCOR). So, not sure how we could initialise a state or a constant using a computed constant.

If you have a CellML file that initialises a state or a constant using a computed constant then I would certainly be interested in it and see what the CellML API and OpenCOR do about it...

agarny:
I believe this should now be ready for review. Please have a particular look at the tests and make sure that the expected generated code is indeed what you would expect (it's relatively easy to get things wrong!). Ping @hsorby, @kerimoyle and @nickerso.

agarny:
I believe this PR is now ready for review. Ping @nickerso, @hsorby and @kerimoyle.

(I will quickly go through it one more time tomorrow morning, just in case.)

agarny:
Ok, I am done with my (very quick) double checking. Leave it to you guys (@nickerso, @hsorby and @kerimoyle) to review now. :)

src/api/libcellml/generator.h
```diff
@@ -63,27 +63,40 @@ class LIBCELLML_EXPORT GeneratorVariable
     static GeneratorVariablePtr create() noexcept;
 
     /**
-     * @brief Get the @c Variable for this @c GeneratorVariable.
+     * @brief Get the initial value @c Variable for this @c GeneratorVariable.
```
kerimoyle: maybe instead of "initial value @c Variable" we could have "initialising @c Variable" ?  I found it hard to follow on my first reading?  Same - perhaps - with the function name?

tests/generator/generator.cpp
```diff
@@ -102,7 +102,7 @@ TEST(Generator, twoVariablesOfIntegration)
     EXPECT_EQ(size_t(0), parser->issueCount());
 
     const std::vector<std::string> expectedIsues = {
-        "Variable 'time' in component 'main' of model 'two_variables_of_integration' and variable 'other_time' in component 'sub_sub_sub' of model 'two_variables_of_integration' cannot both be a variable of integration.",
+        "Variable 'time' in component 'main' and variable 'other_time' in component 'sub_sub_sub' cannot both be a variable of integration.",
```
kerimoyle: Can CellML handle only one variable of integration at a time in the entire model (in which case maybe the error message should be "**the** variable of integration", or is it that these are connected variables, in which case the error message should say that?

tests/generator/generator.cpp
```diff
@@ -102,7 +102,7 @@ TEST(Generator, twoVariablesOfIntegration)
     EXPECT_EQ(size_t(0), parser->issueCount());
 
     const std::vector<std::string> expectedIsues = {
```
kerimoyle: Spelling ...

tests/generator/generator.cpp
```diff
@@ -149,7 +149,7 @@ TEST(Generator, nonFirstOrderOdes)
 
     generator->processModel(model);
 
-    EXPECT_EQ_ISSUES_KINDS(expectedIsues, expectedCauses, generator);
+    EXPECT_EQ_ISSUES_CAUSES(expectedIsues, expectedCauses, generator);
```
kerimoyle: Spelling ...

kerimoyle: Actually, do a general search ... there are lots of them!  Should be `expectedIssues` (not ...Isues)

tests/generator/generator.cpp
```diff
@@ -203,7 +203,7 @@ TEST(Generator, variableInitializedTwice)
     EXPECT_EQ(size_t(0), parser->issueCount());
 
     const std::vector<std::string> expectedIsues = {
-        "Variable 'x' in component 'sub' of model 'variable_initialized_twice' and variable 'x' in component 'main' of model 'variable_initialized_twice' are equivalent and cannot therefore both be initialised.",
+        "Variable 'x' in component 'sub' and variable 'x' in component 'main' are equivalent and cannot therefore both be initialised.",
```
kerimoyle: I wonder whether our messages should talk about things that reflect the spec - ie: "connected" vs. "equivalent" or things that reflect our function names (`addEquivalence` etc?)

tests/generator/generator.cpp
```diff
@@ -225,6 +225,72 @@ TEST(Generator, variableInitializedTwice)
...
+TEST(Generator, nonExistingInitialisingVariable)
+{
+    libcellml::ParserPtr parser = libcellml::Parser::create();
+    libcellml::ModelPtr model = parser->parseModel(fileContents("generator/non_existing_initialising_variable.cellml"));
+
+    EXPECT_EQ(size_t(0), parser->issueCount());
+
+    const std::vector<std::string> expectedIssues = {
+        "Variable 'x' in component 'main' is initialised using variable 'k' which cannot be found.",
+    };
```
kerimoyle, line 286: You've used "is not defined anywhere" earlier?  Is there a difference between defining and finding?

src/api/libcellml/generator.h
agarny: I was trying to stick to the CellML linguo, but I agree that we don't need to (and probably shouldn't in fact!). Ok, changed in commit 534d064a.

Otherwise, what do you mean by the function name?

kerimoyle:
A couple of general questions:
- Since our tests will/could/might at some stage become part of a testing suite (?) should we be writing them with best practice models in mind? This would make it easier later on  ...? 
- Do we have consensus on how error messages are to be constructed? 
  - ie: telling the error and/or telling the user how to fix it or what's gone wrong?  For example:
"variable x is not computed" vs. "variable x is not included in any maths block"? 
  - what the jargon will be around the thing that's wrong (eg: "not found", "not defined", "does not exist" etc)?

tests/generator/generator.cpp
agarny: In principle, CellML supports more than one variable of integration (at least in CellML 1.0/1.1, I can't recall for certain in CellML 2.0...!?), but @nickerso and I once had a chat about this and we agreed that we should only support one variable of integration when it comes to code generation.

So, you are correct, it should "**the** variable of integration" and not "**a** variable of integration". Done in commit 04445ba3.

tests/generator/generator.cpp
agarny: Oops, my bad. Done in commit 0d779535.

tests/generator/generator.cpp
agarny: Indeed, done as part of commit 0d779535.

agarny: Copying/pasting... argh! :)

tests/generator/generator.cpp
agarny: We are dealing with libCellML here, not the specs. So, since our API uses `addEquivalence()`, etc., I would stick to "equivalent".

tests/generator/generator.cpp
agarny: No, no difference. I merely used "found" because that's what the code generator tries to do: find the variable. However, I agree that "defined" would be more suitable. Ok, done in commit 217a0c78.

agarny:
> * Since our tests will/could/might at some stage become part of a testing suite (?) should we be writing them with best practice models in mind? This would make it easier later on  ...?

We might be talking about two different things here...? Here, we are testing libCellML as a library while I think you are talking about the test suite that @MichaelClerx is working on? If so, then they are two very different things indeed. When it comes to the former, we don't really care about the way the tests look like (well, I do care, but as a libCellML "community", we don't :)) while when it comes to @MichaelClerx's test suite, I agree that we should use best practices.

> * Do we have consensus on how error messages are to be constructed?
>   * ie: telling the error and/or telling the user how to fix it or what's gone wrong?  For example:
>     "variable x is not computed" vs. "variable x is not included in any maths block"?
>   * what the jargon will be around the thing that's wrong (eg: "not found", "not defined", "does not exist" etc)?

I don't think we have agreed on anything yet. I guess an issue should be created for it.

nickerso:
> > * Since our tests will/could/might at some stage become part of a testing suite (?) should we be writing them with best practice models in mind? This would make it easier later on  ...?
> 
> We might be talking about two different things here...? Here, we are testing libCellML as a library while I think you are talking about the test suite that @MichaelClerx is working on? If so, then they are two very different things indeed. When it comes to the former, we don't really care about the way the tests look like (well, I do care, but as a libCellML "community", we don't :)) while when it comes to @MichaelClerx's test suite, I agree that we should use best practices.

Just to chime in on this, the idea of testing in libCellML and the test suite is to test software. These models will specifically not always follow best practice as we need to test software. Currently they are the only source of models that people might look at to see how to do things, but they should not be used to demonstrate best practice (other than best practices in terms of software testing).

Documentation, tutorials, etc., are the places where it is important to ensure appropriate best practices are being promoted and encouraged.

agarny:
> Just to chime in on this, the idea of testing in libCellML and the test suite is to test software. These models will specifically not always follow best practice as we need to test software. Currently they are the only source of models that people might look at to see how to do things, but they should not be used to demonstrate best practice (other than best practices in terms of software testing).
> 
> Documentation, tutorials, etc., are the places where it is important to ensure appropriate best practices are being promoted and encouraged.

Damn, I actually agree with that! :)

agarny:
@hsorby: the test failed on Windows because of the directory having (once again) not been properly deleted. I don't know how the deletion of that directory is done, but I know that I sometimes have the same problem on Windows with OpenCOR. My "fix" is to call my deletion script a few times. So, is that something that would be worth doing here?

In the meantime, I am going to retrigger my latest commit...

kerimoyle:
There is so much here!  I figure most of the changes to test resources have come from changes to printer/generator output order stuff rather than substance?  But I'm losing the plot a little ... which test is the one that has been fixed now?  Or isn't there one?

agarny:
There is indeed quite a lot in here. To summarise:
- The original issue was to handle the scaling of compatible variables (e.g. `V` is expressed in millivolts in one component and in volts in another, hence a scaling factor of `1000`/`0.001`). This has resulted in a couple of changes in the API for the `GeneratorVariable` class:
  - the `component()` method has been dropped (since we can always retrieve the component of a variable!); and
  - the `initialisingVariable()` method has been added since we can have a variable (e.g. a state variable) that is initialised somewhere (`initialisingVariable()` will return that variable) but computed somewhere else (`variable()` will return that variable).
- `src/debug.cpp` now contains `printAst()` which is used to print the AST of an equation. This proved quite useful when working on the original issue and thought I would keep it here, just in case we need to do some more "tricky" work on equations. This involved rearranging some of the code, e.g. we now have `src/generatorequationast.h`.
- `src/utilities.cpp`: I have removed the `entityName()` method which was only used by the generator, but not consistently. So, now, I directly use `NamedEntity::name()`. Otherwise, I added `owningComponent()` to match what we have with `owningModel()` and thus avoid having those longish casting statements which make the code difficult to read (for me, at least).
- `src/generator.cpp`: unsurprisingly, this is where the "important" stuff is. It (obviously) addresses the original issue (see `scalingFactor()`, `scaleAst()`, `scaleEquationAst()`), but also a few more that popped up while working on this. For example:
  - Initialisation of a variable using another variable (e.g. see `generateDoubleOrConstantVariableNameCode()` and the bit in `processEquationAst()` where there is a comment that reads "The initial value is not a double, so it has to be an existing variable of constant type").
  - Make sure that the "main" variable for the VOI or any of its equivalent variables is not initialised (see the test where `mVoi == nullptr` in `processEquationAst()`).
- There are a lot of tests covering all of the above.

agarny:
@hsorby, @kerimoyle and @nickerso: it would be nice to have this PR reviewed and then merged in...

I wouldn't want to go through the same process as the one I went through last week, i.e. I had completely forgotten about this PR and some of my work on PRs #598 and #599 involved writing "new" code, which I had actually written for this PR. Then, when those two PRs got merged in, I had several conflicts that took me some time to address. So, a lot of time has been wasted on this PR and it would be nice not to have this happen again... 😉

kerimoyle:
Welcome to my world ... :|

I started going through this last week, but got lost because there was so much going on!  I'll finish that up now.

agarny:
Thanks Keri, I appreciate that this is a big PR, but hopefully https://github.com/cellml/libcellml/pull/572#issuecomment-629869919 will help a bit with the reviewing process.

src/utilities.cpp
```diff
@@ -431,6 +421,11 @@ ModelPtr owningModel(const EntityPtr &entity)
     return model;
 }
 
+ComponentPtr owningComponent(const VariablePtr &variable)
+{
+    return std::dynamic_pointer_cast<Component>(variable->parent());
+}
+
```
kerimoyle, line 428: Isn't it cleaner to have the `variable->parent() `return a `ComponentPtr`?  The only reason to have the `owningModel` as a separate function for components is because sometimes the parent of a component is a `Model` and sometimes a `Component`?

tests/test_utils.cpp
```diff
@@ -42,9 +43,9 @@ std::string fileContents(const std::string &fileName)
 
 void printIssues(const libcellml::LoggerPtr &l, bool headings, bool causes, bool rule)
 {
+    int width = int(floor(log10(l->errorCount())));
```
kerimoyle, line 46: Wowee ... this is scary ... we're expecting more than 999 errors!! 😮

kerimoyle (changes requested):
Just the comment re parents/owning consistency ...?

src/utilities.cpp
agarny: A `Variable` is a `NamedEntity`, which itself is an `Entity`, and `Variable::parent()` is actually `Entity::parent()`, so it returns an `EntityPtr`. However, I guess we could override `Entity::parent()` and have `Variable::parent()` that returns a `ComponentPtr`, but if so then we probably ought to do something similar with `Component` (from which `Model` inherits; this has always baffled me, but that's another debate) and have `Component::parent()` return a `ComponentPtr` and `Component::modelParent()` return a `ModelPtr`

Views? @hsorby, @nickerso?

agarny:
> Just the comment re parents/owning consistency ...?

??

kerimoyle:
> Just the comment re parents/owning consistency ...?
> ??

https://github.com/cellml/libcellml/pull/572#discussion_r428317362

tests/test_utils.cpp
agarny: Maybe not, but I don't like hard-coded solutions, so... :)

src/utilities.cpp
nickerso: I like the parent concept relating to entities. If anything it would be a new method `Variable::component()`...

agarny: Ok, so we could have `Variable::component()` and `Component::model()`, and ditch `owningModel()` and `owningComponent()`? I think I would be ok with that. However, if we were to go ahead with that, then we should probably create a new issue for it.

kerimoyle: > I like the parent concept relating to entities. 

@nickerso  It seems that whenever we use it we have to cast/alter/do something else to it?  What's its value on its own?

agarny: Yes, 99% of cases, you will need to cast it, which I find it to be a PITA, hence we have utilities like `owningModel()` and `owningComponent()`, which is more like writing C code rather than C++ code, but... meh.

kerimoyle: I understand the desire to use inherited classes, but in this case the inheritance causes more work than if they were defined locally: components having both a ::model() and ::component() class, and variables having a ::component() class.  If inheritance is a must, why not reuse the component() one in the Entity class and add a model() class to the Component!

kerimoyle: Resolving in the tentative hope that this will be addressed in #615

kerimoyle (dismissed):

src/utilities.h
```diff
@@ -380,6 +368,16 @@ std::string entityName(const EntityPtr &entity);
...
 
+/**
+ * @brief Get the @c Component that the variable is owned by.
+ *
+ * Return the owning component for this variable.
+ *
+ * @param entity The variable to get the owning component for.
+ * @return The owning @c Component of this variable.
+ */
+ComponentPtr owningComponent(const VariablePtr &variable);
```
hsorby: Should be an `EntityPtr` here.

agarny: Done in commit 947a641.

kerimoyle (approved):

tests/generator/generator.cpp
```diff
@@ -44,32 +44,48 @@ TEST(Generator, emptyModel)
...
+TEST(Generator, initializedVariableOfIntegration)
 {
     libcellml::ParserPtr parser = libcellml::Parser::create();
-    libcellml::ModelPtr invalidModel = parser->parseModel(fileContents("generator/initialized_variable_of_integration.cellml"));
+    libcellml::ModelPtr model = parser->parseModel(fileContents("generator/initialized_variable_of_integration.cellml"));
+
+    EXPECT_EQ(size_t(0), parser->issueCount());
+
+    const std::vector<std::string> expectedIssues = {
+        "Variable 'time' in component 'my_component' of model 'initialized_variable_of_integration' cannot be both a variable of integration and initialised.",
```
nickerso, line 55: Not sure about this message. Its more that the generator in libCellML doesn't know what to do in this case. There is nothing saying this can't happen :) Not sure the below suggestion is any better...

`Variable 'time' in component 'my_component' of model 'initialized_variable_of_integration' is both a variable of integration and initialised; this is not currently supported.`
 
Not a show stopper...

tests/generator/generator.cpp
```diff
@@ -44,32 +44,48 @@ TEST(Generator, emptyModel)
     EXPECT_EQ(EMPTY_STRING, generator->implementationCode());
 }
 
-TEST(Generator, generatorIssues)
+TEST(Generator, initializedVariableOfIntegration)
```
nickerso, line 47: A real mix of `z` and `s` throughout here...

tests/generator/generator.cpp
```diff
@@ -101,8 +117,8 @@ TEST(Generator, twoVariablesOfIntegration)
 
     EXPECT_EQ(size_t(0), parser->issueCount());
 
-    const std::vector<std::string> expectedIsues = {
-        "Variable 'time' in component 'main' of model 'two_variables_of_integration' and variable 'other_time' in component 'sub_sub_sub' of model 'two_variables_of_integration' cannot both be a variable of integration.",
+    const std::vector<std::string> expectedIssues = {
+        "Variable 'time' in component 'main' of model 'two_variables_of_integration' and variable 'other_time' in component 'sub_sub_sub' of model 'two_variables_of_integration' cannot both be the variable of integration.",
```
nickerso, line 121: Another one where I'm not a fan of the wording in the message here. They certainly can both be the variable of integration, its just a case that is not currently supported by the generator.

tests/generator/generator.cpp
```diff
@@ -202,15 +218,81 @@ TEST(Generator, variableInitializedTwice)
...
+
+TEST(Generator, nonConstantInitialisingVariable)
+{
+    libcellml::ParserPtr parser = libcellml::Parser::create();
+    libcellml::ModelPtr model = parser->parseModel(fileContents("generator/non_constant_initialising_variable.cellml"));
+
+    EXPECT_EQ(size_t(0), parser->issueCount());
+
+    const std::vector<std::string> expectedIssues = {
+        "Variable 'x' in component 'main' of model 'my_model' is initialised using variable 'k2', but it is not a constant.",
```
nickerso, line 252: so the error message here is not particularly helpful, but more importantly is why is this case not supported? `k2` is set equal to `k1` which is a constant, so `k2` should also be a constant, or computed constant, right? But regardless, is it not possible to initialise a variable using the initial value of a non-constant variable?

(fine if its not, but again a case of the generator not currently supporting this case which I think should be clear in the message)

tests/resources/generator/cellml_unit_scaling_voi_direct/model.c
```diff
@@ -0,0 +1,57 @@
...
+    states[1] = 5.0;
+}
+
+void computeComputedConstants(double *variables)
+{
+}
+
+void computeRates(double voi, double *states, double *rates, double *variables)
+{
+    rates[0] = voi/1.0;
```
nickerso, line 51: a lot of excessive `/1.0` here :)

nickerso (approved):
Looks pretty good to me, a few minor comments on the error messages that are nothing to do with the main point of this PR :)

tests/generator/generator.cpp
agarny: This code will be reworked as part of our forthcoming `Analyser` class (see issue #499), so probably best to leave it for then.

tests/generator/generator.cpp
agarny: From a pure CellML viewpoint, I agree with you. However, our future `Analyser` class (see issue #499) won't allow this since we don't support it when it comes to code generation. So, maybe the analyser will need an option so that we can either generate an error or a warning (hint?) depending on whether we want to allow several VOIs. By default, we will probably only allow one VOI since this is going to be our main use case (for the generator). Anyway, something for issue #499 anyway.

nickerso: exactly. A hint if it looks like a missing equivalence could fix the issue...

tests/generator/generator.cpp
agarny: The message is about `x` being initialised using `k2`. However, `k2` is not a (true) constant, hence the code generator (analyser in the future) doesn't allow it.

tests/resources/generator/cellml_unit_scaling_voi_direct/model.c
agarny: That's because of the CellML file that contains those divisions by `1.0` because of units conversions. I agree that it doesn't look great, but as a first step, I think it's fine. We can certainly "optimise" the code generator later on, if really needed (since a C compiler will most likely skip the division for us).

tests/generator/generator.cpp
kerimoyle: Hmmm ... so we need a place to explain these kinds of things in more detail (nudge, nudge) ;)

tests/generator/generator.cpp
agarny: Sounds reasonable indeed.

tests/generator/generator.cpp
agarny: That will be part of issue #499. In addition to the issue message, we can have a URL that points to a detailed description (with an example, if needed) of the problem.

tests/generator/generator.cpp
agarny: Yes, that's going to be done as part of issue #487... :)

PR #573, opened 2020-03-11
I571 scaling factor bug

kerimoyle:
Here's the fix for the #571 bug.


kerimoyle:
If the tests pass here then it's ready for review please @agarny @hsorby

src/units.cpp
```diff
@@ -594,8 +602,8 @@ bool Units::compatible(const UnitsPtr &units1, const UnitsPtr &units2)
 
 bool Units::equivalent(const UnitsPtr &units1, const UnitsPtr &units2)
 {
-    // Units must be compatible to return a non-zero scaling factor.
-    return Units::scalingFactor(units1, units2) == 1.0;
+    // Units must be compatible and return a scaling factor of 1.0.
+    return Units::scalingFactor(units1, units2, true) == 1.0;
```
agarny: Why pass `true` to `Units::scalingFactor()` since it's the default value for that parameter... or am I missing something?...

agarny (changes requested):
Looks good to me pending [my question](https://github.com/cellml/libcellml/pull/573#discussion_r391300450).

src/units.cpp
kerimoyle: Oops, I thought it was false as a default? There was a test failing until I added this (though if you're right then it was for other reasons ... !). Either way the comment needed updating :). Feel free to remove the default.

agarny: Done.

agarny (dismissed):

src/units.cpp
```diff
@@ -157,22 +157,17 @@ bool Units::UnitsImpl::isBaseUnit(const std::string &name) const
...
  * @param direction The direction to update multiplier. Either 1 or -1.
  *
  * @return Either @c true or @c false, depending if the units were successfully updated.
  */
-bool updateUnitMultiplier(double &multiplier,
-                          const UnitsPtr &units,
-                          double uExp, double logMult,
-                          int direction)
+
+bool updateUnitMultiplier(double &multiplier, const UnitsPtr &units, int direction)
```
hsorby: Have the output parameter last, this is standard practice for C++

hsorby (changes requested):
Other than the single comment I'm happy with this.

src/units.cpp
kerimoyle: ... so the `multiplier` at the end?  Is that what you mean?  or the `update` bool last in the comments?  Sorry, I just deleted the middle ones, so I'm not sure which bit to fix!

kerimoyle: OK, got it, thanks for doing that.

hsorby (approved):

src/units.cpp
hsorby: np

agarny (approved):

PR #577, opened 2020-03-17
Utilities: have `convertToDouble()` return `inf` or `-inf` if a conversion cannot be done

### LINKED ISSUE #576: Utilities: `convertToDouble()` should return `NaN` not `Inf` if the conversion cannot be done
Right now, we return `Inf` if `convertToDouble()` fails. However, I would expect `NaN` to be returned instead.

agarny:
Addresses issue #576.


tests/parser/parser.cpp
```diff
@@ -595,7 +595,7 @@ TEST(Parser, modelWithInvalidUnits)
         "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
         "<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"model_name\">\n"
         "  <units name=\"fahrenheitish\">\n"
-        "    <unit exponent=\"inf\" units=\"kelvin\"/>\n"
+        "    <unit exponent=\"nan\" units=\"kelvin\"/>\n"
```
nickerso: Why is the `@exponent` changing in this case? the source value is `35.0E+310` which is presumably `inf` not `nan`?

agarny: We use `std::stod()` to convert a `std::string` to a `double`, and either the conversion is successful or it is not.

If it is not, it is either because:
 - an invalid argument was passed; or because
 - the passed value is out of range.

This is likely to be the latter here, but then we can't tell whether it should be `+inf` or `-inf` unless we were to check the sign of the passed value...?

So, maybe we could do the following: try to convert using `std::stod()` and if we get no exception then it's all good otherwise if we get:
 - an `invalid_argument` exception then we return `nan`; or if we get
 - an `out_of_range` exception then we check whether the passed value starts with `-`. If it does then we return `-inf` otherwise we return `inf`.

Would that be better?

nickerso (dismissed):
I like that a lot better, thanks! I'm not sure how coverage is kept at 100% given there is no test added for the NaN case, but that would be the only suggestion I have to improve things.

nickerso:
Oops, knew I should have read the docs before approving. The docs are pretty clear that this method expects strings to be valid real numbers, so I'm not sure why we need to worry about the different return paths. I can see why we need to catch an exception to avoid the dreaded seg-fault, but given that in normal usage this should never happen, can't we just fall back on the existing code to return an arbitrary `NaN` or `inf`?

If this method is going to do all the checking, then we probably don't also need the `isCellMLReal` method and the documentation should be updated to reflect this. But given `isCellMLReal` does more checking then it might be better to make use of that...

agarny:
Hmm... indeed, I clearly didn't RTFM either and I don't like that it expects a string to be a valid real number. It makes its use cumbersome in some cases (as in our code generator, which is where the 'nan' bit was captured).

Anyway, from what I understand, we don't need to worry about `convertToDouble()` having to return a `nan`. However, we may still want to distinguish between `inf` and `-inf`? Also, I should update my use of `convertToDouble()` in the code generator and combine that with a call to `isCellMLReal()` wherever needed (that's what I mean by cumbersome!). Ugly, but it will agree with the documentation. Yuck! :(

kerimoyle:
... RTFM? 
AEWHTDWFTMA! (as engineers we have to deal with far too many acronyms!)

agarny:
> ... RTFM?

https://en.wiktionary.org/wiki/RTFM

> AEWHTDWFTMA! (as engineers we have to deal with far too many acronyms!)

Now, I would have been really impressed if that acronym did actually exist! Disappointed that it doesn't though.

nickerso (approved):
I'm not opposed to this change going in, but I can't see the value in checking for +/- infinity. The method should never be returning `inf` as its an internal method for which usage is tightly controlled. Will add an approve in case a second approval can be found. The "unrelated" change in `units.cpp` should likely be using the `isCellMLReal` method rather than straight out assigning the value of the string.

agarny:
> I'm not opposed to this change going in, but I can't see the value in checking for +/- infinity. The method should never be returning `inf` as its an internal method for which usage is tightly controlled.

Ideally, it would never return `inf`, but it may happen in a case like the one in the test. So, I think there is still a need for `convertToDouble()` to return `inf`. Then, because we should handle the `inf` case, we should also handle the `-inf` case.

> Will add an approve in case a second approval can be found. The "unrelated" change in `units.cpp` should likely be using the `isCellMLReal` method rather than straight out assigning the value of the string.

Re `units.cpp`, I haven't checked the code in details, but I believe that `u.mExponent` and `u.mMultiplier` are known to be "valid" real numbers? Probably the reason the `develop` code calls `std::stod()` and the reason I replaced it with a call to `convertToDouble()` in case we are outside the `double` range.

hsorby (approved):

PR #580, opened 2020-03-17
Utilities: have `convertToInt()` return `0` if a conversion is unsuccessful

### LINKED ISSUE #579: Utilities: `convertToInt()` should return `0` if a conversion is unsuccessful
Right now, `convertToInt()` crashes libCellML if we pass to it a value that is not convertible to an integer. Ideally, we would do something similar to what we have done / are doing in issue #577, but we can't do that since both `nan` and `inf` are `double`s.

So, all we can probably do is to return `0` if we cannot convert a `std::string` to an `int`?

agarny:
Addresses issue #579.


nickerso:
as with #577, `convertToInt` by definition can not fail so this should not be an issue for internal methods. The issue in this case is that `isCellMLInteger` is only checking for an integer value whereas `convertToInt` is assuming the more restrictive `int` type.

The solution should likely be to ensure `isCellMLInteger` is consistent with `convertToInt` so that the appropriate `*IsSet` flags can be set correctly, rather than relying on a 0 return value which is usually a valid value.

agarny:
`convertToInt()` can and will fail if you give it a valid integer value that is out of range of the `int` type.

agarny:
@nickerso and @hsorby: so, how shall we handle out-of-range integer numbers?...

nickerso:
adding a `isInt()` method that supplements the `isCellMLInteger()` by restricting the range to that supported by the actual integer type used in the libCellML implementation. Or just add that check directly into `isCellMLInteger()` and adjust any documentation and error messages accordingly?

agarny:
@kerimoyle actually made a similar suggestion (see [here](https://github.com/cellml/libcellml/issues/579#issuecomment-600338598)).

Thinking of it, both @kerimoyle's options would require updating the specs to make it clear that an integer should be within the 32/64-bit `int` range, and the same for a real number (having to be within the 32/64-bit `double` range). This is what her first option says, but what her second option also implies since validation can only generate errors.

nickerso:
The spec is different from the implementation in libCellML. It is perfectly valid for the spec to define what it means as an `integer`, which it does, and the validation should test against those rules. And it is also fine that an implementation has to tie this down to a specific data type such as `int` which is likely more restrictive than the specification. There is no need to restrict the specification to specific data types - that is an implementation issue.

agarny:
> The spec is different from the implementation in libCellML. It is perfectly valid for the spec to define what it means as an `integer`, which it does, and the validation should test against those rules. And it is also fine that an implementation has to tie this down to a specific data type such as `int` which is likely more restrictive than the specification. There is no need to restrict the specification to specific data types - that is an implementation issue.

That sounds reasonable to me. Ok, I am closing this PR (and the corresponding issue) then and I am going to create a new issue for `convertToInt()` and `convertToDouble()` to take the above into account.

agarny:
See issue #583.

PR #584, opened 2020-03-22
Utilities: have `convertToInt()` and `convertToDouble()` check whether the value to convert is out of range

### LINKED ISSUE #583: Utilities: have `convertToInt()` and `convertToDouble()` check whether the value to convert is out of range
This issue is a result of some discussions in issue #579 and PR #356 

Basically, `convertToInt()` and `convertToDouble()` assume that the string that is passed to them represents a valid integer and a valid real number, respectively. That is fine. However, a valid integer/real number may be out of range for the `int`/`double` type, in which case we need the caller to know about it.

Right now, the methods have the following signature:

```
int convertToInt(const std::string &candidate);
double convertToDouble(const std::string &candidate);
```

We could therefore change them to:

```
bool convertToInt(const std::string &from, int &to);
bool convertToDouble(const std::string &from, double &to);
```

The idea is that if `convertToInt()`/`convertToDouble()` was successful then the method returns `true` and `to` has the expected value. On the other hand, if `convertToInt()`/`convertToDouble()` failed because `from` represents a value that is out of range for the `int`/`double` type then it would return `false` and `to` would be ignored.

Does that sound reasonable @hsorby, @nickerso and @kerimoyle?

agarny:
Addresses issue #583.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -626,7 +626,7 @@ void Parser::ParserImpl::loadUnit(const UnitsPtr &units, const XmlNodePtr &node)
         } else if (attribute->isType("exponent")) {
             if (isCellMLReal(attribute->value())) {
                 if (!convertToDouble(attribute->value(), exponent)) {
-                    // TODO This value won't be saved for validation later, so it does need to be reported now
+                    // TODO This value won't be saved for validation later, so it does need to be reported now.
                     ErrorPtr err = Error::create();
                     err->setDescription("Unit referencing '" + node->attribute("units") + "' in units '" + units->name() + "' has an exponent with the value '" + attribute->value() + "' that is a representation of a CellML real valued number, but out of range of the 'double' type.");
                     err->setUnits(units);
@@ -634,7 +634,7 @@ void Parser::ParserImpl::loadUnit(const UnitsPtr &units, const XmlNodePtr &node)
                     mParser->addError(err);
                 }
             } else {
-                // TODO This value won't be saved for validation later, so it does need to be reported now
+                // TODO This value won't be saved for validation later, so it does need to be reported now.
                 ErrorPtr err = Error::create();
                 err->setDescription("Unit referencing '" + node->attribute("units") + "' in units '" + units->name() + "' has an exponent with the value '" + attribute->value() + "' that is not a representation of a CellML real valued number.");
                 err->setUnits(units);
@@ -644,7 +644,7 @@ void Parser::ParserImpl::loadUnit(const UnitsPtr &units, const XmlNodePtr &node)
         } else if (attribute->isType("multiplier")) {
             if (isCellMLReal(attribute->value())) {
                 if (!convertToDouble(attribute->value(), multiplier)) {
-                    // TODO This value won't be saved for validation later, so it does need to be reported now
+                    // TODO This value won't be saved for validation later, so it does need to be reported now.
                     ErrorPtr err = Error::create();
                     err->setDescription("Unit referencing '" + node->attribute("units") + "' in units '" + units->name() + "' has a multiplier with the value '" + attribute->value() + "' that is a representation of a CellML real valued number, but out of range of the 'double' type.");
                     err->setUnits(units);
@@ -652,7 +652,7 @@ void Parser::ParserImpl::loadUnit(const UnitsPtr &units, const XmlNodePtr &node)
                     mParser->addError(err);
                 }
             } else {
-                // TODO This value won't be saved for validation later, so it does need to be reported now
+                // TODO This value won't be saved for validation later, so it does need to be reported now.
                 ErrorPtr err = Error::create();
                 err->setDescription("Unit referencing '" + node->attribute("units") + "' in units '" + units->name() + "' has a multiplier with the value '" + attribute->value() + "' that is not a representation of a CellML real valued number.");
                 err->setUnits(units);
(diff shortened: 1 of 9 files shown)


src/parser.cpp
```diff
@@ -636,9 +643,16 @@ void Parser::ParserImpl::loadUnit(const UnitsPtr &units, const XmlNodePtr &node)
             }
         } else if (attribute->isType("multiplier")) {
             if (isCellMLReal(attribute->value())) {
-                multiplier = convertToDouble(attribute->value());
+                if (!convertToDouble(attribute->value(), multiplier)) {
+                    // TODO This value won't be saved for validation later, so it does need to be reported now
```
kerimoyle: Fullstop here too.

src/parser.cpp
```diff
@@ -625,7 +625,14 @@ void Parser::ParserImpl::loadUnit(const UnitsPtr &units, const XmlNodePtr &node)
             prefix = attribute->value();
         } else if (attribute->isType("exponent")) {
             if (isCellMLReal(attribute->value())) {
-                exponent = convertToDouble(attribute->value());
+                if (!convertToDouble(attribute->value(), exponent)) {
+                    // TODO This value won't be saved for validation later, so it does need to be reported now
```
kerimoyle: Fullstop ;)

src/parser.cpp
kerimoyle: (and below)

tests/validator/validator.cpp
```diff
@@ -946,13 +956,15 @@ TEST(Validator, integerStrings)
         "Reset in component 'component' referencing variable 'variable' has a non-integer order value ''.",
         "Reset in component 'component' referencing variable 'variable' has a non-integer order value '-'.",
         "Reset in component 'component' referencing variable 'variable' has a non-integer order value 'odd'.",
+        "Reset in component 'component' referencing variable 'variable' has an out of range integer order value '12345678901'.",
     };
```
kerimoyle, line 960: I thought that the integer ceiling was a whole lot higher than this ... ? Or is libcellml going to be 32-bit?

src/units.cpp
```diff
@@ -370,12 +370,12 @@ void Units::unitAttributes(size_t index, std::string &reference, std::string &pr
     reference = u.mReference;
     prefix = u.mPrefix;
     if (!u.mExponent.empty()) {
-        exponent = convertToDouble(u.mExponent);
+        convertToDouble(u.mExponent, exponent);
     } else {
```
kerimoyle: If the conversion to double fails, what do you expect the exponent to be set to? In one line it's defaulting to 1.0, in the other it's failing silently and giving an unknown value to the returned exponent ... I think it might be better to test the return bool here too, or is that caught somewhere else ...?

src/utilities.h
```diff
@@ -212,37 +212,38 @@ static const std::map<Variable::InterfaceType, const std::string> interfaceTypeT
     {Variable::InterfaceType::PUBLIC_AND_PRIVATE, "public_and_private"}};
 
 /**
- * @brief Convert the @p candidate @c std::string to a @c double.
+ * @brief Convert the @p from @c std::string to the @p to @c double.
```
kerimoyle: I very much prefer "input" and "output" (or something similar) than "from" and "to", just because it makes reading the documentation a little less mind-bending.  Very minor, and personal preference only though!

kerimoyle (changes requested):
Looking good, minor changes only...

src/parser.cpp
agarny: Yes, my bad for just copying/pasting things... :) Done in commit 66e27d61.

src/parser.cpp
agarny: Done in commit 66e27d61.

tests/validator/validator.cpp
agarny: libCellML is 64-bit, but an `int` is always 32-bit (see [here](https://en.cppreference.com/w/cpp/language/types)).

src/utilities.h
agarny: Fair enough, I am now using `in` and `out` (shorter than `input` and `output` :)). Done in commit 94b2a7e2.

src/units.cpp
agarny: Good question! It's not my code in the first place, so I merely updated it, that's all. I mean, the very original code was `exponent = std::stod(u.mExponent);`, so I am assuming that at this stage we know that `u.mExponent` is valid...?

Ok, I am going to double check this...

tests/validator/validator.cpp
kerimoyle: In that case we should probably update the test for out of range prefix in the Units module to a lower number?

agarny: We could, but it doesn't matter... or does it? The most important is that the value is indeed out of range, and it is.

kerimoyle: Eh, I think it does ... or at least the comment there should change.  It's my bad - I thought we were 64-bit :)

src/units.cpp
agarny: Ok, that method is used in a lot of places. So, if the conversion fails, we now default to `1.0` (see commit 529b2ab5). Not sure if that's the best way to go about this though. @hsorby?

tests/validator/validator.cpp
agarny: Oops, I hadn't taken into account the comment. Agreed. Done in commit 3a810dd1.

src/units.cpp
kerimoyle: In line with the earlier design changes, I don't think it should be defaulting to anything!  I do think that the `false` return should be caught and passed back to the calling function for error handling though ... ?

agarny: I would probably agree with that, but that method is used so many times that I would rather leave it to the person who wrote that code in the first place. Also, this goes beyond the purpose of this PR.

agarny:
I believe this is ready for re-reviewing.

hsorby (approved):
Looks good to me.

kerimoyle:
Yep, happy too.  Can we make an issue to check the other usages of the "default" value though?

kerimoyle (approved):

agarny:
> Yep, happy too. Can we make an issue to check the other usages of the "default" value though?

I imagine you are talking about `Units::unitAttributes()`? If so then it would probably be a good idea indeed. However, as mentioned before, I am not familiar with that code, so happy for someone else to create such an issue, if really needed.

Issue #590, opened 2020-05-05
Cannot print a model with imports?


kerimoyle:
Say I have two models, one of which imports the other (see below).  According to [2.15.1 ](https://cellml-specification.readthedocs.io/en/latest/reference/formal_and_informative/specB15.html)and [3.4.2.2](https://cellml-specification.readthedocs.io/en/latest/reference/formal_and_informative/specC04_component_reference.html#specc-component-reference), I'm allowed to have a map_variables item that points to an imported component.  How can I get this output using the API?
- if I serialise the model without flattening, there can be no equivalences stored as the variables are not instantiated;
- if I serialise the model after flattening, then it's no longer an import situation ... ?

```
<?xml version="1.0" encoding="UTF-8"?>
<model xmlns="http://www.cellml.org/cellml/2.0#" name="import_example">
  <import xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="importMe.cellml">
    <component component_ref="importedComponent" name="importMe"/>
  </import>
  <component name="localComponent">
    <variable name="x" units="dimensionless" interface="public_and_private"/>
  </component>
  <connection component_1="localComponent" component_2="importedComponent">
    <map_variables variable_1="x" variable_2="x"/> ///**** NOTE THIS BIT!!
  </connection>
</model>
```
Inside importMe.cellml:
```
<?xml version="1.0" encoding="UTF-8"?>
<model xmlns="http://www.cellml.org/cellml/2.0#" name="import_me">
  <component name="importMe">
    <variable name="x" units="dimensionless" interface="public_and_private"/>
  </component>
</model>
```

nickerso:
Maybe throw together a test which demonstrates this behaviour, that would be the easiest way to look at this. Cause as it stands this sounds like a huge bug.

hsorby:
small bug

kerimoyle:
I've written two small tests in https://github.com/cellml/libcellml/pull/591 ... the second one (in variable tests) won't compile, but it shows what I'd expect to be able to do?

nickerso:
So as we remember how this was designed to work, there is actually no bug (see discussion on #591). But there could be new API to make the solution more convenient.

kerimoyle:
> So as we remember how this was designed to work, there is actually no bug (see discussion on #591). But there could be new API to make the solution more convenient.

[...]

PR #591, opened 2020-05-05
Printer to print local children of imported components

### LINKED ISSUE #609: Feature or bug? Local component child of imported parent not printed?
Parsing a model and then printing it gives different results:
**Parsing this:**
```
<?xml version="1.0" encoding="UTF-8"?>
<model xmlns="http://www.cellml.org/cellml/2.0#" name="import_example">
  <import xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="importedModel.cellml">
    <component component_ref="importMe" name="importedComponent"/>
  </import>
  <component name="localComponent">
    <variable name="x" units="dimensionless" interface="public_and_private"/>
  </component>
  <encapsulation>
    <component_ref component="importedComponent">
      <component_ref component="localComponent"/>
    </component_ref>
  </encapsulation>
</model>
```
... gives this model (NB: components are included in their encapsulation structure)
```
 MODEL: 'import_example'
    UNITS: 0 custom units
    COMPONENTS: 1 components
        [0]: importedComponent
            VARIABLES: 0 variables
            COMPONENT importedComponent has 1 child components:
                [0]: localComponent
                    VARIABLES: 1 variables
                        [0]: x [dimensionless]
```

... which is serialised to this:
```
<?xml version="1.0" encoding="UTF-8"?>
<model xmlns="http://www.cellml.org/cellml/2.0#" name="import_example">
  <import xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="importedModel.cellml">
    <component component_ref="importMe" name="importedComponent"/>
  </import>
  <encapsulation>
    <component_ref component="importedComponent">
      <component_ref component="localComponent"/>
    </component_ref>
  </encapsulation>
</model>
```
... ie: the `localComponent` definition is missing from the printed version.  The parsed model is valid, and neither the parser nor the printer have any issues.  You can see this behaviour in #591 ... but it's now to do with the printer not the parser ...

kerimoyle:
Addresses #590.

Additional to this test, I'd also expect to be able to specify this relationship (equivalence between imported variables) through the API.  This isn't possible at the moment?

Also closes #609


tests/variable/variable.cpp
```diff
@@ -1699,3 +1699,38 @@ TEST(Variable, variableInterfaceDontDowngrade)
...
+
+    auto importer = libcellml::ImportSource::create();
+    importer->setUrl("importedModel.cellml");
+    importedComponent->setImportSource(importer);
+    importedComponent->setImportReference("importMe");
+
+    // Create equivalent variables: THIS BIT I DON'T KNOW HOW TO DO!
+    // The importedComponent is not instantiated, so can't point to any of its variables...
+    // libcellml::Variable::addEquivalence(x, importedComponent->variable("x"));
+    libcellml::Variable::addEquivalence(localComponent, "x", importedComponent, "x");
```
hsorby: Why aren't you using the reference to 'x' that you have for the local component?  If you believe that imported component also has a variable 'x' you can create that variable and add it to the component.  With that you can use the API.

I could see us having the API (component, variable, component, string)

kerimoyle: And what about when both components are imported? Then you'd need (component, string, component, string)

kerimoyle: And if I add a variable to the importedComponent, isn't that a bit weird? Then I have two variables with the same name?

hsorby: Having two variables with the same name isn't weird, there are loads of duplicate variable names.

hsorby: I think then that you would make the whole set:

- (component, variable, component, string)
- (component, string, component, string)
- (component, string, component, variable)
- (component, string, component, string)

kerimoyle: Duplicate variable names, sure, but not in the same component?
But yes to the whole set of combinations, though you have a duplicate in there ... :)

kerimoyle: or, better:
variable, variable
variable, component, string,
(possibly?) component, string, variable (?)
component, string, component, string

nickerso: I'd thought last time that I'd talked with @hsorby about this issue, the idea was that you would simply add variables to the imported component and then the equivalences work as normal. But then when printing the model the imported component would not have the variables printed (as they make no sense) and but the equivalences (connections) would still be printed.

hsorby: Yes but I can't see any reason why you couldn't add the string form for convenience.

This is possible at the moment but you can only do it one way. I think if adding a string form would make it easier for users and doesn't make problems and tests still pass at the moment I cannot say it is wrong to add the additional API.

hsorby:
To be clear it is possible to do this at the moment.

tests/variable/variable.cpp
kerimoyle: @nickerso I don't quite follow the logic of adding the variable to the component when it's really already there in the import?  Do you also have to add units for the model (before flattening) to pass validation?  Do you have to also have the units in your instantiated model too?  I'm missing something ...

nickerso: > Yes but I can't see any reason why you couldn't add the string form for convenience.

Yep, totally agree.

nickerso: > it's really already there in the import

Its only there once the import is resolved and instantiated. For exactly this case of handling CellML connections as variable equivalences it is sometimes necessary to make these dummy variables. They are never used for anything else (I think) so units/interfaces/etc. are not needed.

nickerso: and since this would be an imported component these dummy variables are never used in flattening, validation, printing...

kerimoyle:
So is this a thing? What should I do in the tutorials? Use dummy variables or wait for an overloaded function?

nickerso:
> What should I do in the tutorials? Use dummy variables or wait for an overloaded function?

Either push ahead with the dummy variable method in the tutorial (this works now) or implement the proposed new API for the convenience methods and use them in the tutorials.

Personally, I'm not sure how much of a convenience that these methods would be so I'd continue on the tutorials with the dummy variables.

tests/variable/variable.cpp
kerimoyle: The units are used in validation because we now test for unit matching across equivalent variables.  :( I get an error:
```Validator error[0]:
     Description: Variable 'g_Na' in component 'sodiumChannel' has units of 'mS_per_cm2' and an equivalent variable 'g_Na' in component 'sodiumChannelInit' with non-matching units of ''. The mismatch is: 
    See section X.Y.Z in the CellML specification.```

... where `sodiumChannel` is a local component, and `sodiumChannelInit` is an import.  Is this expected? Or should the validator not test for units in a component which has a non-null import source?

Comment (kerimoyle) [reply tests/variable/variable.cpp]: Also interface errors from the dummy variables: 
```
Description: Variable 'm' in component 'mGateInit' has no interface type set. The interface type required is 'public'.
```

**Comment by kerimoyle** [comment]
I'm going to revert the second commit, as discussion about workarounds there has obscured the actual bug ... I'll add the other as a "nice to have maybe" issue on its own.

**Comment by kerimoyle** [comment]
Wow, this PR went all over the show!  In the end the bug was in the `Printer`, not the `Parser` as I'd originally thought.  This PR has three aspects:
- Fixed the bug in the printer that was missing local children of imported components (src/printer.cpp).  This is the main bit!
- Improved the debug printing of the model structure (test_utils.cpp)
- Added tests for printing/parsing of imports, encapsulations, and connections (tests and resources files)

**Comment by kerimoyle** [comment]
Ping @agarny @hsorby @nickerso

**Review by hsorby — DISMISSED** [review]

**Review by agarny — DISMISSED** [review]
Looks good. I have cleaned up various things though. One thing that I would like to see however is for the `connectionsPersistAfterImporting`, `connectionsAsymmetricWithRepeatedUnits` and `connectionsSymmetricWithMismatchedUnits` to check the serialisation of the model. Right now, we test that this or that component/variable exists, etc. which makes the test rather difficult to read/understand (for me, at least).

File: tests/test_utils.cpp [hunk 2]
```diff
@@ -60,44 +60,27 @@ void printIssues(const libcellml::LoggerPtr &l, bool headings, bool causes, bool
...
 
-    std::cout << spacer << "COMPONENTS: " << model->componentCount() << " components" << std::endl;
-    for (size_t c = 0; c < model->componentCount(); ++c) {
-        auto component = model->component(c);
-        prettyPrintComponentToStdout(component, c, spacer + spacer, includeMaths);
-    }
-}
-
-void prettyPrintComponentToStdout(const libcellml::ComponentPtr &component, size_t const c, std::string const spacer, bool includeMaths)
+void printComponent(const libcellml::ComponentPtr &component, size_t c, const std::string &indent, bool includeMaths)
```
hsorby: This was recently changed to 'prettyPrint*' are we done with that now?

tests/test_utils.cpp
```diff
@@ -60,44 +60,27 @@ void printIssues(const libcellml::LoggerPtr &l, bool headings, bool causes, bool
...
-    std::cout << spacer << "[" << c << "]: " << component->name();
     if (component->id() != "") {
-        std::cout << " id: " << component->id();
+        std::cout << ", id: " << component->id();
     }
-    std::cout << std::endl;
 
-    std::cout << spacer << local << "VARIABLES: " << component->variableCount() << " variables" << std::endl;
+    std::cout << std::endl;
+    std::cout << indent << INDENT << "VARIABLES: " << component->variableCount() << " variables" << std::endl;
```
hsorby: Mixing use of 'indent' and 'INDENT' here and in other places.

tests/printer/printer.cpp
```diff
@@ -308,13 +309,14 @@ TEST(Printer, printModelImportingModelParentComponent)
     auto printer = libcellml::Printer::create();
     auto serialisedModel = printer->printModel(model);
 
-    EXPECT_EQ(serialisedModel, fileContents("importingModelParentComponent.cellml"));
+    EXPECT_EQ(serialisedModel, modelContents);
```
hsorby: These entries should be the other way around, we expect 'modelContents' to be equal to what we actually got 'serialiseModel' after our manipulations.

hsorby (dismissed):
A few minor things.  Importantly why change the name of the test utility printing functions again?

tests/test_utils.cpp
agarny: The utility was originally named `printComponent()` and I don't see why it was renamed to something unnecessarily long and that I would never remember, hence I reverted the name of the method to what it is in `develop`.

tests/test_utils.cpp
agarny: No, no mixing. `INDENT` is a constant used to indent while `indent` is an indent that we need for calling `printComponent()` recursively.

tests/printer/printer.cpp
agarny: Good point. Done in commit 2af7803c.

kerimoyle:
I can't quite tell who can approve this now that three of us have worked on it? @hsorby @agarny @nickerso ?

agarny:
> I can't quite tell who can approve this now that three of us have worked on it? @hsorby @agarny @nickerso ?

I don't know about @hsorby's changes, but mine were purely cosmetic. So, I think it's fine for me to still review this PR.

agarny:
> Looks good. I have cleaned up various things though. One thing that I would like to see however is for the `connectionsPersistAfterImporting`, `connectionsAsymmetricWithRepeatedUnits` and `connectionsSymmetricWithMismatchedUnits` to check the serialisation of the model. Right now, we test that this or that component/variable exists, etc. which makes the test rather difficult to read/understand (for me, at least).

Note that I would like to see the above... unless someone can convince that it's not necessary... :)

kerimoyle:
> Note that I would like to see the above... unless someone can convince that it's not necessary... :)

@agarny! It's not necessary! (Did that work?) 

I disagree ... I think in general it's better to have nicely isolated tests (ie: covering as little of the API as possible), and I find it hard to read big chunks of raw CellML, when all the details of the test are quite small.  A couple of the other tests here use the parser/printer, but only where that's relevant to what the test is demonstrating.  The majority do not.  What do others think? @hsorby? @nickerso?

hsorby:
I am very pro building minimal tests for unit tests.  Only building what is actually needed to exercise the API under test.

kerimoyle:
Edit: Ignore me.  All good.

~I've found another bug ... so this needs to be reworked :( and could possibly address #652 at the same time as they're related to printing imports.  We need to:~
- ~add some recursion into the printer when forming lists of imports (at present it only checks direct children)~
- ~include imported units in the same forming function (at present only collects components).~

kerimoyle:
Booooo ... @hsorby the gremlins are back ... CI is building on f32d900 and the fix in 2fb347d means the tests should pass :(

tests/test_utils.cpp
```diff
@@ -60,87 +60,104 @@ void printIssues(const libcellml::LoggerPtr &l, bool headings, bool causes, bool
...
-    }
-
-    // 2.a    Print any custom units of the model
-    std::cout << "The model defines " << model->unitsCount()
-              << " custom units:" << std::endl;
-    for (size_t u = 0; u < model->unitsCount(); ++u) {
-        std::cout << "  Units[" << u << "] is '" << model->units(u)->name() << "'"
-                  << std::endl;
-    }
+static const std::string INDENT = "    ";
```
hsorby: To reduce confusion with later uses of `indent` I think this variable should be renamed to `FIXED_INDENT`.

tests/variable/variable.cpp
```diff
@@ -1700,6 +1700,115 @@ TEST(Variable, variableInterfaceDontDowngrade)
...
+
+    auto x1 = model->component("importedComponent")->component("child1")->variable("x");
+    auto x2 = model->component("importedComponent")->component("child2")->variable("x");
+    EXPECT_EQ(size_t(1), x1->equivalentVariableCount());
+    EXPECT_EQ(size_t(1), x2->equivalentVariableCount());
+    EXPECT_EQ(x1, x2->equivalentVariable(0));
+    EXPECT_EQ(x2, x1->equivalentVariable(0));
+}
+
+TEST(Variable, connectionsAsymmetricWithRepeatedUnits)
```
hsorby: I don't understand what this test is doing.  If I remove it I don't fail any tests nor reduce coverage below 100%.

tests/variable/variable.cpp
```diff
@@ -1700,6 +1700,115 @@ TEST(Variable, variableInterfaceDontDowngrade)
...
+
+    auto x1 = model->component("c1")->variable("x");
+    auto x2 = model->component("c2")->variable("x");
+    EXPECT_EQ(size_t(1), x1->equivalentVariableCount());
+    EXPECT_EQ(size_t(1), x2->equivalentVariableCount());
+    EXPECT_EQ(x1, x2->equivalentVariable(0));
+    EXPECT_EQ(x2, x1->equivalentVariable(0));
+}
+
+TEST(Variable, connectionsSymmetricWithMismatchedUnits)
```
hsorby: I don't understand what this test is doing.  If I remove it I don't fail any tests nor reduce coverage below 100%.

hsorby (dismissed):

tests/test_utils.cpp
kerimoyle: f0f249dc

tests/variable/variable.cpp
kerimoyle: 1) Removing a test kinda means that it's not gonna fail ... ;) 
2) Test coverage should test all the behaviours, not just all the lines of code, no? 100% coverage should be the bare minimum ...
3) It's been so long I'm struggling to remember why they were written.  I suspect it was to do with mismatched units causing an equivalence not to be added, hence the two tests where one (this one) has valid units between the connected variables, and the other where they're not the same (ms and mm).  This PR came through a maze of issues though, so I'll delete the tests as I'm pretty sure the behaviour is covered elsewhere, and the changes to the `Importer` class will overwrite much of this in any case.  8114ded1

tests/variable/variable.cpp
kerimoyle: 8114ded1

hsorby (dismissed):

agarny:
This PR seems to be broken now.

hsorby:
Just minor API change required I think.

hsorby (dismissed):

tests/printer/printer.cpp
```diff
@@ -294,6 +294,40 @@ TEST(Printer, printModelWithStandardUnitsAdded)
     EXPECT_EQ(e, printer->printModel(model));
 }
 
+TEST(Printer, printModelImportingModelParentComponent)
+{
+    auto parser = libcellml::Parser::create();
+    auto modelContents = fileContents("importingModelParentComponent.cellml");
+    auto model = parser->parseModel(modelContents);
+    EXPECT_EQ(size_t(0), parser->errorCount());
```
agarny, line 302: Any particular reason why it's `errorCount()` and not `issueCount()`?

tests/printer/printer.cpp
```diff
@@ -294,6 +294,40 @@ TEST(Printer, printModelWithStandardUnitsAdded)
...
+    auto serialisedModel = printer->printModel(model);
+
+    EXPECT_EQ(modelContents, serialisedModel);
+}
+
+TEST(Printer, printModelImportingModelChildComponent)
+{
+    auto parser = libcellml::Parser::create();
+    auto modelContents = fileContents("importingModelChildComponent.cellml");
+    auto model = parser->parseModel(modelContents);
```
agarny, line 318: I guess it's "fine" not to test that the parser didn't generate any errors...?

tests/resources/importedModelWithCustomUnits1.cellml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<model xmlns="http://www.cellml.org/cellml/2.0#" name="import_me">
+  <units name="myUnits">
+    <unit units="second" prefix="milli" />
+  </units>
+  <component name="importMe">
+    <variable name="x" units="myUnits" interface="public_and_private"/>
+  </component>
+</model>
```
agarny: This CellML file is not being used anymore. Delete it.

tests/resources/importedModelWithCustomUnits2.cellml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<model xmlns="http://www.cellml.org/cellml/2.0#" name="import_me">
+  <units name="myUnits2">
+    <unit units="metre" prefix="milli" />
+  </units>
+  <component name="importMe">
+    <variable name="x" units="myUnits2" interface="public_and_private"/>
+  </component>
+</model>
```
agarny: This CellML file is not being used anymore. Delete it.

agarny (changes requested):
Looks good apart from a few minor requests.

tests/printer/printer.cpp
kerimoyle: In general we should be using what we want to be checking.  I think there's even an open issue for exactly that.  As the Issue class becomes more widely used, we'll start to see information items stored in the logger, so it's better to check for errors when you mean to check for errors, and issues when you mean to check for anything.

tests/printer/printer.cpp
kerimoyle: We don't check everything everywhere every single time ... so yeah, I reckon it's fine.

tests/resources/importedModelWithCustomUnits2.cellml
kerimoyle: 47e287cd

tests/resources/importedModelWithCustomUnits1.cellml
kerimoyle: 47e287cd

agarny (approved):

hsorby (approved):

PR #593, opened 2020-05-07
Added function to return string for debugging variable connections.

kerimoyle:
As discussed yesterday, here's what I'd like to include (or something similar) for debugging connections.

### DIFF
--- a/src/api/libcellml/variable.h
+++ b/src/api/libcellml/variable.h
@@ -121,6 +121,21 @@ public:
     static bool addEquivalence(const VariablePtr &variable1, const VariablePtr &variable2, const std::string &mappingId, const std::string &connectionId = "");
 
     /**
+     * @brief Return a vector of VariablePtrs within the equivalent variable set of the @p variable.
+     *
+     * @param variable The variable to track equivalence from.
+     */
+    static std::string traceEquivalentVariableSet(const VariablePtr &variable);
+
+    /**
+     * @brief Recursive function to trace and return the set of variables equivalent to @p variable.
+     *
+     * @param variable The variable to trace.
+     * @param variableList A vector of variables within the equivalent set.
+     */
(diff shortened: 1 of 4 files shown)


agarny (commented):
This looks far too much like debugging methods and therefore has nothing to do in the API. First, I would rename your `traceEquivalentVariableSet()` to something like `equivalentVariableSet()` and I would have it return a proper set not a string. I mean, what can we do with the string that is currently returned? Nothing, just "view" it for... debugging purposes indeed.

As for `listEquivalentVariables()` it is only used internally, so definitely has nothing to do in the API.

kerimoyle:
Yeah, I'm not sold on the implementation either!  It's just to demonstrate what I want to be able to do.  A designy question then ... how/what kind of tools can we supply through the API that will let users debug their models (ie: not their *code*, their *models*)?  I would have thought that something like this would sit within the Checker/Debugger/WhateverWeCallIt?

agarny:
I think it's fine to have a method like `equivalentVariableSet()` in the `Variable` class, and have it return a set of equivalent variables. People may want to retrieve that set for whatever reason, including debugging purposes.

kerimoyle:
I started with that, yeah.  But then I wondered whether returning a vector of those variables went against what @hsorby was saying about using an index-based system instead?  So we'd have to store that total list and counts, and then provide functions to retrieve them one at a time?

agarny:
Something to check with him, but what is certain is that returning a string is a big no-no for me.

agarny:
FWIW, I also need to retrieve all the equivalent variables of a given variable for my work on PR #572. So, I am going to have something like what you have done, except that in my case I am going to return a vector of variables. Right now, it's going to be "internal" to the `Generator`, but ideally it would be part of the `Variable` class. It's just that I understand that returning a vector may not be the way forward (since we use a horrible and very unpractical index-based system).

kerimoyle:
Closing as not (really) needed now the `Analyser` class is on the way.  It should be built on to that instead.

Issue #597, opened 2020-05-12
Generator: items in the VariableInfoWithType array may not be in the correct order


agarny:
This was indirectly reported by @kerimoyle while using the Hodgkin-Huxley model. From the `VariableInfoWithType` array, we get the impression that the index of `i_Stim` is `8` while it actually is `5`.

PR #598, opened 2020-05-12
Demonstrating errors in code generation

nickerso:
I have added code generation for my "go to" test model (sine approximations with imports) to demonstrate an issue that I am seeing in the code generation. There is some evidence that this is related to #597, but as well as the ordering being incorrect, the `VARIABLE_INFO` is getting the name of the variables wrong. While the variable info correctly identifies the "source" component, it is using the name of the variable from the top-level component in the encapsulation and import hierarchy.

You can see these differences in specifically in https://github.com/nickerso/libcellml/commit/41db32b632b909bfea0fba5a04b114da841f8770. This is a problem in both the C and Python profiles.


agarny:
Good catch! FWIW, I would indeed expect the behaviour to be the same indepdent on the profile chosen. Anyway, it should be an easy fix. I will do it tomorrow once PRs #596 and #599 have been merged in.

nickerso:
> once PRs #596 and #599 have been merged in

Surely the fix can be done without a merge...this is completely blocking me at present...I'm building my own libcellml so not worried about changes being merged in...pretty please :)

agarny:
I think the problem (besides issue #597) is not with the generator itself, but rather with the flattening.

agarny:
> I think the problem (besides issue #597) is not with the generator itself, but rather with the flattening.

I take it back. It is indeed an issue with the generator. Argh!

nickerso:
> I take it back. It is indeed an issue with the generator.

phew! was pretty sure I checked the flattened model and it looked good. This issue came up as I try to navigate through the collection of equivalent variables to find the one that is used in the generated code and I can never get a match as the component and variable names are out of sync.

agarny:
> > I take it back. It is indeed an issue with the generator.
> 
> phew! was pretty sure I checked the flattened model and it looked good. This issue came up as I try to navigate through the collection of equivalent variables to find the one that is used in the generated code and I can never get a match as the component and variable names are out of sync.

Yes, I printed the flattened model and that's where I realised my mistake. :) The problem is buried in our generator, but I am slowly getting there, so you will have a fix before the end of the day. Definitely.

agarny:
Ok, I am done with this. Happy for you, @nickerso, @hsorby and @kerimoyle, to review and merge this in.

agarny:
@nickerso, since you can't review it as such. I am happy to "review" it on your behalf...

nickerso:
Mild panic as this didn't fix my issue, but then realised that I have the ordering fix from #597 also merged into my local source. So ignoring that, yes, this fixes the bug in the variable names and I would happily approve this if I hadn't started it :)

agarny:
Eh, it's not the time for mild panic! :) Anyway, I am going to approve it on your behalf and leave it to @hsorby or @kerimoyle to confirm.

agarny (approved):
This is a brilliant PR. I am all in favour of it. No reservations whatsoever! (If only all PRs could be of that quality. One can only dream...)

src/generator.cpp
```diff
@@ -136,22 +139,25 @@ struct GeneratorInternalVariable
...
-
-    if (!variable->initialValue().empty()) {
+    if (checkInitialValue && !variable->initialValue().empty()) {
         // The variable has an initial value, so it can either be a constant or
         // a state. By default, we consider it to be a constant and, if we find
         // an ODE for that variable, we will know that it was actually a state.
 
         mType = Type::CONSTANT;
+
+        mInitialValueVariable = variable;
```
kerimoyle, line 156: A variable that is reset will have an initial value too? Neither a constant nor a state?

kerimoyle (commented):
Which is the bit that fixes the bug? There are lots of changes that seem a bit arbitrary?

tests/resources/deriv_approx_sin.xml
```diff
@@ -3,7 +3,7 @@
 
   <component name="sin" id="sin">
     <variable id="x" units="dimensionless" name="x" interface="public_and_private"/>
-    <variable id="sin" units="dimensionless" name="sin" initial_value="sin_initial_value" interface="public_and_private"/>
+    <variable id="sin" units="dimensionless" name="sin" initial_value="0.0" interface="public_and_private"/>
```
kerimoyle, line 6: This is now testing something different ... even if we are phasing out the initialisation with variable thing, we should keep the tests?

src/generator.cpp
agarny: We are not yet generating code for resets, so we will see what needs to be done when we do.

agarny:
> Which is the bit that fixes the bug? There are lots of changes that seem a bit arbitrary?

This PR contains some code clean up (as a result of working on this PR), but otherwise there are a couple of key places that fix the bug. I am going to annotate them.

tests/resources/deriv_approx_sin.xml
agarny: This was @nickerso's change and I think it's fine since we are not currently handling the initialisation of a variable from another variable.

src/generator.cpp
```diff
@@ -490,18 +505,26 @@ bool GeneratorEquation::check(size_t &equationOrder, size_t &stateIndex,
...
-        variable->mComponent = mComponent;
+        for (size_t i = 0; i < mComponent->variableCount(); ++i) {
+            VariablePtr localVariable = mComponent->variable(i);
+
+            if (sameOrEquivalentVariable(variable->mVariable, localVariable)) {
+                variable->setVariable(localVariable, false);
+
+                break;
+            }
+        }
```
agarny, line 527: 1) This is the change that directly fixes @nickerso's issue. Before we were only updating the component information, but not the variable information which we now do through our call to `setVariable()`. Note, however, that we pass it `false` since we don't want to check for the initial value.

src/generator.cpp
```diff
@@ -136,22 +139,25 @@ struct GeneratorInternalVariable
...
         // an ODE for that variable, we will know that it was actually a state.
 
         mType = Type::CONSTANT;
+
+        mInitialValueVariable = variable;
     }
+
+    mVariable = variable;
+    mComponent = std::dynamic_pointer_cast<Component>(variable->parent());
 }
```
agarny, line 161: 2) Our new `setVariable()` method now also updates `mVariable`, but also keeps track of the variable object that contains the initial value of a variable (`mInitialValueVariable`).

src/api/libcellml/generator.h
```diff
@@ -78,8 +78,7 @@ class LIBCELLML_EXPORT GeneratorVariable
      * @c Component in which the @c Variable is first defined (in the case of
      * the variable of integration), initialised (in the case of a constant) or
      * actually computed (in the case of a state, computed constant or algebraic
-     * variable). It may or may not be the same @c Component as the parent
-     * component of the @c Variable.
+     * variable).
```
agarny, line 81: 3) Because of points 1 and 2, the `component()` is always going to be the parent component of `variable()`, which makes our API better than before.

src/generator.cpp
```diff
@@ -3300,7 +3300,7 @@ std::string Generator::GeneratorImpl::generateCode(const GeneratorEquationAstPtr
 
 std::string Generator::GeneratorImpl::generateInitializationCode(const GeneratorInternalVariablePtr &variable)
 {
-    return mProfile->indentString() + generateVariableNameCode(variable->mVariable) + " = " + generateDoubleCode(variable->mVariable->initialValue()) + mProfile->commandSeparatorString() + "\n";
+    return mProfile->indentString() + generateVariableNameCode(variable->mVariable) + " = " + generateDoubleCode(variable->mInitialValueVariable->initialValue()) + mProfile->commandSeparatorString() + "\n";
```
agarny, line 3303: 4) This is a side effect of the main change where we now have to retrieve the initial value from `mInitialValueVariable` rather than `mVariable`.

tests/resources/deriv_approx_sin.xml
kerimoyle: Yes we are ... that's still permitted, just to be noted in the tutorials and informative spec that it isn't recommended and will be removed soon.  If it's changed here, we need to check that we still have tests for initialisation from a variable until it's actually removed properly ...

kerimoyle:
I've got some concerns about the "can't initialise a variable with another variable" thing.  We've just finished writing the spec that says that we do support this (at present) ... shouldn't we be maintaining tests and functionality in line with that?  I have no problem with this PR other than this sneaking idea that we might be missing stuff by making changes in the background ... ? @nickerso @hsorby @agarny Thoughts?

nickerso:
> I've got some concerns about the "can't initialise a variable with another variable" thing

CellML allows this, the `develop` branch of libCellML does not yet support this for code generation. As code generation support for this improves this will be added back in. For this particular test I used the existing sine approximations model to save time, happy for the PR to be updated to use a different model to demonstrate this issue (which I think there were others that @agarny fixed as part of this so they could be sufficient).

tests/resources/deriv_approx_sin.xml
agarny: I only meant that the code generator doesn't currently support the initialisation of a variable using another variable. This is clearly something that we need to add to the code generator. I have just created an issue for it (see issue #602).

kerimoyle: Ah, ok, that makes more sense.  I thought we were removing functionality rather than just waiting for it :)

kerimoyle (approved):
OK with me.

agarny:
> I've got some concerns about the "can't initialise a variable with another variable" thing. We've just finished writing the spec that says that we do support this (at present) ... shouldn't we be maintaining tests and functionality in line with that?

This needs to be done in another PR for which I have created an issue (see issue #602). Will work on that issue once we have merged in this PR and PR #599.

Issue #594, opened 2020-05-10
Validator should not check units or interfaces for dummy variables


kerimoyle:
At the moment when you create dummy variables so you can connect to an imported component, the validator checks your dummy variables for equivalent units and interface types.  This check should be removed.

See discussion here: https://github.com/cellml/libcellml/pull/591#discussion_r421194441

nickerso:
yep, sounds good. If the imported component has been resolved the validation could proceed as normal but for unresolved imports the validator should not be checking any variables that might exist.

PR #605, opened 2020-05-27
i594 Validator should not check dummy variables in imported components

kerimoyle:
Addresses #594 

When you create a connection to a dummy variable prior to flattening a model with imports, the dummy variables are being checked for interface types and units.  They should be skipped ...

- [x] Test to show behaviour
- [x] Fix it all :)
- [x] Check coverage


kerimoyle:
Ready for review please.  @agarny @hsorby @nickerso .  Should be pretty quick and easy ... :)

kerimoyle:
Updated and ready for review please @agarny @nickerso @hsorby

src/validator.cpp
```diff
@@ -1060,19 +1060,26 @@ void Validator::ValidatorImpl::validateEquivalenceUnits(const ModelPtr &model, c
     std::string hints;
     for (size_t index = 0; index < variable->equivalentVariableCount(); ++index) {
         auto equivalentVariable = variable->equivalentVariable(index);
+        // If the parent component of the variable is imported, don't check it.
```
agarny: Replace with (or something similar):
```
// If the equivalent variable doesn't have a parent component or its parent component is imported then don't check it.
```

src/validator.cpp
```diff
@@ -1060,19 +1060,26 @@ void Validator::ValidatorImpl::validateEquivalenceUnits(const ModelPtr &model, c
...
+        auto equivalentComponent = std::dynamic_pointer_cast<Component>(equivalentVariable->parent());
```
agarny: Nice candidate for our forthcoming `owningComponent()`. :) (Same below.)

src/validator.cpp
```diff
@@ -1060,19 +1060,26 @@ void Validator::ValidatorImpl::validateEquivalenceUnits(const ModelPtr &model, c
...
+        if (equivalentComponent == nullptr) {
+            continue;
+        }
+        if (equivalentComponent->isImport()) {
+            continue;
+        }
```
agarny: I would replace this with:
```
if ((equivalentComponent == nullptr) || equivalentComponent->isImport()) {
    continue;
}
```

tests/validator/validator.cpp
```diff
@@ -383,6 +383,41 @@ TEST(Validator, importComponents)
...
+    dummyComponent->setSourceComponent(importer, "component_in_that_model");
+    auto dummyVariable = libcellml::Variable::create("I_dont_need_units_or_interface");
+    dummyComponent->addVariable(dummyVariable);
+    model->addComponent(dummyComponent);
+
+    validator->validateModel(model);
+    EXPECT_EQ(size_t(0), validator->issueCount());
+
+    // As soon as a connection is made between the dummyVariable and the variable, the validator
+    // checks the dummyVariable and fails the model.
```
agarny: I must be missing something: how is that failing the model? `dummyVariable` is imported so it won't be checked during validation, right? And, indeed, the validator doesn't have any issues. So, is the comment wrong here?

agarny (changes requested):
Looks good.

tests/validator/validator.cpp
kerimoyle: Leftover ... this is the behaviour that was fixed. 43c9e0a3

src/validator.cpp
kerimoyle: 43c9e0a3

src/validator.cpp
kerimoyle: 43c9e0a3

tests/validator/validator.cpp
```diff
@@ -383,6 +383,39 @@ TEST(Validator, importComponents)
...
+    auto dummyComponent = libcellml::Component::create("dummy_component");
+    dummyComponent->setSourceComponent(importer, "component_in_that_model");
+    auto dummyVariable = libcellml::Variable::create("I_dont_need_units_or_interface");
+    dummyComponent->addVariable(dummyVariable);
+    model->addComponent(dummyComponent);
+
+    validator->validateModel(model);
+    EXPECT_EQ(size_t(0), validator->issueCount());
+
+    libcellml::Variable::addEquivalence(variable, dummyVariable);
```
agarny, line 413: Might want to add a quick comment that although we add an equivalence, the equivalent variable still doesn't get checked because it's imported.

Otherwise, it might be good to add `dummyVariable` to `dummyComponent` after the call to `addEquivalence()`, so that we also test the case when the equivalent variable doesn't have a component parent?

kerimoyle: > Might want to add a quick comment that although we add an equivalence, the equivalent variable still doesn't get checked because it's imported.

It's in the comment a few lines higher?

> Otherwise, it might be good to add dummyVariable to dummyComponent after the call to addEquivalence(), so that we also test the case when the equivalent variable doesn't have a component parent?

That's a little out of scope for this PR?  It's nothing to do with importing ... and we already have `Validator, validateInvalidConnectionsVariableWithoutParentComponent` and `Connection, componentlessVariableInvalidConnection` for that purpose.

src/validator.cpp
```diff
@@ -1060,19 +1060,23 @@ void Validator::ValidatorImpl::validateEquivalenceUnits(const ModelPtr &model, c
...
         double multiplier = 0.0;
         if (!unitsAreEquivalent(model, variable, equivalentVariable, hints, multiplier)) {
             VariablePair reversePair = std::make_pair(equivalentVariable, variable);
             auto it = std::find(alreadyReported.begin(), alreadyReported.end(), reversePair);
             if (it == alreadyReported.end()) {
                 VariablePair pair = std::make_pair(variable, equivalentVariable);
-                ComponentPtr parent1 = std::dynamic_pointer_cast<Component>(variable->parent());
-                ComponentPtr parent2 = std::dynamic_pointer_cast<Component>(equivalentVariable->parent());
+                ComponentPtr parentComponent = std::dynamic_pointer_cast<Component>(variable->parent());
```
agarny: Should now use `owningComponent()` (same elsewhere).

kerimoyle: This will be true all over the place ... so should be checked as a separate issue.

agarny:
For some reasons, GitHub doesn't allow me to reply to https://github.com/cellml/libcellml/pull/605#discussion_r438460031, so I am doing it here:

This is the reason I have created issue #623. However, this doesn't undermine the fact that this PR should use `owningComponent()` since it's now part of the `develop` branch.

kerimoyle:
PRs should be specific, otherwise we're never going to get anywhere! (also the reason that you haven't updated all the error messages in the earlier PR? ;) )

agarny:
This PR adds new code. I would therefore expect that new code to use whatever is available in the `develop` branch. Simple as that. Anyway, will leave it to @nickerso and @hsorby to accept this PR, no problems.

kerimoyle:
@agarny That's fair enough - I've changed the new code now.  I don't think this PR should have to change all of them though (if anything, that should have been in the earlier one which created the function), and the others will be addressed in your #623

agarny:
> @agarny That's fair enough - I've changed the new code now. I don't think this PR should have to change all of them though (if anything, that should have been in the earlier one which created the function), and the others will be addressed in your #623

Agreed, only the new code of this PR should use the `owningComponent()` method. The rest of the code will be updated as part of issue #623. (Agreed too that I could have made the forthcoming changes for issue #623 when I introduced `owningComponent()`.)

src/validator.cpp
```diff
@@ -1111,6 +1115,10 @@ void Validator::ValidatorImpl::validateConnections(const ModelPtr &model) const
     }
 
     for (const VariablePtr &variable : variables) {
+        auto parentComponent = std::dynamic_pointer_cast<Component>(variable->parent());
```
agarny: `owningComponent()` should be used here too.

agarny (approved):

kerimoyle:
Ping @hsorby @nickerso

hsorby (approved):

Issue #623, opened 2020-06-10
Make use of `owningModel()` and `owningComponent()`


agarny:

nickerso:
Issues like this are not very helpful at communicating the reason for why something should be done. It would be good to know what this issue is actually trying to achieve?

agarny:
Fair point. Our codebase has two "new" utilities (`owningModel()` and `owningComponent()`), so this issue is about ensuring that our whole codebase makes use of it wherever possible. This will result in cleaner (no casting all over the place) and more consistent code.

PR #626, opened 2020-06-11
Make use of `owningModel()` and `owningComponent()`

agarny:
Addresses issue #623.

### DIFF
--- a/src/api/libcellml/units.h
+++ b/src/api/libcellml/units.h
@@ -30,6 +30,10 @@ limitations under the License.
 // information.)
 #undef PASCAL
 
+#ifndef SWIG
+template class LIBCELLML_EXPORT std::weak_ptr<libcellml::Units>;
+#endif
+
 namespace libcellml {
 
 /**
--- a/src/api/libcellml/units.h
+++ b/src/api/libcellml/units.h
@@ -37,6 +37,10 @@ namespace libcellml {
  * Class for Units.
  */
 class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
+#ifndef SWIG
+    ,
+                              public std::enable_shared_from_this<Units>
+#endif
 {
 public:
     ~Units() override; /**< Destructor */
@@ -423,7 +427,7 @@ public:
      * @return @c true when these @c Units rely on @c Units which are imported,
      * or @c false otherwise.
      */
-    bool requiresImports() const;
+    bool requiresImports();
 
     /**
      * @brief Return the scaling factor difference between two @c Units.
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -552,7 +552,7 @@ size_t getComponentIndexInComponentEntity(const ComponentEntityPtr &componentPar
 IndexStack reverseEngineerIndexStack(const VariablePtr &variable)
 {
     IndexStack indexStack;
-    ComponentPtr component = std::dynamic_pointer_cast<Component>(variable->parent());
+    ComponentPtr component = owningComponent(variable);
     indexStack.push_back(getVariableIndexInComponent(component, variable));
 
     ComponentEntityPtr parent = component;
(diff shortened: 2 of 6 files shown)


agarny:
@hsorby, @kerimoyle and @nickerso: this is a straightforward PR. Shouldn't take more than a minute to review.

kerimoyle (commented):
I wonder if we can make life easier for users too by putting this functionality somewhere they can access it?  At present the user still has to do the casting themselves ... eg in this test:

```
TEST(Units, parentOfUnits)
{
    libcellml::ModelPtr model = libcellml::Model::create();
    model->setName("I_am_a_model");
    libcellml::UnitsPtr u = libcellml::Units::create();
    u->setName("u");
    model->addUnits(u);

    // Would be great to be able to say:
    // auto parent = owningModel(u); instead of the line below ... 
    libcellml::ModelPtr parent = std::dynamic_pointer_cast<libcellml::Model>(u->parent());
    EXPECT_FALSE(parent == nullptr);
    EXPECT_EQ(model, parent);
}
```

agarny:
> I wonder if we can make life easier for users too by putting this functionality somewhere they can access it? At present the user still has to do the casting themselves ... eg in this test:

Yes, I thought the same when I came across that test. I guess this could be part of our super duper toolbox... :)

agarny:
Ping @hsorby, @kerimoyle and @nickerso.

kerimoyle:
Like I said earlier, I think that the `owningModel` and `owningComponent` functions should be made available to users too?  I wouldn't have thought that needed a separate PR?

agarny:
You are asking for a coupe of methods to be added to the API. This PR is about cleaning up the code. So, two different things. Also, I don't think those methods should be part of the API, but rather of our "toolbox".

src/api/libcellml/units.h
```diff
@@ -423,7 +431,7 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
      * @return @c true when these @c Units rely on @c Units which are imported,
      * or @c false otherwise.
      */
-    bool requiresImports() const;
```
hsorby, line 426: Did we have to drop the const here?

agarny: Yes, because `Units::requiresImports()` now calls `shared_from_this()`.

kerimoyle:
Hi again @agarny.  Could you add a catch for a `nullptr` entity argument please?  It will fix a bug I just found elsewhere :)

agarny:
In `owningModel()` I imagine?

kerimoyle:
Actually, in `owningComponent` ... there are spots where it takes an argument of `variable->parent()`, and if the parent is null then it segfaults.  Prob best to put it in both of them though?

agarny:
Ok, I am going to do that check in both and return `nullptr` in such a case.

agarny:
Ok, done in commit bb12545.

agarny:
Ok, I am going to have to revert commit bb12545. Indeed, `owningModel()` and `owningComponent()` are utilities, so we can directly test them with `nullptr` as a parameter value. The only way would be to test them indirectly, but our current code only makes "proper" use of of those utilities, so...

Bottom line: `owningModel()` and `owningComponent()` should only be used when you are 100% certain that the parameter is not `nullptr`. After all, those are utilities, not methods that are part of the API (if they were, then it would be possible to test them).

kerimoyle:
Do you mind if I change these functions in my PR then?  I found this while writing #630 so could get the coverage from undoing my checks there?

agarny:
Where in your PR do you call `owningModel()` and/or `owningComponent()` in such a way that it segfaults? From what I can tell, all your calls to those methods look find (i.e. the entity that is passed is not `nullptr`).

Otherwise, to change those utilities in your PR won't solve the issue of the coverage...

kerimoyle:
It (was) here: https://github.com/cellml/libcellml/pull/630/files#r440560834

kerimoyle (approved):

nickerso (approved):

PR #630, opened 2020-06-16
I624 validator check for duplicate ids

### LINKED ISSUE #624: Validator should check for duplicate entity ids
See https://cellml-specification.readthedocs.io/en/latest/reference/formal_and_informative/specA2.html#xml-id-attributes.  Id uniqueness isn't being checked in the validator at the moment.

### LINKED ISSUE #629: Bug or feature? Mapping id the same within a connected set?
Is this a bug, or is it expected?  Two individual mappings in one component connect to the same variable in another, and they then share the same mapping id?
v1 -> v2a  } These two are the same?
v1 -> v2b  }
v1 -> v3 ... but this one is different
Any thoughts @hsorby @agarny @nickerso ?

```
TEST(Connection, mappingId)
{
    const std::string in = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
                           "<model xmlns=\"http://www.cellml.org/cellml/2.0#\"  name=\"everything\" >\n"
                           "  <component name=\"c1\" >\n"
                           "    <variable name=\"v1\" units=\"units2\" interface=\"private\"/>\n"
                           "  </component>\n"
                           "  <component name=\"c2\">\n"
                           "    <variable name=\"v2a\" units=\"units2\" interface=\"public\"/>\n"
                           "    <variable name=\"v2b\" units=\"units2\" interface=\"public\"/>\n"
                           "  </component>\n"
                           "  <component name=\"c3\">\n"
                           "    <variable name=\"v3\" units=\"units2\" interface=\"public\"/>\n"
                           "  </component>\n"
                           "  <connection component_1=\"c1\" component_2=\"c2\">\n"
                           "    <map_variables variable_1=\"v1\" variable_2=\"v2a\" id=\"id12a\"/>\n"
                           "    <map_variables variable_1=\"v1\" variable_2=\"v2b\" id=\"id12b\"/>\n"
                           "  </connection>\n"
                           "  <connection component_1=\"c1\" component_2=\"c3\">\n"
                           "    <map_variables variable_1=\"v1\" variable_2=\"v3\" id=\"id13\"/>\n"
                           "  </connection>\n"

                           "</model>\n";
    auto model = parser->parseModel(in);
    EXPECT_EQ("id12a",
              libcellml::Variable::equivalenceMappingId(
                  model->component("c1")->variable("v1"),
                  model->component("c2")->variable("v2a")));  // <<< Fails because id is "id12b"
    EXPECT_EQ("id12b",
              libcellml::Variable::equivalenceMappingId(
                  model->component("c1")->variable("v1"),
                  model->component("c2")->variable("v2b"))); 
    EXPECT_EQ("id13",
              libcellml::Variable::equivalenceMappingId(
                  model->component("c1")->variable("v1"),
                  model->component("c3")->variable("v3")));
}
```

### LINKED ISSUE #635: id uniqueness of all mathml child elements ...
Further to #633 and #630, all child elements of math blocks could have ids, so need to be checked at validation time.

kerimoyle:
Addresses #624 and #635 

At present this checks the model scope for duplicate ids, because this is - in effect - no different from checking a parsed document.

Changes in this PR:
- ~added internal utility function to retrieve `id` attribute from `math` strings,~
- ~added functions to API to get id for `Component::mathId()`, `Reset::testValueMathId()`, `Reset::resetValueMathId()`,~
- added tests for new functions,
- added tests for IDs on all items, not just `Entity` items,
- added function to `Validator` to check for duplicate ids in the model and report them.
- added functionality to parse MathML strings in components and resets and validate uniqueness.
- changed minimum compiler to C++17 (sorry, I know it's in PR #636 but it's here too ...)


src/validator.cpp
```diff
@@ -1355,4 +1389,178 @@ void Validator::ValidatorImpl::checkUnitForCycles(const ModelPtr &model, const U
...
+        if (!item->id().empty()) {
+            info = "  - variable '" + item->name() + "' in component '" + component->name() + "'\n";
+            addIdMapItem(item->id(), info, idMap);
+        }
+        // Equivalent variables.
+        for (size_t e = 0; e < item->equivalentVariableCount(); ++e) {
+            auto equiv = item->equivalentVariable(e);
+            if (equiv != nullptr) {
+                auto equivParent = owningComponent(equiv);
+                if (equivParent != nullptr) {
```
kerimoyle: @agarny This is where I used them ... now there are checks, but I think it would be cleaner to be able to pass in a nullptr to owningComponent and owningModel and have it return a null rather than a segfault?

agarny: Sure, but that's besides the point. `owningComponent()` can't be tested with `nullptr` as a parameter. Our tests only have access to the API, not our utilities.

kerimoyle: I think we don't understand each other.  I would rather not have these two nested if test in my PR version - I'd rather be able to call owningComponent() safely.  This means that if I change the owningComponent function in my PR, and also change my code back to how I wanted it, it will be called during testing, cos that's how I found the segfault in the first place.  It doesn't appear in your PR because your PR doesn't have these functions to check ID.  So coverage will be fine ... ?

agarny: Oh, ok, got it. Sure, if your code can somehow result in `owningModel()` and `owningComponent()` to be called with `nullptr` as a parameter, then by all means update it in your PR.

(Phew, we are getting there!  Too much noise around me at the office. Am not used to it anymore... and don't want to get used to it again!)

kerimoyle:
Ready for review please @agarny @hsorby @nickerso .

kerimoyle:
I've added functionality to parse the MathML string for id attributes anywhere, and check them against the rest of the model.  The wording for errors uses "maths block at index 0" etc as I'd assumed we were splitting up the string for separate retrieval (as in #633).  If we're sure that's not going to make it into the API then I'll change the wording to "first", "second" etc ? @hsorby @agarny @nickerso

src/api/libcellml/component.h
```diff
@@ -122,6 +122,13 @@ class LIBCELLML_EXPORT Component: public ComponentEntity, public ImportedEntity
      */
     void removeMath();
 
+    /**
+     * @brief Return the id attribute from the math string, if present.
+     *
+     * @return id The @c std::string if the id attribute is present, or an empty string if not.
+     */
+    std::string mathId() const;
```
nickerso: This should be removed as it is not clear which ID would be returned.

src/validator.cpp
```diff
@@ -291,6 +293,50 @@ struct Validator::ValidatorImpl
     void checkUnitForCycles(const ModelPtr &model, const UnitsPtr &parent,
                             std::vector<std::string> &history,
                             std::vector<std::vector<std::string>> &errorList);
+
+    /** @brief Function to check ID fields set within the model scope for duplicates.
+     *
+     * @param model The model to be checked.
+     */
+    void checkForDuplicateIds(const ModelPtr &model);
```
nickerso: for me, this should be more more like `checkUniqueIds`

src/validator.cpp
```diff
@@ -1355,4 +1404,231 @@ void Validator::ValidatorImpl::checkUnitForCycles(const ModelPtr &model, const U
...
+        if (!item->id().empty()) {
+            info = "  - reset at index " + std::to_string(i) + " in component '" + component->name() + "'\n";
+            addIdMapItem(item->id(), info, idMap);
+        }
+        if (!item->testValueId().empty()) {
+            info = "  - test_value in reset at index " + std::to_string(i) + " in component '" + component->name() + "'\n";
+            addIdMapItem(item->testValueId(), info, idMap);
+        }
+        if (!item->testValueMathId().empty()) {
+            info = "  - MathML child of test_value in reset at index " + std::to_string(i) + " in component '" + component->name() + "'\n";
```
nickerso: so this only checks the `math` element child for an id, not the full MathML string?

kerimoyle: Ah, good point!  I'd forgotten about the resets ... will do that now.

src/api/libcellml/component.h
kerimoyle: d4b98796

src/validator.cpp
kerimoyle: aa2e8d22

src/validator.cpp
kerimoyle: aa2e8d22

kerimoyle:
Ready for re-review please @agarny @hsorby @nickerso

agarny:
I just had a quick look at the code and, like @nickerso (from a comment on Slack, I believe), it looks overcontrived to me when it could be short and sweet using [`XPath`](https://en.wikipedia.org/wiki/XPath). I have never used `XPath`, but [`libxml2`](http://xmlsoft.org/) has an [`XPath` module](http://xmlsoft.org/html/libxml-xpath.html) and it comes with a quick [tutorial](http://xmlsoft.org/tutorial/ar01s05.html). So, it might be worth looking into it.

kerimoyle:
There are two parts to the new code: stuff to check the ids on items which are accessible through libCellML, and stuff to check the XML in the MathML strings.  
1) The iteration through the former means that we can return useful messages to the user.  This is to avoid the "duplicate found but I'm not telling what or where" unhelpful messages which come from libxml2 and its checks.
2) The iteration in the latter uses exactly the same logic as the parser and validator already use to read and manipulate MathML strings - I just followed what we already do.

If you can narrow down which part you're objecting to that might be helpful for me to figure out how to do it better? As it is, the choice I'm seeing is either getting awful error messages back, or redoing the current XML stuff throughout the rest of libCellML too?

agarny:
Based on your current approach, your code looks fine.

However, looking at your `duplicateIdAll` test, I can for instance see that `id1` is duplicated. Fine. Then, I can see several "sub" messages about where `id1` is duplicated. To be honest, to read all those messages and then find where in the XML code `id1` is duplicated (based on the message information) would take me much longer than search for `id1` in the whole XML file. Then, I would have to do the same for `id2`, `id3`, etc.

So, personally, I would just let the user know that `id1`, `id2`, `id3`, etc. are duplicated and that's all.

kerimoyle:
That's fine for users which have an interface (like OpenCOR) which mean that the items can be edited and searched without directly reading the XML.  But I thought that the whole point of libcellml was that users didn't need to look at the raw code - hence adding the message about which objects needed to be fixed.  
I realise that MathML is a little different in that we still treat it as a string, so users will have to create/edit that string somehow, but even that sounds like it will change in the near future anyway.  The shorter error messages (the ones you like) come from duplications inside the same MathML block (and from the validateMath function earlier in the validator), rather than across the entire document (which is what the new function adds).

kerimoyle:
ping @hsorby @agarny @nickerso

agarny (dismissed):
Did a bit of cleaning up and also slightly reformatted the reporting of duplicated ids, but apart from that it looks good to me.

agarny (dismissed):

src/validator.cpp
```diff
@@ -1376,4 +1425,237 @@ void Validator::ValidatorImpl::checkUnitForCycles(const ModelPtr &model, const U
...
+{
+    std::vector<XmlDocPtr> docs = multiRootXml(input);
+
+    for (const auto &doc : docs) {
+        XmlNodePtr node = doc->rootNode();
+        if (node == nullptr) {
+            return;
+        }
+        if (!node->isMathmlElement("math")) {
+            return;
```
nickerso: why is this return instead of ignoring non-math root nodes?

agarny (dismissed):

nickerso (changes requested):
just the one little comment to address I think...

src/validator.cpp
kerimoyle: 7f063dfa

nickerso (dismissed):

nickerso (dismissed):

agarny (approved):

nickerso (approved):

PR #632, opened 2020-06-16
I629 Fixed parser bug for reading mapping ids

### LINKED ISSUE #629: Bug or feature? Mapping id the same within a connected set?

kerimoyle:
Closes #629 

The `Parser` was not saving the id of parsed mappings, but rather using the last encountered `mappingId` for all mappings in a connection.  This is fixed here.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -376,7 +376,7 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
     }
 
     if (!encapsulationNodes.empty()) {
-        loadEncapsulation(model, encapsulationNodes[0]);
+        loadEncapsulation(model, encapsulationNodes.at(0));
         if (encapsulationNodes.size() > 1) {
             IssuePtr issue = Issue::create();
             issue->setDescription("Model '" + model->name() + "' has more than one encapsulation element.");
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -726,9 +726,10 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr 
     // Define types for variable and component pairs, and their ids.
     using NameInfo = std::vector<std::string>;
     using NameInfoMap = std::vector<NameInfo>;
+    using NamePair = std::pair<std::string, std::string>;
 
     // Initialise name pairs and flags.
-    NameInfo componentNameInfo;
+    NamePair componentNamePair;
     NameInfo variableNameInfo;
     NameInfoMap variableNameMap;
     bool mapVariablesFound = false;
@@ -778,7 +779,7 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr 
         mParser->addIssue(issue);
         component2Missing = true;
     }
-    componentNameInfo = {component1Name, component2Name};
+    componentNamePair = std::make_pair(component1Name, component2Name);
 
     XmlNodePtr childNode = node->firstChild();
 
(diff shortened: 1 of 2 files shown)


src/parser.cpp
```diff
@@ -824,6 +824,7 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr
             std::string variable1Name;
             std::string variable2Name;
             XmlAttributePtr childAttribute = childNode->firstAttribute();
+            mappingId = "";
```
agarny: Might want to replace with `mappingId.clear()`.

src/parser.cpp
```diff
@@ -891,24 +892,24 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr
     ComponentPtr component1 = nullptr;
     ComponentPtr component2 = nullptr;
     // Now check the objects exist in the model. TODO Remove as is validation?
-    if (model->containsComponent(componentNamePair.first)) {
-        component1 = model->component(componentNamePair.first);
+    if (model->containsComponent(componentNamePair.at(0))) {
```
agarny: I would replace `.at(0)` with `[0]`. (Same elsewhere.)

Indeed, `.at()` is only really needed if you provide an index value that may be out of rance (in which it will raise an exception). Here, you know that the index value is fine, so no need for the "extra work" that `.at()` does.

agarny (changes requested):
Looks good pending some minor changes.

src/parser.cpp
kerimoyle: 052f1676

src/parser.cpp
kerimoyle: 052f1676

src/parser.cpp
```diff
@@ -376,7 +376,7 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
     }
 
     if (!encapsulationNodes.empty()) {
-        loadEncapsulation(model, encapsulationNodes[0]);
+        loadEncapsulation(model, encapsulationNodes.at(0));
```
agarny: Why?

kerimoyle: Not related to this PR, was accidental when searching for `.at`

agarny: Ok, fair enough, although it didn't harm changing it (since the call to `.at()` doesn't account for a potential exception). Ok, added this to issue #402.

kerimoyle: I know ... but I want this to go through as fast as poss, so I figured it was not the time to address every little thing I found along the way! Glad we have an issue for general stuff, I'll add some things there too.

agarny (dismissed):

src/parser.cpp
```diff
@@ -724,12 +724,12 @@ void Parser::ParserImpl::loadVariable(const VariablePtr &variable, const XmlNode
 void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr &node) const
 {
     // Define types for variable and component pairs.
-    using NamePair = std::pair<std::string, std::string>;
+    using NamePair = std::vector<std::string>;
```
hsorby: This typedef doesn't make sense or give the correct information to what it is typedefing.

hsorby (changes requested):
I don't understand why the component pair was changed from being a pair. It makes good sense that they are a pair because that is what they are not a vector.

src/parser.cpp
kerimoyle: > I don't understand why the component pair was changed from being a pair. It makes good sense that they are a pair because that is what they are not a vector.

It was to reuse the same typedef for the mapping. I can split it up and have four kinds if you'd rather?  Otherwise I need a type for the connection information (with 2 items) and a separate one for the mapping information (with 3 items), and then two separate collections for each of them ... it seemed better this way?

agarny: Agreed about the naming (I completely missed that). If anything, it should be called `NameInfo`.

kerimoyle: Compromised in 27563871

kerimoyle:
> I don't understand why the component pair was changed from being a pair. It makes good sense that they are a pair because that is what they are not a vector.

@hsorby It was to reuse the same typedef for the mapping. I need a type for the connection information (with 2 items) and a separate one for the mapping information (with 3 items), and then two separate collections for each of them ... it just seemed better this way.  I've changed the names in the latest commit so it's clearer, but can split them up into 4 types if you'd rather.

kerimoyle:
Reverted to three types in 75b52c70.  Re-re-re-review please @hsorby @agarny ...

hsorby (approved):
That looks much better to me.

agarny (approved):

Issue #633, opened 2020-06-17
Add functionality to return individual <math> from concatenated <math> blocks, and individual ids


kerimoyle:
The spec says that multiple blocks with `<math>` tags are permitted inside components.  This means that they could all have their own `id` attributes.  We need to be able to retrieve these for validation, and it would make sense to also be able to access each of the blocks individually too.  The status quo API calls should be maintained so that the behaviour doesn't change: ie, `component->math()` will return the entire math string as usual.

nickerso:
Its not clear to me why the `math` element is being treated differently to the rest of the MathML string.  And the API suggested here starts to get us away from always treating MathML as a string in the libCellML API. If we want to go down that road, I'd be much more in favour of moving to exposing our AST from the generator (which I suspect may already be proposed in another issue).

kerimoyle:
Yep, it was written before I knew that ids were possible on the child attributes ... (see #635)

nickerso:
Another one that I think can be closed given that all MathML elements need to be checked, not just the `math` elements.

PR #637, opened 2020-06-17
I633 separate maths

kerimoyle:
Addresses #633 

- [x] `component->math()` functionality retained;
- [x] `component->mathCount()` returns number of individual `<math>` blocks;
- [x] `component->math(index)` returns (index+1)'th math string;
- [x] `component->mathId(index)` returns id attribute of (index+1)'th math string.
- [ ] Fix final math string being stripped of trailing whitespace?


nickerso:
This doesn't look like useful API for anything other than getting the `id` for one of many potential MathML elements. Not at all clear why this should be in the public API, or needed at all.

kerimoyle:
Agreed ... kind of ... We need to get ids out of the maths strings for validation, so this is the compromise between doing a little simple string manipulation to achieve that, and going for the all-out non-string treatment of all the child elements too.  It just seemed logical to expose the intermediate methods like `math(index)` along the way?

nickerso:
The strings are XML, we're building a library based on XML - why reinvent the wheel? There are standard ways to parse and query the XML, even if you just want to pull the `id` attributes out of the entire math string. Splitting on the top level `math` elements doesn't help in any way for checking IDs in the math...

Issue #614, opened 2020-06-03
Annotation tool development for libCellML


kerimoyle:
One of the user types for libCellML is an "annotation tool developer" who needs to be able to:

- Discover annotatable entities in a CellML model;
- Obtain existing IDs for entities in a model;
- Set IDs for things that don’t have one; and
- For a given ID, find out what kind of entity it is for.

PR #639, opened 2020-06-22
Print model overload

kerimoyle:
Start to address #614 by overloading the `Printer::printModel` function with optional arguments to generate automatic `id` attributes both with and without MathML ids.

- [x] tests for behaviour without maths ids
- [x] tests for behaviour with maths ids
- [x] overload on `printModel` with options to automatically generate ids for objects
- ~generate ids for MathML and child elements~
- [x] python working correctly


kerimoyle:
This PR build is failing because I need advice on how to handle setting id attributes within the MathML strings, before the library will behave as the tests expect.  At present we don't deserialise the MathML string, so that means at printing time it's simply printing (albeit a little prettified).  So.  **What's the best way to set ids on child elements of the MathML string?**

1) Deserialise the MathML string into XML, add id attributes (currently forbidden in our implementation because of our restrictions on the `setAttribute` function), or
2) Interact with the MathML string as a string, and add the ids at print time, or
3) Something else?

kerimoyle:

src/utilities.h
```diff
@@ -555,4 +555,32 @@ static inline std::string trimCopy(std::string s)
...
+ * @param mathIds A @c bool flag indicating whether or not to include MathML elements
+ * and their child elements.
+ * @param idList @c IdList collection of found ids.
+ */
+void listComponentIds(const ComponentPtr &component, bool mathIds, IdList &idList);
+
+/**
+ * @brief Creates an id string for a "type" object, unique in the context of @p idList.
+ *
+ * The id format is {type}_{n} where n is an integer, and type is the given string.
```
nickerso: I guess this is the old description :)

src/debug.cpp
```diff
@@ -19,6 +19,8 @@ limitations under the License.
 #include "libcellml/generator.h"
 #include "libcellml/variable.h"
 
+#undef NAN
```
nickerso, line 22: is this needed?

src/utilities.cpp
```diff
@@ -556,4 +559,132 @@ void findAllVariablesWithEquivalences(const ComponentPtr &component, VariablePtr
...
+    // Collect all existing ids in a list and return. NB can't use a map or a set as we need to be able to print
+    // invalid models (with duplicated ids) too.
+
+    std::unordered_set<std::string> idList;
+    // Model.
+    std::string id = model->id();
+    if (!id.empty()) {
+        idList.insert(id);
+    }
+    // Units.
```
nickerso: doesn't there need to be something to pull IDs out of imports for imported units?

nickerso (commented):
A couple of minor comments and one that I think needs to be addressed - or at least an explanation of what I am missing (re existing IDs on imports for units).

src/debug.cpp
agarny: Yes, one of the system headers defines `NAN` as a macro while we use it in one a class enum, so this causes conflicts. This has been discussed somewhere, but can't find the thread in less than a minute, so giving up. :)

src/utilities.cpp
kerimoyle: You tell me :) From our earlier conversations I'd thought that IDs on imports were discarded? See the last "see more" block here: https://cellml-specification.readthedocs.io/en/latest/reference/formal_and_informative/specA2.html
Is the comment on the arrow in the example wrong? If so, I can change it there and add storage for imported ids here?

src/utilities.h
kerimoyle: Oops, yes! Fixed in 86515095

src/utilities.cpp
kerimoyle: And a followup question - if IDs **are** imported, and the requirement for their uniqueness is over the document, not the model, does the flattener check for this when merging document scopes? I need to check ...

kerimoyle: Ignore me ... I get it now :)

kerimoyle: Fixed in 45a1c42b

src/printer.cpp
```diff
@@ -43,27 +43,29 @@ namespace libcellml {
...
-    std::string printEncapsulation(const ComponentPtr &component) const;
-    std::string printVariable(const VariablePtr &variable) const;
-    std::string printReset(const ResetPtr &reset) const;
-    std::string printResetChild(const std::string &childLabel, const std::string &childId, const std::string &math) const;
+    std::string printUnits(const UnitsPtr &units, IdList &idList, bool autoIds);
+    std::string printComponent(const ComponentPtr &component, IdList &idList, bool autoIds);
+    std::string printEncapsulation(const ComponentPtr &component, IdList &idList, bool autoIds);
+    std::string printVariable(const VariablePtr &variable, IdList &idList, bool autoIds);
+    std::string printReset(const ResetPtr &reset, IdList &idList, bool autoIds);
+    std::string printResetChild(const std::string &childLabel, const std::string &childId, const std::string &math, IdList &idList, bool autoIds);
```
hsorby, line 51: Why have these lost there constness?

src/printer.cpp
```diff
@@ -172,17 +176,21 @@ void buildMaps(const ComponentEntityPtr &componentEntity, ComponentMap &componen
     }
 }
 
-std::string Printer::PrinterImpl::printUnits(const UnitsPtr &units) const
+std::string Printer::PrinterImpl::printUnits(const UnitsPtr &units, IdList &idList, const bool autoIds)
```
hsorby: Don't need `const bool` to basic a type to bother with constness.

src/printer.cpp
kerimoyle: Ho hum.  The new argument (`idList`) changes, but it's not part of the class.  I guess it should be const then.  But do you want it on the bools too? I'm confused by your later comment @hsorby ?

kerimoyle: Can't change the idList, sorry.  I've removed the const bools though. 23a39d46

src/printer.cpp
kerimoyle: 23a39d46

src/utilities.h
```diff
@@ -555,4 +555,30 @@ static inline std::string trimCopy(std::string s)
     return s;
 }
 
+/**
+ * @brief Collect all existing id attributes within the given model.
+ *
+ * @param model The @c ModelPtr to interrogate.
+ * and their child elements.
```
nickerso: (forgetting my doxygen) but shouldn't this comment be part of the brief doxy string? or is that `.` the line above misplaced?

tests/model/model.cpp
```diff
@@ -784,11 +784,11 @@ TEST(Model, removeComponentInsensitiveToOrder)
 
     // If the order of operations is switched the behaviour is the same:
 
-    // Get a pointer to the second componet in the parsed model
+    // Get a pointer to the second componet in the parsed model.
```
nickerso: is this the same `componet` fixed in #640? ;-)

nickerso (dismissed):
Looks good to me. Would be great if @hsorby can check the changes due to his earlier review comments...

src/utilities.h
kerimoyle: The . is out of line, thanks.

tests/model/model.cpp
kerimoyle: Totally.  Gah.

src/utilities.h
kerimoyle: 37a7d869

tests/model/model.cpp
kerimoyle: fca4fc90

nickerso (dismissed):

tests/bindings/python/test_printer.py
```diff
@@ -17,7 +17,7 @@ def test_print_model(self):
 
         # std::string printModel(ModelPtr model)
         p = Printer()
-        self.assertIsInstance(p.printModel(Model()), str)
+        self.assertIsInstance(p.printModel(Model(), False), str)
```
agarny: The second parameter (`False`) is optional in C++ and it should also be in Python.

src/utilities.h
```diff
@@ -555,4 +555,30 @@ static inline std::string trimCopy(std::string s)
...
+ */
+IdList listIds(const ModelPtr &model);
+
+/**
+ * @brief Utility class to iterate through components to collect ids.
+ *
+ * @param component The @c ComponentPtr to interrogate.
+ * @param idList @c IdList collection of found ids.
+ */
+void listComponentIds(const ComponentPtr &component, IdList &idList);
```
agarny: Should be removed since it's only being used in `utilities.cpp`.

tests/bindings/python/test_printer.py
kerimoyle: cca111e7

src/utilities.cpp
```diff
@@ -556,4 +559,138 @@ void findAllVariablesWithEquivalences(const ComponentPtr &component, VariablePtr
...
+    // Because the counter starts high enough that it will always have a letter as the first character,
+    // we don't need to prefix it with any other string to be valid.
+    int counter = 0xb4da55;
+    std::stringstream stream;
+    stream << std::hex << counter;
+    std::string id = stream.str();
+    stream.str(std::string());
+
+    while (idList.count(id) != 0) {
+        counter++;
```
agarny: `++counter;` :)

src/utilities.cpp
```diff
@@ -556,4 +559,138 @@ void findAllVariablesWithEquivalences(const ComponentPtr &component, VariablePtr
...
+        if (!id.empty()) {
+            idList.insert(id);
+        }
+        id = component->reset(r)->resetValueId();
+        if (!id.empty()) {
+            idList.insert(id);
+        }
+    }
+
+    // TODO MathML ids on component math blocks and reset blocks are not yet included.
```
agarny: This should be removed. If we want and need to do this in the future then we should create an issue for it.

src/utilities.h
kerimoyle: 5b41e7f4

src/utilities.cpp
```diff
@@ -556,4 +559,138 @@ void findAllVariablesWithEquivalences(const ComponentPtr &component, VariablePtr
...
+
+    for (size_t c = 0; c < component->componentCount(); ++c) {
+        listComponentIds(component->component(c), idList);
+    }
+}
+
+std::string makeUniqueId(IdList &idList)
+{
+    // Because the counter starts high enough that it will always have a letter as the first character,
+    // we don't need to prefix it with any other string to be valid.
```
agarny: May want to make it clear that the counter is converted to an hexdecimal string. I read that comment several times and couldn't understand until I saw `std::hex` below.

tests/printer/printer.cpp
```diff
@@ -295,3 +293,395 @@ TEST(Printer, printModelWithStandardUnitsAdded)
...
+                                          "    </component_ref>\n"
+                                          "  </encapsulation>\n"
+                                          "</model>\n";
+
+    auto parser = libcellml::Parser::create();
+    auto printer = libcellml::Printer::create();
+
+    auto model = parser->parseModel(in);
+    bool includeAutomaticIds = true;
+    auto out = printer->printModel(model, includeAutomaticIds);
```
agarny, line 421: Replace with:
```
auto out = printer->printModel(model, true);
```

tests/printer/printer.cpp
```diff
@@ -295,3 +293,395 @@ TEST(Printer, printModelWithStandardUnitsAdded)
...
```
agarny, line 549: Replace with:
```
auto out = printer->printModel(model, true);
```

agarny (changes requested):
Just some minor comments, otherwise looks good.

tests/printer/printer.cpp
kerimoyle: I'd rather not - it makes it clearer what the "true" is actually doing when it's a variable with a name.

tests/printer/printer.cpp
kerimoyle: I'd rather not - it makes it clearer what the "true" is actually doing when it's a variable with a name.

src/utilities.cpp
kerimoyle: There's already an issue for it I think, but I think it's worth noting here.  I'll compromise and remove the "TODO" :)

src/utilities.cpp
kerimoyle: 5b41e7f4

src/utilities.cpp
kerimoyle: 5b41e7f4

src/utilities.cpp
kerimoyle: 5b41e7f4

agarny (approved):

kerimoyle:
Wahoo ... one down, one to go ... who's gonna bite ... @nickerso @hsorby ?

nickerso (approved):

PR #640, opened 2020-06-24
Create Annotator class

### LINKED ISSUE #614: Annotation tool development for libCellML

kerimoyle:
This is a work in progress to create tools that would support developers of software to annotate CellML models #614 

- [x] retrieve item by id from model (type = std::any),
- [x] retrieve item-of-type from model,
- [x] retrieve item type by id from model,
- [x] set ids on everything (addressed separately in #639 ).
- ~[ ] fix ambiguity around returning connection items (should be specified by component pair not variable pair?)~


kerimoyle:
Failing on the Windows build? What do I do about it @hsorby?

agarny:
> Failing on the Windows build? What do I do about it @hsorby?

It looks like we are using MSVC 2015 and `<any>` is only available in MSVC 2017 (see [here](https://devblogs.microsoft.com/cppblog/c17-features-in-vs-2017-3/)).

tests/annotator/annotator.cpp
```diff
@@ -182,3 +182,446 @@ TEST(Annotator, errorHandling)
...
+        "component_ref_2",
+        "encapsulation_1",
+        "import_1",
+        "import_2",
+        "doesnt_exist_will_cause_an_issue",
+        "map_variables_1",
+        "map_variables_2",
+        "connection_1",
+    };
+    std::vector<std::string> expectedInstructions = {
```
kerimoyle: Hey @nickerso Would these work for your instructions?

kerimoyle:
Work in progress exploring different ways for users to interact with the `Annotator` class:
- retrieving an enum type and `std::any` item; user must cast item manually;
- retrieve an enum type and set of indices; user can retrieve from the index chain; 
```
Annotator::Type::COMPONENT, {1, 2, 3}  <-->  model->component(1)->component(2)->component(3)
Annotator::Type::VARIABLE, {1, 2, 3} <--> model->component(1)->component(2)->variable(3)
... etc ...
```
- retrieve a string containing instructions for how to reach the item.

tests/annotator/annotator.cpp
kerimoyle: Removing after discussion last week.

kerimoyle:
Ready for review please @agarny @hsorby @nickerso All changes made as discussed yesterday.

src/annotator.cpp
```diff
@@ -0,0 +1,361 @@
...
+AnnotatorPtr Annotator::create() noexcept
+{
+    return std::shared_ptr<Annotator> {new Annotator {}};
+}
+
+void listComponentIdsAndItems(const ComponentPtr &component, ItemList &idList)
+{
+    std::string id = component->id();
+    if (!id.empty()) {
+        idList.insert(std::make_pair(id, std::make_pair(0, component)));
```
nickerso: probably missing something, but shouldn't that `0` be `Annotator::Type::COMPONENT` ?

src/annotator.cpp
```diff
@@ -0,0 +1,361 @@
...
+    AnyItem item;
+    if (!mPimpl->isBuilt) {
+        auto issue = libcellml::Issue::create();
+        issue->setDescription("Please call the Annotator::build function before attempting to access items by their id.");
+        issue->setLevel(libcellml::Issue::Level::ERROR);
+        addIssue(issue);
+        item = std::make_pair(Annotator::Type::ISSUE, issue);
+        return item;
+    }
+    if (mPimpl->mIdList.empty()) {
```
nickerso: Not sure this check is needed. If you can't find the requested `id`, it doesn't really matter whether there are no IDs or not. At the very least, I don't see how this is an error, if anything a hint or maybe a warning...but would just remove this block.

src/annotator.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+    if (it == mPimpl->mIdList.end()) {
+        auto issue = libcellml::Issue::create();
+        issue->setDescription("Could not find an item with id='" + id + "' in the model.");
+        issue->setLevel(libcellml::Issue::Level::WARNING);
+        addIssue(issue);
+        item = std::make_pair(Annotator::Type::ISSUE, issue);
+        // Adding the requested id string and this issue to the map, so that future requests (ie: to the .second)
+        // will return the same IssuePtr rather than creating a new one.
+        mPimpl->mIdList.insert(std::make_pair(id, item));
```
nickerso: no, really don't like this approach. Essentially adding a new ID to the model in a not very nice way. But also looking at this, it means that this method will always return an `item` and then the user needs to check if that item is of type `ISSUE` to see if it was successful. That seems very wrong to me...but maybe I'm missing something about those any items...

src/annotator.cpp
```diff
@@ -0,0 +1,361 @@
...
+        return item;
+    }
+    return mPimpl->mIdList[id];
```
nickerso: In most (all?) methods we have where the user requests an object by "name", you get a `nullptr` if that name doesn't exist or an object if it does. Can we not do that here?

src/annotator.cpp
```diff
@@ -0,0 +1,361 @@
...
+}
+
+ComponentPtr Annotator::component(const std::string &id)
```
nickerso, line 527: I can see these methods all look like they should work, but would be keen to know if the try/catch approach is what we should be using rather that testing the item type. @hsorby @agarny

src/api/libcellml/annotator.h
```diff
@@ -0,0 +1,289 @@
...
+     * @return A smart pointer to an @c Annotator object.
+     */
+    static AnnotatorPtr create() noexcept;
+
+    /**
+     * @brief Builds the internal map of ids to items for the model provided.
+     *
+     * @param model A @c ModelPtr model to build map for.
+     */
+    void build(const ModelPtr &model);
```
nickerso: can we leverage the `build` method to add IDs to any objects that don't have them? Seems like that capability should be somewhere in the annotator class rather than (as well as?) the current addition of IDs in the printer.

src/api/libcellml/annotator.h
```diff
@@ -0,0 +1,289 @@
...
+
+    /**
+     * @brief Retrieves an item with the given id string, in the model
+     * for which the internal map was built using @sa build().
+     *
+     * The item returned is a @c std::pair containing:
+     *  - an @c Annotator::Type enum, and
+     *  - an @c std::any item containing the item.
+     *
+     * Possible string labels and their corresponding items returned are:
```
nickerso: type enums rather than string labels, right?

src/api/libcellml/annotator.h
```diff
@@ -0,0 +1,289 @@
...
+     *              .second: A size_t with the index to the Unit item.
+     *          Retrieve the unit description using Units::unitAttributes() function with
+     *   Annotator::Type::UNITS
+     *          A UnitsPtr item with this id.
+     *   Annotator::Type::VARIABLE
+     *          A VariablePtr item with this id.
+          *
+     * @param id A @c std::string representing the @p id to retrieve.
+     *
+     * @return An @c AnyItem item as described above.
```
nickerso: Not at all endorsing the "return an issue type AnyItem on error" - but if that remains it should be clear in the docs here that is what would happen.

src/utilities.cpp
```diff
@@ -556,4 +559,122 @@ void findAllVariablesWithEquivalences(const ComponentPtr &component, VariablePtr
     }
 }
 
+IdList listIds(const ModelPtr &model, bool mathIds)
```
nickerso: seems to be very similar code here and in the annotator...any way to merge them?

src/utilities.h
```diff
@@ -211,6 +211,22 @@ static const std::map<Variable::InterfaceType, const std::string> interfaceTypeT
     {Variable::InterfaceType::PUBLIC, "public"},
     {Variable::InterfaceType::PUBLIC_AND_PRIVATE, "public_and_private"}};
 
+// static const std::map<Annotator::Type, const std::string> annotatorTypeToString = {
```
nickerso: Is this commented code still needed?

src/utilities.h
```diff
@@ -555,4 +571,9 @@ static inline std::string trimCopy(std::string s)
     return s;
 }
 
+// KRM need docstrings.
```
nickerso: yes.  :)

tests/model/model.cpp
```diff
@@ -784,11 +784,11 @@ TEST(Model, removeComponentInsensitiveToOrder)
 
     // If the order of operations is switched the behaviour is the same:
 
-    // Get a pointer to the second componet in the parsed model
+    // Get a pointer to the second componet in the parsed model.
```
nickerso: `componet` ==> `component` ?

tests/printer/printer.cpp
```diff
@@ -295,3 +293,257 @@ TEST(Printer, printModelWithStandardUnitsAdded)
...
+                           "    <map_variables variable_1=\"variable1\" variable_2=\"variable2\"/>\n"
+                           "    <map_variables variable_1=\"variable1\" variable_2=\"variable4\"/>\n"
+                           "  </connection>\n"
+                           "  <encapsulation>\n"
+                           "    <component_ref component=\"component2\">\n"
+                           "      <component_ref component=\"component3\"/>\n"
+                           "    </component_ref>\n"
+                           "  </encapsulation>\n"
+                           "</model>\n";
+    const std::string expectedOutString = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
```
nickerso: what happened to the random/hexadecimal IDs?

src/utilities.cpp
```diff
@@ -556,4 +559,122 @@ void findAllVariablesWithEquivalences(const ComponentPtr &component, VariablePtr
...
+    // TODO
+
+    for (size_t c = 0; c < component->componentCount(); ++c) {
+        listComponentIds(component->component(c), mathIds, idList);
+    }
+}
+
+std::string makeUniqueId(const std::string &type, IdList &idList)
+{
+    // Construct a unique id appropriate to the type, and add it to the idList.
```
nickerso: I thought we went away from these type-based IDs and we just using hexadecimal strings?

nickerso (changes requested):

src/annotator.cpp
kerimoyle: It needs to be there to distinguish between situations where the annotator has been built (the previous check) and where zero items with ids exist in the model.  Otherwise, we can't tell the difference between an empty idList.  It doesn't need to be an error, but it does need to be a return path.

src/annotator.cpp
kerimoyle: No, that's not how it works - there's no ID added to the model at all.  This is used so that we can keep track of the same request for nonexistent IDs, and only return one issue for them.  Otherwise we'd be generating a duplicate issue each time.

src/annotator.cpp
kerimoyle: Alas, no, for two reasons.  I tried this originally, and I couldn't cast a nullptr into the required return type that the compiler would accept.  Given that, it seemed better to return something that would be helpful to the user, and prevent duplicate issues being generated (as above).

tests/printer/printer.cpp
kerimoyle: They are in a different PR, see https://github.com/cellml/libcellml/pull/639

src/utilities.cpp
kerimoyle: We are, but that is a different PR.  See https://github.com/cellml/libcellml/pull/639

src/annotator.cpp
kerimoyle: Yes, weirdness.  Thanks!

src/annotator.cpp
kerimoyle: If they request a specific item type they will get a `nullptr` (eg: annotator->component("iDontExist") will return `nullptr`.  It's only the `item` function that is different.

src/utilities.h
kerimoyle: Nope :)

tests/model/model.cpp
kerimoyle: Serves me right for adding the missing full-stop ... ;)

src/annotator.cpp
kerimoyle: 83ce6a11

src/annotator.cpp
kerimoyle: 83ce6a11

src/utilities.h
kerimoyle: These are slightly different in the PR which uses them more #639 (we've now dropped testing/adding mathml ids, for example).  Should I merge that one with this one first?  It seems odd to change it in two places (which will have knock-on effects throughout the printer too).

tests/model/model.cpp
kerimoyle: 83ce6a11

src/utilities.h
kerimoyle: Will be done in the other PR instead.

src/utilities.h
kerimoyle: 83ce6a1

src/annotator.cpp
nickerso: but why do we need to distinguish the case when there are no items with `id`s the model vs the requested `id` is not used? There is absolutely nothing wrong with that, so I'm not sure why we need to care?

This use of `any` is really looking less and less attractive, as it seems to be forcing us to do things that just don't make sense...

src/annotator.cpp
nickerso: yeah, sorry, not really added to the model, but if you just look at the id list in the annotator as the list of items in the current model that can be annotated the user is forced to check that each item is at least not an `issue` before being able to use it...again, this is starting to look like the use of `any` is less than optimal.

nickerso (commented):
tentative approval :)

I'm not happy with the way the `item()` method is working, but it looks like that might just be something we need to live with for "version 1" of this functionality. Really seems that trying to make use of those `any` objects are forcing us down a path that is less that ideal...but it is better than what we have and provides useful functionality.

Happy to approve this subject to those C++ question(s) that I've put around the place...

nickerso:
(resolved some of my comments just cause they are not essential to address now, can be refined as we gain experience in using the Annotator)

src/annotator.cpp
kerimoyle: There's not a lot of difference, other than telling the user that there's no point looking for any ids, as there are none, rather than just not having the single one it's looking for.  I can lump them together if it's clearer.

kerimoyle: 83ce6a11

src/annotator.cpp
kerimoyle: It's nothing to do with the use of `any`? That's only so that we can return a pointer reasonably efficiently (rather than having a list of each different pointer type in the argument list of the function).  It's easy enough to remove!  We could just as easily have an `Annotator::Type::NOT_FOUND` enum or similar, with a `nullptr` as the object.  It would still need the user to check for type though, same as before.  But a side note - is viewing/printing a list of ids something a user would want?  I don't do that at the moment?

src/bindings/interface/annotator.i
```diff
@@ -6,21 +6,19 @@
...
 %import "types.i"
 
+%include <std_pair.i>
+%include <std_vector.i>
+%include <std_string.i>
+%include <stdint.i>
+
 %feature("docstring") libcellml::Annotator
-"Validates CellML objects.";
+"Interacs with CellML objects using their id attribute.";
```
agarny: Typo.

kerimoyle: 95ddcc31

src/annotator.cpp
nickerso: could be other reasons for the cast to fail beyond just the item type. So this seems like a reasonable way to go and will ensure that the casting exceptions would ever propagate out from the library

nickerso (dismissed):

nickerso (dismissed):

nickerso (dismissed):

kerimoyle:
Hi @agarny @nickerso @hsorby This is ready for review now.  A couple of things:
- There is (necessarily) a lot of code duplication in the overloaded functions.  This is so that the users can call one function with anything and have it work as they'd expect - we have to do all the monkey business behind the scenes to make that possible.  The typecast restrictions mean that these have to be separate functions; I can't see any way to reasonably condense them.
- There's an awkward situation with the binding interface stuff because SWIG doesn't currently support the `std::any` type, but Python doesn't care ...  Thus, the `annotator.i` file defines some C++ functions with a `ForSWIG` extension in their names, which are there to provide an alternative to the real C++ functions in a format that can be digested by SWIG (and only by SWIG).  If there's a way to make these internal/private please let me know. At present they've got the same status as any other bound function.

kerimoyle:
ping @agarny @hsorby @nickerso

tests/annotator/tests.cmake
```diff
@@ -0,0 +1,16 @@
+# Set the test name, 'test_' will be prepended to the
+# name set here
+set(CURRENT_TEST annotator)
+# Set a category name to enable running commands like:
+#    ctest -R <category-label>
+# which will run the tests matching this category-label.
+# Can be left empty (or just not set)
+set(${CURRENT_TEST}_CATEGORY io)
```
hsorby: I don't think the annotator belongs in the `io` category, maybe `tools`.

hsorby (commented):
I initially intended to just do some indentation fixes and it just got out of control from there.

src/api/libcellml/enums.h
```diff
@@ -0,0 +1,46 @@
...
+    MATHML,
+    MODEL,
+    RESET,
+    RESET_VALUE,
+    TEST_VALUE,
+    UNDEFINED,
+    UNIT,
+    UNITS,
+    VARIABLE,
+};
```
kerimoyle, line 48: Will we add `XML` and `INVALID` to this list so it can be used elsewhere?

src/api/libcellml/types.h
```diff
@@ -16,20 +16,24 @@ limitations under the License.
...
 
+#include "libcellml/enums.h"
+
 namespace libcellml {
 
 // General classes.
 class Annotator; /**< Forward declaration of Annotator class. */
-using AnnotatorPtr = std::shared_ptr<Annotator>; /**< Type definition for shared annotator pointer. */
+using AnnotatorPtr = std::shared_ptr<Annotator>; /**< Type definition for @c std::shared Annotator pointer some othe text. */
 class Generator; /**< Forward declaration of Generator class. */
```
kerimoyle, line 41: typo "othe" ... and also ... huh?

src/api/libcellml/types.h
```diff
@@ -16,20 +16,24 @@ limitations under the License.
...
 using GeneratorPtr = std::shared_ptr<Generator>; /**< Type definition for shared generator pointer. */
 class GeneratorProfile; /**< Forward declaration of GeneratorProfile class. */
 using GeneratorProfilePtr = std::shared_ptr<GeneratorProfile>; /**< Type definition for shared generator variable pointer. */
 class GeneratorVariable; /**< Forward declaration of GeneratorVariable class. */
 using GeneratorVariablePtr = std::shared_ptr<GeneratorVariable>; /**< Type definition for shared generator variable pointer. */
-class Importer; /**< Forward declaration of Importer class. */
+class Importer; /*!< Forward declaration of Importer class. */
```
kerimoyle: Why the ! ?

src/api/libcellml/types.h
```diff
@@ -44,7 +48,7 @@ using ValidatorPtr = std::shared_ptr<Validator>; /**< Type definition for shared
 
 // CellML entities.
 class Component; /**< Forward declaration of Component class. */
-using ComponentPtr = std::shared_ptr<Component>; /**< Type definition for shared component pointer. */
+using ComponentPtr = std::shared_ptr<Component>; /**< Type definition for shared component pointer. @details Some extra definition for detailed description. */
```
kerimoyle: ... is this you just playing with doxygen translation stuff ... ?

src/bindings/interface/annotator.i
```diff
@@ -104,13 +105,10 @@
...
 
-%feature("docstring") libcellml::Annotator::typeString
-"Translate the given Annotator::Type enumeration into a string.";
-
 %feature("docstring") libcellml::Annotator::duplicateCount
 "Return the number of items in the stored model with the given id.";
 
-%feature("docstring") libcellml::Annotator::isBuilt
+%feature("docstring") libcellml::Annotator::hasModel
 "Report whether the annotator index is up-to-date with its stored model state.";
```
kerimoyle, line 123: Docstring should be changed to reflect what the function does?

src/bindings/python/__init__.py
```diff
@@ -51,22 +51,27 @@ class Object:
...
-        'ISSUE',
         'MAP_VARIABLES',
+        'MATHML',
         'MODEL',
         'RESET',
         'RESET_VALUE',
         'TEST_VALUE',
+        'UNDEFINED',
         'UNIT',
         'UNITS',
```
kerimoyle: Also `XML` and `INVALID` ?

src/bindings/python/__init__.py
```diff
@@ -51,22 +51,27 @@ class Object:
...
         'VARIABLE'
-        ])
+], libcellml)
+
+# The Annotator class relies on the CellMLElmement declaration above.
+from libcellml.annotator import Annotator
+
```
kerimoyle: I don't understand why this is here?

src/internaltypes.h
```diff
@@ -50,9 +49,17 @@ using VariablePtrs = std::vector<VariablePtr>; /**< Type definition for list of
...
 using IdList = std::unordered_set<std::string>; /**< Type definition for list of ids. */
 
-using VariablePair = std::pair<VariablePtr, VariablePtr>;
-using UnitItem = std::pair<UnitsPtr, size_t>;
-using AnyItem = std::pair<std::uint64_t, std::any>; // Clang-tidy suggested this type instead of the Annotator::Type enum.
-using ItemList = std::multimap<std::string, AnyItem>;
+using ComponentWeakPtr = std::weak_ptr<Component>; /**< Type definition for weak component pointer. */
+//using ComponentEntityWeakPtr = std::weak_ptr<ComponentEntity>; /**< Type definition for weak component entity pointer. */
+//using EntityWeakPtr = std::weak_ptr<Entity>; /**< Type definition for weak entity pointer. */
+//using ImportedEntityWeakPtr = std::weak_ptr<ImportedEntity>; /**< Type definition for weak imported entity pointer. */
```
kerimoyle: delete if not used?

tests/annotator/annotator.cpp
```diff
@@ -427,40 +445,24 @@ TEST(Annotator, errorHandling)
     auto model = libcellml::Model::create("model");
     auto annotator = libcellml::Annotator::create();
 
-    // Searching before the map is built should return an error.
-    EXPECT_EQ(libcellml::Annotator::Type::ISSUE, annotator->item("i_dont_exist").first);
-    EXPECT_EQ(size_t(2), annotator->issueCount());
-    EXPECT_EQ(annotator->issue(0)->description(), "No model is stored; no action has been taken.");
-    EXPECT_EQ(annotator->issue(1)->description(), "The model index is out-of-date.  Please (re)build it before attempting to access items by their id.");
```
kerimoyle: I don't understand why you're not testing which error is returned but only the total?

tests/annotator/annotator.cpp
```diff
@@ -562,15 +564,19 @@ TEST(Annotator, getItemBySpecificTypeDuplicateId)
...
+    annotator->setModel(model);
 
     // Expect that the errors have been cleared.
     EXPECT_EQ(size_t(0), annotator->errorCount());
 
-    EXPECT_EQ(nullptr, annotator->model("duplicateId"));
-    EXPECT_EQ(nullptr, annotator->encapsulation("duplicateId"));
+    EXPECT_NE(nullptr, annotator->model("duplicateId"));
+    EXPECT_NE(nullptr, annotator->encapsulation("duplicateId"));
+
```
kerimoyle, line 310: I reeeaaaaeally don't agree with this change in behaviour!  I think it's really dodgy to not alert the user that the item found has a non unique id, and just return the first one we happen to come to.  This is not great ... :(

tests/annotator/annotator.cpp
```diff
@@ -1446,6 +1468,27 @@ TEST(Annotator, automaticIdAllItemsWrongModel)
...
+    auto component = model->component(0);
+    libcellml::AnyItem itemComponent = std::make_pair(libcellml::CellMLElement::COMPONENT, model->component(0));
+    EXPECT_EQ("component_1", component->id());
+
+    EXPECT_TRUE(annotator->hasModel());
+    model.reset();
+    EXPECT_FALSE(annotator->hasModel());
+
+    EXPECT_EQ("b4da55", annotator->assignId(itemComponent));
```
kerimoyle: What happens when this component is added to a model? or another component without a parent model is passed to the annotator?  I think if the item passed in for id isn't a part of the stored model (or there is no stored model!) it should return an error instead of assigning an id which can't be said to be unique as it has no context.

tests/annotator/annotator.cpp
```diff
@@ -1513,10 +1556,10 @@ TEST(Annotator, automaticIdAllItemsEntityType)
         model->component("component2")->variable("variable2")->equivalentVariable(0),
         model->component("component2")->variable("variable2"));
 
-    EXPECT_EQ("b4da62", annotator->assignId(libcellml::Annotator::Type::CONNECTION, connection2));
+    EXPECT_EQ("b4da62", annotator->assignId(connection2));
```
kerimoyle: I don't think we should have a default for variable pair stuff (connections and mappings).  And if we did, it shouldn't be the connection as we never actually interact with a connection directly.  But my strong preference would be there is no default at all.

tests/annotator/annotator.cpp
```diff
@@ -1534,30 +1577,60 @@ TEST(Annotator, assignAllIds)
...
+    auto annotator = libcellml::Annotator::create();
 
     annotator->clearAllIds(model);
     EXPECT_EQ(size_t(0), annotator->errorCount());
 
-    auto retrievalError = annotator->item("duplicateId");
-    EXPECT_EQ(libcellml::Annotator::Type::ISSUE, retrievalError.first);
-    EXPECT_EQ(expectedErrors[1], std::any_cast<libcellml::IssuePtr>(retrievalError.second)->description());
+    libcellml::AnyItem item = annotator->item("duplicateId");
+    EXPECT_EQ(libcellml::CellMLElement::UNDEFINED, item.first);
```
kerimoyle, line 1573: Should also test that error is logged

tests/annotator/annotator.cpp
```diff
@@ -1911,42 +1984,21 @@ TEST(Annotator, hashChangesAndUpdates)
...
 
-    EXPECT_FALSE(annotator->isBuilt());
-
-    annotator->buildModelIndex(model);
-    EXPECT_TRUE(annotator->isBuilt());
+    EXPECT_FALSE(annotator->hasModel());
 
-    model->setId("iHaveChangedSinceTheLastBuild");
-    EXPECT_FALSE(annotator->isBuilt());
+    // Model is set and the id map is not created.
```
kerimoyle, line 2015: The id map is **not** created?? I thought it was built when the model was set??

tests/annotator/annotator.cpp
```diff
@@ -1911,42 +1984,21 @@ TEST(Annotator, hashChangesAndUpdates)
...
 
-    // Can set id automatically, but this does not trigger a rebuild.
-    EXPECT_EQ("b4da55", annotator->assignId(libcellml::Annotator::Type::COMPONENT, model->component(0)));
-    EXPECT_FALSE(annotator->isBuilt());
+    // Can set id manually, this does not trigger a rebuild of the id map.
+    model->setId("iHaveChanged");
 
-    // Shouldn't be able to retrieve anything since the build is out of date.
+    // The id map is built when a query is made.
     EXPECT_EQ(nullptr, annotator->model("model_1"));
```
kerimoyle, line 2022: This test is a bit odd now?  The id has changed, so you should rather test that the new id can now be found (ie: that the index has successfully detected the change and been rebuilt)?

tests/annotator/annotator.cpp
```diff
@@ -1911,42 +1984,21 @@ TEST(Annotator, hashChangesAndUpdates)
...
-    EXPECT_EQ(nullptr, annotator->variable("variable_3"));
-    EXPECT_EQ(nullptr, annotator->variable("variable_4"));
-    EXPECT_EQ(nullptr, annotator->connection("connection_1").first);
-    EXPECT_EQ(nullptr, annotator->mapVariables("map_variables_1").first);
-    EXPECT_EQ(nullptr, annotator->mapVariables("map_variables_2").first);
+
+    // Can set id automatically, this does not trigger a rebuild of the id map.
+    EXPECT_EQ("b4da55", annotator->assignId(model->component(0)));
+
+    EXPECT_NE(nullptr, annotator->encapsulation("encapsulation_1"));
```
kerimoyle, line 2027: I don't understand this?  The old id should still be present and able to be retrieved since it's only the model id which was changed outside of the annotator?

tests/annotator/annotator.cpp
```diff
@@ -1957,11 +2009,11 @@ TEST(Annotator, hashUpdatedWithAllAutomaticIds)
...
     auto annotator = libcellml::Annotator::create();
 
-    annotator->buildModelIndex(model);
-    EXPECT_TRUE(annotator->isBuilt());
+    annotator->setModel(model);
+    EXPECT_TRUE(annotator->hasModel());
 
     annotator->assignAllIds();
-    EXPECT_TRUE(annotator->isBuilt());
+    EXPECT_TRUE(annotator->hasModel());
```
kerimoyle, line 2042: I understand the function names when `hasModel` is testing whether or not you have a model.  But I don't like it in this situation where it's actually testing that the stored model is up to date with the stored index.  Can we name it something else?

tests/annotator/annotator.cpp
```diff
@@ -1970,78 +2022,64 @@ TEST(Annotator, autoIdOnOutOfDateBuild)
...
+    EXPECT_EQ("b4da56", annotator->assignId(model->component("component2"), libcellml::CellMLElement::COMPONENT_REF));
     EXPECT_EQ("b4da56", model->component("component2")->encapsulationId());
 
     libcellml::VariablePair connection = std::make_pair(
         model->component("component2")->variable("variable1"),
         model->component("component2")->variable("variable1")->equivalentVariable(0));
     libcellml::Variable::setEquivalenceConnectionId(connection.first, connection.second, "changed");
-    EXPECT_FALSE(annotator->isBuilt());
-    EXPECT_EQ("b4da57", annotator->assignId(libcellml::Annotator::Type::CONNECTION, connection));
+    EXPECT_EQ("b4da57", annotator->assignId(connection));
```
kerimoyle: I think there shouldn't be a default for variable pair items (or if there is, it should be the map_variables one).  But better without any.

tests/annotator/annotator.cpp
```diff
@@ -1970,78 +2022,64 @@ TEST(Annotator, autoIdOnOutOfDateBuild)
...
     model->setId("changed");
-    EXPECT_FALSE(annotator->isBuilt());
-    EXPECT_EQ("b4da5b", annotator->assignId(libcellml::Annotator::Type::MODEL, model));
+    EXPECT_EQ("b4da5b", annotator->assignId(model));
     EXPECT_EQ("b4da5b", model->id());
 
     model->component("component2")->reset(0)->setId("changed");
-    EXPECT_FALSE(annotator->isBuilt());
-    EXPECT_EQ("b4da5c", annotator->assignId(libcellml::Annotator::Type::RESET, model->component("component2")->reset(0)));
+    EXPECT_EQ("b4da5c", annotator->assignId(model->component("component2")->reset(0), libcellml::CellMLElement::RESET));
```
kerimoyle, line 2096: This should/could be with the RESET as default?

src/api/libcellml/enums.h
hsorby: Well, no I hope not.  The XML one is a problem but it is not a CellML element.  The `INVALID` should be covered by `UNDEFINED` no?

src/api/libcellml/types.h
hsorby: Playing around with Doxygen documentation.

src/api/libcellml/types.h
hsorby: Again playing with Doxygen documentation.

src/api/libcellml/types.h
hsorby: Yes.  It should all be removed.

src/bindings/interface/annotator.i
hsorby: Definitely.

src/bindings/python/__init__.py
hsorby: Because of the transformation of how SWIG does enums and how we want to make them more friendly in Python.

src/internaltypes.h
hsorby: Yes.

tests/annotator/annotator.cpp
hsorby: If the documentation of the method explains the behaviour then I think this is the correct.  The only thing I could see it doing is raising a warning that this identifier is not unique.

tests/annotator/annotator.cpp
hsorby: Well if the component has no model then it is unique for the given context.  If you go and add it to a model the context has changed, nothing the library can do about that.

I don't see why it should return an error.  Someone has given me a component and they want an id assigned to it.  I can do that so I do, no problem.

tests/annotator/annotator.cpp
hsorby: Why would you not have a default?

src/annotator.cpp
```diff
@@ -0,0 +1,1741 @@
...
+}
+
+void Annotator::setModel(const ModelPtr &model)
+{
+    removeAllIssues();
+    mPimpl->mModel = model;
+    mPimpl->buildIdList();
+    mPimpl->mHash = mPimpl->generateHash();
+
+    for (auto it = mPimpl->mIdList.begin(), end = mPimpl->mIdList.end(); it != end; it = mPimpl->mIdList.upper_bound(it->first)) {
```
nickerso: Why are duplicate ids being checked here? seems like a lot of work for something that the user hasn't asked for.

If it does stay here, then the documentation definitely needs to mention that this is happening that then the user should be checking for issues after calling this method.

tests/annotator/annotator.cpp
kerimoyle: A default is useful for real CellML entities, where the parent item (reset) is the default, and owns something else (the resetVale).  Here, there is no CellML entity, there is no ownership, and the thing which people interact with (the explicit mapping of **variables** to each other, not the implicit connecting of **connections** to each other) isn't the default.  If you insist on having one of them, as a default, it should be the mapping, not the connection.  But even that is a stretch for me.

tests/annotator/annotator.cpp
kerimoyle: The only reason to assign an id using the Annotator is to assign something unique.  As soon as there's no model, it can't make it unique.  So it's kinda pointless.

src/annotator.cpp
```diff
@@ -0,0 +1,1741 @@
...
+        mPimpl->mIdList.clear();
+        mPimpl->mHash = 0;
+    } else {
+        mPimpl->addIssueNoModel();
+    }
+}
+
+void Annotator::clearAllIds(ModelPtr &model)
+{
+    mPimpl->mModel = model;
```
nickerso, line 923: won't this break everything? if anything, it should be calling `setModel` with the given `model`, right?

tests/annotator/annotator.cpp
kerimoyle: I think this will end up causing all kinds of problems.  It's naive and assumes that the users know our internal index order at which their components/variables/everything is stored ... which is kinda ridiculous as we ourselves don't ever rely on it!  Being able to interact with duplicated ids is a convenience function so that users can fix their own mistakes - I think it's dangerous to just hide the fact that it's duplicated from the user.  If they want to retrieve duplicated things they should use explicit functions to do that so that everyone knows exactly what they're dealing with.

src/bindings/python/__init__.py
kerimoyle: ... but does it need to be in this file?  Now that the enum for Type has been moved out of the Annotator class?

src/api/libcellml/enums.h
kerimoyle: Ah, I thought yesterday you wanted an invalid type.  I'm fine with undefined to cover both.

src/api/libcellml/annotator.h
```diff
@@ -0,0 +1,1054 @@
...
+     * @param index The position of an item within the list of items with the given @p id to retrieve.
+     *
+     * @return A @c ResetPtr.
+     */
+    ResetPtr resetValue(const std::string &id, size_t index);
+
+    /**
+     * @brief Assign an id string to every item in the model which is already 
+     * stored in this annotator, excluding its MathML items.
+     *
```
nickerso: I thought previously this method would be adding ids to items that didn't already have one. If the user wants to assign to everything they would first clear and then assign. Now there is no method to easily make sure that everything has an id without losing existing ones.

src/annotator.cpp
kerimoyle: It doesn't make sense to call `setModel` which will build the expensive index, and then call `clearAllIds` to wipe it out.  I thought this was cleaner.

src/api/libcellml/annotator.h
kerimoyle: Gah!! You told me to take that away!! I used to only assign a new id to things whose id was missing or invalid.  But, from our discussion, you said anyone wanting to assign ids automagically was prepared to nuke it all ... :(

src/annotator.cpp
kerimoyle: That's a hangover from previous stuff I think, so it could go ... I'm still not very happy with how we're dealing with duplicates though.

src/api/libcellml/annotator.h
nickerso: hmmm...that doesn't sound like something I would say :) is there still a print method that will add ids to things? maybe that is the one that will do what I am thinking here...

but at the moment I'm struggling to think why it would ever be ok to ignore ids that have already been set.... I can see a case for wanting to specifically nuke them all and start from scratch, but not that it would be the only method to make sure everything has an id. It would be a fairly typical use-case to want to simply make sure that an existing model has ids everywhere so that, for example, a user can annotate the model without having to edit the model.

kerimoyle: Haha, It's easy enough to change that back, but I'll wait until Hugh has finished his stuff.

src/annotator.cpp
kerimoyle: 57725d21

src/api/libcellml/annotator.h
kerimoyle: Actually, nothing needs changing. These functions all have an `something.empty()` test in them, so existing ids are not changed? I've changed the documentation so it's clearer.  Is that enough or did you find something that was being changed?

kerimoyle: 8d4307d2

src/annotator.cpp
```diff
@@ -252,108 +435,80 @@ void Annotator::buildModelIndex(const ModelPtr &model)
...
+bool Annotator::isUnique(const std::string &id)
 {
-    if (mPimpl->mIdList.count(id) > 1) {
-        if (raiseError) {
-            mPimpl->makeIssueNonUnique(id);
-        }
-        return false;
-    }
-    return true;
+    return mPimpl->mIdList.count(id) == 1;
```
kerimoyle, line 489: This needs to go back to the `> 1` test rather than `== 1`.  Otherwise we get a return of `false` when the id doesn't exist, which doesn't make sense...

tests/annotator/annotator.cpp
kerimoyle: I will change the documentation so that it's clear that the model must be set first.

tests/annotator/annotator.cpp
kerimoyle: Will change so that we have a default but to the `map_variables` instead of the connection.

tests/annotator/annotator.cpp
hsorby: This is with RESET as default I am testing here that you can pass RESET in with out any problems.

cmake/environmentchecks.cmake
```diff
@@ -33,19 +33,30 @@ find_program(LLVM_PROFDATA_EXE NAMES ${PREFERRED_LLVM_PROFDATA_NAMES} llvm-profd
...
+    endif()
+    if(PYTHON_COVERAGE_TEST_PERFORMED AND NOT PYTHON_COVERAGE_FOUND AND HAVE_COVERAGE)
+      set(_PRINT_SUMMARY_MESSAGE TRUE)
+      message(STATUS "Performing Test HAVE_COVERAGE")
+    endif()
+    if(NOT DEFINED PYTHON_COVERAGE_TEST_PERFORMED OR _PRINT_SUMMARY_MESSAGE)
+      message(STATUS "Performing Test HAVE_COVERAGE - ${_RESULT_TEXT}")
     endif()
+    set(PYTHON_COVERAGE_FOUND ${HAVE_COVERAGE} CACHE INTERNAL "Python coverage was found." FORCE)
+    set(PYTHON_COVERAGE_TEST_PERFORMED TRUE CACHE INTERNAL "Testing for Python coverage has occured.")
```
agarny: Why do we have all of this which has nothing to do with creating an `Annotator` class?

tests/annotator/tests.cmake
agarny: We don't seem to have anything in the `tools` category, but we have some tests in the `utils` category, which I think would be fine for the annotator. Done in commit 5adb854b.

src/bindings/python/__init__.py
agarny: What is the status of this file? I was going to remove the commented-out code, but then came across this unresolved conversation.

hsorby:
I cannot fathom why these recent changes have been added to this pull request @agarny.  Clearly they are separate changes with nothing to do with the Annotator class.

src/api/libcellml/enums.h
kerimoyle: We still need to rename this so that it includes `XML` and is named `ItemType` so that it can fit with the rest of the code... That will have knock-on effects throughout the rest of this PR.

src/api/libcellml/module/libcellml
```diff
@@ -29,6 +29,7 @@ limitations under the License.
 #include "libcellml/analyserequationast.h"
 #include "libcellml/analysermodel.h"
 #include "libcellml/analyservariable.h"
+#include "libcellml/annotator.h"
 #include "libcellml/component.h"
```
kerimoyle, line 33: Does this need `enums.h` too?

src/bindings/interface/annotator.i
```diff
@@ -0,0 +1,217 @@
...
+
+%feature("docstring") libcellml::Annotator::component
+"Return the component with the given id.";
+
+%feature("docstring") libcellml::Annotator::encapsulation
+"Return the model with the given encapsulation id.";
+
+%feature("docstring") libcellml::Annotator::model
+"Return the model assigned to this Annotator or the model with the given id.";
+
```
kerimoyle, line 30: Is this too ambiguous?  Should we have separate functions for these two things?

src/bindings/interface/annotator.i
```diff
@@ -0,0 +1,217 @@
...
+%feature("docstring") libcellml::Annotator::resetValue
+"Return the ResetPtr whose reset_value has the given id.";
+
+%feature("docstring") libcellml::Annotator::componentRef
+"Return the ComponentPtr with the given component_ref id.";
+
+%feature("docstring") libcellml::Annotator::unit
+"Return the UnitsPtr and index whose Unit is with given id.";
+
```
kerimoyle, line 54: Is this still true?  Or do we have a single structure that's returned now?

src/bindings/interface/annotator.i
```diff
@@ -0,0 +1,217 @@
...
+%feature("docstring") libcellml::Annotator::connection
+"Return a pair of VariablePtrs that define a connection with given id.";
+
```
kerimoyle, line 57: Ditto here - doesn't it now return a `VariablePair` item rather than two `VariablePtr`s?

src/bindings/interface/annotator.i
```diff
@@ -0,0 +1,217 @@
...
+"Set all items of the given type in the stored model to automatically generated, unique strings.";
+
+%feature("docstring") libcellml::Annotator::assignId
+"Set the id of the given item to an automatically generated, unique string.";
+
+%feature("docstring") libcellml::Annotator::clearAllIds
+"Clear all the id strings in the given or stored model.";
+
+%feature("docstring") libcellml::Annotator::isUnique
+"Return `true` if the given string is not duplicated in the stored model, or `false` otherwise.";
```
kerimoyle: Can we reword this?  "Not duplicated" includes "does not exist".  Should rather be:
"Return `true` if the given string occurs exactly once as an id in the stored model, or `false` otherwise."

src/bindings/interface/annotator.i
```diff
@@ -0,0 +1,217 @@
...
+"Return a list of id strings which are duplicated within the stored model.";
+
+%feature("docstring") libcellml::Annotator::assignComponentId
+"Assign an automatically generated, unique id to the given component.";
+
+%feature("docstring") libcellml::Annotator::assignComponentRefId
+"Assign an automatically generated, unique id to the given component's encapsulation.";
+
+%feature("docstring") libcellml::Annotator::assignConnectionId
+"Assign an automatically generated, unique id to the connection between the given variables.";
```
kerimoyle: "Assign an automatically generated, unique id to the connection between the parent components of the given variables.";

agarny (commented):
> I cannot fathom why these recent changes have been added to this pull request @agarny. Clearly they are separate changes with nothing to do with the Annotator class.

Which ones are you talking about exactly? Those changes are the result of my reviewing this PR. So, all I have done is to refactor, clean up, etc. code that is in this PR.

src/api/libcellml/module/libcellml
agarny: `enums.h` is included by `types.h` which in turn is included by most headers in this module. So, no need to include `enums.h`.

cmake/environmentchecks.cmake
hsorby: That is indeed true, it also shouldn't be in there.

src/bindings/interface/annotator.i
hsorby: 'Tis the beauty of function overloading.

kerimoyle: Yeah .. but it sure ain't beautiful here :| The two functions are doing two totally different things!

src/api/libcellml/module/libcellml
agarny: If we are to add `enums.h` then why don't we add `types.h`?

hsorby:
As far as I am concerned this work is ready for a full and proper review now.  (Assuming I haven't missed any comments) All comments have been addressed.

agarny:
> As far as I am concerned this work is ready for a full and proper review now. (Assuming I haven't missed any comments) All comments have been addressed.

Great, I will resume my review sometime today then. 👍

src/bindings/interface/enums.i
```diff
@@ -0,0 +1,26 @@
...
+
+%include "libcellml/enums.h"
+
+%pythoncode %{
+from enum import IntEnum
+def redo(prefix):
+    tmpD = {k:v for k,v in globals().items() if k.startswith(prefix + '_')}
+    tmpD = {k[len(prefix)+1:]:v for k,v in tmpD.items()}
+    globals()[prefix] = IntEnum(prefix,tmpD)
+redo('CellMLElement')
```
kerimoyle, line 23: Needs renaming to `ItemType` to include XML.

src/bindings/interface/types.i
```diff
@@ -126,3 +149,15 @@ Only meant to be included, shouldn't be passed to cmake as a module!
...
   }
 }
+
+%typemap(in) libcellml::Annotator::Type (int val, int ecode) {
+  ecode = SWIG_AsVal(int)($input, &val);
+  if (!SWIG_IsOK(ecode)) {
+    %argument_fail(ecode, "$type", $symname, $argnum);
+  } else {
+    if (val < %static_cast($type::COMPONENT, int) || %static_cast($type::VARIABLE, int) < val) {
+      %argument_fail(ecode, "$type is not a valid value for the enumeration.", $symname, $argnum);
```
kerimoyle, line 158: Needs to extend to use XML as last type

kerimoyle (commented):
I've only had a quick look, but my biggest concern is that we'll have to redo the `CellMLElement` enum immediately so that it doesn't entirely duplicate the `ItemType` one.  My comments from yesterday also seem to have been missed?

src/bindings/interface/enums.i
hsorby: This doesn't need to be done here.  It can come in with Issues.

src/bindings/interface/types.i
hsorby: XML isn't required by the Annotator class so it should only come in when needed.

src/api/libcellml/annotator.h
```diff
@@ -0,0 +1,1115 @@
...
+    /**
+     * @brief Get the number of items with the given @p id.
+     *
+     * Return the number of items with the given @p id.
+     *
+     * @param id A @c std::string for the id to find.
+     * 
+     * @return the number of items with the given @p id.
+     */
+    size_t duplicateCount(const std::string &id);
```
nickerso: perhaps not the best name for this method :) but can be fixed in future.

src/api/libcellml/enums.h
nickerso: is this part of #696? so we can resolve things on this PR and leave it to that one to sort out?

tests/annotator/annotator.cpp
nickerso: is this one resolved? the methods seem to be documented to behave this way, so I'm assuming this is all good and we could come back and refactor this behaviour if needed...

nickerso (dismissed):
I think this one is looking good and behaves as I think it should be. A couple of outstanding comments from @kerimoyle that should be resolved, but otherwise good to go.

src/api/libcellml/annotator.h
hsorby: Pretty easy to fix now if you have one ready to go.

kerimoyle: It has to end in `Count()`
It's returning the number of ids which are duplicated ... so ... `duplicateIdCount`?

agarny: Why not simply `idCount()`?

tests/annotator/annotator.cpp
```diff
@@ -0,0 +1,2131 @@
...
+            switch (item.first) {
+            case libcellml::CellMLElement::COMPONENT:
+            case libcellml::CellMLElement::COMPONENT_REF:
+                EXPECT_EQ(std::any_cast<libcellml::ComponentPtr>(expectedItems[id][index].second), std::any_cast<libcellml::ComponentPtr>(item.second));
+                break;
+            case libcellml::CellMLElement::CONNECTION:
+            case libcellml::CellMLElement::MAP_VARIABLES:
+                pair = std::any_cast<libcellml::VariablePair>(expectedItems[id][index].second);
+                testPair = std::any_cast<libcellml::VariablePair>(item.second);
+                // TODO not sure how best to test these.
```
kerimoyle: remove?

src/api/libcellml/annotator.h
kerimoyle: ... because that's not what it's doing!  It's returning the number of ids in the model which are duplicated.  Not the number of ids in the model.

agarny: I understand that, but you are passing an id. So, if I was to have a call like `idCount("my_id")` then I would understand it as being the number of times `my_id` is used as an id. Alternatively, I think I would be fine with `duplicatedIdCount()` .

kerimoyle: Maybe `itemCount` then? To me id count is counting the number of different ids.

nickerso: I vote for `itemCount`

agarny: > Maybe `itemCount` then? To me id count is counting the number of different ids.

Fair point, I would personally be happy with `itemCount()`.

tests/annotator/annotator.cpp
hsorby: I think so it is now back to the original setting, @kerimoyle can correct me if i'm wrong here.

src/api/libcellml/types.h
kerimoyle: Are we leaving this as is ...? or changing it back?

tests/annotator/annotator.cpp
kerimoyle: Yes, I'm happier now, thanks.

src/api/libcellml/annotator.h
```diff
@@ -0,0 +1,1115 @@
...
+     * @return A @ref ModelPtr to a model, or a @c nullptr.
+     */
+    ModelPtr model() const;
+
+    /**
+     * @brief Retrieves an item with the given id string.
+     *
+     * The item returned is a @c std::pair containing:
+     *  - a @ref CellMLElement enum, and
+     *  - a @c std::any item containing the item.
```
agarny, line 79: From PR #696 (see https://github.com/cellml/libcellml/pull/696#discussion_r499883292), it would make more sense to do swap the items. IOW, I would have:
```
     * The item returned is a @c std::pair containing:
     *  - a @c std::any item containing the item; and
     *  - a @ref CellMLElement enum.
```

src/api/libcellml/annotator.h
agarny: Are we ok with `itemCount()`? I am happy to modify it as part of my reviewing.

src/annotator.cpp
```diff
@@ -0,0 +1,1678 @@
...
+{
+    // Helper function so that the users don't have to build a model library when they're
+    // about to overwrite it all anyway.
+    if (model == nullptr) {
+        auto issue = Issue::create();
+        issue->setDescription("The Model supplied is a nullptr. No action has been taken.");
+        issue->setLevel(Issue::Level::ERROR);
+        return false;
+    }
+    mPimpl->mModel = model;
```
agarny: I have seen the comment at https://github.com/cellml/libcellml/pull/640/files#r482607731, but does it really apply here?

src/api/libcellml/types.h
```diff
@@ -70,4 +77,49 @@ using UnitsPtr = std::shared_ptr<Units>; /**< Type definition for shared units p
...
 
+/**
+ * @brief Type definition for VariablePtr pair.
+ *
+ * A VariablePair is a pair containing two @ref VariablePtrs.
+ * It may be used to define:
+ *  - a connection between parent components; or
+ *  - an equivalence between two variables;
+ */
+using VariablePair = std::pair<VariablePtr, VariablePtr>; /**<  */
```
agarny, line 88: What is that trailing `/**<  */`?

src/bindings/interface/annotator.i
agarny: Any final decision on this?

src/bindings/interface/annotator.i
agarny: I believe this has been resolved.

src/bindings/interface/annotator.i
agarny: I believe this has been resolved.

src/bindings/interface/annotator.i
agarny: This has been resolved.

src/variable.cpp
```diff
@@ -509,20 +509,23 @@ void Variable::setEquivalenceMappingId(const VariablePtr &variable1, const Varia
 void Variable::setEquivalenceConnectionId(const VariablePtr &variable1, const VariablePtr &variable2, const std::string &connectionId)
```
agarny, line 509: In other places, we use a `VariablePair` when wanting to do something for two variables. Why not here?

tests/annotator/annotator.cpp
```diff
@@ -0,0 +1,2131 @@
...
+
+    // Encapsulation structure.
+    auto c1 = model->component("c1");
+    auto c2 = model->component("c2");
+    auto v3 = libcellml::Variable::create("v3");
+    auto v4 = libcellml::Variable::create("v4");
+    c1->addVariable(v3);
+    c2->addVariable(v4);
+    auto v1v2 = std::make_pair(c1->variable("v1"), c2->variable("v2"));
+    auto v3v4 = std::make_pair(c1->variable("v3"), c2->variable("v4"));
```
agarny, line 113: Exactly the kind of thing I wouldn't expect a libCellML user to have to do (i.e. use something like `std::make_pair()`).

tests/annotator/annotator.cpp
```diff
@@ -0,0 +1,2131 @@
...
+    reset->setResetValueId("reset_value_id");
+    reset->setTestValueId("test_value_id");
+    c1->addReset(reset);
+
+    annotator->setModel(model);
+
+    // Unit items are returned as name string and pair of UnitsPtr parent and index:
+    EXPECT_EQ(libcellml::CellMLElement::UNIT, annotator->item("unit1_id").first);
+    EXPECT_EQ(units, std::any_cast<libcellml::UnitItem>(annotator->item("unit1_id").second).first);
+    EXPECT_EQ(size_t(0), std::any_cast<libcellml::UnitItem>(annotator->item("unit1_id").second).second);
```
agarny, line 141: Concrete example of what I mentioned about the use of `std::pair`. Here, we have no idea what `x.first` and `x.second` are about. We can tell from the context, but it's not ideal.

tests/annotator/annotator.cpp
```diff
@@ -0,0 +1,2131 @@
...
+
+    // Reset children.
+    auto reset = libcellml::Reset::create();
+    reset->setResetValueId("reset_value_id");
+    reset->setTestValueId("test_value_id");
+    c1->addReset(reset);
+
+    annotator->setModel(model);
+
+    // Unit items are returned as name string and pair of UnitsPtr parent and index:
```
agarny: What do you mean by "name string" (same below)? From what I can tell, it's a `class enum`, no?

tests/annotator/annotator.cpp
```diff
@@ -0,0 +1,2131 @@
...
+    EXPECT_EQ(variable, std::any_cast<libcellml::VariablePtr>(annotator->item("variable_id").second));
+    EXPECT_EQ(libcellml::CellMLElement::RESET, annotator->item("reset_id").first);
+    EXPECT_EQ(reset, std::any_cast<libcellml::ResetPtr>(annotator->item("reset_id").second));
+    EXPECT_EQ(libcellml::CellMLElement::UNITS, annotator->item("units_id").first);
+    EXPECT_EQ(units, std::any_cast<libcellml::UnitsPtr>(annotator->item("units_id").second));
+    EXPECT_EQ(libcellml::CellMLElement::IMPORT, annotator->item("import_id").first);
+    EXPECT_EQ(importSource, std::any_cast<libcellml::ImportSourcePtr>(annotator->item("import_id").second));
+}
+
+TEST(Annotator, getNonEntityFromId)
```
agarny: What is meant by non-entity from id?

tests/annotator/annotator.cpp
```diff
@@ -0,0 +1,2131 @@
...
+
+    // Test that repeated calls to the same unfound id generate new errors:
+    EXPECT_EQ(libcellml::CellMLElement::UNDEFINED, annotator->item("i_dont_exist").first);
+    EXPECT_EQ(size_t(2), annotator->issueCount());
+}
+
+TEST(Annotator, duplicateIdBehaviour)
+{
+    std::vector<std::string> expectedErrors = {
+        "The id 'duplicateId' occurs 3 times in the model so a unique item cannot be located."};
```
agarny: I feel like the error message should read "... model, so..." (i.e. missing comma?).

tests/annotator/annotator.cpp
```diff
@@ -0,0 +1,2131 @@
...
+
+    // Doing the extra set of connections and mappings.
+    libcellml::VariablePair r1 = std::make_pair(model->component("component2")->variable("variable1"), model->component("component4")->variable("variable1"));
+    libcellml::VariablePair r2 = std::make_pair(model->component("component2")->variable("variable2"), model->component("component4")->variable("variable2"));
+    libcellml::AnyItem itemConnection2 = std::make_pair(libcellml::CellMLElement::CONNECTION, r1);
+    libcellml::AnyItem itemMapVariables2 = std::make_pair(libcellml::CellMLElement::MAP_VARIABLES, r2);
+    EXPECT_EQ("b4da62", annotator->assignId(itemConnection2));
+    EXPECT_EQ("b4da63", annotator->assignId(itemMapVariables2));
+
+    EXPECT_TRUE(annotator->hasModel());
```
agarny: Why test `annotator->hasModel()` so many since it's always `true`?

tests/annotator/annotator.cpp
```diff
@@ -0,0 +1,2131 @@
...
+    annotator->clearAllIds(model);
+    EXPECT_EQ(size_t(0), annotator->errorCount());
+
+    libcellml::AnyItem item = annotator->item("duplicateId");
+    EXPECT_EQ(libcellml::CellMLElement::UNDEFINED, item.first);
+}
+
+TEST(Annotator, assignIdInvalidType)
+{
+    const std::string e = "The given type 'reset' is invalid for this method.";
```
agarny: This error message doesn't help me. I would expect to know why it is invalid.

tests/annotator/annotator.cpp
agarny: Yes. Done it as part of my reviewing of this PR.

agarny (commented):
Looks ok overall, but still have a few more questions to which it would be nice to have an answer before approving. :)

tests/annotator/annotator.cpp
hsorby: That is a search/replace special `hasModel` was originally something else and I reappropriated it.

src/api/libcellml/types.h
hsorby: Should have been deleted.

agarny: It has gone anyway, so resolved.

src/bindings/interface/annotator.i
hsorby: I am happy as it currently is.

kerimoyle: What happens if the current model in the annotator doesn't have the id specified?  Which function is it calling then?  I really don't like this.

tests/annotator/annotator.cpp
hsorby: Something that isn't a libCellML entity: VariablePair.

tests/annotator/annotator.cpp
kerimoyle: I prefer it without.

src/variable.cpp
kerimoyle: A discussion for another time ... I've made an issue for it at https://github.com/cellml/libcellml/issues/745

tests/annotator/annotator.cpp
kerimoyle: It is now :)

tests/annotator/annotator.cpp
kerimoyle: b386607e

tests/annotator/annotator.cpp
kerimoyle: b386607e

src/annotator.cpp
kerimoyle: I don't understand what you mean?

kerimoyle: Fixed in 3d9e774a

src/annotator.cpp
```diff
@@ -450,30 +446,33 @@ void Annotator::AnnotatorImpl::addIssueNoModel() const
...
-
-    return exists;
+    return true;
 }
 
 AnyItem Annotator::item(const std::string &id)
 {
     AnyItem retrieved = std::make_pair(CellMLElement::UNDEFINED, nullptr);
-    if (isUnique(id)) {
+    auto num = duplicateCount(id);
```
nickerso: hasn't `duplicateCount` been renamed?

nickerso (dismissed):
Thought we were moving away from `duplicateCount` but that can be resolved in future if its still an issue.

tests/annotator/annotator.cpp
agarny: Then, I woudl expect something like `getVariablePairFromId`. At least, I don't have to guess/ask/etc. what `NonEntity` is...

agarny: Done in commit 2df300f4.

src/annotator.cpp
agarny: Nope, see https://github.com/cellml/libcellml/pull/640/files#r499497547. Some of use are fine with renaming it `itemCount`. So, am happy to make that change. Agreed?

cmake/environmentchecks.cmake
agarny: Ok, reverted this in commit a270779c.

src/api/libcellml/annotator.h
nickerso: 3 of us are happy with `itemCount` and no votes against so I think we're safe to go for it...

src/annotator.cpp
agarny: Done in commit 47839635.

src/api/libcellml/annotator.h
agarny: Done in commit 47839635.

src/api/libcellml/annotator.h
agarny: Does anyone have a view on this? Happy to create a separate issue for it, if needed.

src/bindings/interface/annotator.i
agarny: @hsorby and @kerimoyle: you guys need to come up with some kind of an agreement so that we can resolve this conversation. :)

nickerso: > What happens if the current model in the annotator doesn't have the id specified? Which function is it calling then? I really don't like this.

Isn't this just pointing to the overloaded C++ code, so as with the C++ code if you ask for a model with a given id, its only going to return a non-null pointer if the id matches that of the current model. Right? The only difference here is that there is a fall back option to provide the current model for the annotator...

I don't see a need to separate this into two separate python methods. (assuming I'm understanding this issue correctly)

tests/annotator/annotator.cpp
agarny: Created issue #748 for it.

tests/annotator/annotator.cpp
agarny: Will be addressed in issue #748.

tests/annotator/annotator.cpp
agarny: What do/did you mean by this @kerimoyle?

tests/annotator/annotator.cpp
agarny: Is this comment still relevant? @kerimoyle? @hsorby?

tests/annotator/annotator.cpp
agarny: Is this comment still relevant? @kerimoyle? @hsorby?

tests/annotator/annotator.cpp
agarny: Is this comment still relevant? @kerimoyle? @hsorby?

agarny (dismissed):
There are still a few unresolved conversations, but none that I am concerned with. So, personally, I am happy with the PR as-is.

agarny:
> There are still a few unresolved conversations, but none that I am concerned with. So, personally, I am happy with the PR as-is.

Oops, I have somehow broken this PR. Argh! :)

agarny:
@hsorby, @kerimoyle and @nickerso: any idea why Ihad to revert commit cf5ab33 to get the Python tests to run fine again? I really don't like the idea of having to import `CellMLElement` before anything else for things to work. This is really bad, not to mention that I like, in PyCharm, to reformat my code. So, I will regularly end up with something like in commit cf5ab33, which will then break things as we have seen.

IOW, I would like to have something like commit cf5ab33 back in before we merge this PR.

agarny:
> IOW, I would like to have something like commit [cf5ab33](https://github.com/cellml/libcellml/commit/cf5ab33ac5069b8f06f76e31d4b5d37672848af5) back in before we merge this PR.

Ok, I had overlooked the fact that it's in our `__init__.py`. So, it's not like an end-user will have to remember to import `CellMLElement` first. So, ok, although I don't like the fact that we import `CellMLElement` first, it doesn't affect the end-user, so I guess I am "happy" with the PR as-is.

agarny (approved):

src/bindings/interface/annotator.i
hsorby: You have the same understanding that I have and I am happy with the way it works.

agarny: I am happy with this too, so resolving.

nickerso (approved):

Issue #641, opened 2020-06-24
Empty connection and import ids are unreachable


kerimoyle:
Now that empty connections are valid (see [here](https://cellml-specification.readthedocs.io/en/latest/reference/formal_and_informative/specB15.html)) there is no way to retrieve the id attribute of an empty connection element:

```
<connection component_1="component1" component_2="component2" id="youCantDoNothingWithMe" /> 
```

The way we retrieve a `connection` id attribute is using the child `map_variables` pairs of variables or through other variable-based functions, so where those variables don't exist, we're stuck?
```
    /**
     * @brief Get the equivalent connection id for this equivalence.
     *
     * Get the connection id set for the equivalence defined with the given variables.
     * The variables are commutative.  If no connection id is set the empty string is returned.
     *
     * If the two variables are not equivalent the empty string is returned.
     *
     * @param variable1 Variable one of the equivalence.
     * @param variable2 Variable two of the equivalence.
     * @return the @c std::string connection id.
     */
    static std::string equivalenceConnectionId(const VariablePtr &variable1, const VariablePtr &variable2);

    /**
     * @brief Get the equivalent connection id for this equivalence.
     *
     * Get the connection id set for the equivalence defined by the this variable and the given one.
     * If no connection id is set the empty string is returned.
     *
     * If the two variables are not equivalent the empty string is returned.
     *
     * @param equivalentVariable The variable this variable is equivalent to.
     * @return The @c std::string id of the equivalence if found otherwise returns the empty string.
     */
    std::string equivalentConnectionId(const VariablePtr &equivalentVariable) const;
```

nickerso:
This is really a more general issue for objects in a CellML model that have no direct representation in the libcellml data model, right? `encapsulation` and `import` elements at least, and maybe more...

I'd currently see this as a very low priority issue to resolve.

kerimoyle:
We can get to the ids of `encapsulation` and `import` items, but we can't get to empty `connection` ones:  
```
model->encapsulationId();
model->component(0)->importSource()->id();
model->units(0)->importSource()->id();
```
It's not so much a matter of representation (as you say, there are plenty of things we don't represent at present), just of access to data that's in the CellML file, but that we can't "read" at the moment.

nickerso:
I doubt we can get to empty `import` or `encapsulation` elements either, right?

kerimoyle:
Encapsulations are ok, this test passes:
```
TEST(Parser, emptyEncapsulationWithId)
{
[...]

PR #642, opened 2020-06-25
I641 add warnings for empty parsed blocks

kerimoyle:
Addresses #641 

Adds better warning messages when the parser encounters empty blocks (which are valid but not saved).

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -783,7 +783,7 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr 
 
     XmlNodePtr childNode = node->firstChild();
 
-    if (!childNode) {
+    if (childNode == nullptr) {
         IssuePtr issue = Issue::create();
         std::string des = "Connection in model '" + model->name() + "'";
         if (connectionId.empty()) {
@@ -1160,7 +1160,7 @@ void Parser::ParserImpl::loadImport(const ImportSourcePtr &importSource, const M
     }
     XmlNodePtr childNode = node->firstChild();
 
-    if (!childNode) {
+    if (childNode == nullptr) {
         auto issue = Issue::create();
         if (id.empty()) {
             issue->setDescription("Import from '" + node->attribute("href") + "' is empty and will be disregarded.");
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -785,7 +785,14 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr 
 
     if (!childNode) {
         IssuePtr issue = Issue::create();
-        issue->setDescription("Connection in model '" + model->name() + "' does not contain any 'map_variables' elements.");
+        std::string des = "Connection in model '" + model->name() + "'";
+        if (connectionId.empty()) {
+            des += " does not contain any 'map_variables' elements. It will not be loaded into the model.";
+        } else {
+            des += " has an id of '" + connectionId + "' but does not contain any 'map_variables' elements.";
+            des += " The connection will not be loaded into the model, and the associated id bookmark will be lost.";
+        }
+        issue->setDescription(des);
         issue->setModel(model);
         issue->setCause(Issue::Cause::CONNECTION);
         issue->setReferenceRule(Issue::ReferenceRule::CONNECTION_MAP_VARIABLES);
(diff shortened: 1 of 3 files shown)


src/parser.cpp
```diff
@@ -785,7 +785,14 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr
 
     if (!childNode) {
         IssuePtr issue = Issue::create();
-        issue->setDescription("Connection in model '" + model->name() + "' does not contain any 'map_variables' elements.");
+        std::string des = "Connection in model '" + model->name() + "'";
+        if (connectionId.empty()) {
+            des += " does not contain any 'map_variables' elements. It will not be loaded into the model.";
+        } else {
+            des += " has an id of '" + connectionId + "' but does not contain any 'map_variables' elements.";
+            des += " The connection will not be loaded into the model, and the associated id bookmark will be lost.";
```
nickerso: I wouldn't use the word bookmark. It is an id, that is all.

src/parser.cpp
```diff
@@ -1150,6 +1159,19 @@ void Parser::ParserImpl::loadImport(const ImportSourcePtr &importSource, const M
         attribute = attribute->next();
     }
     XmlNodePtr childNode = node->firstChild();
+
+    if (!childNode) {
+        auto issue = Issue::create();
+        if (id.empty()) {
+            issue->setDescription("Import from '" + node->attribute("href") + "' is empty. It will not be loaded into the model.");
+        } else {
+            issue->setDescription("Import from '" + node->attribute("href") + "' has an id of '" + id + "' but is empty. It will not be loaded into the model, and the associated id bookmark will be lost.");
```
nickerso: as above, wouldn't use the word bookmark.

src/parser.cpp
nickerso: actually, I'd probably stop with `The connection will not be loaded into the model.`

nickerso: or maybe `The connection will be disregarded.`

nickerso (dismissed):
I approve this change. Have made a couple of suggestions to "improve" the messages, but could be fixed up later if needed.

src/parser.cpp
kerimoyle: Yesterday @hsorby wanted me to add the note for items with an id, that that info will be lost, on top of the normal empty warning?

nickerso: To me, that information is clearly implied by saying the connection will be disregarded/not loaded.

kerimoyle: Compromise in cc44b080

src/parser.cpp
kerimoyle: Removed in cc44b080

nickerso (dismissed):

src/parser.cpp
```diff
@@ -1150,6 +1159,19 @@ void Parser::ParserImpl::loadImport(const ImportSourcePtr &importSource, const M
         attribute = attribute->next();
     }
     XmlNodePtr childNode = node->firstChild();
+
+    if (!childNode) {
```
agarny: Replace with `if (childNode == nullptr) {`.

tests/parser/parser.cpp
```diff
@@ -2040,3 +2076,7 @@ TEST(Parser, repeatedMathParsePrintBehaviourWithReset)
 
     EXPECT_EQ(in, output2);
 }
+
+TEST(Parser, parseEmptyEncapsulationWithId)
+{
+}
```
agarny: What is this?...

agarny (changes requested):
Looks good apart from a couple minor things.

src/parser.cpp
kerimoyle: I'm happy to change it, but I want to understand why?  We have a whole load of `while(something)` statements throughout the XML stuff, and I would have thought that `if(!something)` was the corollary of those?

tests/parser/parser.cpp
kerimoyle: Finally, a test that passes!  ;)

agarny: And will always pass no matter what. Best test ever!

src/parser.cpp
agarny: Both are (obviously) valid, equivalent, etc. However, `if (childNode == nullptr) {` is the recommended way by tools like ClangTidy (because it makes the code "clearer"). I am not quite sure why it didn't pick it up here (and elsewhere from what I understand). We will have to check our ClangTidy settings.

agarny: Just to confirm, `while (something)` should also be replaced with `while (something != nullptr)`, `while (something != 0`, etc.

tests/parser/parser.cpp
kerimoyle: 2635f862

src/parser.cpp
kerimoyle: 2635f862

kerimoyle: OK.  I'll make a separate issue then - it's all throughout the code.

agarny (approved):

nickerso (approved):

PR #649, opened 2020-07-03
Add support for clcache/ccache

agarny:
Addresses issue #313.

PR #631, opened 2020-06-16
Added an `Analyser` class

agarny:
Addresses issue #499.

PR #650, opened 2020-07-06
Code generator: add support for external model parameters

### LINKED ISSUE #627: Code generator: add support for external model parameters
Right now, we can generate code for a model where model parameters can be one of:
 - Variable of integration;
 - State;
 - Constant;
 - Computed constant; and
 - Algebraic.

However, there may be cases where we don't want some model parameters to be computed and, instead, their value to be retrieved from elsewhere. A typical case is when we want to apply a voltage clamp protocol to a model of cardiac electrophysiology. In that case, we don't want to compute the membrane potential (i.e. the typical `dV/dt`) and instead provide its value at a given time.

When it comes to the `Analyser` class, we could achieve this by updating our `processModel()` method. It currently takes a model as a parameter, but we could also pass it an optional list of variables for which we want to ignore the computation. Then when it comes to the generated code, we would have a callback in place of the equation that would have normally computed the model parameter. Will need to think a bit more about it, but I image that only non-VOI variables could become "external" variables.

@nickerso, we are going to need this for SPARC at some point, so I am going to work on this after issue #499. Otherwise, I remember us discussing this issue at some point, but I can't recall the details, so feel free to feed in this issue, if you remember anything. 🙂

agarny:
Addresses issue #627.


agarny:
Strange... the latest commit listed in this PR is commit 9bf5abc (or https://github.com/agarny/libcellml/commit/9bf5abc22e5cbb16148fd8aa675ff8e36e10463c in my branch), but in my branch it is commit https://github.com/agarny/libcellml/commit/792fe6997e2a8dbb7bfb0794b534e1e93d397837...?!

@hsorby, @kerimoyle and @nickerso: have you guys ever come across something like this?

agarny:
Otherwise, and FWIW, this PR is ready for (pre)reviewing. I will test things a bit further tomorrow (with regards to usability of the generated code), but I think I am pretty much done with it.

This aside, note that this PR relies on PRs #649 and #631.

agarny:
@hsorby: there are definitely things that don't make sense to me when it comes to the new buildbot:
- Commit 20a225f fails as expected since there are some Clang-Tidy checks that fail.
- Commit 3dde729 passes, but why is that? There is nothing in that commit that addresses the Clang-Tidy checks!
- Commit 6b0b7e9 also passes while it's the same story as commit 3dde729!
- Commit 6d151a3 fails and rightly so since the aforementioned Clang-Tidy checks haven't been fixed. They are in commit c341a31, which hopefully will pass...

src/analyser.cpp
```diff
@@ -0,0 +1,1632 @@
...
+
+            mExternalVariables.assign(uniqueExternalVariables.begin(), uniqueExternalVariables.end());
+
+            for (const auto &multipleExternalVariable : multipleExternalVariables) {
+                auto issue = Issue::create();
+
+                issue->setDescription("Variable '" + multipleExternalVariable->name()
+                                      + "' in component '" + owningComponent(multipleExternalVariable)->name()
+                                      + "' is marked as an external variable more than once.");
+                issue->setLevel(Issue::Level::WARNING);
```
nickerso: note to self: is this really worth a warning?

agarny: I like my code to be clean, so if I do something that is unnecessary then I want to be told about it. It's like when you declare a variable in C/C++, but never end up using it. Your compiler will warn you about it.

nickerso (commented):
Not seeing any documentation here that would tell me how to define/what is expected from the compute external variables method... but also not sure if our code generation is documented at all yet anyway :)

agarny:
> Not seeing any documentation here that would tell me how to define/what is expected from the compute external variables method... but also not sure if our code generation is documented at all yet anyway :)

Yes, there is no documentation at this stage... but shouldn't that be done as part of @kerimoyle's tutorial?

agarny:
Yet another case where buildbot doesn't give me the expected result: commit 32281d2 fails while it passes on my Mac...?!

kerimoyle:
> Yes, there is no documentation at this stage... but shouldn't that be done as part of @kerimoyle's tutorial?

Weeeeellll ... I'm gonna need a whole lot more info ... might be easier if you draft something up that I can work from?  Happy to include it, but I don't know what goes on inside the generator either!

agarny:
> > Yes, there is no documentation at this stage... but shouldn't that be done as part of @kerimoyle's tutorial?
> 
> Weeeeellll ... I'm gonna need a whole lot more info ... might be easier if you draft something up that I can work from? Happy to include it, but I don't know what goes on inside the generator either!

I wasn't suggesting that you should write it. I should clearly be the one writing it, at least the first full draft. I was merely wondering whether information about the code generator should be part of the tutorial or some other document.

agarny:
Ping @hsorby, @kerimoyle and @nickerso.

As one might expect, to generate code for a model with external variables can be done in different ways. In all cases, the analysis must account for the fact that there might be external variables. This means that we need to pass any external variable we may have to `Analyser::processModel()`. For instance:

```
auto analyser = libcellml::Analyser::create();
std::vector<libcellml::VariablePtr> externalVariables;

externalVariables.push_back(model->component("membrane")->variable("V"));
externalVariables.push_back(model->component("sodium_channel")->variable("i_Na"));
externalVariables.push_back(model->component("potassium_channel_n_gate")->variable("alpha_n"));

analyser->processModel(model, externalVariables);
```

If there are no external variables, we will obviously simply have:

```
auto analyser = libcellml::Analyser::create();

analyser->processModel(model);
```

When it comes to the generated code, we currently generate something like:

```
void computeRates(double voi, double *states, double *rates, double *variables, ComputeExternalVariables computeExternalVariables)
{
    computeExternalVariables(voi, states, rates, variables);
    ...
}

void computeVariables(double voi, double *states, double *rates, double *variables, ComputeExternalVariables computeExternalVariables)
{
    computeExternalVariables(voi, states, rates, variables);
    ...
}
```

i.e. we update both `computeRates()` and `computeVariables()` by:
- passing an extra parameter to it (`computeExternalVariables`); and
- calling `computeExternalVariables` at the **beginning** of each method to compute external variables.

A few things to keep in mind though:
- for a given `voi` value, we compute all the external variables at once.
- although we pass `states`, `rates` and `variables` to `computeExternalVariables()`, only the `variables` items of `EXTERNAL` type are expected to be modified.
- `computeExternalVariables()` is called first in case any of the external variables are needed to compute anything in `computeRates()` and/or `computeVariables()`.
- because of the previous point, it means that (despite having access to the value of all states, rates and variables) none of the external variables ought to depend on other states, rates and/or variables.

Regarding our first point, we could perfectly replace:

```
computeExternalVariables(voi, states, rates, variables);
```

with:

```
variables[0] = externalVariable(voi, states, rates, variables, 0);
variables[10] = externalVariable(voi, states, rates, variables, 10);
variables[17] = externalVariable(voi, states, rates, variables, 17);
```

but I am not sure there is much point in doing it...?

Anyway, I am going to work some more on the `hodgkinHuxleySquidAxonModel1952WithExternalVariablesUse` test to mimic the case where the value of an external variable comes from some external data that is linearly intepolated. This should be our main use case, so it will tell us whether the current approach is viable. (Who knows, we may need to pass a `void *` argument to `computeExternalVariables()` to get access to "external" information.)

kerimoyle:
Just a quick thought about user interaction with this - nowhere else do we use vectors to add/remove/define anything.  Would it be more in keeping with the operations elsewhere to have an `addExternalVariable` method instead of passing in a vector of external variables?

```
auto analyser = libcellml::Analyser::create();

analyser->addExternalVariable(model->component("membrane")->variable("V"));
analyser->addExternalVariable(model->component("sodium_channel")->variable("i_Na"));
analyser->addExternalVariable(model->component("potassium_channel_n_gate")->variable("alpha_n"));

analyser->processModel(model);
```

agarny:
> Just a quick thought about user interaction with this - nowhere else do we use vectors to add/remove/define anything. Would it be more in keeping with the operations elsewhere to have an `addExternalVariable` method instead of passing in a vector of external variables?
> 
> ```
> auto analyser = libcellml::Analyser::create();
> 
> analyser->addExternalVariable(model->component("membrane")->variable("V"));
> analyser->addExternalVariable(model->component("sodium_channel")->variable("i_Na"));
> analyser->addExternalVariable(model->component("potassium_channel_n_gate")->variable("alpha_n"));
> 
> analyser->processModel(model);
> ```

I have to agree that this would indeed be more in line with the rest of our API. @hsorby and @nickerso, happy with @kerimoyle's suggestion?

nickerso:
> I have to agree that this would indeed be more in line with the rest of our API. @hsorby and @nickerso, happy with @kerimoyle's suggestion?

I would agree.

Some other thoughts having taken a quick look at this one earlier today (so I probably missed something!)... Currently the user would need to handle their own mapping from the external variable they specify to the actual variable present in the generated code, right? So presumably when the `computeExternalVariables` is called, the user will need to iterate through all the variable information to find the external variables and then see which one maps to the actual variable they set as external. Extending the `addExternalVariable` method with something like a string or integer key/index which is then passed through to the `computeExternalVariables` method could make this much easier on the user. And may also suggest that the 

> ```
> variables[0] = externalVariable(voi, states, rates, variables, 0);
> variables[10] = externalVariable(voi, states, rates, variables, 10);
> variables[17] = externalVariable(voi, states, rates, variables, 17);
> ```

could be the way to go?

> to mimic the case where the value of an external variable comes from some external data that is linearly intepolated. This should be our main use case

I would also add the case of an external variable dependent on the state (thinking force-Ca relationships as a very common one here) as a good test case.

agarny:
> > I have to agree that this would indeed be more in line with the rest of our API. @hsorby and @nickerso, happy with @kerimoyle's suggestion?
> I would agree.

Ok, I am going to add an `Analyser:: addExternalVariable(const VariablePtr &variable)` method then. However, for completeness, I thought I would have a quick look at `ComponentEntity::addComponent()`... and if we were to be truly complete (which we probably aren't in other parts of our API...?) then it means that the following methods would need to be added to the `Analyser` class (yeah!):
- `bool addExternalVariable(const VariablePtr &variable)`;
- `bool removeExternalVariable(size_t index)`;
- `bool removeExternalVariable(const std::string &componentName, const std::string &variableName)`;
- `bool removeExternalVariable(const VariablePtr &variable)`;
- `void removeAllExternalVariables()`;
- `bool containsExternalVariable(const std::string &componentName, const std::string &variableName) const`;
- `bool containsExternalVariable(const VariablePtr &variable) const`;
- `VariablePtr externalVariable(size_t index) const`;
- `VariablePtr externalVariable(const std::string &componentName, const std::string &variableName) const`;
- `VariablePtr takeExternalVariable(size_t index)`;
- `VariablePtr takeExternalVariable(const std::string &componentName, const std::string &variableName)`;
- `bool replaceExternalVariable(size_t index, const VariablePtr &variable)`;
- `bool replaceExternalVariable(const std::string &componentName, const std::string &variableName, const VariablePtr &variable)`;
- `bool replaceExternalVariable(const VariablePtr &oldVariable, const VariablePtr &newVariable)`;
- `size_t externalVariableCount() const`;

Let me know which ones, if any, you think we could drop... at the risk of not being complete... :)

> Some other thoughts having taken a quick look at this one earlier today (so I probably missed something!)... Currently the user would need to handle their own mapping from the external variable they specify to the actual variable present in the generated code, right?

That is correct. The code with the `externalVariable()` method below is what I originally had. However, I then tried to think about some possible use cases and I could think of a tool like OpenCOR using libCellML and someone "manually" using libCellML.

In the case of a tool, that tool could easily retrieve the index of external variables in the generated code. So, it should be easy for that tool to generate code for either `computeExternalVariables()` or `externalVariable()`, although we might need that `void *` argument I mentioned earlier.

In the case of someone "manually" using libCellML, that person will likely end up editing the generated code to add the corresponding `computeExternalVariables()` or `externalVariable()` methods. In that case, that person can see (read) the index value for the different external variables from the generated code.

Anyway, that's my current thinking...

> So presumably when the `computeExternalVariables` is called, the user will need to iterate through all the variable information to find the external variables and then see which one maps to the actual variable they set as external. Extending the `addExternalVariable` method with something like a string or integer key/index which is then passed through to the `computeExternalVariables` method could make this much easier on the user. And may also suggest that the
> 
> > ```
> > variables[0] = externalVariable(voi, states, rates, variables, 0);
> > variables[10] = externalVariable(voi, states, rates, variables, 10);
> > variables[17] = externalVariable(voi, states, rates, variables, 17);
> > ```
> 
> could be the way to go?

Remember that upon processing a model, we can retrieve an `AnalyserModel` object (by calling `Analyser::model()`), which can be used to retrieve the index (in the generated code) of a variable of `EXTERNAL` type. So, no need to extend the `addExternalVariable()` method.

So, at this stage, I still feel like `computeExternalVariables()` would be sufficient for our needs, although I will get confirmation once I have implemented our use cases below.

> > to mimic the case where the value of an external variable comes from some external data that is linearly intepolated. This should be our main use case
> 
> I would also add the case of an external variable dependent on the state (thinking force-Ca relationships as a very common one here) as a good test case.

Right now, `computeExternalVariables()` / `externalVariable()` can use the value of any state, rate or variable. However, we compute external variables before computing anything else, which means that we don't currently account for the fact that an external variable may have a dependency on (an)other variable(s), whatever their type.

We could certainly add that feature though, but then if we want to be in line with the rest of our API then it probably means that we would need to introduce an `AnalyserExternalVariable` class, so that we can `addDependency()` to it (as well as `removeDependency()`, `removeAllDependencies()`, etc.).

is that what we want? To allow an external variable to have dependencies is very attractive to me, so I would certainly be in favour of it. Just quite a bit more work involved when it comes to the API side of things. (When it comes to code generation, it will be a bit more work, but it would also result in more sensible generated code.) This also means that we would have no choice but to drop `computeExternalVariables()` in favour of `externalVariable()` (FWIW, this is what I implemented in the first place, so it would be relatively straightforward to put it back in).

Views?

kerimoyle:
> bool removeExternalVariable(size_t index);

I would expect this index to be relative to a component?

> bool containsExternalVariable(const std::string &componentName, const std::string &variableName) const;
bool containsExternalVariable(const VariablePtr &variable) const;

I don't quite understand these - does a `Variable` "contain" an external variable? or just "is" one?

I'm wondering whether "externalVariable" is a designation added to an existing `Variable`, rather like an interface type, so that you could "set" a variable to be external or not, rather than adding and removing them?  But then I guess we'd still want the take/remove/add/replace family as well as the set/unset/is family.  Not sure, sorry :\

agarny:
> > bool removeExternalVariable(size_t index);
> I would expect this index to be relative to a component?

?? An external variable is a `Variable` object which is the child of an `Analyser` object. So, that index will be specific to `Analyser`, nothing to do with a component.

> > bool containsExternalVariable(const std::string &componentName, const std::string &variableName) const;
> > bool containsExternalVariable(const VariablePtr &variable) const;
> 
> I don't quite understand these - does a `Variable` "contain" an external variable? or just "is" one?

An external variable (in the context of an `Analyser`) is just a `Variable`.

> I'm wondering whether "externalVariable" is a designation added to an existing `Variable`, rather like an interface type, so that you could "set" a variable to be external or not, rather than adding and removing them? But then I guess we'd still want the take/remove/add/replace family as well as the set/unset/is family. Not sure, sorry :\

I guess we will discuss this later today at our CellML meeting. :)

agarny:
Currently working on `Analyser::addExternalVariable(const VariablePtr &variable)` and... this is proving to be trickier than originally anticipated. Indeed, if we consider the following code:

```c++
auto analyser = libcellml::Analyser::create();

analyser->addExternalVariable(model->component("environment")->variable("time"));
analyser->addExternalVariable(model->component("membrane")->variable("V"));
analyser->addExternalVariable(model->component("sodium_channel")->variable("V"));
analyser->addExternalVariable(other_model->component("membrane")->variable("V"));

analyser->processModel(model);
```

The problem with the above is that we are adding variables as external variables of the analyser, BUT... at this stage, we don't know about the model that we want to analyse. So:
1. That addition should fail because we are trying to add the VOI. However, `model` hasn't been analysed, so we can tell that `environment.time` is the VOI.
2. That addition should work (default behaviour).
3. That addition should fail because `membrane.V` is equivalent to `sodium_channel.V`, which is already added. That can easily be tested, so that's fine.
4. Not sure what we should expect here since, at this stage, we have one external variable from `model` while here we are trying to add a variable from `other_model`!

IOW, by having `Analyser::processModel(const ModelPtr &model, const std::vector<VariablePtr> &externalVariables)` means that we know what `externalVariables` should be related to `model` (although we don't currently check it; oopsie!).

So, yes, it's really due to the nature of `Analyser` (and `Generator`), which I would like to be generic and not linked to a specific `Model` object (i.e. as we do for the `Validator` class).

So, more food for thoughts for our CellML meeting... Argh!

kerimoyle:
- Is it possible to maintain a vector inside the `Analyser` of externalVariable pointers?  Then, whenever the call comes to process a model, they can be matched against the variables in the model and you go from there?  
- Alternatively, we can add a switch/attribute/flag/setting whatever to the `Variable` class, so that you can then call `setExternalVariable` on variables in the `Model` itself, which would let you pass the model in once (as in validator, printer etc?)

agarny:
> * Is it possible to maintain a vector inside the `Analyser` of externalVariable pointers?  Then, whenever the call comes to process a model, they can be matched against the variables in the model and you go from there?

`Analyser` already has such a vector to track its external variables, so that's not the issue. The issue is that we can't properly test a variable when trying to add it as an external variable to an analyser.

> * Alternatively, we can add a switch/attribute/flag/setting whatever to the `Variable` class, so that you can then call `setExternalVariable` on variables in the `Model` itself, which would let you pass the model in once (as in validator, printer etc?)

No, nothing should be added to the `Variable` class. The concept of external variable is specific to `Analyser`. It refers to a variable that is not set/computed by a model, but externally (hence the name).

kerimoyle:
> The issue is that we can't properly test a variable when trying to add it as an external variable to an analyser.

You can't test anything until you have all the information ... so either you restrict the user to add the model first (then you can test as the variables are added?), or your first test at processing is to go through the variables?

kerimoyle:
And in any case, we have other classes where we don't have a "one stop shop" (eg: the importer), so perhaps that idea is the real stumbling block? You could just return an error if the owning model of the variables and the model to be processed didn't match.

agarny:
> > The issue is that we can't properly test a variable when trying to add it as an external variable to an analyser.
> You can't test anything until you have all the information ... so either you restrict the user to add the model first (then you can test as the variables are added?), or your first test at processing is to go through the variables?

To restrict the user to add the model first means that we wouldn't be consistent. Indeed, the `Analyser` class should behave the same as the `Validator` class, which has a `validateModel()` method while the `Analyser` has a (brand new) `analyseModel()` method. IOW, the `Validator` is generic and so should be the `Analyser` class.

What could certainly be done (but it's ugly) is to allow for any variable to be marked as an external variable (incl. what we know to be a VOI, a duplicate, etc.), and then when before actually analysing the model, we could check the external variables and make sure that they are meaningful. It's, in effect, what we are currently doing.

Ok, I guess I am going to do that... for now, at least.

agarny:
> And in any case, we have other classes where we don't have a "one stop shop" (eg: the importer), so perhaps that idea is the real stumbling block? You could just return an error if the owning model of the variables and the model to be processed didn't match.

I am not currently checking that a variable is indeed in the model that we want to analyse, so that's something that I need to add, and generate a warning if the analyser comes across one such external variable.

kerimoyle:
> IOW, the Validator is generic and so should be the Analyser class.

The only reason the `Validator` is generic is because it doesn't need any other information.  Since the `Analyser` does need other information ...

agarny:
> > IOW, the Validator is generic and so should be the Analyser class.
> 
> The only reason the `Validator` is generic is because it doesn't need any other information. Since the `Analyser` does need other information ...

Sure and that's the same with the analyser... by default (i.e. when there are no external variables). It's our wish to have external variables that "messes up" everything.

nickerso:
I think you may be looking at this the wrong way around. A user tagging variables as external should be forcing that definition when you then analyse a specific model. I don't see anything wrong with a user wanting to specify any variable externally. Maybe the analyser/code generation can't cope with making, for example, the VOI effectively a "constant", but that is a limitation of libcellml, not something that "makes no sense".

I think you want to allow the user to let the analyser know about any variable from any model that should be treated in this way. And then when you process a specific model you can then impose any restrictions on combinations that are not currently supported.

agarny:
> I think you want to allow the user to let the analyser know about any variable from any model that should be treated in this way. And then when you process a specific model you can then impose any restrictions on combinations that are not currently supported.

Yes, that's exactly what I currently have in mind and working on as we "speak".

agarny:
FWIW, regarding https://github.com/cellml/libcellml/pull/650#issuecomment-658419359, I am thinking of dropping `takeExternalVariable()` and `replaceExternalVariable()`. The signature of `bool removeExternalVariable(const std::string &componentName, const std::string &variableName)`, `bool containsExternalVariable(const std::string &componentName, const std::string &variableName) const` and `VariablePtr externalVariable(const std::string &componentName, const std::string &variableName) const` will also need to be modified to cinlude a pointer to a model. IOW, we will have:
```c++
bool addExternalVariable(const VariablePtr &variable);
bool removeExternalVariable(size_t index);
bool removeExternalVariable(const ModelPtr &model, const std::string &componentName, const std::string &variableName);
bool removeExternalVariable(const VariablePtr &variable);
void removeAllExternalVariables();
bool containsExternalVariable(const ModelPtr &model, const std::string &componentName, const std::string &variableName) const;
bool containsExternalVariable(const VariablePtr &variable) const;
VariablePtr externalVariable(size_t index) const;
VariablePtr externalVariable(const ModelPtr &model, const std::string &componentName, const std::string &variableName) const;
size_t externalVariableCount() const;
```

Regarding `externalVariable()`, it will initially return a `VariablePtr`, but it will eventually return an `AnalyserExternalVariablePtr` (needed so that an external variable can have dependencies). Just one step at a time...

agarny:
Another CI annoyance: commit 079cc25 fails while it all works fine on my computer. From what I can tell, CI was done using commit d574dfc rather than 079cc25...

agarny:
@hsorby: another one for you. I keep getting coverage issue with this PR while I have none in my Ubuntu 18.04 VM using gcov (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0. Which version of gcov are we using in our CI?

agarny:
@hsorby, would you mind finding out why GCov coverage doesn't work here? As I mentioned [above](https://github.com/cellml/libcellml/pull/650#issuecomment-676877802), it all works fine on my Ubuntu 18.04 VM. So, I really don't understand why our Linux CI complains about [this line](https://github.com/cellml/libcellml/pull/650/files#diff-db1fe99815e5fd78699d5508bd9307ceR39).

hsorby:
> @hsorby, would you mind finding out why GCov coverage doesn't work here? As I mentioned [above](https://github.com/cellml/libcellml/pull/650#issuecomment-676877802), it all works fine on my Ubuntu 18.04 VM. So, I really don't understand why our Linux CI complains about [this line](https://github.com/cellml/libcellml/pull/650/files#diff-db1fe99815e5fd78699d5508bd9307ceR39).

I'll have a look.

agarny:
Ok, as far as I can tell, this is good to go. Ping @kerimoyle, @hsorby and @nickerso.

src/analyserexternalvariable.cpp
```diff
@@ -0,0 +1,182 @@
...
+std::vector<VariableWeakPtr>::iterator AnalyserExternalVariable::AnalyserExternalVariableImpl::findDependency(const VariablePtr &variable)
+{
+    return std::find_if(mDependencies.begin(), mDependencies.end(), [=](const VariableWeakPtr &v) {
+        return v.lock() == variable;
+    });
+}
+
+AnalyserExternalVariable::AnalyserExternalVariable(const VariablePtr &variable)
+    : mPimpl(new AnalyserExternalVariableImpl(variable))
+{
```
kerimoyle, line 54: If we have a constructor here that takes an object argument (cf just a name elsewhere).  Do we have to specify the similar constructor without an argument, or is that automagic (I see it's used below, but can't see the def?)  And do we need to have constructors elsewhere that take parent arguments too? eg: you can't have a variable without a component, so should we have Variable::create(parentComponent) etc ...?

src/analyserexternalvariable.cpp
```diff
@@ -0,0 +1,182 @@
...
+VariablePtr AnalyserExternalVariable::variable() const
+{
+    return mPimpl->mVariable.lock();
+}
+
+bool AnalyserExternalVariable::addDependency(const VariablePtr &variable)
+{
+    auto pimplVariable = AnalyserExternalVariable::variable();
+
+    if ((pimplVariable != nullptr)
```
kerimoyle, line 76: What are the conditions that would cause `pimplVariable` to be `nullptr` here?  If you've got coverage, clearly it happens ... but I'd be keen to know why/when the simple constructor would fail?

src/analyserexternalvariable.cpp
```diff
@@ -0,0 +1,182 @@
...
+        && !isSameOrEquivalentVariable(variable, pimplVariable)
+        && (owningModel(variable) == owningModel(pimplVariable))
+        && mPimpl->findDependency(variable) == mPimpl->mDependencies.end()) {
```
kerimoyle: Would it be faster to check the owning models before checking for equivalence?

src/analysermodel.cpp
```diff
@@ -41,6 +41,15 @@ AnalyserModel::Type AnalyserModel::type() const
     return mPimpl->mType;
 }
 
+bool AnalyserModel::hasExternalVariables() const
+{
+    if (!isValid()) {
```
kerimoyle, line 46: Why do you check validity here? The function name doesn't imply anything to do with that?

src/api/libcellml/analyser.h
```diff
@@ -58,6 +58,171 @@ class LIBCELLML_EXPORT Analyser: public Logger
...
+    /**
+     * @brief Add an @c AnalyserExternalVariable to this @c Analyser.
+     *
+     * Add the given @c AnalyserExternalVariable to this @c Analyser, but only
+     * if the given @c AnalyserExternalVariable or its primary equivalent has
+     * not already been added.
+     *
+     * @param variable The @c AnalyserExternalVariable to add.
+     *
+     * @return True if the variable was added, false otherwise.
```
kerimoyle: @c true and @c false?

src/api/libcellml/analyser.h
```diff
@@ -58,6 +58,171 @@ class LIBCELLML_EXPORT Analyser: public Logger
      */
     void analyseModel(const ModelPtr &model);
 
+    /**
+     * @brief Add an @c AnalyserExternalVariable to this @c Analyser.
+     *
+     * Add the given @c AnalyserExternalVariable to this @c Analyser, but only
+     * if the given @c AnalyserExternalVariable or its primary equivalent has
+     * not already been added.
```
kerimoyle: This might be explained elsewhere ... but what's a primary equivalent?  The `Variable` that this `AnalyserExternalVariable` came from?

src/api/libcellml/analyser.h
```diff
@@ -58,6 +58,171 @@ class LIBCELLML_EXPORT Analyser: public Logger
...
+     * @param variable The @c AnalyserExternalVariable to add.
+     *
+     * @return True if the variable was added, false otherwise.
+     */
+    bool addExternalVariable(const AnalyserExternalVariablePtr &externalVariable);
+
+    /**
+     * @brief Remove the @c AnalyserExternalVariable at the given @p index.
+     *
+     * Remove the @c AnalyserExternalVariable with the given @p index. @p index
```
kerimoyle: **The** `@p` index ... ?

src/api/libcellml/analyser.h
kerimoyle: Also below ... not sure if this is done throughout the rest of the library or not though!

src/api/libcellml/analyser.h
```diff
@@ -58,6 +58,171 @@ class LIBCELLML_EXPORT Analyser: public Logger
...
+     * @c Analyser.
+     *
+     * Clears all @c AnalyserExternalVariable items that have been added to this
+     * @c Analyser.
+     */
+    void removeAllExternalVariables();
+
+    /**
+     * @brief Tests to see if the @c AnalyserExternalVariable with the given
+     * @p variableName in the @c Component with the given @p componentName in
```
kerimoyle, line 137: Should be imperative throughout C++ docs?  ie: "Test to see whether ..."   Also in "Clears" in previous function and below in this function.

src/api/libcellml/analyser.h
```diff
@@ -58,6 +58,171 @@ class LIBCELLML_EXPORT Analyser: public Logger
...
+     *
+     * @param model The pointer to the @c Model which contains the
+     * @c AnalyserExternalVariable to test.
+     * @param componentName The name of the @c Component which contains the
+     * @c AnalyserExternalVariable to test.
+     * @param variableName The name of the @c AnalyserExternalVariable to test.
+     *
+     * @return @c true if the @c AnalyserExternalVariable is in this @c Analyser
+     * and @c false otherwise.
+     */
```
kerimoyle, line 154: Here we have `@c true` and `@c false` ... cf above.

src/api/libcellml/analyser.h
```diff
@@ -58,6 +58,171 @@ class LIBCELLML_EXPORT Analyser: public Logger
...
+    bool containsExternalVariable(const ModelPtr &model,
+                                  const std::string &componentName,
+                                  const std::string &variableName) const;
+
+    /**
+     * @overload
+     *
+     * @brief Tests to see if the @c AnalyserExternalVariable with the given
+     * pointer is contained within this @c Analyser.
```
kerimoyle, line 163: imperative ...

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+
+namespace libcellml {
+
+/**
+ * @brief The AnalyserExternalVariable class.
+ *
+ * The AnalyserExternalVariable class is for representing an external variable
+ * in the context of a CellML Analyser, i.e. a variable which is not set or
+ * computed by a model, but set externally.
+ */
```
kerimoyle, line 33: Should this be "computed by **this** model", since it would (I think??) be possible to use this functionality to couple two different model simulations together?  ie: this doesn't just take external experimental or clamped conditions, but could take external simulation data too?

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * @brief Create an @c AnalyserExternalVariable object.
+     *
+     * Factory method to create an @c AnalyserExternalVariable. Create an
+     * analyser external variable with::
+     *
+     *   VariablePtr variable = libcellml::Variable::create();
+     *   AnalyserExternalVariablePtr analyserExternalVariable = libcellml::AnalyserExternalVariable::create(variable);
+     *
```
kerimoyle, line 50: Use `auto` here?

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+     * @return A smart pointer to a @c AnalyserExternalVariable object.
+     */
+    static AnalyserExternalVariablePtr create(const VariablePtr &variable) noexcept;
+
+    /**
+     * @brief Get the variable.
+     *
```
kerimoyle, line 58: At the moment none of these lines tell me what's going on, especially as we have "external variables" and "variables" and "primary variables" in terminology in other files? Can this be made more helpful/consistent with elsewhere?

kerimoyle: Also, should "Returns" be "Return" ?

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+     * @brief Add a dependency to this external @c Variable.
+     *
+     * Add the given @c Variable as a dependency of this external @c Variable,
+     * but only if the given @c Variable or its primary equivalent has not
+     * already been added.
+     *
+     * @param variable The @c Variable to add as a dependency.
+     *
+     * @return True if the variable was added, false otherwise.
+     */
```
kerimoyle, line 76: See comment in another file: not sure which we're going with re boolean variables? `@c true` or "true"?

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+    bool addDependency(const VariablePtr &variable);
+
+    /**
+     * @brief Remove the dependency at the given @p index.
+     *
+     * Remove the dependency with the given @p index. @p index must be in the
+     * range [0, \#dependencies).
```
kerimoyle: The @p index?

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+     *
+     * @return True if the dependency was removed, false otherwise.
+     */
+    bool removeDependency(const VariablePtr &variable);
+
+    /**
+     * @brief Remove all dependencies from this external @c Variable.
+     *
+     * Clears all dependencies that have been added to this external
+     * @c Variable.
```
kerimoyle: Clear not clears?

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+     */
+    void removeAllDependencies();
+
+    /**
+     * @brief Tests to see if the dependency with the given @p variableName in
+     * the @c Component with the given @p componentName in the given @p model is
```
kerimoyle, line 135: "Test" not "Tests", also throughout this function ("Return")

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+     * contained within this external @c Variable.
+     *
```
kerimoyle, line 137: Does a variable really "contain" dependencies?  Or does it just have a list of dependencies?  I'm not sure which is true:
- "contained within this external variable"; or
- "contained within this external variable's dependency list"?

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+     * Tests to see if the dependency with the the given @p variableName in the
+     * @c Component with the given @p componentName in the given @p model is
+     * contained within this external @c Variable. Returns @c true if the dummy
+     * is in the external @c Variable and @c false otherwise.
```
kerimoyle: Who's the dummy? 🤣

src/api/libcellml/analyserexternalvariable.h
kerimoyle: Also in overloaded function below.

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+     * with the given @p componentName in the given @p model on success,
+     * @c nullptr on failure.
+     */
+    VariablePtr dependency(const ModelPtr &model,
+                           const std::string &componentName,
+                           const std::string &variableName) const;
+
+    /**
+     * @brief Get the list of @c Variable dependencies.
+     *
```
kerimoyle, line 214: Should we be saying "list" when it's a vector, esp since there's a `std::list` container too ... ?

src/api/libcellml/analyserexternalvariable.h
```diff
@@ -0,0 +1,234 @@
...
+     * Return the list of @c Variable items on which this
+     * @c AnalyserExternalVariable depends.
+     *
+     * @return The list of @c Variable dependencies.
+     */
+    std::vector<VariablePtr> dependencies() const;
+
+    /**
+     * @brief Get the number of dependencies in this external @c Variable.
+     *
```
kerimoyle, line 224: "in" or "of"? see earlier comment about variables "containing" dependencies?

src/api/libcellml/analysermodel.h
```diff
@@ -79,6 +79,17 @@ class LIBCELLML_EXPORT AnalyserModel
      */
     Type type() const;
 
+    /**
+     * @brief Test to determine if @c AnalyserModel has external variables.
+     *
```
kerimoyle, line 84: ... if **this** `AnalyserModel` has ... ?

src/api/libcellml/generatorprofile.h
```diff
@@ -2509,14 +2509,40 @@ class LIBCELLML_EXPORT GeneratorProfile
      * @brief Set the @c std::string for the data structure for the variable
      * type object.
      *
-     * Set the @c std::string for the data structure for the variable type
-     * object.
+     * Set the @c std::string for the data structure for the variable
+     * type object. To be useful, the string should contain the <OPTIONAL_TYPE>
+     * tag, which will be replaced with a type for the external variable type,
```
kerimoyle, line 2520: What happens if no string is given?  I know it's nonsense to pass in something blank ...

src/bindings/interface/analyserexternalvariable.i
```diff
@@ -0,0 +1,70 @@
...
+"Removes a dependency specified by 1) an index, 2) a :class:`Model` object, component name and variable name, or
+3) a :class:`Variable` object. Returns `True` on success.";
+
+%feature("docstring") libcellml::AnalyserExternalVariable::removeAllDependencies
+"Removes all dependencies from this external variable.";
+
+%feature("docstring") libcellml::AnalyserExternalVariable::containsDependency
+"Tests if a dependency, specified by 1) a :class:`Model` object, component name and variable name, or 2) a
+:class:`Variable` object.";
+
```
kerimoyle, line 32: This sentence doesn't make sense?

src/bindings/interface/analyserexternalvariable.i
```diff
@@ -0,0 +1,70 @@
...
+%feature("docstring") libcellml::AnalyserExternalVariable::dependency
+"Returns the dependency, specified by 1) an index, or 2) a :class:`Model` object, component name and variable
+name.";
+
+%feature("docstring") libcellml::AnalyserExternalVariable::dependencies
+"Returns the list of :class:`Variable` objects on which this :class:`AnalyserExternalVariable` depends.";
+
+%feature("docstring") libcellml::AnalyserExternalVariable::dependencyCount
+"Returns the number of dependencies this external variable contains.";
+
```
kerimoyle, line 42: FWIW I like these two descriptions a lot more than the ones on the C++ side :)

src/bindings/interface/generatorprofile.i
```diff
@@ -693,6 +693,12 @@ the <VARIABLE_COUNT> tag, which will be replaced with the number of states in th
 %feature("docstring") libcellml::GeneratorProfile::setVariableTypeObjectString
 "Set the string for the data structure for the variable type object.";
 
+%feature("docstring") libcellml::GeneratorProfile::variableTypeObjectExternalTypeString
+"Return the string for the external type for the data structure for the variable type object.";
+
+%feature("docstring") libcellml::GeneratorProfile::setVariableTypeObjectExternalTypeString
+"Set the string for the external type for the data structure for the variable type object.";
+
```
kerimoyle, line 701: In the other bindings documentation it's all descriptive (returns, sets etc?) but not in this file?

src/analyser.cpp
```diff
@@ -1349,129 +1369,303 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
...
-            } else {
-                // This is the variable of integration, so skip it.
+            for (const auto &externalVariable : mExternalVariables) {
+                if (owningModel(externalVariable->variable()) != model) {
+                    auto issue = Issue::create();
+
+                    issue->setDescription("Variable '" + externalVariable->variable()->name()
+                                          + "' in component '" + owningComponent(externalVariable->variable())->name()
+                                          + "' is marked as an external variable, but it belongs to a different model and will therefore be ignored.");
+                    issue->setLevel(Issue::Level::MESSAGE);
```
kerimoyle: "marked as an external variable, but it belongs to a different model and will therefore be ignored". I think this is confusing ... because it sounds like external variables should be internal ... How about:
"Variable 'blah' in component 'blahblah' cannot be marked as an external variable because it doesn't belong to this model." Or perhaps be worth noting that the component is outside the model instead, since that's the real problem?

src/analyser.cpp
```diff
@@ -1349,129 +1369,303 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
...
+                    issue->setReferenceRule(Issue::ReferenceRule::UNSPECIFIED);
+                    issue->setVariable(externalVariable->variable());
+
+                    mAnalyser->addIssue(issue);
+                } else {
+                    auto internalVariable = AnalyserImpl::internalVariable(externalVariable->variable());
+
+                    primaryExternalVariables[internalVariable->mVariable].push_back(externalVariable->variable());
```
kerimoyle: Earlier (sorry, can't recall exactly where, but I left a comment ... ) you test that the constructor has returned a non-null pointer. Does that need to be done here too?

src/analyser.cpp
```diff
@@ -1349,129 +1369,303 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
...
+                                                    primaryExternalVariable.first)
+                                          != primaryExternalVariable.second.end();
+
+                if (isVoi || (equivalentVariableCount > 1) || !hasPrimaryVariable) {
+                    description += (equivalentVariableCount == 2) ? "Both " : "";
+
+                    for (size_t i = 0; i < equivalentVariableCount; ++i) {
+                        if (i != 0) {
+                            description += (i != equivalentVariableCount - 1) ? ", " : " and ";
+                        }
```
kerimoyle, line 1450: I really dislike loops that use conditionals on the index.  It's faster to just limit the loop to what you know (ie: ditch the conditional altogether), and have the start and finish lines either side.

src/analyser.cpp
```diff
@@ -1349,129 +1369,303 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
...
-                                              internalVariable->mInitialisingVariable,
-                                              internalVariable->mVariable,
-                                              equation);
+                    auto issue = Issue::create();
 
-            variableMappings[equation] = stateOrVariable;
+                    issue->setDescription(description);
+                    issue->setLevel(Issue::Level::MESSAGE);
+                    issue->setReferenceRule(Issue::ReferenceRule::UNSPECIFIED);
+                    issue->setVariable(primaryExternalVariable.first);
```
kerimoyle, line 1501: Should be setting new reference rules for all of these as you find them?  That way we can link it to a URL for the user ...

tests/analyser/analyser.cpp
```diff
@@ -389,24 +434,553 @@ TEST(Analyser, unsuitablyConstrained)
...
+    auto externalVariable = libcellml::AnalyserExternalVariable::create(model->component("membrane")->variable("V"));
+
+    EXPECT_FALSE(analyser->removeExternalVariable(externalVariable));
+
+    analyser->addExternalVariable(externalVariable);
+
+    EXPECT_TRUE(analyser->removeExternalVariable(externalVariable));
+    EXPECT_FALSE(analyser->removeExternalVariable(nullptr));
+}
+
```
kerimoyle, line 566: Do you need a test for where the external variable is deleted after it's been added to the analyser?

tests/analyser/analyser.cpp
```diff
@@ -389,24 +434,553 @@ TEST(Analyser, unsuitablyConstrained)
...
+TEST(Analyser, oneNonPrimaryVoiExternalVariable)
+{
+    auto parser = libcellml::Parser::create();
+    auto model = parser->parseModel(fileContents("generator/hodgkin_huxley_squid_axon_model_1952/model.cellml"));
+
+    EXPECT_EQ(size_t(0), parser->issueCount());
+
+    const std::vector<std::string> expectedIssues = {
+        "Variable 'time' in component 'membrane' is marked as an external variable, but it is equivalent to the primary variable of integration which cannot be used as an external variable.",
+    };
```
kerimoyle, line 685: I don't know what the "primary variable of integration" is?

tests/analyser/analyser.cpp
```diff
@@ -389,24 +434,553 @@ TEST(Analyser, unsuitablyConstrained)
...
+TEST(Analyser, oneNonPrimaryExternalVariable)
+{
+    auto parser = libcellml::Parser::create();
+    auto model = parser->parseModel(fileContents("generator/hodgkin_huxley_squid_axon_model_1952/model.cellml"));
+
+    EXPECT_EQ(size_t(0), parser->issueCount());
+
+    const std::vector<std::string> expectedIssues = {
+        "Variable 'V' in component 'sodium_channel' is marked as an external variable, but it is not a primary variable. Variable 'V' in component 'membrane' is its corresponding primary variable and will therefore be the one used as an external variable.",
+    };
```
kerimoyle, line 843: We need some serious documentation around this!  Primary variable ... external variable ... etc etc etc ... so definitely need something other than the `UNSPECIFIED` type for these references!

kerimoyle (dismissed):
Overall stuff:
- @c true and @c false?  or  true and false?
- imperative voice throughout all documentation (it's a mixture everywhere in the bindings, but previously the C++ was consistent) 
- **The** @p index (missing "The" in several places)

In the generatorprofile.h file, I feel like a lot of `@sa` blocks could be helpful where you're referring to another method?  eg in places where you have things like "To be useful, the string should contain the <OPTIONAL_PARAMETER> tag, which will be replaced with a parameter for the compute external variables method, if needed."  (that's line 3268)

Now that we have levels of issues, it would be good to test for things like blank strings passed to the generator profile where something "useful" is a non-blank string?  This means also that it would be good to set the UNSPECIFIED reference rule to something more (ahem) useful ;) so that we can also tie it to proper documentation and user help.

src/analyser.cpp
agarny: What is the difference between "belongs to a different model" and "doesn't belong to this model"? I mean, the current message does mention that the variable is outside the model, so...?

src/analyser.cpp
agarny: Won't do. This would require duplicating code, which is much worse in my view.

src/analyser.cpp
kerimoyle: The difference was earlier in the sentence where it says that the variable cannot be marked as external, but to be honest, neither are great!  Maybe saying that the component isn't a part of the model, so it and its variables cannot be included in any role of any kind?  It's mostly the language around "external variable" that has connotations which are a little confusing when you're also talking about variables outside the model.

src/analyser.cpp
kerimoyle: It's functionally slower ... for no good reason ... :(

agarny: For one very good reason: not duplicating code.

src/analyser.cpp
agarny: Done in commit 22512f861. However, will need to have a URL for those at some point.

src/analyser.cpp
agarny: I am not sure we want to say that the variables of a model "cannot be included in any role of any kind". This is too vague to me. This PR introduces the concept of "external variable" and the message is related to it. So, we should make use of it. Now, if we think that the name of the concept is too confusing then we should change it.

src/analyser.cpp
agarny: Which constructor are we talking about here?

src/analyserexternalvariable.cpp
agarny: Are we talking about the `AnalyserExternalVariable` class? If so, where is it used to create an instance of `AnalyserExternalVariable` without specifying an argument?

src/analyser.cpp
kerimoyle: It's the variables **outside** the model that can't have any effect on it?

src/analyser.cpp
kerimoyle: Urgh.  A comment.  Somewhere ... :|

src/analyserexternalvariable.cpp
kerimoyle: That's my point.  That even though it might not be used, we have a different form of construction than elsewhere?  eg: we have to create a variable and then add it to the component, as opposed to creating a variable with a component as an argument?  It's just not the same as we've done it elsewhere?

src/analyser.cpp
kerimoyle: Yeah, that's fine.  I have a placeholder all ready and waiting for your Generator and Analyser documentation 😄

src/analyser.cpp
agarny: Just in case, the situation is as follows: we have two models: `m1` and `m2`. Then, we want to analyse `m1` and before doing the analysis itself, we want to let the analyser know that there is an external variable. Such an external variable must belong to `m1`. However, if we try to specify an external variable for `m1` using a variable from `m2` then we will get the above message, as illustrated in the `addExternalVariableFromDifferentModels` test.

src/analyser.cpp
agarny: Oh god, "pressure"... :)

src/analyserexternalvariable.cpp
agarny: It's quite different. An `AnalyserExternalVariable` needs to be linked to a `Variable`, but a `Variable` doesn't need to know (and doesn't care) about the `AnalyserExternalVariable` that is linked to it. However, an `Analyser` needs to know about `AnalyserExternalVariable`s. This is the reason we have `Analyser::addExternalVariable()` in the same way that we have `Component::addVariable()`.

tests/analyser/analyser.cpp
agarny: Done in commit 22512f861 and documentation will indeed need to be added.

tests/analyser/analyser.cpp
agarny: Done in commit 748b17092.

src/api/libcellml/analyser.h
agarny: Done in commit 569486bdf.

src/api/libcellml/analyser.h
agarny: Done in commit 569486bdf.

src/api/libcellml/analyserexternalvariable.h
agarny: Done in commit 569486bdf.

src/api/libcellml/analyserexternalvariable.h
agarny: I would leave it as-is. Indeed, if you check `internaltypes.h`, we have `NameList` which is an `std::vector` (as is `VariablePtrs` which really ought to be renamed to `VariableList`, I think) and `IdList` which is an `std::unordered_set`.

src/analyserexternalvariable.cpp
agarny: Two things:
 - `AnalyserExternalVariable::variable()` is not a constructor. It's defined just above as:
```
VariablePtr AnalyserExternalVariable::variable() const
{
    return mPimpl->mVariable.lock();
}
```
   So, here, we are protecting ourselves against the case where `mVariable` is not available anymore.
 - We have 100% coverage, but we have an `A && B && C && D` test, so `return false` only occurs when `A`, `B`, `C` or `D` is false. Coverage is clearly not smart enough to cover all possible cases.As it happens, if we had the following code before that test:
```
if (pimplVariable == nullptr) {
    std::cout << ">>> pimplVariable is null!" << std::endl;
}
```
   We never get that message printed out... since we don't have a test that removes a variable before adding it (or one of its equivalent variables) as an external variable. I thought I would add such a test:
```
TEST(AnalyserExternalVariable, addRemovedVariable)
{
    auto model = parser->parseModel(fileContents("generator/hodgkin_huxley_squid_axon_model_1952/model.cellml"));

    auto component = model->component("membrane");
    auto variable = component->variable("V");
    auto externalVariable = libcellml::AnalyserExternalVariable::create(variable);
    auto dependency = component->variable("Cm");

    component->removeVariable(variable);

    externalVariable->addDependency(dependency);

    EXPECT_EQ(variable, externalVariable->variable());
}
```
   but the message still doesn't get printed out. Indeed,although `variable` is removed, it's still being used twice somewhere else in libCellML...!? So, it looks like we should be using a weak `Variable` pointer in some places...

So, in summary, I feel like leaving things as they stand, just to be on the safe side.

src/analyserexternalvariable.cpp
agarny: Fair point, in fact finding the dependency also ought to be faster, at least most of the time. So, I have replaced the test as follows:
```c++
if ((pimplVariable != nullptr)
    && (owningModel(variable) == owningModel(pimplVariable))
    && (mPimpl->findDependency(variable) == mPimpl->mDependencies.end())
    && !isSameOrEquivalentVariable(variable, pimplVariable)) {
```
Done in commit 76e17c214.

src/analysermodel.cpp
agarny: For the same reason that I have done it for `AnalyserModel::voi()`, `AnalyserModel::stateCount()`, etc. Indeed, if an analyser model is invalid then I would expect `AnalyserModel::hasExternalVariables()` to return `false` (and something similar for the other methods).

src/api/libcellml/analyser.h
agarny: Done in commit 719641fcb.

src/api/libcellml/analyser.h
agarny: Done in commit 46d341cf4.

src/api/libcellml/analyser.h
agarny: Done in commit 46d341cf4.

src/api/libcellml/analyserexternalvariable.h
agarny: Done in commit 46d341cf4.

src/api/libcellml/analyserexternalvariable.h
agarny: Fair point. Done in commit 5282aa749.

src/api/libcellml/analyserexternalvariable.h
agarny: Definitely! Done in commit 42a7d04da.

src/api/libcellml/analyserexternalvariable.h
agarny: Done in commit 36459b803.

src/api/libcellml/analysermodel.h
agarny: Done in commit 3b6d2d0aa.

src/bindings/interface/generatorprofile.i
agarny: Argh, done in commit 74c8ab86b.

src/bindings/interface/analyserexternalvariable.i
agarny: ?? They are effectively the same.

src/analyserexternalvariable.cpp
agarny: Considering this conversation resolved, but feel free to unresolve it, if needed.

src/api/libcellml/analyserexternalvariable.h
agarny: Hopefully improved in commit 1fe12af62.

src/api/libcellml/analyserexternalvariable.h
agarny: Done in commit e6ee73d6d.

src/api/libcellml/analyserexternalvariable.h
agarny: It has a list of dependencies, so your second option. Hopefully, commit c1fd54b0f updates the API documentation accordingly.

src/api/libcellml/analyserexternalvariable.h
agarny: Oopsie! :) Fixed in commit c1fd54b0f.

src/api/libcellml/analyserexternalvariable.h
agarny: Done in commit 79e70ef32.

src/analyser.cpp
agarny: Considering this conversation resolved, but feel free to unresolve it, if needed.

src/api/libcellml/generatorprofile.h
agarny: This is something that you, @hsorby and I quickly discussed at some point. You can currently configure a profile however you like, which means that it can result in generated code that is nonsensical, and there is nothing we can do about it.

However, as much as possible, our `Generator` will check whether something is indeed an empty string and, if so, will skip some code generation. This is what is done in `Generator::GeneratorImpl::addVariableTypeObjectCode()` for the data structure for the variable type object. IOW, we try to limit nonsensical code generation, but there is just so much we can do.

src/bindings/interface/analyserexternalvariable.i
agarny: Indeed, I must have been distracted and didn't finish the description! Fixed in commit d1660bf58.

src/analyser.cpp
agarny: Ok, I now know what you are talking about and, yes, it would be safer to test it here too (and somewhere else too). Done in commit 46ee71eba.

tests/analyser/analyser.cpp
agarny: Do you mean having the following:
```
EXPECT_TRUE(analyser->removeExternalVariable(externalVariable));
EXPECT_FALSE(analyser->removeExternalVariable(externalVariable));
EXPECT_FALSE(analyser->removeExternalVariable(nullptr));
```
rather than the above? If so, done in commit 1416e92ce.

src/api/libcellml/analyser.h
agarny: It wasn't explained, but you in fact spotted a mistake in the documentation. We check against the primary equivalent variable during the analysis, not when adding an external variable. So, I have updated the description to remove "or its primary equivalent". I have also updated the documentation for `AnalyserVariable::variable()` since it is the method that returns what I call the primary variable. (So, primary equivalent variable is a primary variable that is equivalent to something.) All done in commit a9731860a.

agarny:
I believe this is now ready to be re-reviewed, @kerimoyle, @hsorby and @nickerso.

src/analyserexternalvariable.cpp
kerimoyle: @agarny I left a comment somewhere asking about testing for constructor success ... this is the place I meant.  Since it's not a constructor (my bad) that other comment can be disregarded, but I still don't understand how we have coverage for something that you can't ... well ... cover?

agarny: As far as I can tell, it's a limitation of the coverage tools we are using. Remember, we have an `if` statement that goes something like:
```
if (A && B && C && D) {
    // Do something...
    return true;
}
return false;
```
with `A` being `(pimplVariable != nullptr)`. To properly test the case where `A` is `false`, we would need to rewrite the `if` statement to read something like:
```
if (A) {
    if (B) {
        if (C) {
            if (D) {
                // Do something...
                return true;
            } else {
                return false;
            }
        } else {
            return false;
        }
    } else {
        return false;
    }
} else {
    return false;
}
```
I don't think we want to go that route, not least because I tried to test the case where `A` is `false` and it failed due to our not using weak pointers in all possible places.

kerimoyle: Hmmm ... maybe our coverage tools are different.  Mine gives me an error if I don't have all parts genuinely covered.  If I had your first statement, and `A` was always false, I'd get a coverage error.

agarny: It's not 100% clear to me what you have tried, so can't really comment.

agarny:
Ok, we are back on track. IOW, this PR is now ready for (re-)reviewing. Ping @hsorby, @kerimoyle and @nickerso.

src/analyser.cpp
```diff
@@ -1349,129 +1373,322 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
...
-        // Make our internal variables available through our API.
+        if (!mExternalVariables.empty()) {
+            // Check whether an external variable belongs to the model being
+            // analysed, or whether it is marked as an external variable more
+            // than once through equivalence or is (equivalent to) the variable
+            // of integration.
 
-        auto stateIndex = MAX_SIZE_T;
-        auto variableIndex = MAX_SIZE_T;
+            std::map<VariablePtr, std::vector<VariablePtr>> primaryExternalVariables;
```
kerimoyle, line 1397: Should this be a declared type instead of a combo of std containers?

tests/test_utils.cpp
```diff
@@ -52,7 +52,7 @@ std::chrono::steady_clock::time_point timeNow()
 
 int elapsedTime(const std::chrono::steady_clock::time_point &startTime)
 {
-    return std::chrono::duration_cast<std::chrono::milliseconds>(timeNow() - startTime).count();
+    return int(std::chrono::duration_cast<std::chrono::milliseconds>(timeNow() - startTime).count());
 }
```
kerimoyle: This is in another PR too? #761 ?

agarny: Yes, I merged PR #761 into this one.

src/analyser.cpp
agarny: This is an internal type which is only used here. So, I wouldn't bother with creating a type for it.

agarny:
> Overall stuff:
> 
> * @c true and @c false?  or  true and false?

These were previously addressed in this PR.

> * imperative voice throughout all documentation (it's a mixture everywhere in the bindings, but previously the C++ was consistent)

(Further?) addressed in commit 4a4c87d.

> * **The** @p index (missing "The" in several places)

(Further?) addressed in commit c53771c.

> In the generatorprofile.h file, I feel like a lot of `@sa` blocks could be helpful where you're referring to another method? eg in places where you have things like "To be useful, the string should contain the <OPTIONAL_PARAMETER> tag, which will be replaced with a parameter for the compute external variables method, if needed." (that's line 3268)

I have added some `@sa`s here and there in commit 2569f04.

> Now that we have levels of issues, it would be good to test for things like blank strings passed to the generator profile where something "useful" is a non-blank string? This means also that it would be good to set the UNSPECIFIED reference rule to something more (ahem) useful ;) so that we can also tie it to proper documentation and user help.

Hmm, maybe, but this is beyond the scope of this PR. Feel free to create a new issue for it, if you want.

agarny:
Ping @hsorby, @kerimoyle and @nickerso: good to go.

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.computed.constant.c
```diff
@@ -0,0 +1,108 @@
+/* The content of this file was generated using a modified C profile of libCellML 0.2.0. */
+
+#include "model.computed.constant.h"
+
+#include <math.h>
+#include <stdlib.h>
+
+const char VERSION[] = "0.1.0.post0";
```
hsorby: I think this version number here should be `0.2.0` because you have changed the content.

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.computed.constant.py
```diff
@@ -0,0 +1,116 @@
+# The content of this file was generated using the Python profile of libCellML 0.2.0.
+
+from enum import Enum
+from math import *
+
+
+__version__ = "0.1.0"
```
hsorby: This is a different version number to the `c` and still the same version number as the previous version.

hsorby (changes requested):
I don't think version numbering for the generated code is correct the Python and C versions are different and I think both should now be `0.2.0` with the proposed changes.

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.computed.constant.c
agarny: Done in commit 930df02.

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.computed.constant.py
agarny: Done in commit 930df02.

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.computed.constant.c
agarny: Just in case, the `.post0` is because the C profile was modified to have a sensical interface file name.

hsorby: I'm not sure I understand the reason behind the `.post0` it also appears on some c files and not in others.  I can't see why or how the `.post0` is used or required for a file name.

agarny: That's because the (original C) profile was modified. For `tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.computed.constant.c `, it's because we called `GeneratorProfile::setInterfaceFileNameString()` in the `hodgkinHuxleySquidAxonModel1952WithComputedConstantAsExternalVariable ` test in `tests/analyser/analyser.cpp`:
```c++
...

auto profile = generator->profile();

profile->setInterfaceFileNameString("model.computed.constant.h");

EXPECT_EQ(fileContents("generator/hodgkin_huxley_squid_axon_model_1952/model.computed.constant.h"), generator->interfaceCode());
EXPECT_EQ(fileContents("generator/hodgkin_huxley_squid_axon_model_1952/model.computed.constant.c"), generator->implementationCode());

...
```

agarny: We could certainly remove `profileContents += mLockedProfile->interfaceFileNameString();` in `Generator::GeneratorImpl::modifiedProfile()`, if we wanted to.

hsorby: Yes I think that would be best. For me the version is defined by having the name of the file there but not what the actual name is.

agarny: Done in commit d25ac7e.

hsorby (approved):
I'm approving.

hsorby:
I have approved this but I think once this has gone in I would like to see a bit more work for testing if the profile has been changed or not.  That will be a new issue I would have once this is merged.

agarny:
> I have approved this but I think once this has gone in I would like to see a bit more work for testing if the profile has been changed or not. That will be a new issue I would have once this is merged.

Fair enough, it's true that, at this stage, anything that we change in the profile will result in the profile being considered modified. However, there are things that we might not include (like the interface file name).

kerimoyle (approved):

Issue #652, opened 2020-07-07
Parser/Printer can read a valid model and print an invalid one: ids are duplicated.


kerimoyle:
See test below.  I'm worried about how we store the `ImportSource` items, as we currently duplicate them across their children.  This means that when you have a unique id in the parsed CellML, we print that id multiple times later on ...

```
TEST(Parser, parsePrintMultipleChildrenOfImports)
{
    std::string in = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
                     "<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"everything\" id=\"model_1\">\n"
                     "  <import xmlns:xlink=\"http://www.w3.org/1999/xlink\" xlink:href=\"some-other-model.xml\" id=\"import_1\">\n"
                     "    <component component_ref=\"a_component_in_that_model\" name=\"component1\" id=\"component_1\"/>\n"
                     "    <component component_ref=\"another_component_in_that_model\" name=\"component2\" id=\"component_2\"/>\n"
                     "    <units units_ref=\"a_units_in_that_model\" name=\"units1\" id=\"units_1\"/>\n"
                     "    <units units_ref=\"another_units_in_that_model\" name=\"units2\" id=\"units_2\"/>\n"
                     "    <units units_ref=\"yet_another_units_in_that_model\" name=\"units3\" id=\"units_3\"/>\n"
                     "  </import>\n"
                     "</model>";
    auto parser = libcellml::Parser::create();
    auto model = parser->parseModel(in);
    auto printer = libcellml::Printer::create();
    EXPECT_EQ(in, printer->printModel(model));
}
```

What's printed is this ... so the components share the import, but the units do not (neither with one another, nor with the components):
```
<?xml version="1.0" encoding="UTF-8"?>
<model xmlns="http://www.cellml.org/cellml/2.0#" name="everything" id="model_1">
  <import xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="some-other-model.xml" id="import_1">
    <component component_ref="a_component_in_that_model" name="component1" id="component_1"/>
    <component component_ref="another_component_in_that_model" name="component2" id="component_2"/>
  </import>
  <import xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="some-other-model.xml" id="import_1">
    <units units_ref="a_units_in_that_model" name="units1" id="units_1"/>
  </import>
[...]

Issue #618, opened 2020-06-08
Fix all docstrings


kerimoyle:
Just a general note that I've noticed a function description that's not quite true, and there are probably others ...
(a `UnitsPtr` is not the name, it's the pointer ...)
```
    /**
     * @brief Get the name of the units for this variable.
     *
     * Get the name of the units for this variable. If no units are set
     * an empty @c std::string is returned.
     *
     * @sa setUnits
     * @sa removeUnits
     *
     * @return The @c UnitsPtr of the units for this variable.
     */
    UnitsPtr units() const;
```

kerimoyle:
We also need to make a schema for docstrings and follow it:
- when referencing a parameter, we use the @p prefix and the actual variable name (not the description of the parameter)?
- when referencing an object that has a pointer, do we use the pointer type or the pointed-to type? ie: when passing a model to the printer, should it be discussed as a  @c ModelPtr or a  @c Model  or just a model?
- when just talking about the conceptual thing rather than the code-representation, should we be using any kind of markup at all? 
- how do we talk about null pointers?
- how do we talk about the range of vectors?

PR #608, opened 2020-05-28
i530 Create Importer class to enable import error logging

kerimoyle:
Addresses #530 and replaces #531 

- [x] Move existing tests into `Importer`
- [x] Add tests for import error/warning logging: circular imports, files not found, etc
- [x] Move `flatten()` functionality out of the `Model` class and into the `Importer`
- [x] Move `resolveImports` out of `Model` and into the `Importer` 
- [x] Multiple model resolution without duplicating imports
- [x] Manually add local model instances to import library for resolving against
- [x] Interrogate model for list of dependencies, and return requirements to user.
- [x] Clean up `ModelPtr::hasUnresolvedImports` functions ... redundancies there at the moment
- [x] Find better way for reporting cyclic imports.
- [x] Fix bug #644

PR #653, opened 2020-07-07
ImportSource changes

kerimoyle:
Demonstrates #652


kerimoyle:
This is a work in progress.  I've changed:
- the Model keeps track of all ImportSources that it needs,
- the ImportSources keep track of the things which depend on them,
- the Printer bug has been fixed so that all imports which share a source are grouped together.  

I need to know whether there's a need to separate imports for any reason, and what a mechanism for that could be.  At present, if they have the same url but different ids then they're not considered the same (so that annotation is preserved).  Is there more that should be done?

All the unit tests pass, but I've changed two that required printing imports separately.  They'll fail here as I haven't done the bindings/coverage yet.  Thoughts/advice welcome ...

tests/model/component_import.cpp
```diff
@@ -201,14 +188,14 @@ TEST(ComponentImport, multipleImportAndParse)
 
     libcellml::PrinterPtr printer = libcellml::Printer::create();
     std::string a = printer->printModel(m);
-    EXPECT_TRUE((e1 == a) || (e2 == a));
+    EXPECT_EQ(e, a);
```
nickerso, line 194: This is wrong, `c3` should be a in a separate import to `c1` and `c2`.

tests/model/component_import.cpp
```diff
@@ -201,14 +188,14 @@ TEST(ComponentImport, multipleImportAndParse)
...
 
     // Parse
     libcellml::ParserPtr parser = libcellml::Parser::create();
-    libcellml::ModelPtr model = parser->parseModel(e2);
+    libcellml::ModelPtr model = parser->parseModel(e);
     EXPECT_EQ(size_t(3), model->componentCount());
     a = printer->printModel(model);
-    EXPECT_TRUE((e1 == a) || (e2 == a));
+    EXPECT_EQ(e, a);
```
nickerso, line 201: This is correct, as you get back the same model as parsed, but this is a different model to the one programmatically constructed above...

tests/model/component_import.cpp
kerimoyle: Because it was created with separate ImportSource pointers? Or is there another way to tell?

nickerso: yep, that is the only way to tell I think

agarny: Agreed and that means that we will somehow need to keep track of that fact when parsing a CellML file.

kerimoyle: OK, I'll take out the check-for-duplicate-url bit, easy enough.

kerimoyle: Done in fca5344

tests/model/component_import.cpp
kerimoyle: Done in fca5344

kerimoyle:
I've reinstated the storage of duplicate import sources and changed the tests as appropriate.  Note that this means we can no longer retrieve an import by its `url` as it's not unique any more.  My thoughts at the moment are:
- In this PR we're keeping imported items separate, which seems to be is in opposition to what the Importer is doing in https://github.com/cellml/libcellml/pull/608? Should the importer also be importing things multiple times too? 
- I'd like to provide a convenience method so that if people have created multiple duplicates of imports, but don't actually need them, that the imports can be collapsed.  Since we've previously (erroneously) been printing individual units imports, and making them separate from all component imports, I'm guessing that the PMR is full of unintended duplicates ... Any thoughts?

(ps - I've left in lots of commented sections, will remove when I'm sure they're for the bin)

Edit: Other thoughts:
- When adding/removing a component/units item from an import source, should the import source be also added/removed from that component/units item in return? ie: a two-way relationship should be maintained?  The downstream effects are that removing a component from an import source will set it to return `isImport()` false until a replacement import source is added.

kerimoyle:
Ignore this for now ... I'm rethinking how it should be done.  Thoughts on users' required/expected functionality welcome though :)

kerimoyle:
Hi @agarny @hsorby @nickerso As far as I can tell (without the CI build working properly) this is ready for review please.  It builds, passes tests and coverage on my machine.
Things to note:
- Where separate `<import>` blocks are parsed or `ImportSource` objects added using the API which use the same source URL, these are kept separate.  I will add a `consolidate()` or similar function to the `Importer` class as there will be loads of models out there which, because of the printer bug, have unintended separate import sources.
- Where different import children (components/units) exist in a single `ImportSource`, these are kept together (fixing the bug that this PR started from).
- When a mixture of components and units are in the same import source, they are printed in order of components first and then units.  
- An empty import source is printed (because it's valid according to https://cellml-specification.readthedocs.io/en/latest/reference/formal_and_informative/specB02.html )

kerimoyle:
Booo ... the CI is building based on an outdated commit again :( @hsorby

kerimoyle:
I've now included a `model->consolidateImports()` function to this PR.  The motivation for that is the bug in the `Printer` will have created a multitude of separate import sources throughout the PMR, and I wanted to give users an easy way to rationalise them.  I've played with the idea of moving this into the `Importer` class, but since it doesn't actually need anything to _be_ imported, I decided to leave it in the `Model` class.  Happy to move it if there's objection, but that would require this PR and #608 to be merged first.

nickerso:
> I've now included a `model->consolidateImports()` function to this PR. 

This doesn't seem like useful API and I'd suggest not including it.

kerimoyle:
> This doesn't seem like useful API and I'd suggest not including it.

Over time the number of import groups will have multiplied because of the printer bug ... this is just a way to let people fix that (easily) (if they want to).  I don't really understand the objection @nickerso? Too many bells and whistles or something else?

nickerso:
The only possible use case for this API is to "fix" a bug in code that realistically no one is using to produce CellML models. There is no way to implement a "consolidate imports" method that will avoid future confusion and introduction of buggy models as it seems like a perfectly reasonable method to call that will have loads of unintended consequences.

kerimoyle:
> The only possible use case for this API is to "fix" a bug in code that realistically no one is using to produce CellML models. 

Fair enough, I can change it back.

> There is no way to implement a "consolidate imports" method that will avoid future confusion and introduction of buggy models as it seems like a perfectly reasonable method to call that will have loads of unintended consequences.

... I don't follow this bit though? They only get merged when the URL is identical, so there should be no effect on the meaning of the instantiated model.  I guess it doesn't matter since I'll remove it anyway, but I like to understand what's going on (where possible ...) :)

nickerso:
> I don't follow this bit though? They only get merged when the URL is identical, so there should be no effect on the meaning of the instantiated model. I guess it doesn't matter since I'll remove it anyway, but I like to understand what's going on (where possible ...) :)

Two components in separate imports with the same URL is different to one import with two components in it. These are the types of differences that I don't think we are really testing at present.

kerimoyle:
> Two components in separate imports with the same URL is different to one import with two components in it. 

Wow, ok, I didn't know that!  How is it different?  Because we treat them exactly the same?  And we print them out together? And ... etc??  The only difference I can imagine is having different id attributes, but that doesn't fundamentally affect the semantics of the model itself?

src/importedentity.cpp
```diff
@@ -18,6 +18,8 @@ limitations under the License.
 
 namespace libcellml {
 
+// using ImportSourceWeakPtr = std::weak_ptr<ImportSource>;
```
nickerso: this can be removed?

src/importedentity.cpp
```diff
@@ -43,11 +45,13 @@ ImportedEntity::~ImportedEntity()
 
 bool ImportedEntity::isImport() const
 {
+    // return !mPimpl->mImportSource.expired();
```
nickerso: random question, but these changes are due to the new use of `shared_from_this` for the import sources...or something, right?

src/parser.cpp
```diff
@@ -126,13 +126,13 @@ struct Parser::ParserImpl
      * components or units to the @p model. If any attributes exist in @p import source
      * with names matching those in @p node, they will be overwritten. Likewise,
      * any imported components or units of the same name already in @p model will
-     * be overwritten by those parsed from @p node.
+     * be overwritten by those parsed from @p node.cmake
```
nickerso: `cmake` ???

tests/import_source/import_source.cpp
```diff
@@ -0,0 +1,400 @@
...
+    for (size_t index = 0; index < m1->componentCount(); ++index) {
+        auto c1 = m1->component(index);
+        auto c2 = m2->component(index);
+        compareComponent(c1, c2, m2);
+    }
+}
+
+TEST(ImportSource, createImportSource)
+{
+    auto imp1 = libcellml::ImportSource::create();
```
nickerso, line 25: shouldn't there be an `EXPECT...` something here?

tests/model/component_import.cpp
```diff
@@ -201,14 +191,14 @@ TEST(ComponentImport, multipleImportAndParse)
 
     libcellml::PrinterPtr printer = libcellml::Printer::create();
     std::string a = printer->printModel(m);
-    EXPECT_TRUE((e1 == a) || (e2 == a));
+    EXPECT_EQ(e, a);
```
nickerso, line 194: just to check my understanding, because we now track imports in the models there is no longer any ambiguity in the order in which these get printed and hence no need to check both ordering options?

tests/printer/printer.cpp
```diff
@@ -295,3 +295,52 @@ TEST(Printer, printModelWithStandardUnitsAdded)
...
+                     "</model>\n";
+
+    auto parser = libcellml::Parser::create();
+    auto model = parser->parseModel(in);
+    auto printer = libcellml::Printer::create();
+    EXPECT_EQ(in, printer->printModel(model));
+}
+
+TEST(Printer, printMultipleChildrenOfImportsMixedOrder)
+{
```
nickerso, line 641: interesting test :)

nickerso (dismissed):
A few minor things, but this is looking good to me.

src/importedentity.cpp
kerimoyle: Probably :)

src/importedentity.cpp
kerimoyle: Yes.  We now have a strong connection between `ImportedEntity` items and their `ImportSource`, and a weak connection between the `Model` and the `ImportSource`s it contains.

tests/model/component_import.cpp
kerimoyle: Yep, that's it.

tests/printer/printer.cpp
kerimoyle: I ran out of interestingness ... :)

src/parser.cpp
kerimoyle: Ooops! Typed in the wrong window I guess ... 6e455f8c

tests/import_source/import_source.cpp
kerimoyle: Yes 6e455f8c

src/importedentity.cpp
kerimoyle: I've removed the comment in 6e455f8c

src/importedentity.cpp
kerimoyle: 6e455f8c

nickerso (dismissed):

kerimoyle:
Conflicts fixed.  Ready for review please @agarny @nickerso @hsorby

src/api/libcellml/component.h
```diff
@@ -414,6 +416,16 @@ class LIBCELLML_EXPORT Component: public ComponentEntity, public ImportedEntity
      */
     ComponentPtr clone() const;
 
+    /**
+     * @brief Set the import source of this component.
+     *
+     * If this component is already located in a Model instance, then the
```
agarny, line 422: Should be "... @c Model instance..." Same on the line below.

src/api/libcellml/importsource.h
```diff
@@ -109,6 +117,161 @@ class LIBCELLML_EXPORT ImportSource: public Entity
      */
     ImportSourcePtr clone() const;
 
+    /**
+     * @brief Add a component to this import source item.
+     *
+     * The added @p component will be treated as an import, and
+     * will be sourced from the combination of url of this import source
```
agarny: "URL", not "url".

src/api/libcellml/importsource.h
```diff
@@ -109,6 +117,161 @@ class LIBCELLML_EXPORT ImportSource: public Entity
...
+     * item, and the component's own import reference.
+     *
+     * @param component The @c ComponentPtr to add.
```
agarny, line 127: "Component", not "ComponentPtr".

src/api/libcellml/component.h
kerimoyle: This is inconsistent everywhere ... and should be `@c ModelPtr` or `@c Model` ??

src/api/libcellml/importsource.h
```diff
@@ -109,6 +117,161 @@ class LIBCELLML_EXPORT ImportSource: public Entity
...
+     * @return @c true if the component was added successfully, @c false
+     * if it already exists in this import.
+     */
+    bool addComponent(const ComponentPtr &component);
+
+    /**
+     * @brief Remove the component at the given @p index.
+     *
+     * Remove the component from this import source at the given @p index.
+     * @p index must be in the range [0, \#component).
```
agarny: "components", not "component" (I believe this is what we do in other places, i.e. use plural).

src/api/libcellml/importsource.h
```diff
@@ -109,6 +117,161 @@ class LIBCELLML_EXPORT ImportSource: public Entity
...
+     *
+     * @param index The index of the component to remove.
+     * @param setEmpty Whether or not to set the removed Component's
```
agarny, line 141: "@c Component"

src/api/libcellml/importsource.h
kerimoyle: ... but the parameter isn't a `Component`, it's a `ComponentPtr`?

src/api/libcellml/importsource.h
```diff
@@ -109,6 +117,161 @@ class LIBCELLML_EXPORT ImportSource: public Entity
...
+     *        importSource to nullptr.  True by default.
```
agarny, line 142: We should something have "`nullptr`" rather than "nullptr".

src/api/libcellml/importsource.h
agarny: Same elsewhere...

agarny: Actually, the documentation is not consistent. See the one for `removeComponent()`.

src/api/libcellml/importsource.h
agarny: Same elsewhere.

src/api/libcellml/importsource.h
agarny: Same elsewhere.

src/api/libcellml/importsource.h
```diff
@@ -109,6 +117,161 @@ class LIBCELLML_EXPORT ImportSource: public Entity
...
+     *
+     * @return True if the component was removed, false otherwise.
+     */
+    bool removeComponent(ComponentPtr &component, bool setEmpty = true);
+
+    /**
+     * @brief Remove all components from the import source.
+     *
+     * Note that components which are removed from the import will have their
```
agarny, line 166: Need a "proper" description of the method before that note.

src/api/libcellml/importsource.h
```diff
@@ -109,6 +117,161 @@ class LIBCELLML_EXPORT ImportSource: public Entity
...
+     * import source cleared.  They thus become locally
+     * concrete instances in the model, and will return @c false to the
+     * function isImport().
```
agarny, line 169: Need to properly reference `isImport()`.

src/api/libcellml/importsource.h
kerimoyle: I don't understand? Doesn't the brief comment say it?

src/api/libcellml/importsource.h
kerimoyle: I don't understand?

src/api/libcellml/importsource.h
agarny: I.e. "@c nullptr" rather than "nullptr".

src/api/libcellml/importsource.h
agarny: Same elsewhere.

src/api/libcellml/importsource.h
```diff
@@ -109,6 +117,161 @@ class LIBCELLML_EXPORT ImportSource: public Entity
...
+    bool addUnits(const UnitsPtr &units);
+
+    /**
+     * @brief Remove the units at the given @p index.
+     *
+     * Remove the units from this import source at the given @p index.
+     * @p index must be in the range [0, \#units).
+     *
+     * @param index The index of the units to remove.
+     * @param setEmpty Whether or not to set the removed Units'
```
agarny, line 219: "@c Units"

src/api/libcellml/importsource.h
agarny: Same for `removeAllUnits()`.

src/api/libcellml/model.h
```diff
@@ -366,6 +366,89 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
      */
     bool fixVariableInterfaces();
 
+    /**
+     * @brief Add an import source item to this model.
+     *
+     * Add import source by reference to the model. If the import
```
agarny, line 372: Would replace with "Add the @p importSource to the model" and maybe rephrase the rest of that paragprah accordingly.

src/api/libcellml/model.h
```diff
@@ -366,6 +366,89 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
+     * source item was previously in a different model, it is moved
+     * to this one, and the previous model's import source list is updated.
+     *
+     * The function will return @c false and no action is taken if:
+     *  - The @p imp pointer already exists in this model; or
+     *  - The @p imp is @c nullptr.
```
agarny: Replace "imp" with "importSource".

agarny: Same below.

src/api/libcellml/model.h
```diff
@@ -366,6 +366,89 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
+     *
+     * @param imp The import source to add.
+     *
+     * @return @c true if the import source item is added or @c false otherwise.
+     */
+    bool addImportSource(const ImportSourcePtr &imp);
+
+    /**
+     * @brief Get the number of import source items in the model.
```
agarny, line 387: Why "import source items" and not simply "import sources"?

agarny: Same elsewhere.

src/api/libcellml/model.h
kerimoyle: It has a @p for parameter, so needs to be the same as is listed in the declaration?

src/api/libcellml/model.h
```diff
@@ -366,6 +366,89 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
+     * @return The number of import source items.
+     */
+    size_t importSourceCount() const;
+
+    /**
+     * @brief Get the import source item at the given @p index.
+     *
+     * Returns a reference to an import source at the given @p index.  If the @p index
+     * is not valid a @c nullptr is returned, the range of valid values for the
+     * index is [0, \#import sources).
```
agarny: I would replace wih "#importSources".

agarny: Same elsewhere.

src/api/libcellml/model.h
agarny: ... and elsewhere.

src/api/libcellml/model.h
kerimoyle: Because - subtly - they're not the same thing.  We can have one import source, and many import source items.

src/api/libcellml/units.h
```diff
@@ -491,6 +492,16 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
      */
     UnitsPtr clone() const;
 
+    /**
+     * @brief Set the import source of these units.
+     *
+     * If these units are already located in a Model instance, then the
```
agarny, line 498: "@c Model". Same below.

src/bindings/interface/importsource.i
```diff
@@ -31,6 +31,36 @@ unset).";
 %feature("docstring") libcellml::ImportSource::clone
 "Create a copy of this import source.";
 
+%feature("docstring") libcellml::ImportSource::addComponent
+"Add a dependent Component to this ImportSource.";
```
agarny: I would remove "dependent".

agarny: Also, would use "component" not "Component". (This is what we do elsewhere in our interfaces.) Same below.

kerimoyle: Why? It has knock-on effects for the component, so it is a dependent component?

agarny: Same with "ImportSource" which should be replaced with "import source".

src/bindings/interface/importsource.i
```diff
@@ -31,6 +31,36 @@ unset).";
...
+"Return the number of imported Components from this ImportSource.";
+
+%feature("docstring") libcellml::ImportSource::removeAllComponents
+"Remove all of the imported Components from this ImportSource.";
+
+%feature("docstring") libcellml::ImportSource::component
+"Retrieve the imported Component at the given index.";
+
+%feature("docstring") libcellml::ImportSource::addUnits
+"Add a dependent Units item to this ImportSource.";
```
agarny: "units", not "Units". Same below.

src/bindings/interface/model.i
```diff
@@ -76,6 +76,25 @@ determine the full path to the source model relative to this one.";
 %feature("docstring") libcellml::Model::fixVariableInterfaces
 "Fix variable interfaces throughout the model.";
 
+%feature("docstring") libcellml::Model::addImportSource
+"Manually add an ImportSource to the Model.";
```
agarny, line 80: "import source" and "model", not "ImportSource" and "Model". Same below.

src/bindings/interface/model.i
```diff
@@ -76,6 +76,25 @@ determine the full path to the source model relative to this one.";
...
+
+%feature("docstring") libcellml::Model::importSourceCount
+"Return the number of ImportSource items in the Model.";
+
+%feature("docstring") libcellml::Model::importSource
+"Retrieve an ImportSource by index or url.";
```
agarny: Reminder: "URL". Same below.

src/importsource.cpp
```diff
@@ -14,12 +14,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#include "libcellml/importsource.h"
+#include <algorithm> // Adding because Ubuntu build complained?
```
agarny: Remove comment.

src/importsource.cpp
```diff
@@ -14,12 +14,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-#include "libcellml/importsource.h"
```
agarny, line 17: This shouldn't have been removed (this is how we do things now: add the `.cpp`'s corresponding `.h` file first, then system headers then API headers, with an empty line between each category), and an empty line added.

src/importsource.cpp
```diff
@@ -14,12 +14,20 @@ See the License for the specific language governing permissions and
...
+#include <algorithm> // Adding because Ubuntu build complained?
+#include <vector>
 
+#include "libcellml/component.h"
+#include "libcellml/importedentity.h"
+#include "libcellml/importsource.h"
```
agarny: Remove this include (since it should be included at the top).

src/importsource.cpp
```diff
@@ -83,4 +96,211 @@ ImportSourcePtr ImportSource::clone() const
...
+
+    if (it == mPimpl->mImports.end()) {
+        return false;
+    }
+
+    mPimpl->removeItem(it);
+
+    if (setEmpty) {
+        ImportSourcePtr empty = nullptr;
+        component->setImportSource(empty);
```
agarny, line 151: As previously agreed/discussed, this should be replaced with `component->setImportSource(nullptr);`. Same below.

kerimoyle: It's not possible to do this (as I think we did discuss last time!) The compiler complains about the non-const lvalue reference.

src/importsource.cpp
kerimoyle: 81328612

src/importsource.cpp
kerimoyle: 81328612

src/importsource.cpp
kerimoyle: 81328612

src/bindings/interface/model.i
kerimoyle: 81328612

src/api/libcellml/model.h
kerimoyle: I don't understand the reason.  Elsewhere we use lowercase descriptions? It would be better to ditch the `#` altogether as it's weird to use this kind of shorthand here.  Better to spell it out properly, or provide a reference to the `importSourceCount()` function.

src/importsource.cpp
```diff
@@ -83,4 +96,211 @@ ImportSourcePtr ImportSource::clone() const
...
+    }
+    // Loop through units vector and decrease its indices too.
+    i = int(mUnits.size() - 1);
+    while ((i >= 0) && (int(mUnits[size_t(i)]) > index)) {
+        mUnits[size_t(i)]--;
+        i--;
+    }
+    // Remove the item from the imports.
+    mImports.erase(it);
+}
```
agarny: Definitely not keen on all that casting, not least since `size()` returns a `size_t`, so to cast it to an `int` is "risky" although we are very unlikely to ever have more than 2,147,483,647 components/units. Anyway, I would rewrite this method to read:
```
void ImportSource::ImportSourceImpl::removeItem(std::vector<ImportedEntityWeakPtr>::iterator &it)
{
    // Find current index for the item to remove.
    auto index = size_t(std::distance(mImports.begin(), it));
    auto itU = std::find(mUnits.begin(), mUnits.end(), index);
    if (itU != mUnits.end()) {
        mUnits.erase(itU);
    } else {
        auto itC = std::find(mComponents.begin(), mComponents.end(), index);
        if (itC != mComponents.end()) {
            mComponents.erase(itC);
        }
    }

    // Iterate through components index vector and decrement all indices above the item's index.
    static const size_t MAX_SIZE_T = std::numeric_limits<size_t>::max();
    for (size_t i = mComponents.size() - 1; (i != MAX_SIZE_T) && (mComponents[i] > index); --i) {
        --mComponents[i];
    }
    // Loop through units vector and decrease its indices too.
    for (size_t i = mUnits.size() - 1; (i != MAX_SIZE_T) && (mUnits[i] > index); --i) {
        --mUnits[i];
    }
    // Remove the item from the imports.
    mImports.erase(it);
}
```

kerimoyle:
Thanks for the review @agarny.  I've made some of the changes as suggested, but I'm not prepared to spend time doing the docstrings until we've got some kind of agreement with @hsorby and @nickerso about what they should be everywhere.  And even then, there is so much inconsistency throughout I think they should be addressed in a separate PR which fixes up everything properly.

src/parser.cpp
```diff
@@ -1175,6 +1176,8 @@ void Parser::ParserImpl::loadImport(const ImportSourcePtr &importSource, const M
     while (childNode) {
         if (childNode->isCellmlElement("component")) {
             ComponentPtr importedComponent = Component::create();
+            // importedComponent->setImportSource(importSource);
+
```
agarny: Delete.

src/parser.cpp
```diff
@@ -1198,6 +1201,7 @@ void Parser::ParserImpl::loadImport(const ImportSourcePtr &importSource, const M
             }
         } else if (childNode->isCellmlElement("units")) {
             UnitsPtr importedUnits = Units::create();
+            // importedUnits->setImportSource(importSource);
```
agarny: Delete.

src/importsource.cpp
kerimoyle: a8276f79

tests/model/component_import.cpp
```diff
@@ -329,3 +319,62 @@ TEST(ComponentImport, complexImportAndParse)
...
+    EXPECT_EQ(imp2, model->component("importComponent2")->importSource());
+    EXPECT_EQ(imp2, model->component("importComponent4")->importSource());
+
+    // Add an already-present import source:
+    EXPECT_FALSE(model->addImportSource(imp3));
+
+    // Add a new import source with a url that's already in the list:
+    // NB This has changed so that the new import source
+    // object will trigger an additional import block, even though the
+    // url (and therefore the imported object) already exists.
```
agarny: URL

src/parser.cpp
kerimoyle: dc8c5607

src/parser.cpp
kerimoyle: dc8c5607

tests/model/component_import.cpp
kerimoyle: done already

tests/model/units_import.cpp
```diff
@@ -215,3 +215,65 @@ TEST(UnitsImport, importModifyAndParse)
...
+    std::string url1 = "http://www.example.com#hello";
+    imp1->setUrl(url1);
+
+    auto units1 = libcellml::Units::create("importUnits1");
+    model->addUnits(units1);
+    units1->setImportSource(imp1);
+
+    EXPECT_EQ(size_t(1), model->importSourceCount());
+    EXPECT_EQ(imp1, model->importSource(0));
+    // EXPECT_EQ(imp1, model->importSource(url1));
```
agarny: Delete.

agarny: Same below.

tests/import_source/import_source.cpp
```diff
@@ -0,0 +1,401 @@
...
+        auto u2 = m2->units(index);
+        compareUnits(u1, u2, m2);
+    }
+
+    for (size_t index = 0; index < m1->componentCount(); ++index) {
+        auto c1 = m1->component(index);
+        auto c2 = m2->component(index);
+        compareComponent(c1, c2, m2);
+    }
+}
```
agarny: Shouldn't these be in `test_utils`? Might come handy for some other tests at some point?

tests/import_source/import_source.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+    // Add another with the same url:
+    auto imp2 = libcellml::ImportSource::create();
+    imp2->setUrl(url1);
+
+    EXPECT_TRUE(model->addImportSource(imp2));
+    EXPECT_EQ(size_t(2), model->importSourceCount());
+    EXPECT_EQ(imp2, model->importSource(1));
+
+    // Add another with a new url:
```
agarny: URL

tests/import_source/import_source.cpp
```diff
@@ -0,0 +1,401 @@
...
+    EXPECT_EQ(size_t(1), imp->unitsCount());
+    EXPECT_EQ(units1, imp->units(0));
+    EXPECT_EQ(nullptr, imp->units(99));
+
+    // Add the other direction, expect false as it's already there.
+    EXPECT_FALSE(imp->addUnits(units1));
+
+    // Add the other units so it's an import.
+    units2->setImportSource(imp);
+
```
agarny: Delete empty line.

tests/import_source/import_source.cpp
```diff
@@ -0,0 +1,401 @@
...
+    EXPECT_FALSE(component->isImport());
+    EXPECT_FALSE(units->isImport());
+
+    component->setSourceComponent(imp1, "other_component");
+    units->setSourceUnits(imp1, "other_units");
+
+    EXPECT_TRUE(component->isImport());
+    EXPECT_TRUE(units->isImport());
+    EXPECT_EQ(size_t(1), imp1->componentCount());
+    EXPECT_EQ(size_t(1), imp1->componentCount());
```
agarny: Did you mean `EXPECT_EQ(size_t(1), imp1->unitsCount());`?

tests/import_source/import_source.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+    component->setImportSource(imp2);
+    units->setImportSource(imp2);
+    EXPECT_EQ(size_t(0), imp1->componentCount());
+    EXPECT_EQ(size_t(0), imp1->componentCount());
```
agarny: `EXPECT_EQ(size_t(0), imp1->unitsCount());`

tests/import_source/import_source.cpp
```diff
@@ -0,0 +1,401 @@
...
+    EXPECT_EQ(size_t(1), imp2->componentCount());
+    EXPECT_EQ(size_t(1), imp2->componentCount());
```
agarny, line 177: `EXPECT_EQ(size_t(1), imp2->unitsCount());`

agarny (changes requested):
Various minor changes, but otherwise looks good.

src/api/libcellml/component.h
agarny: My personal preference is for `@c Model`.

src/api/libcellml/importsource.h
agarny: Sure, it does, but my understanding is that this section should contain a more detailed description, which is missing here.

src/api/libcellml/importsource.h
agarny: Sorry, I meant something like `@sa isImport`?

src/api/libcellml/model.h
agarny: Yes, I would expect the declaration to use `importSource` rather than `imp`, which in turn means that we should here use `@p importSource` rather than `@p imp`.

src/api/libcellml/model.h
agarny: Oops, my bad. I stand corrected!

src/bindings/interface/importsource.i
agarny: Just because the method is called `addComponent`. If it was called `addDependentComponent` (which I wouldn't want) then it would make sense for me to have "dependent", but not as it stands. Just my 2c.

src/importsource.cpp
agarny: Oh, that's because there is no `const` in `setImportSource()`. Ok, got it.

src/api/libcellml/model.h
agarny: Well, we do have "[0, \#equivalentVariables)" in `variable.h`. I am "fine" with whatever we go for, as long as we are being consistent and here we are not.

agarny:
> Thanks for the review @agarny. I've made some of the changes as suggested, but I'm not prepared to spend time doing the docstrings until we've got some kind of agreement with @hsorby and @nickerso about what they should be everywhere. And even then, there is so much inconsistency throughout I think they should be addressed in a separate PR which fixes up everything properly.

Sure, happy for this to be done in a separate PR.

tests/model/units_import.cpp
kerimoyle: 2f009081

tests/import_source/import_source.cpp
kerimoyle: 2f009081

tests/import_source/import_source.cpp
kerimoyle: 2f009081

tests/import_source/import_source.cpp
kerimoyle: 2f009081

tests/import_source/import_source.cpp
kerimoyle: 2f009081

tests/import_source/import_source.cpp
kerimoyle: 2f009081

tests/import_source/import_source.cpp
kerimoyle: 2f009081

src/api/libcellml/model.h
kerimoyle: It's not great to use variable names which shadow function names, which is why I chose to use `imp` and not `importSource`?

src/api/libcellml/model.h
kerimoyle: 7a544255

src/api/libcellml/model.h
kerimoyle: TBD in #618

src/api/libcellml/units.h
kerimoyle: TBD in #618

src/bindings/interface/importsource.i
kerimoyle: TBD in #618

src/bindings/interface/model.i
kerimoyle: TBD in #618

src/api/libcellml/model.h
agarny: ?? The method is called `addImportSource()`, so where is the problem?

src/bindings/interface/importsource.i
kerimoyle: Eh.  I'd hope that docstrings were allowed to have more info than just rewriting their function name!  I've changed it for now 7a544255 but hopefully #618 can improve what we provide to users.

src/api/libcellml/model.h
```diff
@@ -420,9 +420,7 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
     /**
      * @overload
      *
-     * @brief Remove the import source with the given pointer.
-     *
-     * Remove the import source with the pointer @p imp.
+     * @brief Remove the import source at the given reference.
```
agarny: Why has the "detailed" description been removed?

src/api/libcellml/model.h
kerimoyle: If you declare a variable called `importSource` in the header, then you have a variable called `importSource` in the function.  Since we have a function called `importSource` in this class, then also having a variable with the same name isn't great?

src/api/libcellml/model.h
kerimoyle: Because it was the same as the brief one.  No point in duplicating it.

agarny: Well, that breaks consistency (and not sure how it will look like in the documentation), but I have voiced my view, so I will leave it to @hsorby and @nickerso to comment.

src/api/libcellml/model.h
agarny: I have no problems with that, not least because this is exactly what is currently being done in `api/libcellml/importedentity.h` and in `api/libcellml/issue.h` where a method `importSource()` is defined and `importSource` is used as a parameter name for `setImportSource()`.

src/api/libcellml/component.h
kerimoyle: Will be addressed in #618

src/api/libcellml/importsource.h
kerimoyle: Will be addressed in #618

src/api/libcellml/importsource.h
kerimoyle: Will be addressed in #618

src/api/libcellml/importsource.h
kerimoyle: Will be addressed in #618

src/api/libcellml/importsource.h
kerimoyle: Will be addressed in #618

src/api/libcellml/model.h
kerimoyle: Heh, I would rather change those ones too then ... I would have thought that there would be a clang flag for this kind of thing? Why do we only use `-Wshadow` in gcc? I'll change it to a longer variable name, but I really don't like shadowing others, so it won't be `importSource`.

kerimoyle: f85f512e

src/api/libcellml/importsource.h
kerimoyle: Done (somewhere...)

src/api/libcellml/importsource.h
kerimoyle: e92e73bb

src/api/libcellml/model.h
nickerso: No point having a full description that is the same as the brief description (doxygen doesn't care, it will use the brief as needed). In my view the full description is only needed if the brief description is not sufficient.

nickerso:
> Sure, happy for this to be done in a separate PR.

agree.

nickerso (approved):

src/api/libcellml/model.h
agarny: We are in a democracy, so... fine. :)

agarny (approved):
I really don't like commit f85f512. It's not consistent with the rest of our codebase and that's also the reason I originally advocated parameter names to start with `p` to avoid problems like this, and also to make it clear whether a variable is a member variable (e.g. `mVariable`), a parameter variable (`pVariable`) or a local variable (`variable`). Anyway...

Issue #655, opened 2020-07-12
Python: make sure that we can find Python 3.8.x


agarny:
On systems with multiple pythons installed, our CMake configuration will not currently choose Python 3.8.x.

kerimoyle:
Why does it need to be 3.8? Why not 3.somethinglessthan8?

agarny:
Our CMake script can find Python 3.0 to 3.7, but not 3.8, hence this issue and https://github.com/cellml/libcellml/pull/656/files#diff-01d7c6dd961002e07a74b697f4967894 that addresses it.

nickerso:
See https://github.com/cellml/libcellml/pull/656#discussion_r453405819 for explanation of this issue.

@agarny - also please note that empty issues are not very useful...

agarny:
> @agarny - also please note that empty issues are not very useful...

Agreed, I was a bit quick on this one.

PR #656, opened 2020-07-13
Python: make sure that we can find Python 3.8.x in systems with other versions

agarny:
Addresses issue #655.


cmake/FindPython/Support.cmake
```diff
@@ -17,7 +17,7 @@ if (NOT DEFINED _${_PYTHON_PREFIX}_REQUIRED_VERSION_MAJOR)
   message (FATAL_ERROR "FindPython: INTERNAL ERROR")
 endif()
 if (_${_PYTHON_PREFIX}_REQUIRED_VERSION_MAJOR EQUAL 3)
-  set(_${_PYTHON_PREFIX}_VERSIONS 3.7 3.6 3.5 3.4 3.3 3.2 3.1 3.0)
+  set(_${_PYTHON_PREFIX}_VERSIONS 3.8 3.7 3.6 3.5 3.4 3.3 3.2 3.1 3.0)
```
nickerso, line 20: is this actually used anywhere? Things have been working fine with 3.8 for a while now...

hsorby (approved):

cmake/FindPython/Support.cmake
nickerso: *fine....on windows at least.

agarny: > is this actually used anywhere? Things have been working fine with 3.8 for a while now...

Really? I have Python 2.7 (system version) and Python 3.8 (from homebrew) and libCellML was always picking up Python 2.7. With that change, it now picks up Python 3.8.

hsorby: This is required when the FindPython script can find multiple Pythons.  If multiple Pythons are available then these versions are used to test for other Pythons.  This means that Python 3.8 cannot be found unless you add it as an option here.

nickerso (approved):

Issue #665, opened 2020-07-20
CMake: get a lot of warnings when building our Python bindings using Python 3.8


agarny:
Say that you have Python 3.8 installed, then to build the Python bindings will result in tens of messages like:
```
src/bindings/python/swig_intermediaries/versionPYTHON_wrap.cxx:1780:7: warning: 'tp_print' is deprecated [-Wdeprecated-declarations]
      0,                                    /* tp_print */
      ^
/usr/local/Cellar/python@3.8/3.8.4/Frameworks/Python.framework/Versions/3.8/include/python3.8/cpython/object.h:260:5: note: 'tp_print' has been explicitly marked deprecated here
    Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
    ^
/usr/local/Cellar/python@3.8/3.8.4/Frameworks/Python.framework/Versions/3.8/include/python3.8/pyport.h:515:54: note: expanded from macro 'Py_DEPRECATED'
#define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
                                                     ^
```

This is a known issue (see [here](https://github.com/cython/cython/issues/3474)).

agarny:
FWIW, we get those warnings on macOS, but not on Windows using MSVC 2017. As for Ubuntu (18.04), I don't know since the default Python 3 is 3.6...

PR #666, opened 2020-07-20
CMake: prevent a lot of warnings when building our Python bindings using Python 3.8 on macOS

agarny:
Addresses issue #665.


nickerso (commented):
not sure how to review this as I don't have a Mac and this change shouldn't affect the CI...

agarny:
> not sure how to review this as I don't have a Mac and this change shouldn't affect the CI...

I know... :) Still, that issue has been bugging me ever since Homebrew upgraded Python to Python 3.8, so... it would be nice to have it in... :) At the end of the day, this PR does something similar to what we are already doing for Clang-Tidy (see [here](https://github.com/cellml/libcellml/blob/4eda134eac1232695ce6a9b06e060810a7bd2020/cmake/common.cmake#L91-L123)).

nickerso (approved):
CI is happy, so probably safe to approve this one.

kerimoyle:
Even having a Mac I can't test this as I don't run Python 3.8! Does @hsorby have it? Should we have at least one person actually able to see if it works before approving ...?

agarny:
> Even having a Mac I can't test this as I don't run Python 3.8! Does @hsorby have it? Should we have at least one person actually able to see if it works before approving ...?

How was your copy of Python installed? Using Homebrew? If so, you can upgrade it using `brew upgrade` and then you can do `brew cleanup -s` to remove any unnecessary files, if you want (to save disk space). From there, you should test the `develop` branch to confirm the warnings, and then test my `issue665` branch to confirm that it doesn't generate warnings anymore.

hsorby (approved):

Issue #669, opened 2020-07-26
Add auto-id functionality to the Annotator class


kerimoyle:
At present we have the option to create automatic ids when serialising a model.  We need to have this functonality in the `Annotator` class too.

kerimoyle:
Contemplating the different ways that this could be done, with different outcomes!

1) Use the current serialisation functionality that's already in the `Printer` and simply print-parse the given model using the `autoIds` option.  This is super easy, but means that we need to return a new copy of the model from the `Annotator` as the current printer functionality creates an entirely new model, and that the `Annotator` would need to reprocess the model.

2) Traverse the model and make use of the (reasonably expensive) mapping inside the `Annotator` between current ids and entities.  This is doable, but there's a chance that `id` strings could differ between those automatically created in the `Printer` and those automatically created in the `Annotator`.  If this is a big deal, it could be written such that they were the same.

3) Provide the ability to set automatic ids by type, for all components, all variables etc as well as for the entire model.  On the plus side it means that you don't get a model that's overly bloated with ids when you only want to annotate units, for example.  On the minus side, the automatic ids would certainly be different from those in the `Printer`.

4) ... something else?

Thoughts? @nickerso @hsorby @agarny ?

nickerso:
Thinking a bit more about how this would actually be used and I'm starting to think that the `autoId` option on the `Printer` is fine for the bulk case of "add ids to everything that doesn't have one". For programmatic access, it is more likely that a tool developer just wants to get an id for some object and have it generate an id and assign it if needed. e.g., something along the lines of `std::string Annotator::getId(AnyObject&)` (or all the overloaded methods for entities and non-entities).

That would give the basic building block for any future requirements that we might want to support. I think.

kerimoyle:
Thanks @nickerso.  After thinking about it more this afternoon I realised it made more sense to do it inside the `Annotator` class rather than the `Printer`, so that the id map in the annotator could be maintained, and so that the model itself didn't need to change (getting an entirely new model with entirely new child pointers is really clumsy, especially since the `Parser` doesn't really update anything, only appends to).  The "group" setting of ids is in #671 already, though an overloaded function `Annotator::setAutomaticIds()` (does everything in the model) and `Annotator::setAutomaticIds(Annotator::Type &type)` (does everything of that `type`).  Extending that to individual items is straightforward - I'll do that tomorrow.

PR #671, opened 2020-07-27
Annotator auto ids

kerimoyle:
Addresses #669.

Needs to be merged **after** #640 as it's built on top of that one.  New code begins from commit c369057.


src/annotator.cpp
```diff
@@ -0,0 +1,1285 @@
...
+        issue->setLevel(libcellml::Issue::Level::WARNING);
+        addIssue(issue);
+        item = std::make_pair(Annotator::Type::ISSUE, issue);
+        return item;
+    }
+
+    // Check whether the id has been duplicated in the model.
+    if (mPimpl->mIdList.count(id) > 1) {
+        auto issue = libcellml::Issue::create();
+        issue->setDescription("The id '" + id + "' occurs " + std::to_string(mPimpl->mIdList.count(id)) + " times in the model. Please use the repairNonUniqueIds() function to fix them.");
```
nickerso: I don't think we should be calling out API methods in the issue messages, right? Are we doing that anywhere else?

kerimoyle: Not sure if we do it elsewhere.  It just seemed like the most useful info I could give!

src/annotator.cpp
```diff
@@ -0,0 +1,1285 @@
...
+}
+
+void Annotator::AnnotatorImpl::doSetAllAutomaticIds()
+{
+    std::string id;
+    // Traverse the model's entities and add ids where they are currently missing.
+    if (mModel->id().empty()) {
+        id = makeUniqueId();
+        mModel->setId(id);
+        mIdList.insert(std::make_pair(id, std::make_pair(Annotator::Type::MODEL, mModel)));
```
nickerso: `doSetModelId`

kerimoyle: (... I don't get it ...?)

nickerso: this is duplicating the code that is the `doSetModelId()` method...

src/annotator.cpp
```diff
@@ -0,0 +1,1478 @@
...
+        doSetTestValueIds(mModel->component(c));
+        doSetConnectionIds(mModel->component(c));
+        doSetMapVariablesIds(mModel->component(c));
+        doSetComponentRefIds(mModel->component(c));
+    }
+
+    doSetEncapsulationIds();
+}
+
+std::string Annotator::AnnotatorImpl::makeUniqueId()
```
nickerso, line 1097: this seems like it could start to get quite inefficient when dealing with large models. Can the counter not be class variable that is tracked rather than having to iterate over all values every time a new ID is needed?

kerimoyle: Yep, done in de000b7

src/annotator.cpp
kerimoyle: Moved to docstring.

kerimoyle:
Sigh.  More CI gremlins here @hsorby :( Build reports are looking at something earlier than 2644704 (which is where the fixes were added).

kerimoyle:
@hsorby This is building on an out-dated commit again :(  The errors are fixed in e844ce3?

kerimoyle:
Closing as has been merged into #640

Issue #672, opened 2020-07-28
Create INFORMATION level for Issue class


kerimoyle:
So that we can give error-neutral information to the user, we need a new `INFORMATION` level within the `Issue` class.

PR #673, opened 2020-07-28
Issue: add INFORMATION as a new issue level

agarny:
Addresses issue #672.


src/api/libcellml/issue.h
```diff
@@ -103,15 +103,15 @@ class LIBCELLML_EXPORT Issue
         COMPONENT,
         CONNECTION,
         ENCAPSULATION,
+        GENERATOR,
```
kerimoyle, line 106: GENERATOR should not be here, this is for entities only

src/bindings/interface/types.i
```diff
@@ -25,24 +25,24 @@ Only meant to be included, shouldn't be passed to cmake as a module!
 
 // Shared typemaps
```
kerimoyle, line 27: I'm guessing that the changes here are just adding new functions similar to the old ones, rather than actual changes??

src/logger.cpp
```diff
@@ -36,6 +36,7 @@ struct Logger::LoggerImpl
     std::vector<size_t> mErrors;
     std::vector<size_t> mWarnings;
     std::vector<size_t> mHints;
+    std::vector<size_t> mInformations;
```
kerimoyle, line 39: Urgh.  Not a fan of the "informations" thing.  Maybe it could be a type of `LOG`? which would then give us `logCount()`, `mLogs` etc and a more natural language to describe it?

kerimoyle (changes requested):
After reading through this I wonder if we should change the enum to be `LOG` instead of `INFORMATION`. The only motivation for this is that we avoid the horrible language and variable names of pluralising "informations"?  Only other change is that `GENERATOR` should not be an issue type.

src/api/libcellml/issue.h
agarny: Agreed, but I was merely putting it in the right place so that the listing is alphabetically ordered, as it should have been in the first place. `GENERATOR` is properly removed in PR #631 (see [here](https://github.com/cellml/libcellml/pull/631/files#diff-d7a946aa2f3f6a627b853150866642f8)).

src/logger.cpp
agarny: Not keen on log. A log is something that I would use to output debug information, which is not what we are after here.

src/bindings/interface/types.i
agarny: Yes, I have added one for `libcellml::Issue::Level `. As for the rest, I have cleaned things up and reordered things.

FWIW, for this PR, I started from PR #650 (since this is where I had introduced `INFORMATION`) and removed all the non-relevant bits. When it comes to `src/bindings/interface/types.i`, I can't recall the details, but I know that I couldn't get something to work unless I was using `$type` instead of say `libcellml::Units::Prefix`. Also, I reordered things as it makes it easier to maintain.

src/logger.cpp
nickerso: agree log is not ideal - in the logger class log would usually mean the whole thing...to me at least. And I'm not too worried about "natural language" :) If anything I'd slightly prefer `INFO` over `INFORMATION`.

nickerso (approved):
Looks fine apart from the unresolved conversation between @kerimoyle and @agarny. As always, I'd plea for unrelated code changes to be kept separate...but no point splitting it up at this point.

agarny:
> As always, I'd plea for unrelated code changes to be kept separate...but no point splitting it up at this point.

I know... Argh!

src/logger.cpp
agarny: Ok, if you would only slightly prefer `INFO` over `INFORMATION` then personally I would stick with `INFORMATION`, just for consistency (i.e. use full words rather than a short version of them).

@kerimoyle, up to you now... :)

kerimoyle (approved):

src/logger.cpp
nickerso: see #674.

Issue #682, opened 2020-08-09
Validator should check that the <degree> tag is only inside root or diff


kerimoyle:
See discussion [here](https://github.com/cellml/cellml-specification/issues/335)

PR #683, opened 2020-08-09
Validator to check position of <degree> tag in MathML

kerimoyle:
Addresses #682


kerimoyle:
@hsorby The gremlins are back :( The stuff in commit 684227b passes everything on my machine - this is being built on the older commit 54a571e.

kerimoyle:
Ready for review please @hsorby @agarny @nickerso.  This is a tiny PR so hopefully can be done quickly!

src/validator.cpp
```diff
@@ -967,6 +967,15 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
         }
+        if (childNode->name() == "degree") {
+            if ((node->name() != "root") && (node->name() != "diff") && (node->name() != "bvar")) {
+                IssuePtr issue = Issue::create();
+                issue->setDescription("Math has a 'degree' element in a parent '" + node->name() + "' element; 'degree' is supported for 'root' or 'diff' elements only.");
+                issue->setComponent(component);
+                issue->setCause(Issue::Cause::MATHML);
+                mValidator->addIssue(issue);
+            }
+        }
```
agarny: Replace with:
```
} else if ((childNode->name() == "degree")
            && (node->name() != "root") && (node->name() != "diff") && (node->name() != "bvar")) {
    issue->setDescription("Math has a 'degree' element in a parent '" + node->name() + "' element; 'degree' is supported for 'root' or 'diff' elements only.");
    issue->setComponent(component);
    issue->setCause(Issue::Cause::MATHML);
    mValidator->addIssue(issue);
}
```

tests/validator/validator.cpp
```diff
@@ -846,6 +846,136 @@ TEST(Validator, validMathMLCiAndCnElementsWithCellMLUnits)
...
+
+    c->addVariable(v1);
+    c->addVariable(v2);
+    c->addVariable(v3);
+    c->setMath(math);
+    m->addComponent(c);
+
+    v->validateModel(m);
+    EXPECT_EQ(size_t(1), v->errorCount());
+    EXPECT_EQ(e, v->error(0)->description());
```
agarny, line 901: I would use `EXPECT_EQ_ISSUES()`.

tests/validator/validator.cpp
```diff
@@ -846,6 +846,136 @@ TEST(Validator, validMathMLCiAndCnElementsWithCellMLUnits)
...
+
+TEST(Validator, validMathMLDegreeInsideRoot)
+{
+    auto math = "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
+                "  <apply><eq/>\n"
+                "    <ci>A</ci>\n"
+                "    <apply>\n"
+                "      <root/>\n"
+                "      <degree><ci>B</ci></degree>\n"
+                "      <ci>C</ci>\n"
```
agarny, line 912: If you swap lines 901 and 902, your test will still be valid while it shouldn't be... since your new code in `Validator::ValidatorImpl::validateMathMLElements()` won't be executed. Same, if line 901 was duplicated. The test would still pass while it shouldn't. Etc.

agarny (changes requested):
I made a couple of comments, but in the end I feel like this PR is wrong. `Validator::ValidatorImpl::validateMathMLElements()` tests whether `degree` is the first child of a `root`, `diff` or `bvar` element. Yet, this is not what we are after.

If `root` is the first child of an `apply` element then the `apply` element can have two or three children. If it has three children then the second child has to be a `degree` element.

If `diff` is the first child of an `apply` element then `apply` must have three children and its second child must be a `bvar` element. In turn, that `bvar` element must have one or two children. If it has two children then its second child must be a `degree` element.

That's roughly the rules for the `degree` element (to be double checked, confirmed, etc.). So, the error message that is currently generated is wrong and so is the test done in `Validator::ValidatorImpl::validateMathMLElements()`, not to mention that he doesn't account for what I mentioned above...

src/validator.cpp
kerimoyle: d21e070

tests/validator/validator.cpp
kerimoyle: The validator we checks against the spec only, and all the spec says is that `degree` must be a child of `root` or `diff`.  It doesn't matter whether the MathML makes sense or not; if the XML says that there's a node called "degree" as a child of "root" then it's ok ... it's not sensible, but, according to the spec, it's ok?  Or are we now introducing stuff that says that the MathML also has to be reasonably constructed in order to be valid? ... because that's an entirely different issue!

tests/validator/validator.cpp
kerimoyle: I just copied from the code above ... you know, for consistency ;)

kerimoyle:
> If root is the first child of an apply element then the apply element can have two or three children. If it has three children then the second child has to be a degree element.

Not mentioned in the spec, therefore not checked in the Validator.  

> If diff is the first child of an apply element then apply must have three children and its second child must be a bvar element. In turn, that bvar element must have one or two children. If it has two children then its second child must be a degree element.

As above.

If we're now going to start checking syntax and meaning of the MathML, then this needs to be a much wider conversation, and is probably better done with a mechanism like those in the Analyser which build the AST for checking.  I'm not sure it makes a whole lot of sense to do that process twice/in two separate places?  Either way, checking placement of different element types is something we haven't done at all - this PR is about the one time when a placement is mentioned in the spec, so needed to be covered off by the validator.

tests/validator/validator.cpp
agarny: Agreed that it would be an entirely different issue, which is why I I would just stick with the DTD validation. It has its limitations, but so does this PR, hence I wouldn't bother with it.

agarny:
> If we're now going to start checking syntax and meaning of the MathML, then this needs to be a much wider conversation, and is probably better done with a mechanism like those in the Analyser which build the AST for checking. I'm not sure it makes a whole lot of sense to do that process twice/in two separate places? Either way, checking placement of different element types is something we haven't done at all - this PR is about the one time when a placement is mentioned in the spec, so needed to be covered off by the validator.

I haven't double checked, but I imagine our spec says that MathML must be valid? Right now, we do this through DTD validation, but maybe we should do more than just DTD validation?

What is certain is that, upon validation, the analyser assumes that **everything** is fine with the math while I now realise that it may not be the case, and I imagine that the analyser might crash as a result (to be confirmed).

hsorby:
I think this PR request can be closed without merging.  The CellML spec. seems inconsistent with the MathML spec. here.  I cannot see a requirement for the supplementary restriction on the degree node.

kerimoyle:
I've created an issue in the spec repo that will make this PR redundant.  https://github.com/cellml/cellml-specification/issues/337

PR #685, opened 2020-08-12
Importer instantiates only what it needs

### LINKED ISSUE #645: Should the importer instantiate all objects in an imported model, or just what it needs?
At present, an imported item triggers the import of the entirety of its parent model.  Should it?  See [here](https://github.com/cellml/libcellml/pull/608/files/7349c363d5cef96be92913b41aa037cd7fcdf474#r448045188).

### LINKED ISSUE #668: Do we need to distinguish "external dependencies" in the Importer class?
The `Importer` class proposed in #608 distinguishes between models that have been added to the library directly and those that were added via the `resolveImports` method. This distinction exists due to the import resolution originally being part of the model class and the external dependencies being used there to track the imported models, whereas in the Importer we directly have the library of models performing this role.

This issue is to remind us to revisit this distinction once #608 is merged in.

### LINKED ISSUE #681: Importer: externalDependency() should return a pointer to a class rather than a `std::pair`
`Importer::externalDependency()` currently returns a `std::pair<std::string, std::string>` object, which is not neat. Ideally, we would return a pointer to a class that has only two methods to retun each of the strings in the original pair. (This is what is done in PR #631 with the `AnalyserModel` class for example.)

See https://github.com/cellml/libcellml/pull/608/files#r466043964.

kerimoyle:
This is built on top of #630 so should be reviewed and merged after that one.

It addresses:
- #645 so that imports are only resolved to the extent that they need to be, rather than resolving each item in an imported model whether it's needed or not;
- #668 by removing the external dependency distinction ...
- which also fixes #681 as well.


kerimoyle:
ping @hsorby @agarny @nickerso

kerimoyle:
Just to say there was a bug in the importer where second-generation encapsulated imports were being skipped.  It's fixed as of 
5289c41.

src/api/libcellml/importedentity.h
```diff
@@ -95,6 +95,26 @@ class LIBCELLML_EXPORT ImportedEntity
...
+    /**
+     * @brief Set the resolution status of this imported entity.
+     *
+     * Set the resolution status of this imported entity.  When @c true, this 
+     * indicates that the item and all its dependent children have been resolved.
+     * Otherwise, @c false.
+     * 
+     * @param status A boolean indicating import resolution status.
+     */
+    void setResolved(bool status);
```
agarny, line 116: See issue #724.

src/component.cpp
```diff
@@ -432,4 +432,30 @@ ComponentPtr Component::clone() const
...
+{
+    if (isImport()) {
+        return true;
+    }
+    for (size_t c = 0; c < componentCount(); ++c) {
+        if (doRequiresImport(component(c))) {
+            return true;
+        }
+    }
+    return false;
```
nickerso: can't this all be replaced with `doRequiresImport(this)` or something like that?

src/importer.cpp
```diff
@@ -150,130 +125,258 @@ bool Importer::ImporterImpl::doResolveImports(ModelPtr &model, const std::string
...
+    }
+
+    // Check that the model instance in the library has resolved all of the required dependencies.
+    auto sourceModel = importComponent->importSource()->model();
+    auto sourceComponent = sourceModel->component(importComponent->importReference());
+    if (sourceComponent != nullptr) {
+        // Check whether the sourceComponent is itself an import. Note that the file path passed in
+        // here must be the path to the sourceModel, rather than the path to the previous import, so that
+        // the chain always tests local to the importing model first.
+        auto newBase = baseFile + directoryPath(importComponent->importSource()->url());
```
nickerso, line 221: shouldn't this be using `resolvePath` ? doesn't make sense to add a directory to a file path?

nickerso (dismissed):
couple of minor comments, but liking the now correct behaviour of these tests.

nickerso (dismissed):

agarny (dismissed):

src/component.cpp
agarny: Done in commit 739d5864.

src/importer.cpp
agarny: Actually, `baseFile` is not a file path, but a file directory. `baseFile` is used in quite a few places, so I leave it to @kerimoyle to change it, if really needed.

agarny (approved):

nickerso (approved):

PR #693, opened 2020-08-20
Include Importer in python tests

kerimoyle:
Super tiny PR to include the importer python tests in the build.  Also fix that those test showed ...

### DIFF
--- a/src/bindings/interface/importer.i
+++ b/src/bindings/interface/importer.i
@@ -2,11 +2,11 @@
 
 #define LIBCELLML_EXPORT
 
+%include <std_string.i>
+
 %import "createconstructor.i"
 %import "logger.i"
 %import "types.i"
-
-%include <std_string.i>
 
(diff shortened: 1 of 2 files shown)


src/bindings/interface/importer.i
```diff
@@ -6,6 +6,8 @@
 %import "logger.i"
 %import "types.i"
 
+%include <std_string.i>
+
```
agarny: This should be moved after `#define LIBCELLML_EXPORT` as we do in our other interface files.

agarny (changes requested):
Just one minor change.

src/bindings/interface/importer.i
kerimoyle: 8f04142a

hsorby (approved):

agarny (approved):

PR #696, opened 2020-08-25
Issue class to make use of std::any type

### LINKED ISSUE #689: Should the Issue class use std::any
We currently store a whole load of different pointer types in the Issue class:

```
struct Issue::IssueImpl
{
    std::string mDescription; /**< The string description for why this issue was raised. */
    Issue::Cause mCause = Issue::Cause::UNDEFINED; /**< The Issue::Cause enum value for this issue. */
    Issue::Level mLevel = Issue::Level::ERROR; /**< The Issue::Level enum value for this issue. */
    Issue::ReferenceRule mReferenceRule = Issue::ReferenceRule::UNDEFINED; /**< The Issue::ReferenceRule enum value for this issue. */
    ComponentPtr mComponent; /**< Pointer to the component that the issue occurred in. */
    ImportSourcePtr mImportSource; /**< Pointer to the import source that the issue occurred in. */
    ModelPtr mModel; /**< Pointer to the model that the issue occurred in. */
    UnitsPtr mUnits; /**< Pointer to the units that the issue occurred in. */
    VariablePtr mVariable; /**< Pointer to the variable that the issue occurred in. */
    ResetPtr mReset; /**< Pointer to the reset that the issue ocurred in. */
};
```

These could be replaced by a single `AnyItem` instance? That would in turn mean it was more straightforward to retrieve an item from the `Issue` it's attached to?

kerimoyle:
Addresses #689


src/api/libcellml/enums.h
```diff
@@ -0,0 +1,49 @@
...
+#include <memory>
+
+namespace libcellml {
+/**
+ * @brief The ItemType enum class.
+ *
+ * 
+ */
+// KRM not sure how to get this working with SWIG.
+enum class LIBCELLML_EXPORT ItemType
```
hsorby: I would suggest that you don't export an enum.

kerimoyle: Yeah, without this I get a different set of errors.  Hang on a mo ...

tests/issue/issue.cpp
```diff
@@ -604,3 +615,154 @@ TEST(Issue, undefinedIssueUrl)
     EXPECT_EQ(e, issue->url());
     EXPECT_EQ(e, issue->referenceHeading());
 }
+
+#endif
+
+TEST(Issue, setGetByObjectOrItem)
```
kerimoyle: @agarny Here's what I'm struggling with! Inside the `Issue::Impl` is two fields:
- `mAnyItem` which is an `std::any`
- `mItemType` which is one of the `ItemType` enums.
At the moment I'm not doing any switching, it's all written as if I will only ever store `Component` types inside the any class (for my debugging sanity).  So.  

I can add and retrieve an item (here it's a `Component`) into the `mPimpl->mAnyItem` item, which is a `std::any` class.  When you add anything, they're all stored in the `mAnyItem` slot.
When you retrieve anything specifically (eg: `issue->component()`) it will do the casting first and return a component pointer.  But you can also choose to retrieve the std::any item itself, so that you can do the casting later based on the value of the `mItemType`.  It's this external casting which is failing.  I've checked that the same pointers and values are being stored, but when I call `std::any_cast` in the test (ie: outside the `Issue`) class, it fails.  The operation is identical to what's happening internal to the `Issue` class in order to return the component directly.

agarny: Just a look and... it seems to be working fine for me on macOS. I mean that line 644 down there is fine... So... what am I missing here?

kerimoyle: :O Really?? I'm not sure if that's good news or bad news ... Let me look again ...

kerimoyle:
@hsorby Buildbot is still using old commits ... 🤦‍♀️

kerimoyle:
I think this is ready for review please @hsorby @agarny @nickerso (as soon as the buildbot catches up ... all the tests pass on my machine).

kerimoyle:
I'm not sure what's happening here @hsorby ?  The coverage data doesn't make any sense?

kerimoyle:

agarny:
FWIW, @hsorby, the Windows CI passes, but looking at the log, there is [a warning](https://autotest.bioeng.auckland.ac.nz/buildbot/libcellml/#/builders/3/builds/402/steps/6/logs/warnings__1_). I thought that warnings were treated as errors?

agarny:
> FWIW, @hsorby, the Windows CI passes, but looking at the log, there is [a warning](https://autotest.bioeng.auckland.ac.nz/buildbot/libcellml/#/builders/3/builds/402/steps/6/logs/warnings__1_). I thought that warnings were treated as errors?

Oh, that's probably because it's in the Python bindings.

kerimoyle:
(Just a note to myself to wait for @hsorby's changes in #640 before doing the python coverage needed here - they should use the same approach)

src/api/libcellml/enums.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <memory>
+
+namespace libcellml {
+
+/**
+ * @brief The ItemType enum class.
+ *
+ */
+enum class ItemType
```
hsorby: I'm not a fan of the name for this enumeration.  I think it should be `ReferenceType` or `ReferenceCellMLType`.  With documentation stating that it is a reference that identifies a CellML 2.0 specification element information item.

I can't decide if the `CellML` is needed in the name or not.

kerimoyle: I don't think that CellML can be in the name.  It needs to have an XML class, and that's outside CellML ... I also don't think "reference" fits either.  It's the type of thing which is stored in the issue.  Whether that's a CellML element or something else, I thought that "item" was pretty broad ...

hsorby: I don't think `XML` needs to be in this enum.  To me it is telling us what sort of thing is stored in the issue thus `XML` shouldn't be here as it is not something that is stored in the issue.  For me this has all become very muddled up and that the `ReferenceType` should describe the reference that is being held be it a `Variable`, `Component`, `VariablePair`, `UnitItem`, etc.  The `ReferenceRule` would specify what is wrong which is where the `XML` comes in.  `ReferenceRule` already has an `UNDEFINED` value but it could also have `NOT_SPECIFIED` to indicate that the reference rule that is broken is not being specified.

kerimoyle: It's there because the old "cause" and "item type" are 99% the same ... but if you want to be able to report a "cause" being something related to XML in the `Parser` then you need this category.  I think that a not specified entry isn't helpful, especially when we actually do know where the problem is!

hsorby: I'm thinking that the `cause` should go and we simply use `referenceRule` to do that job.

kerimoyle: Yeah, except that these `Issue` items need to be relevant to all of the `Logger` types, not just the `Validator`.  Is there another name that doesn't imply it's validation related?

hsorby: The `Validator` is restricted to the issues that it can raise.  Which is a subset of the reference rules.  We already have reference rules that are not available to the validation.  This was always the vision, the `Validator` validates against the specification and uses reference rules to identify exactly where in the specification the error arrises from.  Reference rules would be extended for other reference material.  For an issue that doesn't come from a defined reference we have `UNDEFINED`.

agarny: What's the situation with this?

src/api/libcellml/issue.h
```diff
@@ -310,121 +283,279 @@ class LIBCELLML_EXPORT Issue
...
-     * @param component A pointer to the component that this component issue is relevant to.
+     * @param component A @c ComponentPtr relevant to this issue.
      */
     void setComponent(const ComponentPtr &component);
 
     /**
-     * @brief Get the component for this issue.
-     *
-     * Get the component that this issue is relevant to.
+     * Get the component relevant to this issue.
```
agarny: Didn't @hsorby say that we must always use `@brief` even if it's the same as the "full" description?

src/api/libcellml/issue.h
```diff
@@ -310,121 +283,279 @@ class LIBCELLML_EXPORT Issue
...
+
+    /**
+     * @brief Set an @c std::any item relevant to this issue.
+     *
+     * Set an @c std::any item relevant to this issue.
+     *
+     * @param item An @c std::any item relevant to this issue.
+     * @param type An @c ItemType enum.
+     */
+    void setItem(ItemType type, const std::any &item);
```
agarny: I would find it more logical to have the following signature instead:
```
void setItem(const std::any &item, ItemType type);
```

src/validator.cpp
```diff
@@ -1389,7 +1379,6 @@ void Validator::ValidatorImpl::validateNoUnitsAreCyclic(const ModelPtr &model)
                 des += issues[issues.size() - 1] + "'";
                 issue->setDescription("Cyclic units exist: " + des);
                 issue->setModel(model);
```
kerimoyle: This should be `issue->setUnit` so that we store the unit not the model?

src/importer.cpp
```diff
@@ -343,7 +343,7 @@ void Importer::ImporterImpl::makeIssueCyclicDependency(const ModelPtr &model,
     auto issue = Issue::create();
     issue->setDescription(msg);
     issue->setLevel(Issue::Level::WARNING);
-    issue->setCause(ItemType::IMPORT);
+    issue->setReferenceRule(Issue::ReferenceRule::IMPORT_EQUIVALENT);
     mImporter->addIssue(issue);
```
kerimoyle: Need to set an item here - probably the import source?

src/parser.cpp
```diff
@@ -817,7 +809,6 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr
                     IssuePtr issue = Issue::create();
                     issue->setDescription("Connection in model '" + model->name() + "' has an invalid non-whitespace child text element '" + textNode + "'.");
                     issue->setModel(model);
```
kerimoyle, line 822: needs reference rule?

src/api/libcellml/enums.h
kerimoyle: XML is gone now.

kerimoyle:
Hey @hsorby I'm OK with removing the `Cause` stuff, but I wonder if it should be replaced with use of the reference rules everywhere?  Particularly in the parser, we don't have anything that could be tied to a URL?

src/api/libcellml/issue.h
```diff
@@ -310,121 +283,279 @@ class LIBCELLML_EXPORT Issue
...
+    ComponentPtr math() const;
+
+    /**
+     * @brief Set the connection relevant to this issue.
+     *
+     * The internal type will be set to @c ItemType::CONNECTION.
+     *
+     * @param pair The @c VariablePair whose connection is relevant to this issue.
+     */
+    void setConnection(const VariablePair &pair);
```
agarny: There are a few methods that require an `std::pair` as a parameter (like `setConnection()` indeed) or return an `std::pair`, and I am really not keen on it (and the same elsewhere in the API if we have something similar).

`std::pair` is a pure C++ construct, so I wouldn't expect something using libCellML having to create a pair and pass it to a libCellML method. This is, for instance, why we have all those `create()` methods, so that end-users don't have to "manually" do something like `std::shared_ptr<Issue> {new Issue {}}`.

In the same way, when we retrieve a pair, we have to do something like `x.first` and `x.second`, which is not very meaningful to me. Instead, I would expect to do something like `x.variable1()` and `x.variable2()` (when it comes to a mapping pair, for instance).

agarny (commented):
Looks good overall. Just a few comments before I can approve. :)

src/api/libcellml/issue.h
hsorby: yup he did.

src/api/libcellml/issue.h
hsorby: You think we should wrap a simple class around std::pair?

src/api/libcellml/issue.h
hsorby: I find it more logical as is.  Does anyone else want to offer an opinion?

src/api/libcellml/issue.h
agarny: Yes, I think this would make things cleaner from an API viewpoint. Also, when it comes to calling something like `setConnection()`, I think it would be nice to have two signatures:
1. `void setConnection(const VariablePairPtr &variablePair);`; and
2. `void setConnection(const VariablePtr &variable1, const VariablePtr &variable2);`.

src/api/libcellml/issue.h
agarny: Maybe you do, but besides the fact that I don't, it is also inconsistent with our different `Issue::create()` methods, e.g.
```
static IssuePtr create(const ModelPtr &model, ItemType type = ItemType::MODEL) noexcept;
```

src/importer.cpp
hsorby: There is no import source available at this point.

kerimoyle: Yes there is, it just needs to be passed in through the arguments of this function.  Otherwise this becomes an orphaned issue, but we do have things that it could be tied to.

src/api/libcellml/issue.h
nickerso: Is there an issue to track this one going forward? not something we need to worry about in this PR, right?

agarny: Yes, I guess this shouldn't stop this PR from being merged in. Ok, I am going to create an issue for it.

agarny: See issue #744.

src/api/libcellml/issue.h
nickerso: I agree with @agarny that this seems opposite from the create methods, but here we are dealing with an `any` item whereas the create objects have a concrete type - and you need to know the type before the item can be useful so I do think its more logical to have the type first.

agarny: The problem is that if we leave it as-is, we take the risk of end-users getting confused. I certainly would be and would think it is wrong. Whether we like it or not, we have established some kind of a convention with our `create()` methods. So, I really feel like we ought to stick to it.

src/issue.cpp
```diff
@@ -171,70 +236,299 @@ Issue::ReferenceRule Issue::referenceRule() const
...
+    return std::any_cast<VariablePair>(mPimpl->mItem);
 }
 
 void Issue::setVariable(const VariablePtr &variable)
 {
-    mPimpl->mVariable = variable;
-    mPimpl->mCause = Issue::Cause::VARIABLE;
+    if (variable != nullptr) {
+        setItem(ItemType::VARIABLE, variable);
+    } else if (mPimpl->mItemType == ItemType::VARIABLE) {
```
nickerso: I'm not understanding why all these `set*()` methods will only clear the current item if the given object is a `nullptr` and the current type matches. I'm not seeing why any call to `set*()` with a `nullptr` would not cause the current item to be cleared...

tests/issue/issue.cpp
```diff
@@ -616,3 +626,697 @@ TEST(Issue, undefinedIssueUrl)
...
+}
+
+TEST(Issue, cannotCrossClearStoredItem)
+{
+    auto model = libcellml::Model::create();
+    auto issue = libcellml::Issue::create(model);
+    EXPECT_EQ(model, issue->model());
+    EXPECT_EQ(libcellml::ItemType::MODEL, issue->itemType());
+
+    // Cannot clear by setting to nullptr through another type.
```
nickerso, line 1323: this is the behviour I am not sure I understand. Why does using a non-null object override whatever is there but a null object will respect the existing type. (Hoping this isn't opposite to anything I've said earlier!)

nickerso (dismissed):
Not convinced the `Issue::set*()` methods are working in a way I understand, and there still seem to be a few unresolved discussions, but not seeing anything that must be resolved now.

src/api/libcellml/issue.h
kerimoyle: The signature mimics the definition of the `AnyItem` type, which is a pair with the first item as the item type and the second is the item.  This just means that the user doesn't have to call `std::make_pair` in order to pass something in as an AnyItem type.

agarny: As mentioned before, I don't like the official use of `std::pair`. That use should somehow be hidden (through the use of a class for instance, as suggested by @hsorby). This means that we can then be consistent when it comes to the order of parameters.

src/api/libcellml/issue.h
```diff
@@ -188,21 +195,27 @@ class LIBCELLML_EXPORT Issue
...
         MAP_VARIABLES_VARIABLE1,
         MAP_VARIABLES_VARIABLE2,
         MAP_VARIABLES_UNIQUE,
         MAP_VARIABLES_AVAILABLE_INTERFACE,
 
         // Issues not present in the normative specification:
         MAP_VARIABLES_IDENTICAL_UNIT_REDUCTION,
+
+        // Placeholder for further references:
+        TODO
```
nickerso: not a fan of `TODO` - anything that is todo should not be going into the code :) otherwise why not `UNDEFINED`?

src/parser.cpp
nickerso: has one now, looks like.

nickerso (commented):
Not sure if the new `TODO` is an indication this is still work in progress...

src/api/libcellml/issue.h
kerimoyle: It needs to be something separate from `UNDEFINED` ... maybe `NOT_DEFINED_YET`? It's just a placeholder until we've written the list of links to the docs and issue types.

nickerso: so that is just a reference that doesn't exist, right? so its undefined. once its defined it becomes defined...

kerimoyle: I interpret `undefined` to mean `I don't know what on earth you've done but it's way too freaky to have its own error type`.  What this enum is meant to be is `legit mistake, here's the docs, oh oops we haven't written that yet`.

tests/issue/issue.cpp
hsorby: I'll have another look at that.

src/api/libcellml/issue.h
kerimoyle: I've changed it to `NOT_SPECIFIED` in 54c49aa4

src/api/libcellml/issue.h
```diff
@@ -195,21 +195,27 @@ class LIBCELLML_EXPORT Issue
...
         MAP_VARIABLES_VARIABLE1,
         MAP_VARIABLES_VARIABLE2,
         MAP_VARIABLES_UNIQUE,
         MAP_VARIABLES_AVAILABLE_INTERFACE,
 
         // Issues not present in the normative specification:
         MAP_VARIABLES_IDENTICAL_UNIT_REDUCTION,
+
+        // Placeholder for further references:
+        NOT_SPECIFIED
```
agarny: I would call it `UNSPECIFIED` rather than `NOT_SPECIFIED`, so that we are being consistent with `UNDEFINED` (rather than `NOT_DEFINED`).

If agreed, then I am happy to update it.

src/api/libcellml/issue.h
agarny: Done in commit 6d858624.

hsorby:
> Fair point. I would be fine with with `UnitChild`. @nickerso, @hsorby: is that fine with you?

Not for me, it is a reference to a unit therefore `UnitReference` it is not a `UnitChild` if anything it would have to be a `UnitsChild`.

agarny:
> > Fair point. I would be fine with with `UnitChild`. @nickerso, @hsorby: is that fine with you?
> 
> Not for me, it is a reference to a unit therefore `UnitReference` it is not a `UnitChild` if anything it would have to be a `UnitsChild`.

Yes, `UnitsChild`, indeed.

src/api/libcellml/issue.h
agarny: Done in commit e5943638.

src/api/libcellml/enums.h
```diff
@@ -0,0 +1,46 @@
...
+#include <memory>
+
+namespace libcellml {
+
+/**
+ * @brief The CellMLReferenceType enum class.
+ *
+ */
+enum class CellMLReferenceType
+{
```
kerimoyle, line 35: Gah!  Why are we randomly changing names of stuff?  These are not references, they're elements ...? I don't get it ...

hsorby: The underlying things are elements but the what we are doing here is describing  how we reference CellML types.

tests/issue/issue.cpp
nickerso: I'm liking the update!

nickerso (dismissed):
subject to the change in the `CellMLReferenceType` discussed offline, this is looking all good to .

src/api/libcellml/enums.h
agarny: That new name should please you then, @kerimoyle. :)

agarny (dismissed):

src/api/libcellml/types.h
```diff
@@ -70,4 +74,133 @@ using UnitsPtr = std::shared_ptr<Units>; /**< Type definition for shared units p
...
+    static UnitPtr create(const UnitsPtr &units, size_t index) noexcept;
+
+    /**
+     * @brief Get the unit.
+     *
+     * Get the unit in the @ref Units.
+     *
+     * @return The unit.
+     */
+    UnitsPtr units() const;
```
kerimoyle, line 119: Why is this function called `units` when it returns a `unit` ?

kerimoyle: Or ... what does this function really do?  Return the pointer to the `Units` which this `Unit` depends on? Either way, the docs don't match what the function does.

agarny: It returns a `Units`, so if anything the documentation should be fixed to read: `@return The units.`?

agarny (approved):

nickerso (approved):

Issue #703, opened 2020-09-03
Fix cloning: cloned ownership links should mimic the source links


kerimoyle:
At present cloning Variables creates orphaned, duplicated Units items which are not added to the parent model.  
Cloning models does not handle imports well either.  Both need to be fixed before #702 can be addressed.

PR #704, opened 2020-09-03
Fix cloning behaviour

kerimoyle:
Addresses #703

### DIFF
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -531,22 +531,6 @@ ModelPtr Model::clone() const
         m->addComponent(component(index)->clone());
     }
 
-    for (size_t index = 0; index < importSourceCount(); ++index) {
-        m->addImportSource(importSource(index)->clone());
-    }
-
-    // Generate equivalence map starting from the models components.
-    EquivalenceMap map;
-    IndexStack indexStack;
-    for (size_t index = 0; index < componentCount(); ++index) {
-        indexStack.push_back(index);
-        auto c = component(index);
-        recordVariableEquivalences(c, map, indexStack);
-        generateEquivalenceMap(c, map, indexStack);
-        indexStack.pop_back();
-    }
-    applyEquivalenceMapToModel(map, m);
-
     for (size_t index = 0; index < m->componentCount(); ++index) {
         fixComponentUnits(m, m->component(index));
     }
@@ -562,6 +546,18 @@ ModelPtr Model::clone() const
         fixImportSourceUnits(i1, i2);
         fixImportSourceComponents(i1, i2);
     }
+
+    // Generate equivalence map starting from the models components.
+    EquivalenceMap map;
+    IndexStack indexStack;
+    for (size_t index = 0; index < componentCount(); ++index) {
+        indexStack.push_back(index);
+        auto c = component(index);
+        recordVariableEquivalences(c, map, indexStack);
+        generateEquivalenceMap(c, map, indexStack);
+        indexStack.pop_back();
+    }
+    applyEquivalenceMapToModel(map, m);
 
     return m;
 }
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -492,6 +492,28 @@ void fixComponentUnits(const ModelPtr &model, const ComponentPtr &component)
     }
 }
 
+void fixImportSourceUnits(const ImportSourcePtr &i1, ImportSourcePtr &i2)
+{
+    auto m2 = owningModel(i2);
+
+    // Go through all the imported units in this import source and update their sources.
+    for (size_t index = 0; index < i1->unitsCount(); ++index) {
+        auto u1 = i1->units(index);
+        auto u2 = m2->units(u1->name());
+        u2->setImportSource(i2);
+    }
+}
+
+void fixImportSourceComponents(const ImportSourcePtr &i1, ImportSourcePtr &i2)
+{
+    auto m2 = owningModel(i2);
+    for (size_t index = 0; index < i1->componentCount(); ++index) {
+        auto c1 = i1->component(index);
+        auto c2 = m2->component(c1->name(), true);
+        c2->setImportSource(i2);
+    }
+}
+
 ModelPtr Model::clone() const
 {
     auto m = create();
@@ -509,6 +531,10 @@ ModelPtr Model::clone() const
         m->addComponent(component(index)->clone());
     }
 
+    for (size_t index = 0; index < importSourceCount(); ++index) {
+        m->addImportSource(importSource(index)->clone());
+    }
+
     // Generate equivalence map starting from the models components.
     EquivalenceMap map;
     IndexStack indexStack;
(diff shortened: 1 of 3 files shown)


kerimoyle:
Ready for review please @agarny @hsorby @nickerso Should be short and sweet 😄

src/model.cpp
```diff
@@ -475,6 +475,45 @@ bool Model::hasImports() const
     return importsPresent;
 }
 
+void fixComponentUnits(const ModelPtr &model, const ComponentPtr &component)
+{
+    for (size_t v = 0; v < component->variableCount(); ++v) {
+        auto variable = component->variable(v);
+        if (variable->units() != nullptr) {
+            // Find the units in the model and switch out.
+            auto units = model->units(variable->units()->name());
```
nickerso, line 484: Not keen on using `name` to define the "same" units. Need to actually compare the units definitions...pretty sure there is existing functionality that does that.

tests/test_utils.cpp
```diff
@@ -345,6 +366,12 @@ void compareModel(const libcellml::ModelPtr &m1, const libcellml::ModelPtr &m2)
         auto c2 = m2->component(index);
         compareComponent(c1, c2, m2);
     }
+
+    for (size_t index = 0; index < m1->importSourceCount(); ++index) {
+        auto i1 = m1->importSource(index);
+        auto i2 = m2->importSource(index);
+        compareImportSource(i1, i2, m2);
```
nickerso, line 374: what happens if `m1` and `m2` have different numbers of import sources?

kerimoyle: m2 is the cloned model from m1.

kerimoyle: I guess we could assert they're the same first ...

src/model.cpp
```diff
@@ -475,6 +475,45 @@ bool Model::hasImports() const
...
+            if (units != nullptr) {
+                variable->setUnits(units);
```
agarny, line 486: What happens to the original `variable->units()`?

src/model.cpp
```diff
@@ -475,6 +475,45 @@ bool Model::hasImports() const
...
+            }
+        }
+    }
+    for (size_t c = 0; c < component->componentCount(); ++c) {
+        fixComponentUnits(model, component->component(c));
+    }
+}
+
+void fixImportSourceUnits(const ImportSourcePtr &i1, ImportSourcePtr &i2)
```
agarny, line 495: Why `ImportSourcePtr &i2` and not `const ImportSourcePtr &i2`?

tests/test_utils.cpp
kerimoyle: 2385edb6

src/model.cpp
kerimoyle: I wondered about that, except that according to the spec you can't have duplicate names, but you can have duplicate definitions.  Since the units are added to variables in the CellML format by their names, I figured this was the safer way to do it?

src/model.cpp
```diff
@@ -475,6 +475,45 @@ bool Model::hasImports() const
...
+{
+    auto m2 = owningModel(i2);
+
+    // Go through all the imported units in this import source and update their sources.
+    for (size_t index = 0; index < i1->unitsCount(); ++index) {
+        auto u1 = i1->units(index);
+        auto u2 = m2->units(u1->name());
```
agarny, line 502: Similar to https://github.com/cellml/libcellml/pull/704/files#r482676716.

src/model.cpp
kerimoyle: It disappears into the ether.  No references ... created inline ... so that if people called this as a `variable->clone()` function, they can still access the units if they want to.  But if it's come from a `model->clone()` call, then they get fixed up properly.

src/model.cpp
kerimoyle: Because it's changing.

agarny: Where? As far as I can tell, it's used to retrieve its owning model and set the import source of `u2`, nothing else.

src/model.cpp
agarny: Ok.

src/model.cpp
agarny: Some may decide to clone something in a not-yet-valid model, no?

src/model.cpp
```diff
@@ -475,6 +475,45 @@ bool Model::hasImports() const
...
+        u2->setImportSource(i2);
+    }
+}
+
+void fixImportSourceComponents(const ImportSourcePtr &i1, ImportSourcePtr &i2)
```
agarny, line 507: See https://github.com/cellml/libcellml/pull/704/files#r482683037.

src/model.cpp
```diff
@@ -475,6 +475,45 @@ bool Model::hasImports() const
...
+{
+    auto m2 = owningModel(i2);
+    for (size_t index = 0; index < i1->componentCount(); ++index) {
+        auto c1 = i1->component(index);
+        auto c2 = m2->component(c1->name(), true);
```
agarny, line 512: See https://github.com/cellml/libcellml/pull/704/files#r482676716.

agarny: I know we are dealing with a component here, and not a unit, but it's the same story: not sure we can only rely on the name of a component?

src/model.cpp
kerimoyle: Sure.  But that doesn't mean identifying units by their contents is correct either.  Since we actually use the names to identify things, I think it's better to continue with that philosophy. Duplicates or missing names should raise a warning though.

src/model.cpp
kerimoyle: in setting the import source of u2, the import source itself is updated.  Two way street ... see the setImportSource functions in the Component and Units classes.

src/model.cpp
kerimoyle: What else would you use that's safer?

src/model.cpp
kerimoyle: And see https://github.com/cellml/libcellml/pull/704/files#r482685749

src/model.cpp
agarny: Was being lazy, so thanks for pointing me in the right direction.

src/model.cpp
agarny: Yeah, not sure to be honest. I was just thinking about the case where you have a (currently invalid) model with two components with the same name.

agarny (commented):
Looks fine although I am not quite sure what can be done about https://github.com/cellml/libcellml/pull/704/files#r482676716.

src/model.cpp
nickerso: in a valid model you can't have multiple units with the same name...

nickerso:
I guess part of this is looking at those functions as general utilities - but in fact they are (currently) only going to be used in the very specific case of cloning a model. Which is fine as long as we don't come back in future and pull those methods out as general utilities :)

nickerso (approved):
Does what it says in the box - fixes the current cloning behaviour.

src/model.cpp
agarny: Sure, but here I am talking about a model that 1) is not yet valid, 2) has two units with the same name, and 3) then you do the cloning.

src/model.cpp
nickerso: I'm thinking that in this case we don't worry about invalid models. There seems to be no easy way to fix up these issues with cloning and having imported components or units with duplicate names is just going to screw up all efforts at trying to handle cloning a model correctly. Happy to leave that as a know feature and submit an issue to keep track of it. Assuming this one goes in like it is now.

agarny: Ok, fair enough. I am happy with this PR (besides the case where two units/components have the same name).

src/model.cpp
nickerso: I agree, but I can't see how that issue can be easily addressed. While this PR will fix the issue of cloning resulting in multiple copies of the same units (making a valid model invalid).

agarny (approved):

src/model.cpp
nickerso: For completeness, #705 is the issue to track this for future developers to fix.

PR #716, opened 2020-09-23
Validator should only check for units by name

### LINKED ISSUE #715: Should imported units be "found" by the validator?
```
TEST(ImportSource, KRM)
{
    auto component = libcellml::Component::create("component");
    auto importSource = libcellml::ImportSource::create();
    importSource->setUrl("somewhere.cellml");
    model->addImportSource(importSource);

    auto mV = libcellml::Units::create("mV");
    mV->setImportReference("mV");
    mV->setImportSource(importSource); // << removing this means that the model is valid ... 
    model->addUnits(mV);

    auto v = libcellml::Variable::create("variable");
    v->setUnits(mV);
    component->addVariable(v);

    EXPECT_EQ(model, mV->parent());
    EXPECT_TRUE(model->hasUnits("mV"));

    auto validator = libcellml::Validator::create();
    validator->validateModel(model);
    printIssues(validator);  // << this says it can't find the units ... :(
}
```

kerimoyle:
Addresses #715


src/validator.cpp
```diff
@@ -635,7 +635,7 @@ void Validator::ValidatorImpl::validateVariable(const VariablePtr &variable, con
     } else if (!isStandardUnitName(unitsName)) {
         ComponentPtr component = owningComponent(variable);
         ModelPtr model = owningModel(component);
-        if ((model != nullptr) && !model->hasUnits(variable->units())) {
+        if ((model != nullptr) && !model->hasUnits(unitsName)) {
```
agarny, line 688: This clearly changes everything and, looking at `Model::ModelImpl::findUnits()` (which is the method effectively used here), we can indeed see that the implementation is completely different depending on whether we pass it a `Units` or a string. I find this very weird and I am not sure I like it.

tests/validator/validator.cpp
```diff
@@ -2625,10 +2625,36 @@ TEST(Validator, refToUnitsByNameNeedsLinkUnitsToValidate)
     t2->setUnits("millisecond");
 
     validator->validateModel(model);
-    EXPECT_EQ(size_t(1), validator->errorCount());
+    EXPECT_EQ(size_t(1), validator->errorCount()); // KRM this will fail now since we *don't* need to link units anymore
```
agarny: I guess this test will become moot if we were to go with your proposed change?

kerimoyle: Exactly - that's why I'm asking.

src/validator.cpp
kerimoyle: I think it's ok to have a `hasUnits` function that also tests content fo the units; just not that it's used here as it doesn't come into validation.

hsorby (dismissed):

hsorby:
I think the `Validator` should only be checking that the reference is valid.

hsorby (changes requested):
I think the implementation is correct but the tests need to change to show the new understanding.

kerimoyle:
> I think the implementation is correct but the tests need to change to show the new understanding.

Cool, ta.  I'll remove that test and resubmit - it's not needed any more.

kerimoyle:
I've had to make a couple of other tweaks too and added a couple of small tests.  Ready for review please @agarny @hsorby @nickerso

src/validator.cpp
agarny: Maybe, but when you have a method with two different signatures, I would still expect that method to work the same while here `hasUnits()` does different things (through `findUnits()`) and that's what I find dodgy.

agarny (dismissed):
Apart from my general comment about `hasUnits()`, it looks good to me.

kerimoyle:
> Apart from my general comment about hasUnits(), it looks good to me.
@agarny I do take your point ... but even though that's been highlighted by this PR, it's not something that this PR is meant to address.  I'm just calling a different function in the validator, not changing how we search for anything.  Can we make that discussion a separate issue?

agarny:
> > Apart from my general comment about hasUnits(), it looks good to me.
> > @agarny I do take your point ... but even though that's been highlighted by this PR, it's not something that this PR is meant to address.  I'm just calling a different function in the validator, not changing how we search for anything.  Can we make that discussion a separate issue?

Agreed and that's exactly why I still approved your PR despite my concern about the different implementations of `hasUnits()`, for which I have just created an issue (see issue #719).

kerimoyle:
Great, thanks @agarny :)

src/units.cpp
```diff
@@ -453,18 +456,17 @@ double Units::scalingFactor(const UnitsPtr &units1, const UnitsPtr &units2, bool
     bool updateUnits2 = false;
 
     if ((units1 != nullptr) && (units2 != nullptr)) {
-        double multiplier = 0.0;
+        if ((owningModel(units1) == owningModel(units2)) && (units1->name() == units2->name())) {
+            return 1.0;
+        }
```
hsorby: I don't understand this addition, why should I care about who owns the units.  What does it have to do with the validator only checking units by name.

tests/units/units.cpp
```diff
@@ -2469,3 +2469,52 @@ TEST(Units, addUnitsMultipleTimes)
...
+    u2->addUnit("second");
+    u1->addUnit("second", "milli");
+
+    auto scaling = libcellml::Units::scalingFactor(u1, u5);
+    EXPECT_EQ(0.0, scaling);
+}
+
+TEST(Units, scalingFactorBetweenSameUnits)
+{
+    auto model = libcellml::Model::create("model");
```
hsorby, line 2522: What is model doing here?

tests/units/units.cpp
```diff
@@ -2469,3 +2469,52 @@ TEST(Units, addUnitsMultipleTimes)
...
+    auto u1 = libcellml::Units::create("units");
+    u1->addUnit("second", "milli");
+    model1->addUnits(u1);
+
+    auto model2 = model1->clone();
+    auto u2 = model2->units(0);
+
+    auto scaling = libcellml::Units::scalingFactor(u1, u2);
+    EXPECT_EQ(1.0, scaling);
+}
```
hsorby, line 2544: Where is the co-test for this a non-unit scaling factor for units of different scale?

hsorby (dismissed):

src/units.cpp
kerimoyle: It's perfectly valid to have two units in separate models with the same name.  There was a check lower down the function that returned a scaling factor of 1 if the two units had the same name.  That's only "safe" (ish) when they're from the same model.  It's not directly related to the validation stuff, but when I changed that another test broke, so this is making sure that all our current tests continue to pass even with the change.

tests/units/units.cpp
kerimoyle: See line 196 in src/units.cpp.  In order to check for child unit references, we need to have access to the other units in the model... so it needs a model.

tests/units/units.cpp
kerimoyle: There are loads of tests for scaling factors earlier in this file - I don't quite understand which aspect is missing?  This test was intended to show that even though these two have the same name, they're still tested properly.  I guess changing the  scale away from 1 would show that more clearly - is that what you meant?

src/units.cpp
hsorby: It was further down for a reason right.  After all else is said and done if the units have the same name then they are considered to have a scaling factor of `1.0` given this is the case by definition.

tests/units/units.cpp
kerimoyle: 6008b53

tests/units/units.cpp
hsorby: But `model` is not used in this test?

src/units.cpp
kerimoyle: I'm not understanding two parts of that:
> It was further down for a reason right.

There is no change to the functionality at all - the only thing that's different is that we can skip a whole lot of recursive calculation if we are going to return a value based on the name in any case.  I don't understand how doing those extra calculations is a good reason to have that test further down the function?

> After all else is said and done if the units have the same name then they are considered to have a scaling factor of 1.0 given this is the case by definition.

I don't agree.  Why would they be considered the same? We're checking units definitions here?

tests/units/units.cpp
kerimoyle: Duh! I get it now.  Changed in a8c89fdb

kerimoyle:
Ready for re-review please @agarny @hsorby @nickerso

agarny (dismissed):

agarny:
Ping @hsorby and @nickerso for re-reviewing this PR.

src/units.cpp
nickerso: I think this is still an unresolved issue and this change breaks the way this method should work

agarny (dismissed):

nickerso (dismissed):

agarny (approved):

nickerso (approved):

PR #718, opened 2020-09-24
i717 Model::fixVariableInterfaces should not downgrade public_and_private to private

### LINKED ISSUE #717: Model::fixVariableInterfaces() should not downgrade existing types
The fix function should not change existing `public_and_private` types to be `private`.

kerimoyle:
Addresses #717 

This introduces a minimum requirement when fixing the interface types.  The new `hasMinimumInterfaceType` will return:

     *    Parameter (right) /
     *   Stored value (below) | none | public | private | public_and_private  
     *   ---------------------+------+--------+---------+-------------------
     *                   none | T    | F      | F       | F
     *   ---------------------+------+--------+---------+-------------------
     *                 public | T    | T      | F       | F
     *   ---------------------+------+--------+---------+-------------------
     *                private | T    | F      | T       | F
     *   ---------------------+------+--------+---------+-------------------
     *     public_and_private | T    | T      | T       | T

### DIFF
--- a/src/api/libcellml/variable.h
+++ b/src/api/libcellml/variable.h
@@ -435,12 +435,16 @@ public:
      * Test if this variable permits access through the @p interfaceType. The results
      * will be given according to this truth table:
      * 
-     *    argument : stored interface type = return value
-     *    (anything) : public_and_private = true
-     *    private : public = false
-     *    public : private = false
-     *    public, private, public_and_private : none = false
-     *    none : (anything) = true
+     *    Parameter (right) /
+     *   Stored value (below) | none | public | private | public_and_private  
+     *   ---------------------+------+--------+---------+-------------------
+     *                   none | T    | F      | F       | F
+     *   ---------------------+------+--------+---------+-------------------
+     *                 public | T    | T      | F       | F
+     *   ---------------------+------+--------+---------+-------------------
+     *                private | T    | F      | T       | F
+     *   ---------------------+------+--------+---------+-------------------
+     *     public_and_private | T    | T      | T       | T
      *
      * @param interfaceType The interface type to test for.
      *
--- a/src/api/libcellml/variable.h
+++ b/src/api/libcellml/variable.h
@@ -430,6 +430,25 @@ public:
     bool hasInterfaceType(InterfaceType interfaceType) const;
 
     /**
+     * @brief Test if this variable permits access through the @p interfaceType.
+     *
+     * Test if this variable permits access through the @p interfaceType. The results
+     * will be given according to this truth table:
+     * 
+     *    argument : stored interface type = return value
+     *    (anything) : public_and_private = true
+     *    private : public = false
+     *    public : private = false
+     *    public, private, public_and_private : none = false
+     *    none : (anything) = true
+     *
+     * @param interfaceType The interface type to test for.
+     *
+     * @return @c true if the interface type is acceptable, @c false otherwise.
+     */
+    bool hasMinimumInterfaceType(InterfaceType interfaceType) const;
+
+    /**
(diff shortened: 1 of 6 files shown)


kerimoyle:
How is it that the tests on my machine (mistakenly) pass tests for missing docstrings? I never find them the first time around :(
image

kerimoyle:
Ready for review please.  @agarny @hsorby @nickerso

src/api/libcellml/variable.h
```diff
@@ -429,6 +429,25 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
...
+     *
+     * Test if this variable permits access through the @p interfaceType. The results
+     * will be given according to this truth table:
+     * 
+     *    argument : stored interface type = return value
+     *    (anything) : public_and_private = true
+     *    private : public = false
+     *    public : private = false
+     *    public, private, public_and_private : none = false
+     *    none : (anything) = true
```
agarny: This truth table doesn't match the one in https://github.com/cellml/libcellml/pull/718#issue-492718408. No only the "columns" are inverted, but for a `public` stored interface type, to have a `private` argument gives true in one case and false in the other...?!

So, before reviewing further, it would be good to know what is supposed to be correct. Also, I would reformat your truth table. I find it unnecessarily complicated to read and use. Why not simply something like:

```
                   | none | public | private | public_and_private   <--- Parameter
-------------------+------+--------+---------+-------------------
              none |      |        |         |
-------------------+------+--------+---------+-------------------
            public |      |        |         |
-------------------+------+--------+---------+-------------------
           private |      |        |         |
-------------------+------+--------+---------+-------------------
public_and_private |      |        |         |

        ^
        |
 Stored interface
       type
```

kerimoyle: > Why not simply something like

Because they are a PITA to create ;). OK, I'll check it.

kerimoyle: > but for a public stored interface type, to have a private argument gives true in one case and false in the other...?!
I can't see that anywhere?

kerimoyle: a01c97b5

agarny: I meant that in https://github.com/cellml/libcellml/pull/718#issue-492718408, you originally had:
```
mInterfaceType : parameter -> returns
public : private, public_and_private -> true
```

while from your code above
```
     *    argument : stored interface type = return value
     *    private : public = false
```

kerimoyle: Yep, found it and changed comment above. Thanks.

src/variable.cpp
```diff
@@ -485,6 +485,33 @@ bool Variable::hasInterfaceType(InterfaceType interfaceType) const
...
+    // Stored value (below) | none | public | private | public_and_private
+    // ---------------------+------+--------+---------+-------------------
+    //                 none | T    | F      | F       | F
+    // ---------------------+------+--------+---------+-------------------
+    //               public | T    | T      | F       | F
+    // ---------------------+------+--------+---------+-------------------
+    //              private | T    | F      | T       | F
+    // ---------------------+------+--------+---------+-------------------
+    //   public_and_private | T    | T      | T       | T
+
```
hsorby: We don't need this here.

hsorby (commented):
Some extra comments that aren't necessary.

src/variable.cpp
kerimoyle: b880139

nickerso (dismissed):

hsorby (dismissed):

agarny (approved):

agarny (approved):

nickerso (approved):

Issue #729, opened 2020-09-30
Segfault caused by parser


kerimoyle:
New segfault popping up in the `requiresImport` function.

kerimoyle:
I've tracked it down, and it's something in the `Parser` instead.  Here's my test:

segfault.cellml
```
<?xml version="1.0" encoding="UTF-8"?>
<model xmlns="http://www.cellml.org/cellml/2.0#" name="segfault">
  <import xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="importMe.cellml">
    <component component_ref="importMe" name="importedChild"/>
  </import>
  <component name="parentName!"/> <<<<<<<<< NB invalid name in this component DOES NOT match component_ref below
  <encapsulation>
    <component_ref component="parentName">
      <component_ref component="importedChild"/>
    </component_ref>
  </encapsulation>
</model>
```

importMe.cellml
```
<?xml version="1.0" encoding="UTF-8"?>
<model xmlns="http://www.cellml.org/cellml/2.0#" name="ImportMe">
  <component name="importMe"/>
</model>
```

```
TEST(Importer, KRM) {
    auto parser = libcellml::Parser::create();
    auto model = parser->parseModel(fileContents("importer/KRM/segfaultImports.cellml"));
    auto importer = libcellml::Importer::create();
    importer->resolveImports(model, resourcePath("importer/KRM/")); // <<< Segfaults because a component returns a nullptr
```

The error is coming from the `Parser`, where the imported components are being created, but because of the bad name, the children of that component are being ditched.

PR #730, opened 2020-09-30
Parser bug: components no longer orphaned

kerimoyle:
Addresses #729

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -1034,7 +1034,7 @@ ComponentPtr Parser::ParserImpl::loadComponentRef(const ModelPtr &model, const X
         }
         attribute = attribute->next();
     }
-    if ((!parentComponent) && (parentComponentName.empty())) {
+    if (!parentComponent && parentComponentName.empty()) {
         IssuePtr issue = Issue::create();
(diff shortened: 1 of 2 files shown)


src/parser.cpp
```diff
@@ -1074,9 +1074,15 @@ ComponentPtr Parser::ParserImpl::loadComponentRef(const ModelPtr &model, const X
             mParser->addIssue(issue);
         }
 
-        if ((parentComponent) && (childComponent)) {
+        if (childComponent) {
             // Set parent/child encapsulation relationship.
-            parentComponent->addComponent(childComponent);
+            if (parentComponent) {
```
agarny, line 1079: Should read: `if (parentComponent != nullptr) {`.

agarny (changes requested):
Just one minor change.

src/parser.cpp
kerimoyle: OK ... then I'll also change the line above which reads `if(childComponent) {` too ... !

kerimoyle: Actually ... this is the idiom throughout the whole of the parser :(

agarny: Ok, forget about it then. Will have to be done as part of our big cleaning up.

agarny (approved):

nickerso (approved):

Issue #733, opened 2020-10-01
Importer should clear issues at the time of each resolveImports call.


kerimoyle:
At present we don't clear issues from the importer anywhere.  I think we should do it for subsequent calls to `resolveImports`.

nickerso:
This sounds familiar - didn't we discuss it previously?! :-)

what is the behaviour of the other logging classes? Does the parser clear issues when parsing a new model? Although I'm not convinced that this is really the same case, as you can easily imagine the case where a user wants to resolve imports in a bunch of models and then look for issues - so I'd prefer not automagically clearing issues on a call to `resolveImports`.

kerimoyle:
Yes.  The validator, parser, analyser, everything else clears old errors.  Unless we're prepared to have different families of issues each tied to a different model (which is best done with the model as a logger, and we've certainly been around that maypole), then this is the only way to go?  If the user wanted to resolve a bunch of models, that's fine, they just have to check for issues after each call - same as every other utility class.

agarny:
I am in favour of clearing old issues before doing a particular task. Think of a compiler, you don't need to clear previous issues before recompiling a model.

nickerso:
think of this as the complier clearing issues for each include file for a given source, rather than reporting all issues with compilation of the source file. How is that for abstract! :) 

the user should be able to determine if they are finished with a task and want issues cleared before starting a new task. So at a minimum, I'd like to see a `clearIssues` argument to the resolve method, if not leaving it completely out of resolve and the user just calls the logger clear method directly (which I assume is accessible?).

kerimoyle:
> think of this as the complier clearing issues for each include file for a given source, rather than reporting all issues with compilation of the source file. How is that for abstract! :)

Nooooo ... it's not at all.  Just between calling resolveImports for the whole entire mother-model, not for each individual import dependency.

kerimoyle:
At present we have:
parser->parseModel(modelString1) -> stores issues associated with parsing the first model string
parser->parseModel(modelString2) ->**clears** issues associated with the first operation, stores issues associated with the second.

Ditto:
- the validator
- the printer
- the analyser

There's no reason that we'd expect people to want to resolve a bunch of imported models and keep the errors when we don't let them parse, validate, analyse or print them like that ...

PR #734, opened 2020-10-01
Clear importer issues

kerimoyle:
Addresses #733.  The issues stored in the `Importer` class's logger will be cleared before working through subsequent calls to `resolveImports`.  The call to `resolveImports` will also start from scratch, so that resolved imports are re-fetched at that time.

NB: This is built on top of #696 so should be merged after that one.  The only changes here are from commit: 9e8ae90036c094f9c5d842eda05a6e431a641376 (and onwards, if any).


nickerso:
> Addresses #733. The issues stored in the `Importer` class's logger will be cleared before working through subsequent calls to `resolveImports`. 

I think discussion on #733 needs to be resolved first.

> The call to `resolveImports` will also start from scratch, so that resolved imports are re-fetched at that time.

This concerns me greatly, as it is not only a very different issue to clearing the logs but it also sounds a lot like something we should not be doing. The whole point of the importer is to cache imported models so that they don't always need to be re-fetched each time they are needed.

kerimoyle:
> The whole point of the importer is to cache imported models so that they don't always need to be re-fetched each time they are needed.

The previously imported stuff won't be imported again, it's just that links to models will be refreshed.  That way, we're able to call `resolveImports` more than once (so you can address the issues it finds), and you can be sure that the latest and greatest models are always the ones it uses.

PR #735, opened 2020-10-01
Segfault in analyser :(

kerimoyle:
Help! @agarny I can't find what's wrong with this ... :(  PR just to demonstrate test with segfault in it.

### DIFF
--- a/tests/analyser/analyser.cpp
+++ b/tests/analyser/analyser.cpp
@@ -19,7 +19,7 @@ limitations under the License.
 #include "gtest/gtest.h"
 
 #include <libcellml>
-
+#if 0
 TEST(Analyser, initialisedVariableOfIntegration)
 {
     auto parser = libcellml::Parser::create();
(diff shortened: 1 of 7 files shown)


agarny:
Argh, ok, I am going to to look into it.

agarny:
Ok, FWIW, I have found where the problem is. It's an `AnalyserInternalEquation` object which variable is `nullptr`. So, now, I need to understand why that is the case...

agarny:
Ok, I have found the cause of the problem: you have two equations that compute `i_stim` (in `tests/resources/KRM/Controller.cellml` and `tests/resources/KRM/MembraneModel.cellml`). When the analyser comes across the first equation, it properly considers it as being used to compute `i_stim`. Then, when it comes across the second equation, everything is already known and, therefore, that equation status remains "unknown" (hence the `nullptr` variable). That's clearly a case I didn't account for. I am therefore going to create a proper issue for it and a corresponding PR. Argh!

agarny:
Actually, I do account for that case (since it should result in an overconstained variable, something that the analyser can report), but clearly not properly.

kerimoyle:
Whew, I'm glad you found the problem!  I'll fix up my demo model (which I am rapidly wanting to rename "demon") and keep going ... :)

agarny:
Closing in favour of PR #738.

PR #749, opened 2020-10-07
NEW Issue class to make use of std::any type

### LINKED ISSUE #689: Should the Issue class use std::any

kerimoyle:
This contains the changes from #696 with the merge conflicts fixed with #316 

Addresses #689.

---

The plan is to use this PR instead of PR #696. For this, we need to get this PR to pass CI. If it does then we can drop PR #696. If, for some reasons, we can't then we need to drop this PR and go back to PR #696 (hopefully, this won't be needed!).


agarny:
Deleteing in favour of PR #696.

PR #756, opened 2020-10-19
Updated parser test to include alternative error messages from Ubuntu 20

### LINKED ISSUE #751: Failing `io_unit_test_parser` test
```
99% tests passed, 1 tests failed out of 67

Total Test time (real) =  18.64 sec

The following tests FAILED:
	 15 - io_unit_test_parser (Failed)
Errors while running CTest
make: *** [Makefile:172: test] Error 8
```

kerimoyle:
Closes #751

### DIFF
--- a/tests/parser/parser.cpp
+++ b/tests/parser/parser.cpp
@@ -34,9 +34,10 @@ TEST(Parser, invalidXMLElements)
         "  <Wizard>Gandalf</SomeGuyWithAStaff>\n"
         "  <Elf>\n"
         "</fellows>\n";
-    const std::vector<std::string> expectedIssues = {
-        "Specification mandate value for attribute bearded.",
-        "Specification mandates value for attribute bearded.",
+
+    // Depending on the libXML2 used by different Ubuntu installations, the error messages differ.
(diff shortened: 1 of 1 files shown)


kerimoyle:
@agarny Please can you confirm that this fixes the problem on the Ubuntu 20 build?

agarny:
> @agarny Please can you confirm that this fixes the problem on the Ubuntu 20 build?

I can confirm that everything is now fine on Ubuntu 20.04.

kerimoyle:
Ready for review please @agarny @nickerso @hsorby

agarny (dismissed):

hsorby (dismissed):
I think I would have had the `expectedIssues*` named after the libxml2 version rather than Ubuntu version as this is the source of the difference but I can live with the current names.

agarny:
> I think I would have had the `expectedIssues*` named after the libxml2 version rather than Ubuntu version as this is the source of the difference but I can live with the current names.

I actually had the same thought, so maybe it should be done? I mean, who knows when we will next look into this, so we might as well have the correct information in?

agarny (approved):

hsorby (approved):

PR #757, opened 2020-10-19
Return one issue only from missing or illegal identifiers

### LINKED ISSUE #754: Missing units causes extra validation error
See https://github.com/kerimoyle/libcellml-tutorials/issues/6

Missing unit should only trigger issue 1, not issue 0:
```
The validator has found 8 issue(s).
Issue 0: CellML identifiers must contain one or more basic Latin alphabetic characters.
  reference: 1.3.1.1
  see: https://cellml-specification.readthedocs.io/en/latest/reference/formal_and_informative/specA03.html?issue=1.3.1.1
  stored item type: UNDEFINED

Issue 1: Variable 't' in component 'gateEquations' does not have a valid units attribute. The attribute given is ''.
  reference: 2.8.1.2
  see: https://cellml-specification.readthedocs.io/en/latest/reference/formal_and_informative/specB08.html?issue=2.8.1.2
  stored item type: VARIABLE
```

This might make sense from a libcellml developer point of view, but absolutely doesn't from a users point of view :D
There is no invalid identifier in the model!

kerimoyle:
Closes #754

### DIFF
--- a/src/validator.cpp
+++ b/src/validator.cpp
@@ -70,13 +70,13 @@ void updateBaseUnitCount(const ModelPtr &model,
                          double uExp, double logMult, int direction);
 
 /**
- * @brief Utility function to construct an @c Issue if required for a given CellML identifier string. 
- * 
- * Utility function to construct an @c Issue with the appropriate description for 
- * a given identity string.
- * 
+ * @brief Utility function to construct an @c Issue if required for a given CellML identifier string.
+ *
+ * Utility function to construct an @c Issue with the appropriate description for
+ * a given CellML identifier string.
+ *
  * @param name The @c std::string identifier to check.
- * 
+ *
  * @return An @c IssuePtr if the @p name is illegal, @c nullptr otherwise.
  */
 IssuePtr makeIssueIllegalIdentifier(const std::string &name);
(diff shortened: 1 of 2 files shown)


kerimoyle:
Ready for review please @agarny @nickerso @hsorby

hsorby (dismissed):
I have consolidated the test for a valid CellML identifier into a single function which I think was necessary.  With that change I am approving this PR.

kerimoyle:
> I have consolidated the test for a valid CellML identifier into a single function which I think was necessary. With that change I am approving this PR.

Please can we have discussions about changes before making them?  It seems pretty backward to have to raise and discuss an issue before fixing it, but random changes can be made and then approved ...

hsorby (dismissed):

src/validator.cpp
```diff
@@ -247,8 +277,38 @@ struct Validator::ValidatorImpl
...
+     * @brief Validate the text of a @c cn element.
+     *
+     * Validates that the @c cn element has a units attached.
+     *
+     * @param component The component that the @c cn element belongs to.
+     * @param unitsName The name of the units.
+     * @param textNode The text of the @c cn element.
+     *
+     * @return  @c true if the @c cn units is valid, @c false otherwise.
+     */
```
agarny, line 311: Sometimes, we talk about `ci`/`cn` element and sometimes a `ci`/`cn` node. Would be good to be consistent.

agarny (approved):
Just a minor comment, but wouldn't hold this PR because of it.

hsorby (approved):

PR #761, opened 2020-10-21
CMake: disable some warnings that have been introduced in LLVM 11

agarny:
Addresses issue #760.

PR #763, opened 2020-10-26
Analyser/Generator: cache variable equivalences

### LINKED ISSUE #762: Analyser/Generator: cache variable equivalences
To determine whether two variables are equivalent takes longer using CellML 2.0 than CellML 1.0/1.1 since directionality at the connection level has been dropped in the former. This is clearly a big bottle neck since both the analyser and the generator need to do a lot of those checks.

So, ideally, our `AnalyserModel` class could be used to retrieve whether two variables are equivalent (rather than use the `isSameOrEquivalentVariable()` utility) and cache the results as we go along.

agarny:
Addresses issue #762.


agarny:
Some quick tests show that, on my Mac, `Generator.fabbriFantiniWildersSeveriHumanSanModel2017` is now ~33% faster to run while `Generator.garnyKohlHunterBoyettNobleRabbitSanModel2003` is ~27.5% faster. Better than nothing, I guess, and no doubt that there is room for improvement in other places.

cmake/common.cmake
```diff
@@ -124,9 +124,16 @@ function(configure_clang_and_clang_tidy_settings _TARGET)
 
   if(CLANG_TIDY_AVAILABLE)
```
nickerso: nothing in the commit log to describe why these changes are needed, suggest pulling out into a separate PR along with the issue that these changes are addressing.

nickerso: (not sure how to comment on a specific file, the above comment refers to `cmake/common.cmake` and all the changes in there.)

src/api/libcellml/analysermodel.h
```diff
@@ -461,6 +461,26 @@ class LIBCELLML_EXPORT AnalyserModel
      */
     bool needAcothFunction() const;
 
+    /**
+     * @brief Test to determine if @p variable1 and @p variable2 are the same or
+     * (directly or indirectly) equivalent.
+     *
+     * Test to see if @p variable1 is the same or (directly or indirectly)
```
nickerso: no need for all the (directly or indirectly) - equivalence is equivalence.

src/api/libcellml/analysermodel.h
```diff
@@ -461,6 +461,26 @@ class LIBCELLML_EXPORT AnalyserModel
...
+     *
+     * @param variable1 The @c Variable to test if it is the same or (directly
+     * or indirectly) equivalent to @p variable2.
+     * @param variable2 The @c Variable that is potentially the same or
+     * (directly or indirectly) equivalent to @p variable1.
+     *
+     * @return @c true if @p variable1 is the same or (directly or indirectly)
+     * equivalent to @p variable2 and @c false otherwise.
+     */
+    bool isSameOrEquivalentVariable(const VariablePtr &variable1,
```
nickerso: convince me why this method should be checking "same" and "equivalent"...

cmake/common.cmake
agarny: Agreed with your first comment. It comes from https://github.com/cellml/libcellml/pull/761/files, which I need on my Mac since I am now on LLVM 11. So, if anything, I should have made it clear that this PR relies on PR #761.

src/api/libcellml/analysermodel.h
agarny: Directly or indirectly doesn't refer to directionality. IOW, in `A <-> B <-> C`, `A` is directly equivalent to `B` and indirectly equivalent to `C`.

src/api/libcellml/analysermodel.h
agarny: This PR is not fully ready to review and, indeed, I just refactored that method without changing its name. IOW, this method probably ought to be renamed `areEquivalentVariables()`? Also, it should take two `AnalyserVariablePtr`s as arguments, not two `VariablePtr`s.

agarny: Partly done in commit 1bdb65338. About using passing two `AnalyserVariablePtr`s rather than two `VariablePtr`s, not sure it can be done in the end. Indeed, our generator relies on `AnalyserModel::areEquivalentVariables()` to retrieve the corresponding `AnalyserVariablePtr` of a `VariablePtr`...

However, I am going to have a quick go at what @kerimoyle described at https://github.com/cellml/libcellml/issues/762#issuecomment-716799564.

agarny: Ok, from our discussion at today's libCellML meeting, it looks like this conversation can be considered as resolved.

agarny:
Ready to be (re-)reviewed, @hsorby, @kerimoyle and @nickerso.

src/generator.cpp
```diff
@@ -198,6 +198,7 @@ AnalyserVariablePtr Generator::GeneratorImpl::analyserVariable(const VariablePtr
 double Generator::GeneratorImpl::scalingFactor(const VariablePtr &variable) const
 {
     // Return the scaling factor for the given variable.
+
```
hsorby: I would just delete this the comment adds nothing here this method is self documenting.

hsorby (dismissed):
Minor change for me otherwise happy with this.

src/generator.cpp
agarny: Done in commit 2fc440e6f.

agarny:
Ready to be (re-)reviewed, @hsorby, @kerimoyle and @nickerso.

src/analysermodel.cpp
```diff
@@ -16,10 +16,39 @@ limitations under the License.
...
 
+#include <algorithm>
+
+#include "libcellml/variable.h"
+
 #include "analysermodel_p.h"
 
 namespace libcellml {
 
+bool AnalyserModel::AnalyserModelImpl::haveEquivalentVariables(const Variable *variable1,
```
nickerso: why is this "have" ? is it not checking that variable1 and variable2 are equivalent?

src/analysermodel.cpp
```diff
@@ -348,4 +377,35 @@ bool AnalyserModel::needAcothFunction() const
...
+                                           const VariablePtr &variable2)
+{
+    // We used to have a utilities method which implementation was:
+    //
+    //     return (variable1 == variable2)
+    //            || variable1->hasEquivalentVariable(variable2, true);
+    //
+    // However, a call to Variable::hasEquivalentVariable() can be time
+    // consuming. So, here, we stripped down the implementation of that method
+    // (and of others that it calls) and cache its results for future re-use.
```
nickerso: Is this not a bad thing to do? surely `Variable::hasEquivalentVariable()` should be the only place doing this kind of check? And here is just caching the results of lookups to improve repeated calls?

src/api/libcellml/analysermodel.h
nickerso: nothing to do with directionality - simply trying to say that direct or indirect equivalence makes no difference here so no need to keep mentioning it in the documentation.

nickerso (changes requested):

src/api/libcellml/analysermodel.h
agarny: Argh, fair point. Fixed in commit e168fc038.

src/analysermodel.cpp
agarny: Yes, renamed to `areEquivalentVariables()` in commit c168cc9a1.

src/analysermodel.cpp
agarny: Yes, this is definitely a bad thing to do, but one that I did at some point to gain as much speed up as possible. However, my implementation back then was clearly different since I have now just tested again using the original utility and, sure, it's slower, but somewhat negligible compared to the case where we don't cache the results. So, yes, I have "reverted" things a bit in commits bb742dcd4 and 403112a6b.

agarny:
Ready to be (re-)reviewed, @hsorby, @kerimoyle and @nickerso.

src/api/libcellml/analysermodel.h
```diff
@@ -461,6 +461,25 @@ class LIBCELLML_EXPORT AnalyserModel
      */
     bool needAcothFunction() const;
 
+    /**
+     * @brief Test to determine if @p variable1 and @p variable2 are
+     * equivalent.
+     *
+     * Test to see if @p variable1 is the same or equivalent to @p variable2.
+     * Returns @c true if @p variable1 is equivalent to @p variable2 and
+     * @c false otherwise.
```
nickerso, line 480: Not sure if this needs to be extended to mention that this version of equivalence checking will be using information cached in the `AnalyzerModel` ?

nickerso (dismissed):
minor comment, but suspect its not an issue so happy to approve.

src/api/libcellml/analysermodel.h
agarny: Whether it's cached or not is purely about implementation. When it comes to the API, it doesn't matter in my view. So, would leave it as-is, unless @hsorby and/or @kerimoyle feel otherwise.

agarny:
Please note that this PR currently makes use of PR #761, so maybe that bit of code should be reverted since we have yet to decide on issue #765...? Views?

agarny:
Ok, we are back on track. IOW, this PR is now ready for (re-)reviewing. Ping @hsorby, @kerimoyle and @nickerso.

src/analyser.cpp
```diff
@@ -272,7 +274,7 @@ bool AnalyserInternalEquation::check(size_t &equationOrder, size_t &stateIndex,
         for (size_t i = 0; i < mComponent->variableCount(); ++i) {
             auto localVariable = mComponent->variable(i);
 
-            if (isSameOrEquivalentVariable(variable->mVariable, localVariable)) {
+            if (model->areEquivalentVariables(variable->mVariable, localVariable)) {
```
kerimoyle, line 285: I don't understand why this function needs to be part of a model class?

kerimoyle: (ditto below)

src/analysermodel.cpp
```diff
@@ -348,4 +349,21 @@ bool AnalyserModel::needAcothFunction() const
     return mPimpl->mNeedAcothFunction;
 }
 
+bool AnalyserModel::areEquivalentVariables(const VariablePtr &variable1,
+                                           const VariablePtr &variable2)
+{
+    auto key = reinterpret_cast<intptr_t>(variable1.get()) * reinterpret_cast<intptr_t>(variable2.get());
+    auto cacheKey = mPimpl->mCachedEquivalentVariables.find(key);
+
```
kerimoyle, line 384: Do I have this correct: You're keeping track of pairs of variables by storing the product of their pointers, after the pointers have been cast to integers? How can that be guaranteed to be unique? Most of the time, sure ... but it's a bit weird? How is it faster than caching some other format that would be unique (and a lot cleaner) like a binned combination of indices?  And if you do want to keep it like this, why does it need to be a signed integer pointer cf. unsigned?  I'm so confused ...

src/utilities.h
```diff
@@ -431,23 +431,21 @@ bool isStandardPrefixName(const std::string &name);
...
 
 /**
- * @brief Test to determine if @p variable1 and @p variable2 are the same or
- * (directly or indirectly) equivalent.
+ * @brief Test to determine if @p variable1 and @p variable2 are equivalent.
  *
- * Test to see if @p variable1 is the same or (directly or indirectly)
- * equivalent to @p variable2. Returns @c true if @p variable1 is the same or
- * (directly or indirectly) equivalent to @p variable2 and @c false otherwise.
+ * Test to see if @p variable1 is the same or equivalent to @p variable2.
```
kerimoyle: ... the same AS or equivalent to ... (also below)

src/analysermodel.cpp
```diff
@@ -348,4 +349,21 @@ bool AnalyserModel::needAcothFunction() const
...
+    if (cacheKey != mPimpl->mCachedEquivalentVariables.end()) {
+        return cacheKey->second;
+    }
+
+    bool res = libcellml::areEquivalentVariables(variable1, variable2);
+
+    mPimpl->mCachedEquivalentVariables[key] = res;
+
+    return res;
+}
```
kerimoyle, line 394: Another question: This is the only function which affects the cache?  If so, where is the cache cleared so that new/old/changed variable equivalences are updated?  I realise that the Analyser assumes that the model it's been given is static, but what happens if the Analyser is called twice? or used to process two different models? or one model with a change in variable equivalence?  I have a feeling I'm missing something here :(

kerimoyle (commented):
It must be Monday (or a day ending in Y) because this PR confuses me ... :| See review comments for details.

src/utilities.h
agarny: Fixed in commit f1f8262c4.

src/analyser.cpp
agarny: FWIW, it's not part of the `Model` class, but the `AnalyserModel` class. So, yes, we now have two `areEquivalentVariables()` methods. One in our utilities and another in `AnalyserModel`. The difference between the two versions is that the latter relies on caching (something that can be done since an `AnalyserModel` is a static model unlike a `Model` which is "live"). No, the reason that the second `areEquivalentVariables()` method is in `AnalyserModel`, well... that's the only place I can think of, but happy to consider another location.

src/analysermodel.cpp
agarny: When you analyse a model (see `Analyser::AnalyserImpl::analyseModel()`) we create an instance of `AnalyserModel`, which is then populated with whatever we want the user to access from such an instance. As part of that process, the cache gets automatically built (through calls to `AnalyserModel::areEquivalentVariables()`). This is something that happens only once, so no need for clearing the cache or anything.

src/analysermodel.cpp
agarny: You are correct, we currently use as a key the product of the address of both variables. There is indeed no guarantee of uniqueness. Also, the casting to `intptr_t` is because it's the type that should be used for addresses. That type will increase the risk of non-uniqueness, but should indeed be easily overcome by using another type (even a `double`).

Now, regarding the product of addresses, the idea is to come up with a very fast way of computing the key. I originally had a string of the formwat `<@variable1><@variable2>`, but it was too time consuming to generate (more than 2.5x slower to run the `Generator.fabbriFantiniWildersSeveriHumanSanModel2017` test). FWIW, this is how I was creating that string-based key:
```c++
std::ostringstream strs;
strs << variable1.get() << variable2.get();
auto key = strs.str();
```
rather than:
```c++
auto key = reinterpret_cast<intptr_t>(variable1.get()) * reinterpret_cast<intptr_t>(variable2.get());
```

So, what do you mean exactly by "binned combination of indices"? I am certainly not happy with the product of addresses, so I would be happy to go with another solution as long as it is (roughly) as fast.

src/analysermodel.cpp
kerimoyle: OK, that makes (a little more) sense ... so the caching is **only** to speed up repeated calls to test equivalence between the same two variables?  And those calls are only expected to happen internally to the `Analyser` class?

agarny: That is all correct. Then, when it comes to `AnalyserModel::areEquivalentVariables()`, we make it available through our API because we never know: an end-user might want to take advantage of it.

src/analysermodel.cpp
kerimoyle: Have you thought about a sparse symmetric matrix, where you have an entry only in places where there is an equivalence between variables?  (Depending on how you decided to code the rows/columns it wouldn't necessarily need to be sparse either).  The coordinates of the matrix rows/columns could be an indexed list of the pointers already processed.  Since you already know that models passed to the `Analyser` are valid, you can use a simpler way than the pointer addresses to represent them: component and variable indices, for example? Using the pointers' addresses as if they are strings makes for something that's probably not going to be amenable to fast searching through the map structure.
I'm not sure about how speed would scale - that would need testing - but at least it could guarantee uniqueness.  
It could perhaps be used as a cleaner/faster way to store equivalences outside the analyser too (though I guess the Analyser is the object that "needs" it the most, so ... dunno ).

agarny: I went for a much simpler approach (see commit fd0be4408) which relies on the [Cantor pairing function](https://en.wikipedia.org/wiki/Pairing_function#Cantor_pairing_function).

Either way, thanks for "insisting" on the non-uniqueness of my previous key implementation. I should have been more diligent.

src/analysermodel.cpp
kerimoyle: OK, thanks.  In that case I think I second what @nickerso said about the documentation for this function.  It needs to be really clear that calls to `areEquivalentVariables` on the `AnalyserModel` are fundamentally different from similar calls on the `Variable` and `Model` classes.

agarny: Done in commit 87eb46f6b.

agarny:
Ping @hsorby, @kerimoyle and @nickerso: good to go.

src/analysermodel.cpp
```diff
@@ -348,4 +349,39 @@ bool AnalyserModel::needAcothFunction() const
...
+    // an AnalyserModel object refers to a static version of a model, which
+    // means that we can safely cache the result of a call to that utility. In
+    // turn, this means that we can speed up any feature (e.g., code generation)
+    // that also relies on that utility. When it comes to the key for the cache,
+    // we use the Cantor pairing function with the address of the two variables
+    // as parameters, thus ensuring the uniqueness of the key (see
+    // https://en.wikipedia.org/wiki/Pairing_function#Cantor_pairing_function).
+
+    auto v1 = reinterpret_cast<uint64_t>(variable1.get());
+    auto v2 = reinterpret_cast<uint64_t>(variable2.get());
```
hsorby: This won't work with emscripten. We cannot assume a fixed size of pointer. It is possible we leave this until Javascript compilation gets officially added though.

hsorby (commented):
I have a problem with the assumption of a fixed size because this will not work with emscripten.  I would prefer for this to be fixed now but I am okay with leaving it until emscripten properly joins the codebase.

Otherwise I am happy with the changes.

src/analysermodel.cpp
agarny: I am happy to fix it now, but... why won't it work with emscripten and how should it be fixed?

hsorby: emscripten is only 32 bit so `uint64_t` won't work.

hsorby: As to the how I will have to remind myself.

agarny: What about something like:
```c++
#if defined(__UINT64_TYPE__)
    auto v1 = reinterpret_cast<uint64_t>(variable1.get());
    auto v2 = reinterpret_cast<uint64_t>(variable2.get());
#elif defined(__UINT32_TYPE__)
    auto v1 = reinterpret_cast<uint32_t>(variable1.get());
    auto v2 = reinterpret_cast<uint32_t>(variable2.get());
#else
#    error "Unsupported architecture (only 32-bit and 64-bit architectures are supported)."
#endif
```

On 64-bit, both `__UINT32_TYPE__` and `__UINT64_TYPE__` are defined, but on 32-bit I would only expect `__UINT32_TYPE__` to be defined. So, it ought to work?

hsorby: I think I would prefer to go with `uintptr_t` do you see any reason not to? [integer types](https://en.cppreference.com/w/cpp/types/integer)

agarny: No idea why I didn't use `uintptr_t` in the first place since, indeed, I can't see any reason not to.

agarny: Done in commit e7a40a5.

hsorby (dismissed):
I think this is good to go.

src/analysermodel_p.h
```diff
@@ -60,6 +60,8 @@ struct AnalyserModel::AnalyserModelImpl
     bool mNeedAsechFunction = false;
     bool mNeedAcschFunction = false;
     bool mNeedAcothFunction = false;
+
+    std::map<uint64_t, bool> mCachedEquivalentVariables;
```
hsorby: Umm, `uintptr_t`

hsorby (changes requested):

nickerso (dismissed):

src/analysermodel_p.h
agarny: Oops, done in commit 7a0545d.

nickerso (approved):

hsorby (approved):

Issue #772, opened 2020-11-04
Model::removeUnits / findUnits behaves differently between name and reference arguments


kerimoyle:
This bug is coming from the differences in how we search for units in a model using the `findUnits` function (see below).  I can understand searching for a name match (since this is how we do everything else), but I don't understand the reference search.  What's the rationale for not simply testing that the pointer is found?? Why test name and definition?  
If we do have a need to test the name and definition for units, can we move that to a separate explicit function and have the default behaviour in line with the other classes?  @agarny @hsorby @nickerso what do you think?

```
std::vector<UnitsPtr>::iterator Model::ModelImpl::findUnits(const std::string &name)
{
    return std::find_if(mUnits.begin(), mUnits.end(),
                        [=](const UnitsPtr &u) -> bool { return u->name() == name; });
}

std::vector<UnitsPtr>::iterator Model::ModelImpl::findUnits(const UnitsPtr &units)
{
    return std::find_if(mUnits.begin(), mUnits.end(),
                        [=](const UnitsPtr &u) -> bool { return units->name().empty() ? false : u->name() == units->name() && Units::equivalent(u, units); });
}
```

Here's the test that fails:
```
auto model = libcellml::Model::create();
    auto myImportedComponent = libcellml::Component::create("myImportedComponent");
    auto myConcreteComponent = libcellml::Component::create("myConcreteComponent");
    auto myConcreteUnits = libcellml::Units::create("myConcreteUnits");
    auto myImportedUnits = libcellml::Units::create("myImportedUnits");

    auto import = libcellml::ImportSource::create();
    import->setUrl("import.cellml");
    myImportedComponent->setImportSource(import);
    myImportedUnits->setImportSource(import);

    model->addComponent(myImportedComponent);
    model->addComponent(myConcreteComponent);
    model->addUnits(myConcreteUnits);
    model->addUnits(myImportedUnits);

    auto success = model->removeUnits(myImportedUnits);
    EXPECT_TRUE(success); // <<<<<<<<<< fails, units are not found!
    success = model->removeUnits("myImportedUnits");
    EXPECT_TRUE(success);  // << passes.

```

kerimoyle:
The reason that the find is not working is because the units are imported, so we can't test for equivalence.  That's fine ... except that the pointers are identical, which surely should take precedence over names and definitions!  I suggest that if we want to keep searching by name and contents we either:
[...]

Issue #719, opened 2020-09-24
Model: its `hasUnits()` method does different things depending on the signature that is used


agarny:
`Model::hasUnits()` has two signatures:
 1. `bool hasUnits(const std::string &name) const;`; and
 2. `bool hasUnits(const UnitsPtr &units) const;`.

There is clearly nothing wrong with that (on the contrary), but the problem is that the implementation is different whether we use the first or second signature. Indeed, although they both rely on a call to `Model::ModelImpl::findUnits()`, it happens that `findUnits()` also has two signatures:
 1. `std::vector<UnitsPtr>::iterator findUnits(const std::string &name);`; and
 2. `std::vector<UnitsPtr>::iterator findUnits(const UnitsPtr &units);`.

The problem is that they don't behave the same:
```c++
std::vector<UnitsPtr>::iterator Model::ModelImpl::findUnits(const std::string &name)
{
    return std::find_if(mUnits.begin(), mUnits.end(),
                        [=](const UnitsPtr &u) -> bool { return u->name() == name; });
}

std::vector<UnitsPtr>::iterator Model::ModelImpl::findUnits(const UnitsPtr &units)
{
    return std::find_if(mUnits.begin(), mUnits.end(),
                        [=](const UnitsPtr &u) -> bool { return units->name().empty() ? false : u->name() == units->name() && Units::equivalent(u, units); });
}
```

The bottom line is that I would expect a method that has different signatures to behave the same underneath. Anything else (like here) is a potential source for problems. (See https://github.com/cellml/libcellml/pull/716/files#r493927037 for some background on this issue.)

hsorby:
I don't agree that methods with different signatures should behave the same underneath.  However the documentation should make it very clear about the differences between the methods.

hsorby:
I think the underlying problem here is really an issue with using equivalence to match units.  I think that equality should be used here instead.

agarny:
> I don't agree that methods with different signatures should behave the same underneath. However the documentation should make it very clear about the differences between the methods.

This makes things very confusing for the end-user (independent of the documentation). If we want to keep the current behaviour, then we really ought to rename the method.

nickerso:
I agree with @hsorby - these are both find units methods and the different signature makes it very clear that the behaviour must be different. The switch from checking equivalence to equality would even ensure that these two different methods are behaving even more similar - you can just do a better comparison when you are given more information.

I think it is safe to close this one.

agarny:
[...]

PR #773, opened 2020-11-04
Model::findUnits bug

kerimoyle:
Addresses #772 and #719

### DIFF
--- a/tests/model/model.cpp
+++ b/tests/model/model.cpp
@@ -669,3 +669,32 @@ TEST(Model, removeImportedUnitsByNameOrReference)
     EXPECT_EQ(size_t(0), model->unitsCount());
     EXPECT_EQ(size_t(1), model->importSourceCount());
 }
+
+TEST(Units, removeUnitsObjectIdentical)
+{
+    libcellml::ModelPtr m = libcellml::Model::create();
+
+    libcellml::UnitsPtr u1 = libcellml::Units::create();
+    m->addUnits(u1);
+
+    EXPECT_EQ(size_t(1), m->unitsCount());
+
+    m->removeUnits(u1);
+    EXPECT_EQ(size_t(0), m->unitsCount());
+}
+
+TEST(Units, removeUnitsObjectEquivalent)
+{
+    libcellml::ModelPtr m = libcellml::Model::create();
+
+    libcellml::UnitsPtr u1 = libcellml::Units::create();
+    libcellml::UnitsPtr u2 = libcellml::Units::create();
+
+    EXPECT_TRUE(libcellml::Units::equivalent(u1, u2));
+    m->addUnits(u1);
+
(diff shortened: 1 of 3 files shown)


kerimoyle:
I'm struggling to understand two of the python tests:

in Model.test_replace_units:
```
        # bool replaceUnits(const UnitsPtr &oldUnits, const UnitsPtr &newUnits)
        m = Model()
        a = Units()
        m.addUnits(a)
        b = Units()
        self.assertFalse(m.replaceUnits(b, a))
        self.assertFalse(m.replaceUnits(a, b))   # KRM I don't know why this is expected to fail??
        self.assertFalse(m.replaceUnits(b, a))
        del [m, a, b]
```
in Model.test_remove_units:
```
        # bool removeUnits(const UnitsPtr &units)
        m = Model()
        u1 = Units()
        u2 = Units()
        self.assertFalse(m.removeUnits(u1))
        m.addUnits(u1)
        self.assertFalse(m.removeUnits(u2))
        self.assertFalse(m.removeUnits(u1)) # KRM ... why is this not expected to work??
        del [m, u1, u2]
```

kerimoyle:
Ready for review please (assuming build passes) @agarny @hsorby @nickerso

hsorby:
I have two tests I would like to add to this but I can't. Is the branch is locked?

kerimoyle:
Should be unlocked now?

tests/model/model.cpp
```diff
@@ -669,3 +669,32 @@ TEST(Model, removeImportedUnitsByNameOrReference)
...
+
+TEST(Units, removeUnitsObjectEquivalent)
+{
+    libcellml::ModelPtr m = libcellml::Model::create();
+
+    libcellml::UnitsPtr u1 = libcellml::Units::create();
+    libcellml::UnitsPtr u2 = libcellml::Units::create();
+
+    EXPECT_TRUE(libcellml::Units::equivalent(u1, u2));
+    m->addUnits(u1);
```
kerimoyle, line 694: Comparing two base units without names that are not in the model.  There's no way that `equivalent` should be `true` when there's no information about either??

tests/model/model.cpp
```diff
@@ -669,3 +669,32 @@ TEST(Model, removeImportedUnitsByNameOrReference)
...
+
+    EXPECT_EQ(size_t(1), m->unitsCount());
+
+    m->removeUnits(u2);
+    EXPECT_EQ(size_t(0), m->unitsCount());
```
kerimoyle, line 699: I really don't like this behaviour!  When removing units by reference, it should be *only* the reference that is tested.  When removing units by name, we take the first one that matches ... just as we do for every other item type ... equivalence to other units is irrelevant.

hsorby:
Closing this as it has been incorporated through PR #785.

Issue #771, opened 2020-11-04
Parser creates extra import source items in the model


kerimoyle:
I'm not sure what's expected from this situation: where an imported component is parent to a concrete child?  I didn't expect this though ...  (edit: removed units to make the problem simpler, same behaviour).

```
    std::string modelString = 
        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
        "<model xmlns=\"http://www.cellml.org/cellml/2.0#\">\n"
        "  <import xmlns:xlink=\"http://www.w3.org/1999/xlink\" xlink:href=\"import.cellml\">\n"
        "    <component component_ref=\"myComponent\" name=\"myImportedComponent\"/>\n"
        "  </import>\n"
        "  <component name=\"myConcreteComponent\"/>\n"
        "  <encapsulation>\n"
        "    <component_ref component=\"myConcreteComponent\">\n"
        "      <component_ref component=\"myImportedComponent\"/>\n"
        "    </component_ref>\n"
        "  </encapsulation>\n"
        "</model>";

    auto parser = libcellml::Parser::create();
    auto model = parser->parseModel(modelString);
    EXPECT_EQ(size_t(1), model->importSourceCount()); // <<< model->importSourceCount() = 2!!
```

Issue #770, opened 2020-11-03
removeComponent for an imported component needs to update importSource reference


kerimoyle:
At present removing an imported component (and perhaps units) from a model doesn't also update the list in the import source ... and it should.

```
model:
- import 
       - component1
       - component2
- component3

model->componentCount() = 3 
model->importSourceCount() = 1
model->importSource(0)->componentCount() = 2

model->removeComponent(component1)

model->componentCount() = 2
model->importSource(0)->componentCount() = 2 <<<< should be 1, and removed component should be off the list.
```

hsorby:
Closing this as the shift away from import sources tracking components and units makes this moot.

Issue #760, opened 2020-10-21
CMake: disable some warnings that have been introduced in LLVM 11


agarny:
LLVM 11 was recently released and it introduces a couple of warnings that break our compilation process (on macOS):
 - `misc-no-recursion`; and
 - `cppcoreguidelines-avoid-non-const-global-variables` (occurs in Google Test).

Clearly, we want and need recursion, so we should disable `misc-no-recursion`. As for `cppcoreguidelines-avoid-non-const-global-variables`, we have no choice but to disable it (but only for Google Test).

PR #774, opened 2020-11-05
I770 Update ImportSource when removing/replacing/taking imported items

kerimoyle:
Addresses #770, #771, #772.
This PR is aimed at fixing how the import sources are updated when imported items are passed around the place. There are a few bugs coming from this single source, hence the three issues above.


kerimoyle:
There's a bug in the `Parser` (see #771 ) which is highlighted only when the other bugs around the `ImportSource` are fixed.  That one will probably have to be done as part of this PR too, since it's only manifest when these other things are resolved.

kerimoyle:
Ready for review please @agarny @hsorby @nickerso (build on MacOS only failing because of waiting for #760 to be merged).

src/api/libcellml/model.h
```diff
@@ -350,16 +350,20 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
      * Add import source by reference to the model. If the import
      * source item was previously in a different model, it is moved
      * to this one, and the previous model's import source list is updated.
-     *
+     * if the optional @p merge parameter is true, the contents of the given @p importSrc
```
hsorby: If

src/componententity.cpp
```diff
@@ -222,17 +252,25 @@ ComponentPtr ComponentEntity::takeComponent(const std::string &name, bool search
...
+        auto model = owningModel(oldComponent);
+        if (model != nullptr) {
+            model->addImportSource(newComponent->importSource());
+        }
+    }
+
     if (removeComponent(index)) {
-        mPimpl->mComponents.insert(mPimpl->mComponents.begin() + int64_t(index), component);
-        component->setParent(parent);
+        mPimpl->mComponents.insert(mPimpl->mComponents.begin() + int64_t(index), newComponent);
```
hsorby, line 271: Should this `int64_t` be `uintptr_t`?

hsorby (dismissed):
All tests that go `replaceImportedComponentByNameIndexOrReference` should be separate tests `replaceImportedComponentByName`; `replaceImportedComponentByIndex`; `replaceImportedComponentByReference`.

src/componententity.cpp
kerimoyle: Perhaps, but if so should be a separate PR as it's everywhere ... and not really related to this PR.  I just changed the name of the component handle so it was easier to understand. 
image

src/api/libcellml/model.h
kerimoyle: b82e97c1

kerimoyle:
Review comments all addressed in b82e97c1.  Ready for review please @hsorby @agarny @nickerso.

hsorby:
During my review of this PR I thought that we should reimagine how ImportSources are considered.  I have started a [mini-project](https://github.com/cellml/libcellml/projects/5) to capture this possible transition.

Issue #775, opened 2020-11-08
Should empty import sources be printed?


kerimoyle:
Should the `Printer` write models containing an empty import source like this?

```
  <import xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="importedModel.cellml"/>
```

agarny:
I don't find it "clean" to print something that is meaningless, but I understand that the printer is to print the contents of a model description. So, if the user has created an empty import source then... it should be printed. Indeed, the user may, later on, decide to reopen the CellML model, and add imports to it and, here, s/he could do it without first having to recreate the import source.

kerimoyle:
> Indeed, the user may, later on, decide to reopen the CellML model, and add imports to it and, here, s/he could do it without first having to recreate the import source. 

That was my thinking.  Even though they might be currently "empty", they still contain data (the url), which shouldn't be thrown away by default. It's the same as having empty components.

agarny:
> It's the same as having empty components.

Exactly.

kerimoyle:
After discussions this week, we should make a utility function in the `Model` class that will remove any unused or empty items without changing the underlying model semantics.

PR #778, opened 2020-11-12
Added a Model::clean function to remove unneeded items

kerimoyle:
Addresses #775


kerimoyle:

src/model.cpp
```diff
@@ -539,4 +539,70 @@ bool Model::fixVariableInterfaces()
...
 
+bool isComponentEmpty(const ComponentPtr &component)
+{
+    for (int i = int(component->componentCount() - 1); i >= 0; --i) {
+        if (isComponentEmpty(component->component(size_t(i)))) {
+            component->removeComponent(size_t(i));
+        }
+    }
+
+    return (component->variableCount() + component->resetCount() + component->componentCount()) == 0
```
agarny: This should be `(component->variableCount() + component->resetCount() + component->componentCount() == 0)`.

src/model.cpp
```diff
@@ -547,4 +547,72 @@ bool Model::fixVariableInterfaces()
...
+}
+
+void checkUnits(const ComponentPtr &component, std::vector<UnitsPtr> &unused)
+{
+    for (size_t v = 0; v < component->variableCount(); ++v) {
+        auto u = component->variable(v)->units();
+        if (u != nullptr) {
+            auto it = std::find_if(unused.begin(), unused.end(), [=](const UnitsPtr &t) -> bool {
+                // KRM Not sure whether including names is a good idea or not.  They should be linked by
+                //     pointers only anyway, but the findUnits function is still testing names and equivalence ...
```
agarny: Remove.

src/model.cpp
```diff
@@ -547,4 +547,72 @@ bool Model::fixVariableInterfaces()
...
+                return t == u || u->name() == t->name();
```
agarny: Replace with `return (t == u) || (u->name() == t->name()); `.

tests/model/model.cpp
```diff
@@ -624,3 +624,59 @@ TEST(Model, removeComponentInsensitiveToOrder)
     EXPECT_EQ(size_t(0), modelParsed->componentCount());
     EXPECT_EQ(size_t(2), modelApi->componentCount());
 }
+
+TEST(Model, cleanModel)
+{
+    // Make a model with empty components, unrequired units, empty import sources.
+    auto e = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
```
agarny: Replace `auto e` with `const std::string e`. Same elsewhere in this PR.

agarny (commented):
Some initial comments until we come to some agreement regarding this PR. :)

src/api/libcellml/model.h
```diff
@@ -414,17 +414,25 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
      *
      * @return @c true if the import source is in the model and @c false otherwise.
      */
     bool hasImportSource(const ImportSourcePtr &importSrc) const;
 
+    /**
+     * @brief Remove any empty import sources, empty components, and unused units from the 
+     *        model.
+     * 
+     *  Remove any empty import sources, empty components, and unused units from the model.
```
nickerso: Need to define what empty means in this context.

src/model.cpp
```diff
@@ -547,4 +547,72 @@ bool Model::fixVariableInterfaces()
     return allOk;
 }
 
+static const size_t MAX_SIZE_T = std::numeric_limits<size_t>::max();
+
+bool isComponentEmpty(const ComponentPtr &component)
```
nickerso: I'd rename this something like `findAndRemoveEmptyComponents`

src/model.cpp
```diff
@@ -547,4 +547,72 @@ bool Model::fixVariableInterfaces()
...
+            component->removeComponent(i);
+        }
+    }
+
+    return (component->variableCount() + component->resetCount() + component->componentCount()) == 0
+           && component->math().empty()
+           && !component->isImport();
+}
+
+void checkUnits(const ComponentPtr &component, std::vector<UnitsPtr> &unused)
```
nickerso: This method would also need to check if the units are used in the math (including math in the resets)

nickerso (dismissed):

src/model.cpp
kerimoyle: Yeah ... not until I know what's going on ... For the final thing, sure.

src/model.cpp
kerimoyle: f15e055d

src/api/libcellml/model.h
kerimoyle: f15e055d

tests/model/model.cpp
kerimoyle: f15e055d

src/model.cpp
kerimoyle: f15e055d

src/model.cpp
kerimoyle: f15e055d

src/model.cpp
kerimoyle: You're right, bother.  In order to do that, I need to parse the MathML.  If the MathML isn't valid then it can't be parsed, so errors need to be logged.  There is no logger because this is done as part of the `Model` class.  I can move this into the `Printer` class, since that's where people will be using it anyway, and log errors there.  Thoughts?

nickerso: I can almost see this as an optional switch on `printModel` to clean and then print. But might also be easier to simply not do anything with units. Just clean "empty" things for now, cleaning "unused" things can be a future addition if it would ever be useful.

kerimoyle: In the first round of discussions we talked about that - and concluded somewhere that this should be something separate from the printer in the `Model` class.  I can't remember why!  If I don't check units then it can stay where it is in the `Model`, where it's a little more flexible as you're not constrained to printing (even if that's what most people will do).

kerimoyle: 8837c78c

src/model.cpp
kerimoyle: Moot now as have removed this function 8837c78c

src/model.cpp
kerimoyle: Removed this function in 8837c78c

nickerso (dismissed):
@hsorby and @agarny - I've updated this one to make it clear what I'm thinking cleaning means.

src/api/libcellml/model.h
```diff
@@ -413,17 +413,30 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
      * @return @c true if the import source is in the model and @c false otherwise.
      */
     bool hasImportSource(const ImportSourcePtr &importSource) const;
 
+    /**
+     * @brief Remove any empty components and units from the model.
+     * 
+     *  Remove any empty components and units fom this model.
+     *  In this context, "empty" means:
+     *   - components with no name, id, resets, variables, maths, or non-empty child components; and
```
agarny, line 432: `non-empty child components`? Shouldn't it simply be `child components`?

src/api/libcellml/model.h
```diff
@@ -413,17 +413,30 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
+     *   - units which have no name, id, or child unit's.
```
agarny: `child units` rather than `child unit's`?

src/api/libcellml/model.h
```diff
@@ -413,17 +413,30 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
...
+     *  This is applied recursively. In the situation where a component is only called
+     *  non-empty because it has a child component, but that child is empty, both will be
+     *  removed.
```
agarny: Might want to reword this a bit to read `In the situation where a component has an empty child component, both will be removed. `.

src/model.cpp
```diff
@@ -527,6 +527,44 @@ bool Model::fixVariableInterfaces()
     return allOk;
 }
 
+static const size_t MAX_SIZE_T = std::numeric_limits<size_t>::max();
```
agarny: Might want to move this to `utilities.h` and also remove the one that is defined in `analyser.cpp`?

nickerso (dismissed):
Now with full coverage.

src/model.cpp
```diff
@@ -527,6 +527,44 @@ bool Model::fixVariableInterfaces()
...
+void Model::clean()
+{
+    // Remove empty components.
+    for (size_t i = componentCount() - 1; i != MAX_SIZE_T; --i) {
+        if (findAndRemoveEmptyComponent(component(i))) {
+            removeComponent(i);
+        }
+    }
+
+    // Remove empty units
```
agarny: Missing final point.

tests/bindings/python/test_model.py
```diff
@@ -379,6 +379,41 @@ def test_imports(self):
...
+        model.addUnits(u1)
+        model.addUnits(u2)
+        model.addUnits(u3)
+        
+        c3.addVariable(v)
+
+        self.assertEqual(4, model.componentCount())
+        self.assertEqual(3, model.unitsCount())
+
+        # Call the Model.clean() function to remove empty items.
```
agarny: Might want to mention that the empty items are `c4` and `u3`.

tests/model/model.cpp
```diff
@@ -628,6 +628,106 @@ TEST(Model, removeComponentInsensitiveToOrder)
...
+    u2->setId("u2");
+    auto u3 = libcellml::Units::create();
+    auto v = libcellml::Variable::create("x");
+
+    model->addComponent(c1);
+    model->addComponent(c2);
+    model->addComponent(c3);
+    model->addComponent(c4);
+
+    // test unnamed components in an encapsulation hierarchy
```
agarny: `// Test unnamed components in an encapsulation hierarchy.`

tests/resources/dirty_model.cellml
```diff
@@ -0,0 +1,13 @@
...
+<model xmlns="http://www.cellml.org/cellml/2.0#"
+    xmlns:cellml="http://www.cellml.org/cellml/2.0#" name="dirtyModel">
+  <units name="namedEmptyUnits"/>
+  <units/>
+  <units id="nonEmptyId"/>
+  <component name="namedEmptyComponent"/>
+    <component name="nonEmptyComponent">
+        <variable name="x" units="requiredUnits"/>
+    </component>
+  <component />
```
agarny: Formatting (to make it easier to read/review/maintain/etc.):
```xml
  <component name="nonEmptyComponent">
    <variable name="x" units="requiredUnits"/>
  </component>
  <component/>
```

agarny (dismissed):
A good start. A few minor comments and, more importantly, some missing tests, I think, when it comes to:
 - a unit definition that has child units, e.g.
 ```xml
<units name="child_unit"/>
<units name="empty_child_unit">
    <unit/>
</units>
<units name="my_unit">
    <unit units="child_unit"/>
    <unit units="empty_child_unit"/>
    <unit/>
</units>
 ```
 which, once cleaned, should give us:
 ```xml
<units name="child_unit"/>
<units name="my_unit">
    <unit units="child_unit"/>
</units>
 ```
 - unit not cleared if it is imported;
 - component not cleared if it has resets;
 - component not cleared if it has some math; and
 - component not cleared if it is imported.

src/api/libcellml/model.h
nickerso: I think this is right - its listing the things that an empty component does not have.

src/api/libcellml/model.h
nickerso: d1008b6

src/api/libcellml/model.h
nickerso: 25578c0

src/model.cpp
nickerso: fcb43fc

src/model.cpp
nickerso: 615f237

tests/bindings/python/test_model.py
nickerso: 615f237

tests/model/model.cpp
nickerso: Tests changed, so that comment is no longer relevant.

tests/resources/dirty_model.cellml
nickerso: 615f237

src/model.cpp
```diff
@@ -527,6 +527,42 @@ bool Model::fixVariableInterfaces()
     return allOk;
 }
 
+bool findAndRemoveEmptyComponent(const ComponentPtr &component)
```
hsorby: I really think this method should be called `traverseHierarchyAndRemoveIfEmpty`.

hsorby (changes requested):
The tests don't build up very well (or at all) but apart from the minor change requested I can accept this.

src/model.cpp
nickerso: c1c7470 and 3514c3e

nickerso:
> "descendants" not "descendents". Also, may want to replace "cleaned" with "removed"?

descendents updated in 54cacbc

nickerso:
> "descendants" not "descendents". Also, may want to replace "cleaned" with "removed"?

and "cleaned" part in db2331e

agarny (approved):

hsorby (approved):

Issue #768, opened 2020-11-02
Importer::removeAllModels function


kerimoyle:
It would be really useful to be able to wipe broken models from the importer's library and/or clear it all out.  Since the importer uses cached models before parsing new ones, it's a complicated business to keep track of what's coming from where when you're fixing import bugs.

PR #781, opened 2020-11-18
Create an Importer::removeAllModels function

kerimoyle:
Addresses #768 

This is built on top of #774 so should be merged after that one.


kerimoyle:
Review please @agarny @hsorby @nickerso

Issue #777, opened 2020-11-09
Printer is stripping newlines from MathML output?


kerimoyle:
When you use the `Component::setMath` or `appendMath` functions to set the maths in a component, when the model is serialised for printing (or at some stage before then) the newlines are lost.  

Newlines included in mathml string:
```
    std::string equation =
        "  <apply><eq/>\n"
        "    <apply><diff/>\n"
        "      <bvar><ci>t</ci></bvar>\n"
        "      <ci>X</ci>\n"
        "    </apply>\n" 
        "    <apply><minus/>\n"
        "      <apply><times/>\n"
        "        <ci>alpha_X</ci>\n"
        "        <apply><minus/>\n"
        "          <cn cellml:units=\"dimensionless\">1</cn>\n"
        "          <ci>X</ci>\n"
        "        </apply>\n" 
        "      </apply>\n" 
        "      <apply><times/>\n"
        "        <ci>beta_X</ci>\n"
        "        <ci>X</ci>\n"
        "      </apply>\n" 
        "    </apply>\n" 
        "  </apply>\n"; 

    gateEquations->setMath(mathHeader);
    gateEquations->appendMath(equation);
    gateEquations->appendMath(mathFooter);
```

... but they're not printed:
```
... 
<component name="gateEquations">
    <variable name="t" units="ms" interface="public_and_private"/>
    <variable name="alpha_X" units="per_ms" interface="public"/>
    <variable name="beta_X" units="per_ms" interface="public"/>
    <variable name="X" units="dimensionless" interface="public_and_private"/>
    <math xmlns="http://www.w3.org/1998/Math/MathML" xmlns:cellml="http://www.cellml.org/cellml/2.0#">  <apply><eq/>    <apply><diff/>      <bvar><ci>t</ci></bvar>      <ci>X</ci>    </apply>    <apply><minus/>      <apply><times/>        <ci>alpha_X</ci>        <apply><minus/>          <cn cellml:units="dimensionless">1</cn>          <ci>X</ci>        </apply>      </apply>      <apply><times/>        <ci>beta_X</ci>        <ci>X</ci>      </apply>    </apply>  </apply></math>
  </component>
...
```

kerimoyle:
[...]

PR #782, opened 2020-11-18
Fix bug in prettyPrint for formatting XML

kerimoyle:
Addresses #777


kerimoyle:

src/printer.cpp
```diff
@@ -529,6 +529,7 @@ std::string Printer::printModel(const ModelPtr &model, bool autoIds) const
 
     // Generate a pretty-print version of the model using libxml2.
     XmlDocPtr xmlDoc = std::make_shared<XmlDoc>();
+    xmlKeepBlanksDefault(0); // see: http://www.xmlsoft.org/html/libxml-tree.html#xmlDocDumpFormatMemoryEnc
```
agarny: Don't need the comment. If anything, I would expect further explanation, in your comment at line 530, as why we need to call this method.

tests/printer/printer.cpp
```diff
@@ -768,3 +768,138 @@ TEST(Printer, noChangeToManualIds)
...
+        "    </apply>\n"
+        "  </apply>\n"
+        "</apply>\n";
+    component6->setMath(mathHeader);
+    component6->appendMath(equation6);
+    component6->appendMath(mathFooter);
+
+    std::string expectedOutput = fileContents("printer/pretty_print_mathml.cellml");
+    EXPECT_EQ(expectedOutput, printer->printModel(model));
+}
```
agarny, line 980: I would probably split this test in six (one for each type of pretty printing you want to test) and maybe have the expected result in the `tests/printer/printer.cpp` rather than in `tests/printer/pretty_print_mathml.cellml`? Will leave it to @hsorby to comment on this.

agarny (dismissed):

tests/printer/printer.cpp
kerimoyle: Yeah, I thought about that. But it's a big file and I couldn't be bothered doing all the line endings!  I guess if it's split into several tests they all make the same thing anyway.

agarny: If you were to break things down, you would need to specify the expected output only once (and use it in all six tests) and it would be smaller, easier to understand, maintain (if ever needed), etc.

kerimoyle:
Ah, bother, we do need to use the `regex` after all.  This is not enough to remove unwanted whitespace in the MathML (see line 142 in the cellml file).  I've fixed it and made your changes @agarny and will push now.

tests/printer/printer.cpp
kerimoyle: Done in b43279a

src/printer.cpp
kerimoyle: Done in b43279a

src/printer.cpp
```diff
@@ -128,13 +129,10 @@ std::string printConnections(const ComponentMap &componentMap, const VariableMap
...
 {
-    std::string repr;
-    std::istringstream lines(math);
-    std::string line;
-    while (std::getline(lines, line)) {
-        repr += line;
-    }
-    return repr;
+    std::regex before(">[\\s\n\t]*");
+    std::regex after("[\\s\n\t]*<");
```
agarny: I seem to remember that building a regular expression is time consuming, so we might want to use something like:
```
static const std::regex before(">[\\s\n\t]*");
static const std::regex after("[\\s\n\t]*<");
```

kerimoyle:
Hmmm ... I've pushed a commit to here, but it's not showing up?  420113d0

tests/printer/printer.cpp
```diff
@@ -805,7 +841,20 @@ TEST(Printer, prettyPrint)
...
-    auto component2 = libcellml::Component::create("component2");
+    EXPECT_EQ(prettyModelString, printer->printModel(model));
+}
+
+TEST(Printer, prettyPrintNoSpacesNewlines)
+{
+    auto printer = libcellml::Printer::create();
+
+    std::string mathHeader = "<math xmlns=\"http://www.w3.org/1998/Math/MathML\" xmlns:cellml=\"http://www.cellml.org/cellml/2.0#\">\n";
+    std::string mathFooter = "</math>\n";
```
agarny: You might also want to move those two outside (and make them `const`) and remove the ones in the other tests.

src/printer.cpp
kerimoyle: Does it need to be static?  Can't just be part of the impl struct?

agarny: By having it `static`, we can create as many printers as we want (you never know, an end-user might want to do that!) and have only one instance of the regular expression (and thus save time).

tests/printer/printer.cpp
kerimoyle: fixed in 3c4d1d61

src/printer.cpp
kerimoyle: OK.  Done in 3c4d1d61

src/printer.cpp
```diff
@@ -36,6 +37,9 @@ limitations under the License.
 
 namespace libcellml {
 
+static const std::regex before(">[\\s\n\t]*");
+static const std::regex after("[\\s\n\t]*<");
```
agarny: No need for it to be outside of `printMath()` (unless we were ever to use those regular expressions, but we are not for now, so...

kerimoyle: Ah, I thought that's what you wanted!  OK, will move it back.

kerimoyle: cc6a4619

agarny (dismissed):

tests/printer/printer.cpp
```diff
@@ -768,3 +768,213 @@ TEST(Printer, noChangeToManualIds)
     auto printer = libcellml::Printer::create();
     EXPECT_EQ(in, printer->printModel(model, true));
 }
+
+const std::string mathHeader = "<math xmlns=\"http://www.w3.org/1998/Math/MathML\" xmlns:cellml=\"http://www.cellml.org/cellml/2.0#\">\n";
+const std::string mathFooter = "</math>\n";
+const std::string prettyModelString =
```
hsorby: These should be at the top of the file and be declared in capitals `MATH_HEADER`.

tests/printer/printer.cpp
```diff
@@ -768,3 +768,213 @@ TEST(Printer, noChangeToManualIds)
...
+
+TEST(Printer, prettyPrintSpacesNewlines)
+{
+    auto printer = libcellml::Printer::create();
+    auto model = libcellml::Model::create();
+    model->setName("model");
+
+    auto component1 = libcellml::Component::create("component");
+    model->addComponent(component1);
+    std::string equation1 =
```
hsorby: const

tests/printer/printer.cpp
```diff
@@ -768,3 +768,213 @@ TEST(Printer, noChangeToManualIds)
...
+
+TEST(Printer, prettyPrintNoSpacesNewlines)
+{
+    auto printer = libcellml::Printer::create();
+    auto model = libcellml::Model::create();
+    model->setName("model");
+
+    auto component2 = libcellml::Component::create("component");
+    model->addComponent(component2);
+    std::string equation2 =
```
hsorby: const

tests/printer/printer.cpp
```diff
@@ -768,3 +768,213 @@ TEST(Printer, noChangeToManualIds)
...
+
+TEST(Printer, prettyPrintNoSpacesNoNewlines)
+{
+    auto printer = libcellml::Printer::create();
+    auto model = libcellml::Model::create();
+    model->setName("model");
+
+    auto component3 = libcellml::Component::create("component");
+    model->addComponent(component3);
+    std::string equation3 =
```
hsorby: const

hsorby (changes requested):

tests/printer/printer.cpp
```diff
@@ -768,3 +768,213 @@ TEST(Printer, noChangeToManualIds)
...
+
+TEST(Printer, prettyPrintRandom)
+{
+    auto printer = libcellml::Printer::create();
+    auto model = libcellml::Model::create();
+    model->setName("model");
+
+    auto component5 = libcellml::Component::create("component");
+    model->addComponent(component5);
+    std::string equation5 =
```
hsorby: const

tests/printer/printer.cpp
```diff
@@ -768,3 +768,213 @@ TEST(Printer, noChangeToManualIds)
...
+
+TEST(Printer, prettyPrintSpacesWrongPlaceNewlines)
+{
+    auto printer = libcellml::Printer::create();
+    auto model = libcellml::Model::create();
+    model->setName("model");
+
+    auto component6 = libcellml::Component::create("component");
+    model->addComponent(component6);
+    std::string equation6 =
```
hsorby: const

hsorby (changes requested):

tests/printer/printer.cpp
kerimoyle: 364144e8

tests/printer/printer.cpp
kerimoyle: 364144e8

tests/printer/printer.cpp
kerimoyle: 364144e8

tests/printer/printer.cpp
kerimoyle: 364144e8

tests/printer/printer.cpp
kerimoyle: 364144e8

tests/printer/printer.cpp
kerimoyle: 364144e8

agarny (approved):

hsorby (approved):

Issue #789, opened 2020-12-01
Rename llvm_coverage_missing_lines


hsorby:
With the addition of the `python_coverage` target we also have the `python_coverage_html_report` target.  This issue is about changing the name of the target `llvm_coverage_missing_lines` to `llvm_coverage_html_report`.

agarny:
Closing as a result of PR #790.

PR #790, opened 2020-12-01
Rename target for llvm_coverage_missing_lines

hsorby:
Addresses #789.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -173,7 +173,7 @@ if(LIBCELLML_LLVM_COVERAGE)
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
     COMMENT "Running LLVM coverage tests")
 
-  add_custom_target(llvm_coverage_html_report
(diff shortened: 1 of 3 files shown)


hsorby:
This one should be a quick one to agree or disagree with :)

nickerso (commented):
I thought this one would be about making the LLVM coverage name consistent with the python coverage report...but you are also changing the python name to remove the `html` ? so shouldn't the LLVM coverage just be `llvm_coverage_report` ?

agarny:
> I thought this one would be about making the LLVM coverage name consistent with the python coverage report...but you are also changing the python name to remove the `html` ? so shouldn't the LLVM coverage just be `llvm_coverage_report` ?

Indeed, that's what I would have expected too.

hsorby:
The target names and output directories are now `*_report` and `coverage_report` respectively for both llvm and python.

nickerso (approved):

agarny (approved):

agarny:
Closes issue #789.

Issue #788, opened 2020-12-01
Missing component import when adding import source before adding to model.


hsorby:
When adding a component hierarchy that has an imported component deep inside the hierarchy the imported component does not appear in the model if the component hierarchy is created before it is added to the model.

PR #791, opened 2020-12-01
Fix order dependent adding of component imports.

hsorby:
Addresses #788


nickerso (dismissed):

tests/import_source/import_source.cpp
```diff
@@ -309,3 +309,37 @@ TEST(ImportSource, createLinkedMultiple)
     EXPECT_FALSE(c1->isImport());
     EXPECT_FALSE(u1->isImport());
 }
+
+TEST(ImportSource, addImportComponentBeforeAddingToModel)
```
agarny, line 313: Shouldn't we also have a similar test for adding `c2` to `c1` **after** adding `c1` to the model?

agarny (commented):
Looks good pending a minor comment and tests passing.

tests/import_source/import_source.cpp
hsorby: Addressed in 3b34378.

agarny (approved):

nickerso (approved):

Issue #797, opened 2020-12-07
Validator: don't swap unnecessarily


agarny:
There are three cases in `src/validator.cpp` where we have something like:
```c++
std::vector<std::string>().swap(history);
```
In some cases, there is no need for it at all and, at best, it could be replaced with:
```c++
history.clear();
```

(See https://github.com/cellml/libcellml/pull/767#pullrequestreview-546042950.)

PR #798, opened 2020-12-07
Validator: don't swap unnecessarily (#797)

agarny:
Addresses issue #797.


kerimoyle (approved):
If the tests pass I'm happy.

agarny:
> If the tests pass I'm happy.

It's not only the tests that pass, but it also makes sense. In `Validator::ValidatorImpl::validateNoUnitsAreCyclic()` and `Validator::ValidatorImpl::checkUnitsForCycles()`, `hash` and `childHistory` are local variables that were "swapped" just before going out of scope. So, we were basically wasting computational time for no good reason. As for `history` in `Validator::ValidatorImpl::validateNoUnitsAreCyclic()`, the idea was to clear it and this can be done by calling `history.clear()` rather than `swap()` on `history` (same story as yesterday).

hsorby (approved):

Issue #796, opened 2020-12-07
The Validator should check for cyclical imports, if it can


kerimoyle:
At present the checking for cyclical imports is done in the `Importer` class only.  The `Validator` should use the same functionality to check for cycles, if it's given a base location to resolve against.

PR #799, opened 2020-12-08
Validator checks for cyclical imports

kerimoyle:
This PR includes a bunch of sideline stuff that had to happen along the way in order to address #796 

- the `Validator` now checks imports can be found and opened.
- functions to retrieve index of a child from its parent have been added (eg: `model->componentIndex(myComponent)` )
- a function to return the encapsulation index stack for a given component has been added (eg: `myComponent->encapsulationIndices()`
- the object stored in the `Issue` items that the `Importer` generates always refer to items in the top level model.

There are more tests and an improvement to the error messages to come...


kerimoyle:
Hi @agarny @hsorby @nickerso I had to make a function `componentIndex` as part of this PR.  I went ahead and made similar ones/tests on the `Model` class for `ImportSource` and `Units`, even though they weren't actually required for this PR.  I'm wondering:
- it would be better to only include the required `ComponentEntity`-based one that was required, or
- to also add the same thing to retrieve indices for variables and resets?

agarny:
We should only include the required `ComponentEntity`-based one, and create a new issue for the other ones (and create a PR for it now before we "forget" about it).

kerimoyle:
> We should only include the required ComponentEntity-based one, and create a new issue for the other ones (and create a PR for it now before we "forget" about it).

My bad - I forgot I'd changed how I did things, so all the functions there are needed.  I rely on the indices everywhere now since they're preserved in the cloning process, so no need to account for dodgy names.  I've added a separate issue for others (resets, variables) as #800.

@agarny @nickerso @hsorby This is ready for review please.

NB: I get coverage passing on my machine, but the buildbot isn't being triggered for that test?

hsorby:
I don't like this idea of `encapsulationIndices`, and I cannot see myself approving of it.  For me it doesn't belong in the API.

kerimoyle:
> I don't like this idea of encapsulationIndices, and I cannot see myself approving of it. For me it doesn't belong in the API.

@hsorby I agree ... kind of ... What's the better way to do it?  I need to be able to find one component in a model, and pass that to another model.  Since there are no friend classes, it will be available through the API? Or can I change that somewhere?  I can see it being useful to a user, though I don't really like the name ...!

PR #804, opened 2021-01-24
Revisit import source.

### LINKED ISSUE #801: Pare back ImportSource.
For me the import source has lost its way and it needs to be pared back to its original purpose.  The documentation for the class states 

```
The import source class is for defining a location at which a CellML model is expected to be obtainable (the `source`).
The import source also acts as  a handle to indicate that imported entities originate from the same instance of that imported source.
 ```

I think this documentation describes what the class should be doing.  This issue is to pare back the ImportSource so that it does just that.

hsorby:
Removing the

* addComponent
* removeComponent
* hasComponent
* addUnits
* removeUnits
* hasUnits

type classes from ImportSource.

Addresse #801.


hsorby:
Hmm, I think I have based this off the wrong branch from @kerimoyle I will look into changing it.

Issue #808, opened 2021-01-27
ClangFormat: only run the ClangFormat test if we have at least version 11 of ClangFormat


agarny:
The reason for this is that the formatting that results from ClangFormat version 11+ (vs. version 10-) is slightly differently:

<img width="1324" alt="Screen Shot 2020-10-27 at 20 07 39" src="https://user-images.githubusercontent.com/602265/97268027-1c0e7980-1890-11eb-9e9c-25fb3f7ee6c3.png">

nickerso:
why is this a problem? i.e., it would be good to have some useful information rather than "No description provided".

agarny:
> why is this a problem? i.e., it would be good to have some useful information rather than "No description provided".

Added some description.

PR #809, opened 2021-01-27
CMake: only run the ClangFormat test if we have version 11+ of ClangFormat

agarny:
Addresses issue #808.


cmake/environmentchecks.cmake
```diff
@@ -121,7 +121,17 @@ else()
 endif()
 
 if(CLANG_FORMAT_EXE AND GIT_EXE)
-  set(CLANG_FORMAT_TESTING_AVAILABLE TRUE CACHE INTERNAL "Executables required to run the ClangFormat test are available.")
+  set(CLANG_FORMAT_VERSION_NEEDED 11)
```
hsorby: I would rename this variable to `CLANG_FORMAT_VERSION_MINIMUM`.

hsorby (changes requested):

cmake/environmentchecks.cmake
agarny: Agreed, done in commit 38011d81d.

hsorby (approved):

kerimoyle (approved):

PR #836, opened 2021-03-26
Annotator: replace std::pair with AnyCellmlElement

### LINKED ISSUE #748: Annotator: don't have `item()` return an `std::pair`
In PR #640, `Annotator::item()` returns an `AnyItem` which is an alias for `std::pair<CellMLElement, std::any>`. We shouldn't return an `std::pair` in the API. This makes it awkward to use for the end-user (`x.first` and `x.second` don't mean anything). So, instead, we should have a class that overlaps `std::pair<CellMLElement, std::any>`.

agarny:
Closes #748.


src/api/libcellml/types.h
```diff
@@ -235,6 +236,86 @@ class LIBCELLML_EXPORT VariablePair
...
-using AnyItem = std::pair<CellmlElementType, std::any>;
+class LIBCELLML_EXPORT AnyItem
+{
+public:
+    ~AnyItem(); /**< Destructor. */
+    AnyItem(const AnyItem &rhs) = delete; /**< Copy constructor. */
+    AnyItem(AnyItem &&rhs) noexcept = delete; /**< Move constructor. */
+    AnyItem &operator=(AnyItem rhs) = delete; /**< Assignment operator. */
+
+    static AnyItemPtr create() noexcept;
```
hsorby: Requires Doxygen documentation.

src/api/libcellml/types.h
```diff
@@ -235,6 +236,86 @@ class LIBCELLML_EXPORT VariablePair
...
+    /**
```
hsorby, line 238: New line here.

src/api/libcellml/types.h
```diff
@@ -235,6 +236,86 @@ class LIBCELLML_EXPORT VariablePair
...
+     *
+     * Factory method to create a @ref AnyItemPtr.  Create with @c std::any item
+     * and type with::
+     *
+     *   auto anyItem = libcellml::AnyItem::create(item, type);
+     *
+     * @return A smart pointer to a @ref AnyItem object.
+     */
+    static AnyItemPtr create(CellmlElementType type, const std::any &item) noexcept;
+    // KRM
```
hsorby: Remove

src/api/libcellml/types.h
```diff
@@ -235,6 +236,86 @@ class LIBCELLML_EXPORT VariablePair
...
+    static AnyItemPtr create(CellmlElementType type, const ComponentPtr &item) noexcept;
+    static AnyItemPtr create(CellmlElementType type, const ImportSourcePtr &item) noexcept;
+    static AnyItemPtr create(CellmlElementType type, const ModelPtr &item) noexcept;
+    static AnyItemPtr create(CellmlElementType type, const ResetPtr &item) noexcept;
+    static AnyItemPtr create(CellmlElementType type, const UnitPtr &item) noexcept;
+    static AnyItemPtr create(CellmlElementType type, const UnitsPtr &item) noexcept;
+    static AnyItemPtr create(CellmlElementType type, const VariablePtr &item) noexcept;
+    static AnyItemPtr create(CellmlElementType type, const VariablePairPtr &item) noexcept;
```
hsorby: All of these require Doxygen documentation.

src/bindings/interface/types.i
```diff
@@ -198,7 +224,132 @@ Provides support for shared pointers declared in types.h.
...
+
+  %}
+}
+
 %ignore libcellml::Unit::create;
 %ignore libcellml::VariablePair::create;
+%ignore libcellml::AnyItem::create;
 
-%include "libcellml/types.h"
+%include "libcellml/types.h"
```
hsorby: New line

src/types.cpp
```diff
@@ -113,4 +118,157 @@ bool VariablePair::isValid() const
...
 
+/**
+ * @brief The AnyItem::AnyItemImpl struct.
+ *
+ * The private implementation for the AnyItem class.
+ */
+struct AnyItem::AnyItemImpl
+{
+    std::any mItem = nullptr; /**< std::any item cast for the item.*/
+    CellmlElementType mType = CellmlElementType::UNDEFINED; /**< Type for the item.*/
```
hsorby: space in `.*/` should be `. */`

tests/any_item/any_item.cpp
```diff
@@ -0,0 +1,35 @@
...
+    EXPECT_EQ(libcellml::CellmlElementType::UNDEFINED, anyItem->type());
+}
+
+TEST(AnyItemPtr, createWithArguments)
+{
+    auto model = libcellml::Model::create();
+    auto item = libcellml::AnyItem::create(libcellml::CellmlElementType::MODEL, std::any(model));
+    EXPECT_EQ(libcellml::CellmlElementType::MODEL, item->type());
+    EXPECT_EQ(model, std::any_cast<libcellml::ModelPtr>(item->item()));
+}
```
hsorby: Needs test for all API of the class `setItem` and `setType`.

hsorby (changes requested):
A few things need to be completed.

I would also like to ask should the class be called `AnyCellmlElement` and not `AnyItem`.

src/api/libcellml/types.h
agarny: Done in commit 98458b98.

src/api/libcellml/types.h
agarny: Done in commit 6e7e6831.

src/bindings/interface/types.i
agarny: Done in commit e72e60cb.

src/types.cpp
agarny: Done in commit ffec2304.

agarny:
> A few things need to be completed.

Yes, so far this PR is really just about getting it back on track.

> I would also like to ask should the class be called `AnyCellmlElement` and not `AnyItem`.

Done in commit 63f411db.

tests/any_item/any_item.cpp
agarny: I have removed `setItem()` and `setType()` from the `AnyCellmlElement` API. Indeed, I reckon that a user should only ever consume an `AnyCellmlElement` object, not modify its contents. The latter should only be made possible internally (by the `Annotator` class).

agarny:
Commit 6ed25db contains an awful lot of changes. Something that I really didn't like was that we were returning `std::any` from our API. As we know, this causes issues with SWIG, resulting in various hacks to get around that problem. So, now, although we keep on using `std::any` internally (it's indeed quite convenient), we don't return `std::any` from our API anymore. This should therefore make things easier on the SWIG (yet to be done).

Also, while doing this, I have removed the ability to create an `AnyCellmlElement` object from the API. I just don't understand why we allowed this in the first place. Such an object should only ever be created internally. As it happens, it's the same story with `Issue`. Why would someone using the API want to create an `Issue` object? So, yes, we can't now create those objects from the API. Unsurprisingly, this makes things much easier, both on the C++ front and on the SWIG side of things (at least, it should since I have yet to update that side of the library).

src/api/libcellml/types.h
agarny: Not relevant anymore.

src/api/libcellml/types.h
agarny: Not relevant anymore.

tests/any_item/any_item.cpp
agarny: Not relevant anymore.

agarny:
@hsorby and @nickerso: this now ready for review.

agarny:
@hsorby & @nickerso: I believe this is once again ready for review.

agarny:
Ok, back on track. I have also incorporated some of @hsorby's changes. However:
- I can't see why, in `Validator`, we would use `setVariablePair()` rather than `setMapVariables()` and `setConnection()`. This is internal code and makes it easier for us to read/maintain/etc.
- Why get rid of `resetValue()` and `testValue()` in the `Issue` class? Not surprisingly (?), this breaks our annotator tests.

Anyway, this is now ready for (re-)review.

hsorby:
The main reason for getting rid of `setMapVariables` is so that we don't get it in the API.  I know it is internal but thought it might creep into the API if we kept it.

The issue class should for my thinking return the thing that it is we don't have a `testValue` object so we shouldn't be asking for one and getting a `Reset`.

src/utilities.h
```diff
@@ -656,11 +656,11 @@ bool traverseComponentEntityTreeLinkingUnits(const ComponentEntityPtr &component
  *  Utility function used when linking units names to their corresponding @c Units items.
```
hsorby: @ref Units, not @c.

src/utilities.h
```diff
@@ -630,16 +630,16 @@ std::string makeUniqueId(IdList &idList);
  * Function to support linking units names to their corresponding @c Units items.
  *
  * @param component The component to check.
- * @param issueList A vector of @c IssuePtr items in which unlinked units are recorded for reporting.
+ * @param descriptionList A @c std::vector of @c std::string items in which unlinked units are recorded for reporting.
```
hsorby: This is not what a description list is.

tests/validator/validator.cpp
```diff
@@ -51,10 +51,13 @@ TEST(Validator, unnamedModel)
     const std::vector<std::string> expectedSpecificationHeadings = {
         "2.1.1",
     };
+    const std::vector<std::string> expectedUrls = {
+        "https://cellml-specification.readthedocs.io/en/latest/reference/formal_and_informative/specB01.html?issue=MODEL_NAME",
+    };
     libcellml::ValidatorPtr validator = libcellml::Validator::create();
```
hsorby, line 58: To be consistent we should have an empty line here.

src/utilities.h
```diff
@@ -630,16 +630,16 @@ std::string makeUniqueId(IdList &idList);
  * Function to support linking units names to their corresponding @c Units items.
```
hsorby: @ref Units, not @c.

src/utilities.cpp
```diff
@@ -1292,19 +1282,19 @@ bool linkComponentVariableUnits(const ComponentPtr &component, std::vector<Issue
 
 bool traverseComponentEntityTreeLinkingUnits(const ComponentEntityPtr &componentEntity)
 {
-    std::vector<IssuePtr> issueList;
+    DescriptionList issueList;
```
hsorby: `issueList` is a bad variable name.

src/types.cpp
```diff
@@ -73,14 +68,20 @@ size_t UnitsItem::index() const
 
 bool UnitsItem::isValid() const
 {
-    auto units = mPimpl->mUnits.lock();
-    if (units != nullptr) {
-        return mPimpl->mIndex < units->unitCount();
-    }
-
-    return false;
+    return mPimpl->mIndex < mPimpl->mUnits.lock()->unitCount();
```
hsorby: I think I could cause this to segfault.

src/model_p.h
```diff
@@ -0,0 +1,40 @@
...
+ * distributing the code.
+ */
+struct Model::ModelImpl
+{
+    std::vector<UnitsPtr> mUnits;
+
+    std::vector<UnitsPtr>::const_iterator findUnits(const std::string &name) const;
+    std::vector<UnitsPtr>::const_iterator findUnits(const UnitsPtr &units) const;
+
+    bool equalUnits(const ModelPtr &other) const;
```
hsorby, line 46: Should have Doxygen documentation.

src/importer.cpp
```diff
@@ -193,21 +195,21 @@ bool Importer::ImporterImpl::checkComponentForCycles(const ComponentPtr &compone
...
         auto importedComponent = model->component(component->importReference(), true);
         if (importedComponent == nullptr) {
-            auto issue = Issue::create();
-            issue->setDescription("Component '" + component->name() + "' imports a component named '" + component->importReference() + "' from the model imported from '" + resolvingUrl + "'. The component could not be found.");
-            issue->setLevel(Issue::Level::ERROR);
-            issue->setImportSource(component->importSource());
-            issue->setReferenceRule(Issue::ReferenceRule::IMPORTER_MISSING_COMPONENT);
+            auto issue = Issue::IssueImpl::create();
+            issue->mPimpl->setDescription("Component '" + component->name() + "' imports a component named '" + component->importReference() + "' from the model imported from '" + resolvingUrl + "'. The component could not be found.");
+            issue->mPimpl->setLevel(Issue::Level::ERROR);
```
hsorby: Redundant.

src/importer.cpp
```diff
@@ -193,21 +195,21 @@ bool Importer::ImporterImpl::checkComponentForCycles(const ComponentPtr &compone
...
         auto model = component->importSource()->model();
         if (model == nullptr) {
-            auto issue = Issue::create();
-            issue->setDescription("Component '" + component->name() + "' requires a model imported from '" + resolvingUrl + "' which is not available in the importer.");
-            issue->setLevel(Issue::Level::ERROR);
-            issue->setImportSource(component->importSource());
-            issue->setReferenceRule(Issue::ReferenceRule::IMPORTER_NULL_MODEL);
+            auto issue = Issue::IssueImpl::create();
+            issue->mPimpl->setDescription("Component '" + component->name() + "' requires a model imported from '" + resolvingUrl + "' which is not available in the importer.");
+            issue->mPimpl->setLevel(Issue::Level::ERROR);
```
hsorby: Redundant.

src/importer.cpp
```diff
@@ -155,21 +157,21 @@ bool Importer::ImporterImpl::checkUnitsForCycles(const UnitsPtr &units, History
...
     auto importedUnits = model->units(units->importReference());
     if (importedUnits == nullptr) {
-        auto issue = Issue::create();
-        issue->setDescription("Units '" + units->name() + "' imports units named '" + units->importReference() + "' from the model imported from '" + resolvingUrl + "'. The units could not be found.");
-        issue->setLevel(Issue::Level::ERROR);
-        issue->setImportSource(units->importSource());
-        issue->setReferenceRule(Issue::ReferenceRule::IMPORTER_MISSING_UNITS);
+        auto issue = Issue::IssueImpl::create();
+        issue->mPimpl->setDescription("Units '" + units->name() + "' imports units named '" + units->importReference() + "' from the model imported from '" + resolvingUrl + "'. The units could not be found.");
+        issue->mPimpl->setLevel(Issue::Level::ERROR);
```
hsorby: Redundant.

src/importer.cpp
```diff
@@ -155,21 +157,21 @@ bool Importer::ImporterImpl::checkUnitsForCycles(const UnitsPtr &units, History
...
     auto model = units->importSource()->model();
     if (model == nullptr) {
-        auto issue = Issue::create();
-        issue->setDescription("Units '" + units->name() + "' requires a model imported from '" + resolvingUrl + "' which is not available in the importer.");
-        issue->setLevel(Issue::Level::ERROR);
-        issue->setImportSource(units->importSource());
-        issue->setReferenceRule(Issue::ReferenceRule::IMPORTER_NULL_MODEL);
+        auto issue = Issue::IssueImpl::create();
+        issue->mPimpl->setDescription("Units '" + units->name() + "' requires a model imported from '" + resolvingUrl + "' which is not available in the importer.");
+        issue->mPimpl->setLevel(Issue::Level::ERROR);
```
hsorby: Redundant.

src/api/libcellml/model.h
```diff
@@ -39,6 +39,8 @@ class LIBCELLML_EXPORT Model: public ComponentEntity
                               public std::enable_shared_from_this<Model>
 #endif
 {
+    friend class Parser;
+
```
hsorby: Not required.

hsorby (changes requested):
First review pass.

src/utilities.h
agarny: I am inclined to leave it for another issue? I mean, I just searched for `@c Units` in our codebase and there are many instances of it (and other similar ones).

agarny: Happy to change this particular one though, if you really insist. :)

src/utilities.h
agarny: The description was "wrong" in the first place, as are some other descriptions (just search for `descriptionList` and you will find one for `traverseComponentEntityTreeLinkingUnits()`). Anyway, happy to change the documentation for this (and the other) `descriptionList` if you tell me what you would like to see.

src/utilities.h
agarny: See https://github.com/cellml/libcellml/pull/836#discussion_r634778487.

src/utilities.cpp
agarny: Nothing to do with this PR as such, but ok done in commit b41a9e7.

tests/validator/validator.cpp
agarny: Done in commit a06be2cd.

src/types.cpp
agarny: Not sure who changed the original code to this, but if I did then mea culpa. I agree that this is not good. I have therefore reverted to the original code (see commit 3d375612). However, I am now wondering whether we are still going to get 100% coverage... could it be the reason that someone/I switched to the code above?

src/model_p.h
agarny: Do we seriously need to add documentation every of our single pimpl methods? As the ones writing the code, I would hope that our method and parameter names are clear enough to understand without having to document them. That's also what our review process is about: ensure that our code is clear.

Anyway, done in commit 9084b839.

src/importer.cpp
agarny: Agreed, fixed in commit c2923e4f.

src/importer.cpp
agarny: Should have chcked the whole thing first. Ok, fixed this one and other similar ones in commit 7a25f5ed.

src/importer.cpp
agarny: See commit 7a25f5ed.

src/importer.cpp
agarny: See commit 7a25f5ed.

src/api/libcellml/model.h
agarny: Agreed, fixed in commit bd383374.

src/types.cpp
agarny: I can indeed confirm that we don't have 100% coverage anymore. Ok, need to find a way to "segfault" this now...

agarny: Looking into it, `UnitsItem` is very much an internal type and, yet, it's in our API...

agarny: Ok, maybe not since it's (only) needed by the `Annotator` class.

src/model_p.h
hsorby: In my opinion yes, we should have documentation for every internal method.

src/utilities.h
hsorby: Agreed that we should try and fix this in bulk. Maybe a session during a meeting? perhaps.  I have been fixing them up if I notice them for any functions or methods that I change.  I won't insist but I think it would be better to have it done than not.

agarny: Done in commit b904f477.

src/types.cpp
agarny: Note that coverage has been addressed in commit 9b2745ae. However, this highlights another issue which is that the API allows us to create an `UnitsItem` while it clearly shouldn't. So... we should do what we have for the `Issue` class, i.e. only allow the creation of of `Issue` instances from our internal codebase. Am going to do that...

agarny: First, commit 70796228 implements a better way of testing coverage for `UnitsItem::isValid()`.

src/utilities.h
hsorby: A @c std::vector of variables paired with a description.

tests/units/units.cpp
```diff
@@ -2731,3 +2731,28 @@ TEST(Units, circularImportDeeperLevelBaseUnits)
...
+
+        annotator->setModel(model);
+
+        unitsItem = annotator->item("second_id")->unitsItem();
+
+        EXPECT_TRUE(unitsItem->isValid());
+    }
+
+    EXPECT_FALSE(unitsItem->isValid());
+}
```
hsorby: If this is purely for coverage reasons then it belongs in the coverage test.

src/types.cpp
agarny: I guess that `UnitsItem` is similar to `VariablePair`, i.e. a convenience class. So, ok, I am leaving things as-is then, which means that I have reverted commit 7079622.

src/utilities.h
agarny: Done in commit 4090dfc6.

src/utilities.h
agarny: Done in commit b904f47.

hsorby (changes requested):

agarny:
Ok, I think I have addressed all of @hsorby's comments. So, ready for another round of reviews.

tests/units/units.cpp
agarny: FWIW, this test has been superseded by commit 157189f5. However, it has now been moved to the coverage test file (see commit 172ae199).

src/api/libcellml/annotator.h
```diff
@@ -114,13 +114,13 @@ class LIBCELLML_EXPORT Annotator: public Logger
      * @param id A @c std::string representing the @p id to retrieve.
      * @param index The index of the item to return from the list of items with @p id.
      *
-     * @return An @c AnyItem item as described in item(const std::string &).
+     * @return An @c AnyCellmlElement item as described in item(const std::string &).
```
hsorby: `@ref`

src/api/libcellml/annotator.h
```diff
@@ -114,13 +114,13 @@ class LIBCELLML_EXPORT Annotator: public Logger
...
      */
-    AnyItem item(const std::string &id, size_t index);
+    AnyCellmlElementPtr item(const std::string &id, size_t index);
 
     /**
      * @brief From a list of items in the stored model with the given @p id string,
-     *        this method returns a @c ComponentPtr in the @p index position, if it exists.
+     *        this method returns a @c Component in the @p index position, if it exists.
```
hsorby: `@ref`

src/api/libcellml/annotator.h
```diff
@@ -133,32 +133,13 @@ class LIBCELLML_EXPORT Annotator: public Logger
      * @param id A @c std::string representing the identifier of the item to retrieve.
      * @param index The position of an item within the list of items with the given @p id to retrieve.
      *
-     * @return A @c ComponentPtr on success otherwise @c nullptr.
+     * @return A @c Component on success otherwise @c nullptr.
```
hsorby: `@ref`

src/api/libcellml/annotator.h
```diff
@@ -89,23 +89,23 @@ class LIBCELLML_EXPORT Annotator: public Logger
...
-     *  - <CellmlElementType::UNIT, @ref UnitPtr> - Retrieve the identifier with Units::unitId(size_t) const.
+     *  - <CellmlElementType::UNIT, @ref UnitsItemPtr> - Retrieve the identifier with Units::unitId(size_t) const.
      *  - <CellmlElementType::UNITS, @ref UnitsPtr> - Retrieve the identifier with Units::id().
      *  - <CellmlElementType::VARIABLE, @ref VariablePtr> - Retrieve the identifier with Variable::id().
      *
      * @param id A @c std::string representing the @p id to retrieve.
      *
-     * @return An @c AnyItem item (as described above).
+     * @return An @ref AnyCellmlElement item (as described above).
      */
```
hsorby, line 98: I don't think this documentation is accurate any more.

hsorby (changes requested):
I think the documentation needs a little look.

I have one more thing to check now.

src/api/libcellml/annotator.h
hsorby: Done already.

src/api/libcellml/annotator.h
hsorby: Done already.

src/api/libcellml/annotator.h
hsorby: Done already.

src/api/libcellml/annotator.h
agarny: Done in commit c28817c1.

src/api/libcellml/annotator.h
```diff
@@ -66,46 +66,37 @@ class LIBCELLML_EXPORT Annotator: public Logger
...
+     *
+     * Return the item with the given @p id string. The item returned is an
+     * @ref AnyCellmlElement object containing both:
+     *  - a type (as a @ref CellmlElementType enum); and
+     *  - an item, which can be a @ref Component, @ref ImportSource, @ref Model,
+     *    @ref Reset, @ref Units, @ref UnitsItem, @ref Variable, or
+     *    @ref VariablePair.
+     *
+     * @overload
+     *
```
hsorby: This method is not the overloaded one.

hsorby (changes requested):

src/api/libcellml/annotator.h
agarny: Argh, of course. Fixed in commit 0c635993.

hsorby (dismissed):

src/annotator.cpp
```diff
@@ -437,57 +457,71 @@ void Annotator::setModel(const ModelPtr &model)
...
         addIssueNotFound(id);
-        return false;
+        return 0;
     }
-    return true;
+    return count;
 }
 
-AnyItem Annotator::item(const std::string &id)
+bool Annotator::AnnotatorImpl::canReturnItem(const std::string &id, size_t index) const
```
nickerso: its hard to tell with documentation to know what this method should be doing, but the behaviour of `canReturnItem` is different to the previous `exists` method. Not at all clear why an index is even needed if this is now going to error out with non-unique ids.

nickerso: oops, it doesn't error out but return's `true` yet still sets an issue? This behaviour seems weird...

nickerso (changes requested):
Seems pretty good, but I think I'm missing something. Can you please explain, perhaps by adding some documentation to the `idCount` and `canReturnItem` methods in the annotator implementation, describing why that approach is better than the previous simple `exists` method. I'm not seeing how that is a good change...

src/annotator.cpp
hsorby: I think this is down to the philosophy of the `Annotator`.  The idea is that `id`s should be unique and if they aren't then the user should be told about it.

hsorby: The name `canReturnItem` is more of a hangover from when there was a `std::pair` involved.  I think now it makes more sense to call this method `exists` again.

src/annotator.cpp
```diff
@@ -437,57 +457,71 @@ void Annotator::setModel(const ModelPtr &model)
...
 {
     if (!mAnnotator->hasModel()) {
         addIssueNoModel();
-        return false;
+        return 0;
     }
-    if ((mIdList.count(id) <= index) || (mIdList.count(id) == 0)) {
+
+    auto count = mIdList.count(id);
+    if ((count <= index) || (count == 0)) {
```
hsorby: I believe the `(count == 0)` part of this condition is redundant.

src/annotator.cpp
nickerso: > I think this is down to the philosophy of the Annotator. The idea is that ids should be unique and if they aren't then the user should be told about it.

but this internal API change is not reflected in the current public API documentation or philosophy...or have I missed something?

hsorby: I have made some changes to clarify this.

hsorby (dismissed):

src/api/libcellml/annotator.h
```diff
@@ -70,42 +70,36 @@ class LIBCELLML_EXPORT Annotator: public Logger
...
     /**
      * @brief Return the item at @p index with the @p id.
      *
-     * From a list of items in the stored model with the given @p id string,
-     * this method returns the item in the @p index position.
+     * From a list of items extracted from the stored model find and return the item with
+     * the given @p id string and @p index.  The default @p index is zero if none is given.
      *
-     * See item(const std::string &) for a full breakdown of the @ref AnyCellmlElement return value.
+     * Errors are logged if:
```
nickerso: issues not errors?

nickerso (changes requested):
Thanks @hsorby - these changes definitely make it clearer what that method is doing. But I still can't see how the introduction of this issue for non-unique IDs is an improvement? There shouldn't be an issue created if I ask for any item with an index greater than 0 as I clearly know that there are duplicate IDs. Adding an issue every time I do that just clogs up the annotator with issues that I'd need to check and ignore...or more likely just never check for issues as I know that I'm going to ignore them all every time.

src/annotator.cpp
```diff
@@ -512,22 +517,29 @@ bool Annotator::AnnotatorImpl::exists(const std::string &id, size_t index) const
         return false;
     }
 
+
```
agarny: Why this empty line?

src/annotator.cpp
```diff
@@ -512,22 +517,29 @@ bool Annotator::AnnotatorImpl::exists(const std::string &id, size_t index) const
...
 
-    addIssueNonUnique(id);
     return true;
 }
 
 AnyCellmlElementPtr Annotator::item(const std::string &id, size_t index)
 {
-    return mPimpl->exists(id, index) ? std::move(items(id)[index]) : AnyCellmlElement::AnyCellmlElementImpl::create();
+    auto privateIndex = realIndex(index);
+    return mPimpl->exists(id, index) ? std::move(items(id)[privateIndex]) : AnyCellmlElement::AnyCellmlElementImpl::create();
```
agarny: Why not simply:
```c++
return mPimpl->exists(id, index) ?
           std::move(items(id)[realIndex(index)]) :
           AnyCellmlElement::AnyCellmlElementImpl::create();
```

src/annotator.cpp
```diff
@@ -574,7 +586,8 @@ ComponentPtr Annotator::component(const std::string &id, size_t index)
 {
     mPimpl->update();
     if (mPimpl->exists(id, index)) {
-        auto i = items(id).at(index);
+        auto privateIndex = realIndex(index);
+        auto i = items(id).at(privateIndex);
         return i->component();
```
agarny: Why not simply (same elsewhere):
```c++
return items(id).at(realIndex(index))->component();
```
I know you do, but I really don't see the benefit of all those temporary variables.

hsorby: 5d65515

src/annotator.cpp
hsorby: 5d65515

src/annotator.cpp
hsorby: 5d65515

hsorby (dismissed):

hsorby (dismissed):

hsorby (approved):

nickerso (approved):

nickerso:
@agarny - will let you hit merge if you are still happy with this one.

PR #839, opened 2021-03-29
A collection of DAE models (and their ODE equivalent whenever possible)

### LINKED ISSUE #838: Create a collection of DAE models
We need a collection of DAE examples that we can use to build DAE code generation abilities in libCellML.

agarny:
Closes #838.


hsorby:
How many models (pairs of models) are you expecting to add to this collection?

agarny:
No idea at this stage. There is the Robertson model, which can be seen as our simple test model. Then, there is that Saucerman model that I mentioned in our Zoom call. That one is definitely more involved (from what I remember). It should be added at some point (after having doubled checked its status in PMR).

agarny:
This is now part of PR #942 (see e982f342 which includes the ODE and DAE versions of the Robertson model), so closing as a result.

Issue #845, opened 2021-04-21
Javascript binding fails on Unit


Moop204:
Running any cellml file with 'Unit' element causes a crash. 

```
stoi: no conversion
undefined
RuntimeError: abort(undefined). Build with -s ASSERTIONS=1 for more info.
/home/moop204/Documents/uni/thesis/test/libcellml-test/node_modules/libcellml.js/libcellml.common.js:58
        throw ex 
        ^

RuntimeError: abort(RuntimeError: abort(undefined). Build with -s ASSERTIONS=1 for more info.). Build with -s ASSERTIONS=1 for more info.
    at process.abort (/home/moop204/Documents/uni/thesis/test/libcellml-test/node_modules/libcellml.js/libcellml.common.js:224:7)
    at process.emit (events.js:315:20)
    at processPromiseRejections (internal/process/promises.js:209:33)
    at processTicksAndRejections (internal/process/task_queues.js:98:32)
```
After reformatting libcellml.common I find that it fails in this bit of code:
```
if (ENVIRONMENT_IS_NODE) {
  if (ENVIRONMENT_IS_WORKER) {
    scriptDirectory = require("path").dirname(scriptDirectory) + "/";
  } else {
    scriptDirectory = __dirname + "/";
  }
  read_ = function shell_read(filename, binary) {
    if (!nodeFS) nodeFS = require("fs");
    if (!nodePath) nodePath = require("path");
    filename = nodePath["normalize"](filename);
    return nodeFS["readFileSync"](filename, binary ? null : "utf8");
  };

  readBinary = function readBinary(filename) {
    var ret = read_(filename, true);
    if (!ret.buffer) {
      ret = new Uint8Array(ret);
    }
    assert(ret.buffer);
    return ret;
  };

  if (process["argv"].length > 1) {
    thisProgram = process["argv"][1].replace(/\\/g, "/");
  }
  arguments_ = process["argv"].slice(2);
  process["on"]("uncaughtException", function (ex) {
    if (!(ex instanceof ExitStatus)) {
      throw ex; <-- HERE
    }
  });
  process["on"]("unhandledRejection", abort);
  quit_ = function (status) {
    process["exit"](status);
  };
  Module["inspect"] = function () {
    return "[Emscripten Module object]";
  };
}
```

Code run is:
```
const cellml = require('libcellml.js/libcellml.common');
cellml().then(libcellml=>{
    const p = new libcellml.Parser();
[...]

PR #846, opened 2021-04-22
Units prefix improvements

hsorby:
Looking at avoiding the exception raised by `std::stoi` this causes problems for web assembly binaries.  Hopefully other exceptions are harder to trigger.

Addresses #845.


src/units.cpp
```diff
@@ -328,14 +333,26 @@ void Units::addUnit(const std::string &reference, const std::string &prefix, dou
     UnitDefinition ud;
     ud.mReference = reference;
     // Allow all nonzero user-specified prefixes
-    try {
-        int prefixInteger = std::stoi(prefix);
-        if (prefixInteger != 0.0) {
+
+    bool predefinedPrefix = false;
```
nickerso: can't this use the new `convertPrefixToInt()` method?

hsorby: No because we don't know the state of the prefix in this method.  The `convertPrefixToInt()` method is for when the input is well formed.  That is, it is either a known prefix string or a valid CellML integer.

nickerso (dismissed):
looks good once the clang formatting is kept happy...

nickerso (dismissed):

src/units.cpp
```diff
@@ -181,12 +181,17 @@ bool updateUnitMultiplier(const UnitsPtr &units, int direction, double &multipli
...
 
             if (isStandardPrefixName(pre)) {
                 prefixMult = standardPrefixList.at(pre);
+            } else if (pre.empty()) {
+                prefixMult = 0;
+            } else if (convertToInt(pre, prefixInt)) {
+                prefixMult = prefixInt;
             } else {
                 return false;
             }
```
agarny: I would use our new `convertPrefixToInt()` method for this, which means that `convertPrefixToInt()` would have to be reworked a bit to return `true` or `false` depending on whether the conversion was successful or not. IOW, I don't like "duplication" of code.

agarny (changes requested):
Looks good besides some minor refactoring that I would like to see done.

src/units.cpp
hsorby: 9a0c7b9

src/units.cpp
```diff
@@ -327,15 +329,27 @@ void Units::addUnit(const std::string &reference, const std::string &prefix, dou
 {
     UnitDefinition ud;
     ud.mReference = reference;
+
     // Allow all nonzero user-specified prefixes
```
agarny: Need a full stop.

src/units.cpp
```diff
@@ -185,8 +185,10 @@ bool updateUnitMultiplier(const UnitsPtr &units, int direction, double &multipli
...
 
-            if (isStandardPrefixName(pre)) {
-                prefixMult = standardPrefixList.at(pre);
+            // Make sure convertPrefixToInt is safe to use whilst trying to avoid
+            // causing an exception.
+            if (pre.empty() || isStandardPrefixName(pre) || isCellMLInteger(pre)) {
+                prefixMult = convertPrefixToInt(pre);
             } else {
                 return false;
             }
```
agarny: I was expecting something like (not tested):
```c++
bool ok;

prefixMult = convertPrefixToInt(pre, &ok);

if (!ok) {
    return false;
}
```
with
```c++
int convertPrefixToInt(const std::string &in, bool *ok)
{
    int prefixInt = 0;

    if (ok != nullptr) {
        *ok = true;
    }

    if (isStandardPrefixName(in)) {
        prefixInt = standardPrefixList.at(in);
    } else if (in.empty()) {
        prefixInt = 0;
    } else if (isCellMLInteger(in)) {
        convertToInt(in, prefixInt);
    } else if (ok != nullptr) {
        *ok = false;
    }

    return prefixInt;
}
```

agarny (changes requested):
I would go one extra step on that refactoring.

src/units.cpp
hsorby: My preference is to keep `convertPrefixToInt` as it is.  What do you think @nickerso?

src/units.cpp
hsorby: 90d15a2

src/units.cpp
agarny: You can still call `convertPrefixToInt()` the way you were doing it before, it's just that now you don't have to do those ugly tests before calling it in `updateUnitMultiplier()`. This makes code maintenance unnecessarily "complex".

hsorby: True, but I am also looking at the other locations that this function is used and that proposal doesn't work well for those cases.

nickerso: I've already approved this one twice, so I think its pretty clear I'm happy with it as is....

nickerso (dismissed):

src/units.cpp
agarny: What other locations? Re `convertPrefixToInt()`, if you want the exact same behaviour as your version, then you could have:
```c++
int convertPrefixToInt(const std::string &in, bool *ok)
{
    int prefixInt = 0;

    if (ok != nullptr) {
        *ok = true;
    }

    if (isStandardPrefixName(in)) {
        prefixInt = standardPrefixList.at(in);
    } else if (in.empty()) {
        prefixInt = 0;
    } else if ((ok == nullptr) || isCellMLInteger(in)) {
        convertToInt(in, prefixInt);
    } else if (ok != nullptr) {
        *ok = false;
    }

    return prefixInt;
}
```

hsorby: `convertPrefixToInt` is used in the analyser and validator right.

agarny: Indeed, they use it and I have just tested my implementation at https://github.com/cellml/libcellml/pull/846#discussion_r618845525 and the one at https://github.com/cellml/libcellml/pull/846#discussion_r618896294, and it all works fine. So, yes, I would use my implementation at https://github.com/cellml/libcellml/pull/846#discussion_r618845525.

agarny: FWIW `convertPrefixToInt()` can be optimised a bit (since we initialise `prefixInt` to `0`:
```c++
int convertPrefixToInt(const std::string &in, bool *ok)
{
    int prefixInt = 0;

    if (ok != nullptr) {
        *ok = true;
    }

    if (isStandardPrefixName(in)) {
        prefixInt = standardPrefixList.at(in);
    } else if (isCellMLInteger(in)) {
        convertToInt(in, prefixInt);
    } else if ((ok != nullptr) && !in.empty()) {
        *ok = false;
    }

    return prefixInt;
}
```

hsorby: 7ebbfc9

agarny (dismissed):

nickerso (approved):
Not a fan of the new `convertPrefixToInt()` method as it is now inconsistent with `convertToInt()` and `convertToReal()`. But otherwise this is good to go (assuming my changes to the documentation of that method are correct).

agarny:
> Not a fan of the new `convertPrefixToInt()` method as it is now inconsistent with `convertToInt()` and `convertToReal()`.

Those are internal methods and the new `convertPrefixToInt()` method ensures easier maintenance of our code. If we really want, we could update `convertToInt()` and `convertToReal()` to also take an optional `ok` argument.

> But otherwise this is good to go (assuming my changes to the documentation of that method are correct).

Looks good to me, so approving this PR, but will leave it to @hsorby to merge it in after he has confirmed he's also happy with your change.

agarny (approved):

PR #851, opened 2021-04-25
Allow only out of range exceptions for stoi and stod.

### LINKED ISSUE #850: Avoid all but out of range exceptions when using stoi and stod.
The most common exceptions we face are those raised by `stoi` and `stod`.  If we can minimise these then the Javascript bindings will fail less.

hsorby:
Closes #850.


hsorby:
This is ready for review.  Reasonably small and shouldn't take long. Has major benefits for the javascript bindings.

agarny (approved):

nickerso (approved):

PR #853, opened 2021-04-27
Rename Unit class to UnitsItem class.

### LINKED ISSUE #852: Rename Unit to UnitsItem
Some confusion in the API regarding unit children of units objects (c.f., `unit` elements as children of `units` elements in the CellML XML serialization). 

To help clarify, we should rename the current `Unit` object (pointer to units parent and an index) to be `UnitsItem`.

hsorby:
Closes #852.

### DIFF
--- a/src/api/libcellml/types.h
+++ b/src/api/libcellml/types.h
@@ -75,8 +75,8 @@ class ParentedEntity; /**< Forward declaration of parented entity class. */
 using ParentedEntityPtr = std::shared_ptr<ParentedEntity>; /**< Type definition for shared parented entity pointer. */
 class Reset; /**< Forward declaration of Reset class. */
 using ResetPtr = std::shared_ptr<Reset>; /**< Type definition for shared reset pointer. */
-class Unit; /**< Forward declaration of Unit class. */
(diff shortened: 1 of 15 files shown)


src/api/libcellml/types.h
```diff
@@ -85,31 +85,31 @@ class VariablePair; /**< Forward declaration of VariablePair class. */
 using VariablePairPtr = std::shared_ptr<VariablePair>; /**< Type definition for shared variable pair pointer. */
 
 /**
- * @brief The Unit class
+ * @brief The UnitsItem class
  *
- * The Unit class contains a @ref Units to the parent Units item, and
- * the index to the @ref Unit item within the @ref Units.
+ * The UnitsItem class contains a @ref Units to the parent Units item, and
+ * the index to the @ref UnitsItem item within the @ref Units.
```
nickerso, line 91: This doesn't quite make sense, but can't immediately think of the fix...(also its existing code, not a change introduced here so don't think it needs to be fixed before merging).

nickerso (approved):
looks good assuming testing passes...

agarny (approved):

PR #861, opened 2021-05-14
Update the javascript bindings to use UnitsItem and not Unit.

hsorby:
The javascript bindings are out-of-date with the API they are still using Unit and not the new UnitsItem.

Addresses #845.


nickerso (approved):

agarny (approved):

agarny:
@hsorby: I haven't closed issue #845 in case there are still things that are need fixing, but happy for someone to close it if we think it is / should be fine.

PR #873, opened 2021-06-09
Make resolve imports method more robust

### LINKED ISSUE #592: resolvePath(...) relies on Unix-style path separator
Our current "temporary" import resolution code relies on the very naive `resolvePath()` function to resolve relative import source locations. It does this with a search for the last `/` character and adding the relative path to that (see https://github.com/cellml/libcellml/blob/develop/src/model.cpp#L403). This kind of works fine, until you forget this is what it is doing and you try to use Python's `os.path` module to handle file paths - in which case you end up with no `/` and the import resolution silently falls over :) But it doesn't segfault, so that is good!

Mostly just reporting this so that we keep this in mind when working on the Importer class #531 to try and make this more generally useful.

hsorby:
Resolved #592.


tests/importer/importer.cpp
```diff
@@ -1070,3 +1070,39 @@ TEST(Importer, importingCommonUnitsDefinitions)
     importer->resolveImports(model, resourcePath("importer/"));
     EXPECT_EQ(size_t(0), importer->issueCount());
 }
+
+TEST(Importer, baseFileEndingInBackSlash)
```
agarny: The test should be called `baseFileEndingInForwardSlash`.

tests/importer/importer.cpp
```diff
@@ -1070,3 +1070,39 @@ TEST(Importer, importingCommonUnitsDefinitions)
...
+    auto parser = libcellml::Parser::create();
+
+    auto importedUnits = parser->parseModel(fileContents("importer/units_imported.cellml"));
+
+    EXPECT_TRUE(importedUnits->hasUnresolvedImports());
+    importer->resolveImports(importedUnits, resourcePath("importer/"));
+    EXPECT_FALSE(importedUnits->hasUnresolvedImports());
+}
+
+TEST(Importer, baseFileEndingInForwardSlash)
```
agarny: The test should be called `baseFileEndingInBackSlash `.

agarny (changes requested):
Looks good, except for a couple of tests which name is wrong.

tests/importer/importer.cpp
hsorby: 2d8a05a

tests/importer/importer.cpp
hsorby: 2d8a05a

agarny (dismissed):

CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ cmake_minimum_required(VERSION 3.10.2)
 set(PROJECT_NAME libCellML)
 set(PROJECT_URL https://libcellml.org)
 set(_PROJECT_VERSION 0.2.0)
-set(PROJECT_DEVELOPER_VERSION -dev.19)
+set(PROJECT_DEVELOPER_VERSION -dev.21)
```
nickerso, line 20: why the version change in this PR?

src/api/libcellml/importer.h
```diff
@@ -67,15 +67,15 @@ class LIBCELLML_EXPORT Importer: public Logger
...
      *
      * Resolve all @c Component and @c Units imports by loading the models
-     * from local disk through relative URLs.  The @p baseFile is used to determine
-     * the full path to the source model relative to this one.
+     * from local disk through relative URLs.  The @p basePath defines
+     * the full path to the directory that import URLs are relative to.
      *
      * @param model The @c Model whose imports need resolution.
-     * @param baseFile The @c std::string location on local disk of the source @c Model.
+     * @param basePath The (effective) directory on local disk of the source @c Model.
```
nickerso: I don't know what `(effective)` means here. Perhaps clearer to have something more like: The full path used to resolve relative import URLs. ?

CMakeLists.txt
hsorby: I think I forgot to merge the changes back in from the latest dev releases I made.  We need it at that level because that is what we have already used.   The release script is still a work in  progress.

src/api/libcellml/importer.h
hsorby: 49c3f4e

agarny (approved):

nickerso (approved):

PR #875, opened 2021-06-21
Report errors from importing invalid models

### LINKED ISSUE #866: Report on errors encountered from importing invalid models
Currently if an importing model imported a model that was not valid XML or an invalid CellML model there is no error supplied.  The model is set to the `nullptr` or only part of the model that could be read in.

hsorby:
Fixes #866.


src/api/libcellml/logger.h
```diff
@@ -89,6 +103,20 @@ class LIBCELLML_EXPORT Logger
      */
     IssuePtr error(size_t index) const;
 
+    /**
+     * @brief Remove issue of level ERROR at the specified @p index.
+     *
+     * Removes a error at @p index.  If @c true is returned
```
agarny: "Removed a[n] error..."

src/api/libcellml/logger.h
```diff
@@ -67,6 +67,20 @@ class LIBCELLML_EXPORT Logger
...
+     * Removes an issue at @p index.  If @c true is returned
+     * then the issue at @p index was successfully removed.
+     * If @p index is within the range [0, \#issues)
+     * then @c true is returned, otherwise @c false is returned.
+     *
+     * @param index The index of the issue to remove.
+     *
+     * @return @c true if the issue is removed, @c false otherwise.
+     */
+    bool removeIssue(size_t index);
```
agarny: Do we really want this kind of method (same for `removeError()`, `removeWarning()`, and `removeMessage()`) in the API? I wouldn't expect an API user to need to remove a given issue/error/warning/message. If anything, it should only be needed internally...?

src/importer.cpp
```diff
@@ -369,6 +383,28 @@ bool Importer::ImporterImpl::fetchImportSource(const ImportSourcePtr &importSour
     return true;
 }
 
+bool isErrorRelatedToComponent(const IssuePtr &error, const ComponentPtr &component)
+{
+    auto errorType = error->item()->type();
+    if (CellmlElementType::COMPONENT == errorType) {
```
agarny: I understand that for tests, we want to have `expectedResult == errorType `, but couldn't we simply have `errorType == expectedResult` here? (Same below.)

tests/importer/importer.cpp
```diff
@@ -1106,3 +1106,79 @@ TEST(Importer, basePathNotEndingInSlash)
...
+{
+    auto e = "The attempt to import the model at '" + resourcePath("importer/not_even_proper.xml") + "' failed: the file is not valid XML.";
+
+    auto importer = libcellml::Importer::create();
+    auto parser = libcellml::Parser::create();
+
+    auto model = parser->parseModel(fileContents("importer/import_invalid_xml.cellml"));
+
+    importer->resolveImports(model, resourcePath("importer"));
+    EXPECT_EQ(size_t(1), importer->errorCount());
```
nickerso, line 1120: random question for @hsorby - in the original commit of the failing tests to demonstrate this problem, you were expecting 3 errors. Was that just a random number or did you expect there to be other errors that became non-errors as you implemented the fix?

tests/analyser/analyser.cpp
```diff
@@ -874,6 +874,9 @@ TEST(Analyser, threeEquivalentExternalVariablesNotIncludingPrimaryVariable)
                                                                    expectedReferenceRules(expectedIssues.size(), libcellml::Issue::ReferenceRule::ANALYSER_EXTERNAL_VARIABLE_USE_PRIMARY_VARIABLE),
                                                                    expectedUrls(expectedIssues.size(), "https://libcellml.org/documentation/guides/latest/runtime_codes/index?issue=ANALYSER_EXTERNAL_VARIABLE_USE_PRIMARY_VARIABLE"),
                                                                    analyser);
+
+    EXPECT_TRUE(analyser->removeMessage(0));
+    EXPECT_FALSE(analyser->removeMessage(0));
```
agarny: Is the idea to test the new Logger API? If so, and besides the fact that I think it shouldn't exist, then it should be done in a Logger-specific test, not here.

tests/importer/importer.cpp
hsorby: It was semi-random, I was expecting around three errors but in the implementation once I hit an XML error I stopped and just returned the first one.

tests/coverage/coverage.cpp
```diff
@@ -104,3 +108,47 @@ TEST(Coverage, unitsItem)
...
+    const std::string in =
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"model_name\">\n"
+        "  <encapsulation/>\n"
+        "</model>\n";
+
+    const std::vector<std::string> expectedIssues = {
+        "Encapsulation in model 'model_name' does not contain any child elements.",
+    };
+
```
agarny: Not used, so delete.

tests/importer/importer.cpp
```diff
@@ -1106,3 +1106,79 @@ TEST(Importer, basePathNotEndingInSlash)
...
+    EXPECT_EQ(e, importer->error(0)->description());
+}
+
+TEST(Importer, importInvalidUnitsFromCellmlModel)
+{
+    auto e = "Encountered an error when resolving units 'units1_imported' from '" + resourcePath("importer/invalid_model.cellml") + "'.";
```
agarny, line 1126: This message doesn't help me to determine what is wrong when resolving units 'units1_imported'. So, as an end-user, how would I go about fixing the problem?

Same question for tests `importInvalidComponentFromCellmlModelComponentError`, `importInvalidComponentFromCellmlModelVariableError` and `importInvalidComponentFromCellmlModelResetError`.

agarny (changes requested):

tests/importer/importer.cpp
hsorby: A) I think it does. It tells the user that the units they are trying to import as 'units1_imported' has some problem.  The problem was noticed when trying to read in the units from the model defined by the path.
B) You would have to parse the model at the path and look at the errors given and decide if they are errors that effect your use of the model.

src/importer.cpp
hsorby: c645aa8

src/api/libcellml/logger.h
hsorby: ad4108f

tests/coverage/coverage.cpp
hsorby: cf56598

tests/analyser/analyser.cpp
hsorby: cf56598

src/api/libcellml/logger.h
hsorby: cf56598

src/api/libcellml/logger.h
```diff
@@ -146,6 +146,20 @@ class LIBCELLML_EXPORT Logger
...
+     * then the error at @p index was successfully removed.
+     * If @p index is within the range [0, \#errors)
+     * then @c true is returned, otherwise @c false is returned.
+     *
+     * @param index The index of the error to remove.
+     *
+     * @return @c true if the error is removed, @c false otherwise.
+     */
+    bool removeError(size_t index);
+
```
agarny: If, internally, we need to call `removeError()`, then I would expect that method to be part of `Logger`'s private implementation, not here in the `protected` section. (Same for `addIssue()` in fact.)

tests/importer/importer.cpp
agarny: Agreed with **A)**, but I don't see why I would need to do **B)**. The importer tells us that something is wrong, so I would expect to be told exactly what is wrong and not be "forced" to do **B)** to find out exactly what is wrong.

nickerso: the goal of this PR is to simply get an error when something goes wrong with a bit more information than the current `nullptr` burried in the resolved model. This seems like a good first step in that direction without needing to spend time on implementing a full propagation of error handling up model import hierarchies. This now gives the user sufficient information to track down the problem.

In future we could look to propagate all errors in way that is safe, meaningful, and robust across a hirarchy of model imports, but I don't think we need that now.

src/api/libcellml/logger.h
hsorby: Indeed, I think this might be possible with moving `setParent` into a private implementation.  Will look at moving these once that works.

agarny: Just to be sure: is this going to be done as part of this PR or another one? If the latter, do we have an issue for it?

hsorby: This will be done after [internalise set parent](https://github.com/cellml/libcellml/pull/876) is done. No issue for it currently.

agarny (approved):

src/api/libcellml/logger.h
hsorby: Has now #878.

nickerso (approved):

PR #876, opened 2021-06-23
Privatise set parent

### LINKED ISSUE #858: Take setParent out of the public API
The `setParent` method should not be available in the public API and should be made only accessible through the private implementation.

hsorby:
Resolves #858.


src/api/libcellml/component.h
```diff
@@ -435,8 +438,10 @@ class LIBCELLML_EXPORT Component: public ComponentEntity, public ImportedEntity
 
     bool doEquals(const EntityPtr &other) const override; /**< Virtual implementation method for equals, @private. */
 
-    struct ComponentImpl; /**< Forward declaration for pImpl idiom, @private. */
-    ComponentImpl *mPimpl; /**< Private member to implementation pointer, @private. */
+    class ComponentImpl; /**< Forward declaration for pImpl idiom, @private. */
+
+    ComponentImpl *pFunc(); /**< Getter for private implementation pointer, @private. */
+    ComponentImpl const *pFunc() const; /**< Const getter for private implementation pointer, @private. */
```
agarny: Why do we now have `pFunc()` and can't just use `mPimpl` as we used to? Is that because we have a "normal" getter and a `const` one? If so, where do we need the `const` one?

Otherwise, for consistency, we might want to have:
```c++
const ComponentImpl *pFunc() const;
```
rather than:
```c++
ComponentImpl const *pFunc() const;
```

src/api/libcellml/componententity.h
```diff
@@ -31,10 +31,10 @@ class Component;
 class LIBCELLML_EXPORT ComponentEntity: public NamedEntity
 {
 public:
-    ~ComponentEntity() override; /**< Destructor. */
+    ~ComponentEntity() override = default; /**< Destructor. */
     ComponentEntity(const ComponentEntity &rhs) = delete; /**< Copy constructor. */
     ComponentEntity(ComponentEntity &&rhs) noexcept = delete; /**< Move constructor. */
-    ComponentEntity &operator=(ComponentEntity rhs) = delete; /**< Assignment operator. */
+    ComponentEntity &operator=(ComponentEntity rhs) = delete; /**< Assignment operator @private*/
```
agarny: Missing full point (and space).

src/api/libcellml/componententity.h
```diff
@@ -313,11 +313,12 @@ class LIBCELLML_EXPORT ComponentEntity: public NamedEntity
 
     bool doEquals(const EntityPtr &other) const override; /**< Virtual implementation method for equals, @private. */
 
-    ComponentEntity(); /**< ComponentEntity Constructor. */
+    class ComponentEntityImpl; /**< Forward declaration for pImpl idiom, @private. */
 
-private:
```
agarny, line 318: Why is the `private` section gone?

(The descriptions below reference `@private`...?)

src/api/libcellml/entity.h
```diff
@@ -84,11 +84,37 @@ class LIBCELLML_EXPORT Entity
...
+     * @brief Getter for private implementation pointer.
+     *
+     * Getter for private implementation pointer, @private.
+     *
+     * @return A pointer to EntityImpl.
+     */
+    inline EntityImpl *pFunc()
+    {
+        return mPimpl;
+    }
```
agarny: We are not being consistent:
 - Here, we have a description for the `pFunc()` method, but not in the other headers?
 - Here, we include the body of the `pFunc()` method in the header while it really ought to be in the corresponding `.cpp` file.
 - Some times, we have things in the `protected` section sometimes in the `private` method. I would expect things to be in the `private` method.

src/api/libcellml/namedentity.h
```diff
@@ -32,10 +32,10 @@ namespace libcellml {
 class LIBCELLML_EXPORT NamedEntity: public ParentedEntity
 {
 public:
-    ~NamedEntity() override; /**< Destructor. */
+    ~NamedEntity() override = default; /**< Destructor. */
     NamedEntity(const NamedEntity &rhs) = delete; /**< Copy constructor. */
     NamedEntity(NamedEntity &&rhs) noexcept = delete; /**< Move constructor. */
-    NamedEntity &operator=(NamedEntity rhs) = delete; /**< Assignment operator. */
+    NamedEntity &operator=(NamedEntity rhs) = delete; /**< Assignment operator, @private.*/
```
agarny: Missing space at the end of the description.

src/api/libcellml/parentedentity.h
```diff
@@ -32,10 +32,10 @@ namespace libcellml {
 class LIBCELLML_EXPORT ParentedEntity: public Entity
 {
 public:
-    ~ParentedEntity() override; /**< Destructor. */
+    ~ParentedEntity() override = default; /**< Destructor. */
     ParentedEntity(const ParentedEntity &rhs) = delete; /**< Copy constructor. */
     ParentedEntity(ParentedEntity &&rhs) noexcept = delete; /**< Move constructor. */
-    ParentedEntity &operator=(ParentedEntity rhs) = delete; /**< Assignment operator. */
+    ParentedEntity &operator=(ParentedEntity rhs) = delete; /**< Assignment operator, @private.*/
```
agarny: Missing space.

src/api/libcellml/variable.h
```diff
@@ -467,13 +469,15 @@ class LIBCELLML_EXPORT Variable: public NamedEntity
     VariablePtr clone() const;
 
 private:
+    bool doEquals(const EntityPtr &other) const override; /**< Virtual implementation method for equals, @private. */
```
agarny, line 472: This description doesn't match other similar descriptions.

src/component_p.h
```diff
@@ -0,0 +1,51 @@
...
+#include "utilities.h"
+
+namespace libcellml {
+
+/**
+ * @brief The Component::ComponentImpl struct.
+ *
+ * This struct is the private implementation struct for the Component class.  Separating
+ * the implementation from the definition allows for greater flexibility when
+ * distributing the code.
```
agarny: Not a `struct` anymore.

src/componententity.cpp
```diff
@@ -24,47 +24,36 @@ limitations under the License.
...
+    return reinterpret_cast<ComponentEntity::ComponentEntityImpl *>(Entity::pFunc());
+}
+
+inline ComponentEntity::ComponentEntityImpl const *ComponentEntity::pFunc() const
 {
+    return reinterpret_cast<ComponentEntity::ComponentEntityImpl const *>(Entity::pFunc());
 }
 
-ComponentEntity::~ComponentEntity()
+ComponentEntity::ComponentEntity(ComponentEntityImpl *pImpl)
```
agarny, line 54: Might want to move this implementation before `ComponentEntity::pFunc()`.

src/componententity.cpp
```diff
@@ -24,47 +24,36 @@ limitations under the License.
...
+std::vector<ComponentPtr>::const_iterator ComponentEntity::ComponentEntityImpl::findComponent(const ComponentPtr &component) const
 {
     return std::find_if(mComponents.begin(), mComponents.end(),
                         [=](const ComponentPtr &c) -> bool { return c->equals(component); });
 }
 
-// Interface class Model implementation
-ComponentEntity::ComponentEntity()
-    : mPimpl(new ComponentEntityImpl())
+inline ComponentEntity::ComponentEntityImpl *ComponentEntity::pFunc()
```
agarny: Don't bother with `inline` (compilers are clever enough to do the right thing nowadays), not least since we should otherwise `inline` quite a few other "trivial" methods too.

src/componententity_p.h
```diff
@@ -0,0 +1,44 @@
...
+#include "namedentity_p.h"
+
+namespace libcellml {
+
+/**
+ * @brief The ComponentEntity::ComponentEntityImpl struct.
+ *
+ * This struct is the private implementation struct for the ComponentEntity class.  Separating
+ * the implementation from the definition allows for greater flexibility when
+ * distributing the code.
```
agarny: Again, not a `struct`, but a `class`. (Check other places, just in case.)

src/importsource.cpp
```diff
@@ -36,62 +38,72 @@ using ImportedEntityWeakPtr = std::weak_ptr<ImportedEntity>;
  *
  * The private implementation for the ImportSource class.
  */
-struct ImportSource::ImportSourceImpl
+class ImportSource::ImportSourceImpl: public Entity::EntityImpl
```
agarny, line 41: Starting to use inheritance in our private implementation? Nice! :)

src/parentedentity.cpp
```diff
@@ -83,4 +72,9 @@ bool ParentedEntity::hasAncestor(const ParentedEntityPtr &entity) const
     return hasAncestor;
 }
 
+void ParentedEntity::ParentedEntityImpl::setParent(const ParentedEntityPtr &parent)
+{
+    mParent = parent;
+}
```
agarny, line 78: Why are we bothering with this method? I would now expect our code to replace things like:
```c++
rootValueAst->setParent(ast);
component->pFunc()->setParent(newParent);
```
with:
```c++
rootValueAst->mParent = ast;
component->pFunc()->mParent = newParent;
```

agarny (changes requested):

src/api/libcellml/component.h
hsorby: It is there to enforce `const` correctness which we didn't have before (naughty us).  It gets used in any `const` method.

src/parentedentity.cpp
hsorby: I think you still want to encapsulate.

src/componententity.cpp
hsorby: Some aren't, with the mix of building on Windows, macOS, and Ubuntu with building using SWIG and Emscripten the `inline` keyword becomes important.

src/api/libcellml/entity.h
hsorby: I need this one to be declared in the header (I think it is Emscripten that won't compile/link if it isn't).

src/api/libcellml/variable.h
hsorby: Sorry, can't see this one. Looks the same as all the other descriptions to me.

src/componententity_p.h
hsorby: d2b906e

src/component_p.h
hsorby: d2b906e

src/api/libcellml/componententity.h
hsorby: 2cdf2f8

src/api/libcellml/component.h
hsorby: 28bd6d2

src/api/libcellml/parentedentity.h
hsorby: 870b741

src/api/libcellml/namedentity.h
hsorby: 870b741

src/api/libcellml/componententity.h
hsorby: 870b741

nickerso (dismissed):

src/api/libcellml/analyser.h
```diff
@@ -31,10 +31,10 @@ namespace libcellml {
...
 {
 public:
-    ~Analyser() override; /**< Destructor. */
-    Analyser(const Analyser &rhs) = delete; /**< Copy constructor. */
-    Analyser(Analyser &&rhs) noexcept = delete; /**< Move constructor. */
-    Analyser &operator=(Analyser rhs) = delete; /**< Assignment operator. */
+    ~Analyser() override; /**< Destructor, @private. */
+    Analyser(const Analyser &rhs) = delete; /**< Copy constructor, @private. */
+    Analyser(Analyser &&rhs) noexcept = delete; /**< Move constructor, @private. */
+    Analyser &operator=(Analyser rhs) = delete; /**< Assignment operator, @private. */
```
agarny, line 37: Why `@private` since those methods are `public`? The same for (all the?) other classes whether Analyser-related or not.

src/api/libcellml/variable.h
agarny: I meant that the description for `doEquals()` is similar (in style) to that of, for instance, `Variable()` and `pFunc()` while I would have expected it to look similar to that of, say, `clone()`.

src/namedentity.cpp
```diff
@@ -19,44 +19,41 @@ limitations under the License.
...
+
 namespace libcellml {
 
-/**
- * @brief The NamedEntity::NamedEntityImpl struct.
- *
- * The private implementation for the NamedEntity class.
- */
-struct NamedEntity::NamedEntityImpl
+inline NamedEntity::NamedEntityImpl *NamedEntity::pFunc()
```
agarny: Remove `inline` (and do the same in all the other places where we use `inline`, at least those in this PR).

src/componententity.cpp
agarny: We might then want to review our current use of `inline` since we, for instance, use it with some instances of `pFunc()` but not some others.

src/api/libcellml/analyser.h
nickerso: I believe the `@private` is there to encourage doxygen to produce documentation that is useful (since these are public methods that shouldn't appear in the documented API), and maybe also hints to be used as part of the JS binding work...

agarny: Is that the case @hsorby? If so, the I wouldn't be impressed, but whatever works for us, I guess.

hsorby: Yes I am using it to tidy up the Doxygen output.  I am open for alternatives but this is the best compromise that I have at the moment.

agarny: I don't know enough about Doxygen to suggest an alternative (assuming there is one!). So, "happy" for the current status quo.

src/namedentity.cpp
hsorby: a7f9eb8, df859a2

src/api/libcellml/variable.h
hsorby: 54b1dc8 made the `doEquals()` documentation match that used for `doAddComponent()` which is a better match rather than `clone()`.

src/componententity.cpp
hsorby: a7f9eb8

agarny (approved):

nickerso (approved):

PR #888, opened 2021-06-30
Add test where units are expected to be linked after parsing a model

### LINKED ISSUE #887: Units not linking or not reported as linked when parsing a model
Contrary to my understanding when I created a test for the `Analyser` I thought the `Parser` would link all the Units.  My test showed that either this wasn't the case or the `hasUnlinkedUnits` method doesn't give the correct answer.

hsorby:
Closes #887.


tests/parser/file_parser.cpp
```diff
@@ -162,3 +162,14 @@ TEST(Parser, parseModelWithImportedEquivVariables)
 
     EXPECT_EQ(modelContents, serialisedModel);
 }
+
+TEST(Parser, unitsAreLinkedOnParse)
+{
+    libcellml::ParserPtr p = libcellml::Parser::create();
+    auto model = p->parseModel(fileContents("generator/hodgkin_huxley_squid_axon_model_1952/model.cellml"));
+
+    printModel(model, false);
```
agarny: Debug code, so remove it.

agarny (changes requested):

tests/parser/file_parser.cpp
hsorby: 8e16b79

agarny (approved):

nickerso (approved):

nickerso:
@hsorby - the error in the coverage looks like the buildbot might have run out of space?

hsorby:
I don't think so, I think that is a disk failed error.  I think this is due to an error with the virtual disk more than anything.  I'll try a rebuild.

PR #897, opened 2021-07-04
Move ParentedEntities removeParent out of the public API.

### LINKED ISSUE #895: Remove removeParent from the public API
After moving `setParent` out of the public API we should probably also move `removeParent` out as well.

hsorby:
Fixes #895.

### DIFF
--- a/src/api/libcellml/parentedentity.h
+++ b/src/api/libcellml/parentedentity.h
@@ -46,13 +46,6 @@ public:
     ParentedEntityPtr parent() const;
 
     /**
-     * @brief Clear the pointer to the parent entity.
-     *
-     * Clears the pointer to the parent entity.
-     */
-    void removeParent();
-
-    /**
      * @brief Test to see if the given entity has a parent.
      *
      * Tests the given entity to determine if the entity has a parent.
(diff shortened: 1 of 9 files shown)


src/api/libcellml/parentedentity.h
```diff
@@ -83,7 +76,6 @@ class LIBCELLML_EXPORT ParentedEntity: public Entity
     explicit ParentedEntity(ParentedEntityImpl *pImpl); /**< Constructor for derived classes, @private. */
 
 private:
-    ParentedEntityImpl *pFunc(); /**< Getter for private implementation pointer, @private. */
```
agarny, line 86: Why remove it?

agarny (approved):

src/api/libcellml/parentedentity.h
hsorby: Because `ParentedEntity` only uses the const form of pFunc after moving `removeParent` out.

agarny: I thought it might be the case and I was going to say that it doesn't harm having both signatures, but this would obvisouly break our coverage tests. Ok, fair enough then.

nickerso (approved):

PR #903, opened 2021-07-09
Implement javascript tests for Units class.

### LINKED ISSUE #902: Static Units functions missing in JS bindings
I was trying to check if two Units could have a connection to each other with the static compatible function and noticed that it and the other static functions (equivalent and scalingFactor) also do not have Javascript bindings.

hsorby:
Fixes #902.


src/api/libcellml/units.h
```diff
@@ -564,7 +564,7 @@ class LIBCELLML_EXPORT Units: public NamedEntity, public ImportedEntity
      */
     std::string unitId(size_t index) const;
 
-#ifdef DAMN_YOU_EMSCRIPTEN
+#ifdef EMSCRIPTEN_JAVASCRIPT_BINDINGS
```
agarny: Why not just `JAVASCRIPT_BINDINGS`? Whether we use Empscripten should be irrelevant here?

nickerso (approved):

agarny (approved):

PR #921, opened 2021-07-15
CMake: mark `CCACHE_EXE` as advanced

### LINKED ISSUE #920: CMake: mark `CCACHE_EXE` as advanced
When using a tool like `ccmake`, we only want to see the options that are directly related to libCellML. All the other options should be marked as advanced. We do this for all those other options except for `CCACHE_EXE`. So, we should mark `CCACHE_EXE` as advanced, so that by default we only see libCellML-related options.

agarny:
Fixes #920.


agarny:
@hsorby and @nickerso: it should take less than 10 seconds to review. Literally. :)

hsorby (approved):

nickerso (approved):

PR #901, opened 2021-07-06
Generator: refine the generated code for algebraic models

agarny:
Fixes #899.

PR #941, opened 2021-07-20
Generator: further refinements to the generated C code

### LINKED ISSUE #938: Generator: further refinements to the generated C code
- [x] For an algebraic model, we don't need to define the `VariableInfo` structure (this was an oversight on my part in #901).
- [x] In C, to cast the output of `malloc()` is redundant (see [here](https://en.wikipedia.org/wiki/C_dynamic_memory_allocation#Type_safety)). To cast would normally allow our generated C code to be compilable by a C++ compiler, but then again any decent C++ compiler would complain our use of `malloc()`/`free()`. So, we may as well stick to the C standard.

agarny:
Fixes #938.

### DIFF
--- a/src/generator.cpp
+++ b/src/generator.cpp
@@ -603,7 +603,8 @@ std::string Generator::GeneratorImpl::generateVariableInfoObjectCode(const std::
 
 void Generator::GeneratorImpl::addVariableInfoObjectCode()
 {
-    if (!mLockedProfile->variableInfoObjectString().empty()) {
+    if ((mLockedModel->type() == AnalyserModel::Type::ODE)
+        && !mLockedProfile->variableInfoObjectString().empty()) {
         if (!mCode.empty()) {
             mCode += "\n";
         }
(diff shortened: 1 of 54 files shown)


agarny:
Note that I haven't updated the version of the code generator (too much work without a script!).

hsorby:
You don't need to we haven't released that version of the code generator.

agarny:
> You don't need to we haven't released that version of the code generator.

Indeed, so why did we bother in #901?

nickerso:
> > You don't need to we haven't released that version of the code generator.
> 
> Indeed, so why did we bother in #901?

we only need to update the version once a given version has been released. The update from #901 was needed as the 0.2 profile version had been released, but now we are bumped up to 0.3 we're good with changes until the next release.

agarny:
> > > You don't need to we haven't released that version of the code generator.
> > 
> > 
> > Indeed, so why did we bother in #901?
> 
> we only need to update the version once a given version has been released. The update from #901 was needed as the 0.2 profile version had been released, but now we are bumped up to 0.3 we're good with changes until the next release.

Doh! Makes sense indeed... 🙄

nickerso (approved):

hsorby (approved):

PR #944, opened 2021-07-24
CMake: make sure that the version of ClangFormat is properly determined

### LINKED ISSUE #943: CMake: the version of ClangFormat is not properly retrieved
On my Mac, `clang-format --version` returns `Homebrew clang-format version 12.0.1`. From there, our CMake script determines the version of ClangFormat to be `Homebrew 12.0.1`.

agarny:
Fixes #943.

### DIFF
--- a/cmake/environmentchecks.cmake
+++ b/cmake/environmentchecks.cmake
@@ -143,7 +143,7 @@ if(CLANG_FORMAT_EXE AND GIT_EXE)
   execute_process(COMMAND ${CLANG_FORMAT_EXE} -version
                   OUTPUT_VARIABLE CLANG_FORMAT_VERSION
                   ERROR_QUIET OUTPUT_STRIP_TRAILING_WHITESPACE)
-  string(REGEX REPLACE "clang-format version ([.0-9]+).*" "\\1" CLANG_FORMAT_VERSION "${CLANG_FORMAT_VERSION}")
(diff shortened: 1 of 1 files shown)


hsorby (dismissed):

cmake/environmentchecks.cmake
```diff
@@ -143,7 +143,7 @@ if(CLANG_FORMAT_EXE AND GIT_EXE)
   execute_process(COMMAND ${CLANG_FORMAT_EXE} -version
                   OUTPUT_VARIABLE CLANG_FORMAT_VERSION
                   ERROR_QUIET OUTPUT_STRIP_TRAILING_WHITESPACE)
-  string(REGEX REPLACE "clang-format version ([.0-9]+).*" "\\1" CLANG_FORMAT_VERSION "${CLANG_FORMAT_VERSION}")
+  string(REGEX REPLACE "^.*clang-format version ([.0-9]+).*" "\\1" CLANG_FORMAT_VERSION "${CLANG_FORMAT_VERSION}")
```
nickerso: for completeness, shouldn't the regex be `^.*clang-format version ([.0-9]+).*$` in case there is some installation that adds text to the end of the version string?

agarny: All you added is `$` to mark the end of the string, but this is not needed as such since there is already `.*`. This being said, I put `^`, so we probably ought to be consistent, hence commit 7ff197ea.

hsorby (approved):

nickerso (approved):

PR #947, opened 2021-07-26
CMake: make sure that ClangFormat is of the expected version

### LINKED ISSUE #946: CMake: make sure that ClangFormat is of the expected version
On Windows, I have (through MSVC) ClangFormat 6 installed and upon configuring libCellML, I am allowed to format the code (i.e. do something like `ninja format_code`) and this results in ClangFormat complaining about the options in our `.clang-format` file. This is understandable since we expect ClangFormat 11+.

So, we should only create the `format_code` target if ClangFormat 11+ is available.

agarny:
Fixes #946.


cmake/environmentchecks.cmake
```diff
@@ -138,17 +138,21 @@ else()
   endif()
 endif()
 
-if(CLANG_FORMAT_EXE AND GIT_EXE)
```
hsorby, line 141: Nope, you can't take away that.

hsorby (approved):

nickerso (approved):

PR #960, opened 2021-08-29
Add support for `buildcache`

### LINKED ISSUE #959: Add support for `buildcache`
We currently provide support for [`clcache`](https://github.com/frerich/clcache) on Windows (even though it's officially discontinued) and [`ccache`](https://ccache.dev/) on Linux/macOS.

Another compiler caching tool is [`buildcache`](https://github.com/mbitsnbites/buildcache) which works on Windows, Linux and macOS. Some [benchmarks](https://github.com/mbitsnbites/buildcache/blob/master/doc/benchmarks.md) show that it is on par with `ccache` (and better than `clcache` in my personal experience) if used with direct mode enabled (i.e. `BUILDCACHE_DIRECT_MODE` set to `true`). (When it comes to the cache size, like `ccache` it can be compressed, e.g., by having `BUILDCACHE_COMPRESS_FORMAT` set to `ZSTD`.)

So, here I would recommend adding support `buildcache` and making it the preference, should `clcache`/`ccache` also be present.

FWIW, on my MacBook Pro:
 - With no compiler caching tool:
   - `ninja  312.17s user 22.39s system 1074% cpu 31.147 total`
   - `ninja  309.83s user 22.22s system 1056% cpu 31.432 total`
   - `ninja  325.33s user 23.41s system 1061% cpu 32.862 total`
 - With [`buildcache`](https://github.com/mbitsnbites/buildcache):
   - Cold:
     - `ninja  324.32s user 40.01s system 1021% cpu 35.678 total`
     - `ninja  348.76s user 43.44s system 1034% cpu 37.906 total`
     - `ninja  340.35s user 42.02s system 1061% cpu 36.008 total`
   - Warm:
     - `ninja  56.89s user 9.63s system 890% cpu 7.469 total`
     - `ninja  57.38s user 9.74s system 913% cpu 7.345 total`
     - `ninja  57.22s user 9.83s system 954% cpu 7.023 total`

agarny:
Fixes #959.


cmake/environmentchecks.cmake
```diff
@@ -128,6 +130,10 @@ else()
   endif()
 endif()
 
+if(BUILDCACHE_EXE)
+  set(BUILDCACHE_AVAILABLE TRUE CACHE INTERNAL "Executable required to cache compilations.")
+endif()
+
 if(WIN32)
```
nickerso: given the way its used, shouldn't this not bother checking for `clcache` and `ccache` if `buildcache` is found?

agarny: Fair point, done in commit adb22bde.

nickerso:
I don't know how to test this with Visual Studio. I have buildcache and point CMake at it and I can configure fine, but I'm not seeing any obvious change in the visual studio solution or the build time. I can confirm it doesn't break things, but would be good if there was something obvious I'm missing that would tell me this is having the desire impact on build times...

agarny:
Normally, all you need is for `buildcache` to be on your `PATH`. From there, `libCellML`'s build system should pick it up and make use of it.

Having said this, I have just checked on Windows and it's not taking advantage of `buildcache` (while it does on Linux and macOS). I know (or, rather, I believe since I am now doubting myself!) that I got it to work with libOpenCOR, so I tried `buildcache` there and... it's not working either...!?

Ok, will need to look into it in more details... Argh!

agarny:
I had a quick look and everything works fine when building libCellML in release mode:
- Without `buildcache`: ~90s
- With `buildcache`:
  - Cold: ~105s
  - Warm: ~15s

When it comes to debug mode, I believe this is due to CMake using the `/Zi` option (i.e. "enable debugging information") for MSVC by default. I remember having the same issue with OpenCOR when trying to use `clcache`. In the end, I had to replace the `/Zi` option with the `/Z7` option (i.e. "enable old-style debug info") to be able to take advantage of `clcache`. Will look into this tomorrow, but that would also explain why things didn't work for libOpenCOR either...

agarny:
FWIW, to build libCellML with `buildcache` now works fine... from the command line, but... not using a Visual Studio solution.
```batch
REM Initialise VS2017's environment
call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat"

REM Reset buildcache for a cold build
buildcache -C
buildcache -s

REM Or... reset buildcache for a warm build
buildcache -z

REM Configure libCellML
cd [libCellML]
mkdir build
cd build
cmake -G Ninja -DLIBXML2_LIBRARY="C:\Program Files\LibXml2 2.9.6\lib\xml2.lib" ..

REM Build libCellML
ninja

REM Confirm that buildcache has properly cached our libCellML build
buildcache -s
```
Well, Windows/MSVC is always going to be a real PITA, so as far as I am concerned, I am happy with "just" `buildcache` working from the command line. IOW, this is ready for re-review.

CMakeLists.txt
```diff
@@ -207,22 +207,40 @@ endif()
...
+        set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE_EXE})
+      endif()
     endif()
   endif()
 endif()
 
+# On Windows, if we are using MSVC then replace some compiler flags.
+if(WIN32 AND MSVC)
+  # If we are treating warnings as errors then remove the /W3 option since it 
+  # will eventually be replaced with the /W4 and /WX options.
```
nickerso: struggling to understand this logic. If it will eventually be replaced by other options why remove it here?

CMakeLists.txt
```diff
@@ -207,22 +207,40 @@ endif()
...
+  if(LIBCELLML_TREAT_WARNINGS_AS_ERRORS)
+    replace_compiler_flag("/W3" "")
+  endif()
+
+  # Replace the /Zi option (i.e. enable debugging information), if present, with
+  # the /Z7 option (i.e. enable old-style debugging information) otherwise
+  # caching (be it with buildcache or clcache) won't work.
+  replace_compiler_flag("/Z[iI]" "/Z7")
```
nickerso: I guess this should be wrapped with an "if using buildcache" ?

nickerso (commented):
not really going to be able to test this one if it isn't usable in visual studio...

CMakeLists.txt
agarny: Technically speaking, this is only needed when building using `buildcache` or `clcache`. This being said, it doesn't harm having whether we use a compiler cache or not. In fact, I would leave it as-is to be consistent on Windows with MSVC.

CMakeLists.txt
agarny: Because we add `/W4 /WX` to a target using `target_compile_options()` and this doesn't allow to remove `/W3`. So, what we are doing here is to remove `/W3` all possible targets.

agarny:
> not really going to be able to test this one if it isn't usable in visual studio...

I have updated my [previous comment](https://github.com/cellml/libcellml/pull/960#issuecomment-912154996) which should allow you to test things from the command line.

CMakeLists.txt
nickerso: I'm going to assume that I don't want to be using "old style debugging information" in Visual Studio just to enable something that I can't use.

nickerso:
> > not really going to be able to test this one if it isn't usable in visual studio...
> 
> I have updated my [previous comment](https://github.com/cellml/libcellml/pull/960#issuecomment-912154996) which should allow you to test things from the command line.

@nickerso hasn't compiled anything from a command line for a long time...

CMakeLists.txt
agarny: Done in commit ada6e1b0.

CMakeLists.txt
```diff
@@ -238,7 +238,9 @@ if(WIN32 AND MSVC)
   # Replace the /Zi option (i.e. enable debugging information), if present, with
   # the /Z7 option (i.e. enable old-style debugging information) otherwise
   # caching (be it with buildcache or clcache) won't work.
-  replace_compiler_flag("/Z[iI]" "/Z7")
+  if(BUILDCACHE_EXE OR CCACHE_EXE)
```
nickerso: `CLCACHE` I suspect.

agarny: Damn it! I guess that's why we have a reviewing process... 🙄 Done in commit c89cd35.

CMakeLists.txt
```diff
@@ -41,18 +41,18 @@ include(environmentchecks)
 
 # COMPILER_CACHE  ==> LIBCELLML_COMPILER_CACHE
 set(_PARAM_ANNOTATION "Enable compiler cache.")
-if((WIN32 AND CLCACHE_AVAILABLE) OR (NOT WIN32 AND CCACHE_AVAILABLE))
+if(BUILDCACHE_AVAILABLE OR (WIN32 AND CLCACHE_AVAILABLE) OR (NOT WIN32 AND CCACHE_AVAILABLE))
```
hsorby: This should be reduced to `HAVE_COMPILER_CACHE` in environmentchecks.cmake.

hsorby: The condition also shouldn't rely on `WIN32` for clcache it should be `MSVC`.

CMakeLists.txt
```diff
@@ -41,18 +41,18 @@ include(environmentchecks)
...
   set(LIBCELLML_COMPILER_CACHE ON CACHE BOOL "${_PARAM_ANNOTATION}")
 endif()
-if(DEFINED COMPILER_CACHE AND ((WIN32 AND CLCACHE_AVAILABLE) OR (NOT WIN32 AND CCACHE_AVAILABLE)))
+if(DEFINED COMPILER_CACHE AND (BUILDCACHE_AVAILABLE OR (WIN32 AND CLCACHE_AVAILABLE) OR (NOT WIN32 AND CCACHE_AVAILABLE)))
```
hsorby: `HAVE_COMPILER_CACHE`

CMakeLists.txt
```diff
@@ -207,22 +207,43 @@ endif()
...
-      if(RESULT EQUAL 0)
-        set(CMAKE_CXX_COMPILER_LAUNCHER ${CLCACHEWRAPPER})
-      endif()
-    endif()
+  if(BUILDCACHE_EXE)
+    set(CMAKE_CXX_COMPILER_LAUNCHER ${BUILDCACHE_EXE})
   else()
-    if(CCACHE_EXE)
-      set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE_EXE})
+    if(WIN32)
```
hsorby: Clearly this should be `MSVC`.

CMakeLists.txt
```diff
@@ -207,22 +207,43 @@ endif()
...
+    else()
+      if(CCACHE_EXE)
+        set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE_EXE})
+      endif()
     endif()
   endif()
 endif()
 
+# On Windows, if we are using MSVC then replace some compiler flags.
+if(WIN32 AND MSVC)
```
hsorby: I'm not sure where else you would use `MSVC`.

CMakeLists.txt
```diff
@@ -207,22 +207,43 @@ endif()
...
+    replace_compiler_flag("/W3" "")
+  endif()
+
+  # Replace the /Zi option (i.e. enable debugging information), if present, with
+  # the /Z7 option (i.e. enable old-style debugging information) otherwise
+  # caching (be it with buildcache or clcache) won't work.
+  if(BUILDCACHE_EXE OR CLCACHE_EXE)
+    replace_compiler_flag("/Zi" "/Z7")
+  endif()
+endif()
```
hsorby: All of this should be pushed into `common.cmake`.

cmake/environmentchecks.cmake
```diff
@@ -25,10 +25,13 @@ if (EMSCRIPTEN)
 else ()
   find_package(Python ${PREFERRED_PYTHON_VERSION} COMPONENTS Interpreter Development)
 
-  if(WIN32)
-    find_program(CLCACHE_EXE clcache)
-  else()
-    find_program(CCACHE_EXE ccache)
+  find_program(BUILDCACHE_EXE buildcache)
+  if(NOT BUILDCACHE_EXE)
+    if(WIN32)
```
hsorby: `MSVC`

cmake/environmentchecks.cmake
```diff
@@ -70,13 +73,16 @@ else ()
 
   set(CMAKE_REQUIRED_FLAGS ${_ORIGINAL_CMAKE_REQUIRED_FLAGS})
 
-  if(WIN32)
-    mark_as_advanced(CLCACHE_EXE)
-  else()
-    mark_as_advanced(CCACHE_EXE)
+  if(NOT BUILDCACHE_EXE)
+    if(WIN32)
```
hsorby: `MSVC`

cmake/environmentchecks.cmake
```diff
@@ -128,13 +134,17 @@ else()
...
-if(WIN32)
-  if(CLCACHE_EXE)
-    set(CLCACHE_AVAILABLE TRUE CACHE INTERNAL "Executable required to cache compilations.")
-  endif()
+if(BUILDCACHE_EXE)
+  set(BUILDCACHE_AVAILABLE TRUE CACHE INTERNAL "Executable required to cache compilations.")
 else()
-  if(CCACHE_EXE)
-    set(CCACHE_AVAILABLE TRUE CACHE INTERNAL "Executable required to cache compilations.")
+  if(WIN32)
```
hsorby: `MSVC`

hsorby (changes requested):

CMakeLists.txt
agarny: > This should be reduced to `HAVE_COMPILER_CACHE` in environmentchecks.cmake.

Agreed, not least since this is what I did in libOpenCOR (also it's called `COMPILER_CACHING_AVAILABLE`).

> The condition also shouldn't rely on `WIN32` for clcache it should be `MSVC`.

Agreed.

All done in commit 6a452f55.

CMakeLists.txt
agarny: See commit 6a452f55.

CMakeLists.txt
agarny: We don't even need to test anything in fact. See commit 6a452f55.

CMakeLists.txt
agarny: Agreed. See commit 6a452f55.

cmake/environmentchecks.cmake
agarny: See commit 6a452f55.

cmake/environmentchecks.cmake
agarny: See commit 6a452f55.

cmake/environmentchecks.cmake
agarny: See commit 6a452f55.

CMakeLists.txt
agarny: Why? Wouldn't you put in `common.cmake` functions/macros that are called in various places? The above code is only ever executed once, so we might as well leave it here unless you want to keep `CMakeLists.txt` as short as possible... and make maintenance / reading the code unnecessarily "difficult" by forcing people to look for things elsewhere...

CMakeLists.txt
```diff
@@ -41,18 +41,18 @@ include(environmentchecks)
...
 set(_PARAM_ANNOTATION "Enable compiler cache.")
-if((WIN32 AND CLCACHE_AVAILABLE) OR (NOT WIN32 AND CCACHE_AVAILABLE))
+if(COMPILER_CACHE_AVAILABLE)
   set(LIBCELLML_COMPILER_CACHE ON CACHE BOOL "${_PARAM_ANNOTATION}")
 endif()
-if(DEFINED COMPILER_CACHE AND ((WIN32 AND CLCACHE_AVAILABLE) OR (NOT WIN32 AND CCACHE_AVAILABLE)))
+if(DEFINED COMPILER_CACHE AND COMPILER_CACHE_AVAILABLE)
   set(LIBCELLML_COMPILER_CACHE "${COMPILER_CACHE}" CACHE BOOL "${_PARAM_ANNOTATION}" FORCE)
 elseif(COMPILER_CACHE)
   if(WIN32)
```
hsorby: `MSVC`

CMakeLists.txt
```diff
@@ -207,19 +207,34 @@ endif()
...
+    replace_compiler_flag("/W3" "")
+  endif()
+
+  # Replace the /Zi option (i.e. enable debugging information), if present, with
+  # the /Z7 option (i.e. enable old-style debugging information) otherwise
+  # caching (be it with buildcache or clcache) won't work.
+  if(BUILDCACHE_EXE OR CLCACHE_EXE)
+    replace_compiler_flag("/Zi" "/Z7")
   endif()
 endif()
```
hsorby: This should all go into a macro called something like `apply_compiler_cache_settings` and 
 put in `cmake/macros.cmake` .

cmake/environmentchecks.cmake
```diff
@@ -25,10 +25,15 @@ if (EMSCRIPTEN)
...
-    find_program(CLCACHE_EXE clcache)
-  else()
-    find_program(CCACHE_EXE ccache)
+  if(NOT WIN32 OR MSVC)
+    find_program(BUILDCACHE_EXE buildcache)
+  endif()
+  if(NOT BUILDCACHE_EXE)
+    if(MSVC)
+      find_program(CLCACHE_EXE clcache)
+    elseif(NOT WIN32)
```
hsorby: I don't see how this applies, I think you can have `ccache` on Windows with `mingw`.

cmake/environmentchecks.cmake
```diff
@@ -70,9 +75,11 @@ else ()
 
   set(CMAKE_REQUIRED_FLAGS ${_ORIGINAL_CMAKE_REQUIRED_FLAGS})
 
-  if(WIN32)
+  if(BUILDCACHE_EXE)
+    mark_as_advanced(BUILDCACHE_EXE)
+  elseif(MSVC)
     mark_as_advanced(CLCACHE_EXE)
-  else()
+  elseif(NOT WIN32)
```
hsorby: Comment above applies here also.

hsorby (changes requested):

CMakeLists.txt
agarny: Removing the `/W3` compiler flag has nothing to do with it, but ok let's make things more difficult to maintain. Done in commit 23ce671.

cmake/environmentchecks.cmake
agarny: Agreed. Done in commit 3943f635.

cmake/environmentchecks.cmake
agarny: Agreed. Done in commit 3943f635.

CMakeLists.txt
agarny: Oops! Done in commit 735deb85.

CMakeLists.txt
hsorby: I think this makes it infinitely easier to maintain we have decluttered the main entry point and have a descriptive label for what this part contains.

hsorby (dismissed):

cmake/environmentchecks.cmake
```diff
@@ -25,9 +25,11 @@ if (EMSCRIPTEN)
 else ()
   find_package(Python ${PREFERRED_PYTHON_VERSION} COMPONENTS Interpreter Development)
 
-  find_program(BUILDCACHE_EXE buildcache)
+  if(NOT WIN32 OR MSVC)
```
nickerso: not clear what we are trying to exclude here.

nickerso (approved):
Happy as long as CI passes

cmake/environmentchecks.cmake
agarny: Agreed that it's not needed as such, hence removed in commit 627b5162.

hsorby (approved):

PR #961, opened 2021-08-30
Generator: initialise external variables and fix our compute methods

### LINKED ISSUE #958: Analyser: undesired behaviour with external variables
If I have the following model:

```
<?xml version='1.0' encoding='UTF-8'?>
<model name="test_model" xmlns="http://www.cellml.org/cellml/2.0#" xmlns:cellml="http://www.cellml.org/cellml/2.0#">
    <units name="centimeter">
        <unit prefix="centi" units="metre"/>
    </units>
    <units name="centimeter_squared">
        <unit exponent="2" units="centimeter"/>
    </units>
     <units name="microlitre">
        <unit prefix="micro" units="litre"/>
    </units>
    <units name="microlitre_per_centimeter_cubed">
        <unit units="microlitre"/>
        <unit exponent="-3" units="centimeter"/>
    </units>
   <component name="cell_geometry">
        <variable initial_value="0.01" name="L" units="centimeter"/>
        <variable initial_value="0.0011" name="rad" units="centimeter"/>
        <variable name="vcell" units="microlitre" interface="public"/>
        <variable name="vss" units="microlitre" interface="public"/>
            <apply>
                <eq/>
                <ci>vcell</ci>
                <apply>
                    <times/>
                    <cn cellml:units="microlitre_per_centimeter_cubed">1000</cn>
                    <cn cellml:units="dimensionless">3.14</cn>
                    <ci>rad</ci>
                    <ci>rad</ci>
                    <ci>L</ci>
                </apply>
            </apply>
            <apply>
                <eq/>
                <ci>vss</ci>
                <apply>
                    <times/>
                    <cn cellml:units="dimensionless">0.02</cn>
                    <ci>vcell</ci>
                </apply>
            </apply>
        </math>
    </component>
</model>

```

and I run it through the following Python script

```


import libcellml as lc


def main():
    with open('test_model.cellml') as f:
        content = f.read()

    p = lc.Parser()
    m = p.parseModel(content)

    v = lc.Validator()
    v.validateModel(m)

    print(v.errorCount())

    a = lc.Analyser()

    a.addExternalVariable(lc.AnalyserExternalVariable(m.component("cell_geometry").variable("L")))
    a.addExternalVariable(lc.AnalyserExternalVariable(m.component("cell_geometry").variable("rad")))

    a.analyseModel(m)

    am = a.model()

    g = lc.Generator()
    g.setModel(am)

    pr = lc.GeneratorProfile(lc.GeneratorProfile.Profile.PYTHON)
    g.setProfile(pr)

    print(g.implementationCode())


if __name__ == "__main__":
    main()
```

I get the following Python implementation code

```
# The content of this file was generated using the Python profile of libCellML 0.2.0.

from enum import Enum
from math import *


__version__ = "0.3.0"
LIBCELLML_VERSION = "0.2.0"

VARIABLE_COUNT = 4


class VariableType(Enum):
    CONSTANT = 1
    COMPUTED_CONSTANT = 2
    ALGEBRAIC = 3
    EXTERNAL = 4


VARIABLE_INFO = [
    {"name": "L", "units": "centimeter", "component": "cell_geometry", "type": VariableType.EXTERNAL},
    {"name": "rad", "units": "centimeter", "component": "cell_geometry", "type": VariableType.EXTERNAL},
    {"name": "vcell", "units": "microlitre", "component": "cell_geometry", "type": VariableType.ALGEBRAIC},
    {"name": "vss", "units": "microlitre", "component": "cell_geometry", "type": VariableType.COMPUTED_CONSTANT}
]


def create_variables_array():
    return [nan]*VARIABLE_COUNT


def initialise_constants(variables):
    pass


def compute_computed_constants(variables):
    variables[3] = 0.02*variables[2]


def compute_variables(variables, external_variable):
    variables[0] = external_variable(variables, 0)
    variables[1] = external_variable(variables, 1)
    variables[2] = 1000.0*3.14*variables[1]*variables[1]*variables[0]
```

Which I think has the problem in the `compute_computed_constants` function where the equation for setting `variable[2]` does not appear.  Also, I cannot initialise the external variables.

agarny:
Fixes #958.


agarny:
@hsorby and @nickerso: I believe this is now ready to be reviewed.

tests/resources/generator/cell_geometry_model/model.c
```diff
@@ -0,0 +1,44 @@
...
+
+void computeComputedConstants(double *variables)
+{
+    variables[2] = 1000.0*3.14*variables[1]*variables[1]*variables[0];
+    variables[3] = 0.02*variables[2];
+}
+
+void computeVariables(double *variables)
+{
+}
```
hsorby, line 44: Why do we have this empty function but not an empty `computeRates` function?

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.algebraic.c
```diff
@@ -56,7 +56,7 @@ void deleteArray(double *array)
     free(array);
 }
 
-void initialiseStatesAndConstants(double *states, double *variables)
+void initialiseVariables(double voi, double *states, double *variables, ExternalVariable externalVariable)
```
hsorby, line 59: This model file is called `model.algebriac.c` but it has a `voi`?

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.external.py
```diff
@@ -89,20 +92,20 @@ def compute_computed_constants(variables):
 
 
 def compute_rates(voi, states, rates, variables, external_variable):
-    variables[0] = external_variable(voi, states, rates, variables, 0)
+    variables[0] = external_variable(voi, states, variables, 0)
```
hsorby, line 95: How come this external variable is not seen in the `compute_variables` function?

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.external.py
```diff
@@ -89,20 +92,20 @@ def compute_computed_constants(variables):
...
     variables[11] = 0.1*(variables[0]+25.0)/(exp((variables[0]+25.0)/10.0)-1.0)
     variables[12] = 4.0*exp(variables[0]/18.0)
     rates[0] = variables[11]*(1.0-states[0])-variables[12]*states[0]
     variables[13] = 0.07*exp(variables[0]/20.0)
     variables[14] = 1.0/(exp((variables[0]+30.0)/10.0)+1.0)
     rates[1] = variables[13]*(1.0-states[1])-variables[14]*states[1]
-    variables[17] = external_variable(voi, states, rates, variables, 17)
+    variables[17] = external_variable(voi, states, variables, 17)
```
hsorby, line 102: This external variable also doesn't feature in the `compute_variables` function.

tests/resources/generator/cell_geometry_model/model.c
agarny: This is an algebraic model (as opposed to a differential model), so no `computeRates` function is needed here. As for the `computeVariables` function, it's empty because there are no variables to compute, but in theory there could be some in an algebraic model.

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.algebraic.c
agarny: The name of the file is mileading. `algebraic` relates to the external variable, i.e. we originally have an algebraic variable (`variables[5]`) which we, here, decided to flag as an external variable.

Regarding the `voi`, it's here because the model is indeed a differential model, independent of whether we have flagged some variables as being external.

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.external.py
agarny: `compute_variables` is for variables that are states/rates-dependent. So... yes, `variables[0]`and `variables[17]` ought to be in `compute_variables` too. Oversight at my end. Will fix it.

tests/resources/generator/cell_geometry_model/model.c
hsorby: So I guess my question was about clarity of the functions that are visible.  Are the functions based on the type of model that has been identified?  So if a model has been identified as algebraic then it will have methods such and such. And if it has been identified as differential it will have methods this and that.  This is an implicit definition correct?

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.algebraic.c
hsorby: okay, just checking that that was the case.

tests/resources/generator/cell_geometry_model/model.c
agarny: Correct. The functions are based on the type of model we are dealing with (i.e. algebraic or differential) AND on whether there are external variables.

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.external.py
agarny: Done in commit e133304b.

tests/resources/generator/hodgkin_huxley_squid_axon_model_1952/model.external.py
agarny: Done in commit e133304.

src/generator.cpp
```diff
@@ -2053,7 +2053,8 @@ void Generator::GeneratorImpl::addImplementationComputeVariablesMethodCode(std::
             if ((std::find(remainingEquations.begin(), remainingEquations.end(), equation) != remainingEquations.end())
                 || (((equation->type() == AnalyserEquation::Type::ALGEBRAIC)
                      || (equation->type() == AnalyserEquation::Type::EXTERNAL))
```
hsorby: Surely this statement can be dropped?

agarny: Agreed. Done in commit 7b822308.

hsorby:
I'm interested to know why the external variables get printed in reverse numerical order.  Everything else gets printed in numerical order (where possible).  In the HH1952 model for instance in the `compute_variables` method we get
```
variables[17] = external_variable(voi, states, variables, 17)
variable[10] = external_variable(voi, states, variables, 10)
```
Why isn't the order the other way around?

agarny:
> I'm interested to know why the external variables get printed in reverse numerical order. Everything else gets printed in numerical order (where possible). In the HH1952 model for instance in the `compute_variables` method we get
> 
> ```
> variables[17] = external_variable(voi, states, variables, 17)
> variable[10] = external_variable(voi, states, variables, 10)
> ```
> 
> Why isn't the order the other way around?

It's not in reverse numerical order. As you can tell, both `v10` and `v17` have been flagged as external variables, BUT... on top of that, `v10` has been flagged as being dependent on `v17`, hence the order in which they are being computed.

hsorby (approved):

nickerso (approved):
Looks good to me. I wonder (but haven't looked yet) if we have a test similar to `cellGeometryModelWithExternalVariables` but setting variables from a different model as external variables?

agarny:
> Looks good to me. I wonder (but haven't looked yet) if we have a test similar to `cellGeometryModelWithExternalVariables` but setting variables from a different model as external variables?

I am not aware of such a test and, indeed, I believe it would only make sense if we were to actually run a model with external variables, which we don't (at the moment). At the moment, we "just" generate code for models with external variables. So, we don't address how those external variables are "computed".

nickerso:
> I am not aware of such a test and, indeed, I believe it would only make sense if we were to actually run a model with external variables, which we don't (at the moment). At the moment, we "just" generate code for models with external variables. So, we don't address how those external variables are "computed".

was more just to check if the analyser correctly determines the variables are not in the model being analysed so have no impact on the generated code. (Separate issue, just mentioning it here while thinking about this PR)

PR #967, opened 2021-09-17
Update emscripten documentation to catch up with the current zlib build.

### LINKED ISSUE #966: Update documumentation on building Javascript bindings
Since the recent changes to zlib and libxml2 the documentation for the Javascript bindings has been out of date and needs updating.

hsorby:
Fixes #966.

### DIFF
--- a/docs/dev_building_web_assembly.rst
+++ b/docs/dev_building_web_assembly.rst
@@ -1,9 +1,9 @@
 
 ===========================================
-Building `libCellML` for use in the Browser
+Building *libCellML* for use in the Browser
 ===========================================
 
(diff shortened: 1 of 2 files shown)


agarny (approved):
All looks good and works fine for me on macOS. Only thing I would favour is the use of `ninja` rather than `make`. :) (Or, if `make`, then might want to suggest `make -j`, `make -j 4` or something.)

nickerso (approved):
This improves the docs in line with the recent updates for zlib and libxml2, so I think its fine to merge in. I couldn't get things to succeed completely under windows, but the given commands generally worked, so there might need to be some tweaks in future.

PR #978, opened 2021-11-22
CMake: support CMAKE_BUILD_TYPE

### LINKED ISSUE #972: CMake: support CMAKE_BUILD_TYPE
CMake comes with some options (`CMAKE_BUILD_TYPE` and `CMAKE_INSTALL_PREFIX`) for which we have our own version (`BUILD_TYPE` and `INSTALL_PREFIX`). `INSTALL_PREFIX` knows about `CMAKE_INSTALL_PREFIX`, but `BUILD_TYPE` doesn't. This means that if I pass `-DCMAKE_BUILD_TYPE=XXX` to CMake then libCellML will completely ignore it and will always build a debug version of libCellML.

agarny:
Fixes #972.


agarny:
@hsorby and @nickerso: if you agree with this PR then we might want to do something similar for `[CMAKE_]INSTALL_PREFIX` (i.e. only use the value if it is not empty).

hsorby (approved):

nickerso (approved):

Issue #969, opened 2021-10-13
Request:  allow import of CellML 1.0 and 1.1 models


luciansmith:
Once libCellML is stable, I would like to use it to replace my old CellML API code in Antimony to allow CellML translation to and from Antimony and SBML.

As such, it would be great if it could import CellML 1.0 and 1.1 models, as I'm not going to have the bandwidth to write an additional translator, especially one that ferrets out all the referenced models and translates them as well.  I could still use the library without it, but it would have severely reduced functionality, since all existing models are CellML 1.x

Thank you!

agarny:
Hi @luciansmith, libCellML has always been targeted at CellML 2.0, so I can't see it ever allowing the import of CellML 1.0/1.1 models. [libOpenCOR](https://github.com/opencor/libopencor), on the other hand, will allow such an import. I am just starting work on it though... even though this will be one of the first things that I will do, obviously.

luciansmith:
I don't want CellML 1.x objects; I just want to import and upconvert the document and get 2.0 objects.  The spec library is definitely the place to share the ability to convert documents.

luciansmith:
(The context for this is that David mentioned that there was an XSLT that upconverted 1.x to 2.0, and that since libcellml already links libxml2, it might be able to do this in the background if people needed it.  Which I do!)

agarny:
> (The context for this is that David mentioned that there was an XSLT that upconverted 1.x to 2.0, and that since libcellml already links libxml2, it might be able to do this in the background if people needed it. Which I do!)

Indeed, it would be very easy to do. @hsorby and @nickerso, views?

hsorby:
Yes, this is possible.  The problem we face is the XSLT is version 2 and the support from libraries is not great.  I have not looked at this recently with libXml2 but I think libXml2 doesn't fully support version 2 of XSLT.  I currently use Saxon to do the conversion which does work.  I will have to investigate this again as I cannot remember the exact details of what does and doesn't work.

luciansmith:
Which is probably yet another good argument to why I shouldn't be working on converting these documents myself ;-)

hsorby:
In the next release (tentatively 0.4.0) we will support importing CellML 1.0 and CellML 1.1 models.
This is possible by creating a `Parser` and setting the strict mode to false.
The simplest way to do this is create a `Parser` passing the strict flag directly to the object constructor.

For Python this would be done like so:
```python
import libcellml
p = libcellml.Parser(False)
```

[...]

PR #982, opened 2022-01-31
Parse 1.x CellML models transforming them into CellML 2.0

hsorby:
Addresses #969.


src/bindings/javascript/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ target_link_libraries(libcellml PUBLIC javascript-bindings)
 
 # Don't need this anymore (breaks linking)? -s USE_ES6_IMPORT_META=0
 set_target_properties(libcellml PROPERTIES
-  LINK_FLAGS "--bind -s ASSERTIONS=1 -s 'EXPORT_NAME=\"libCellML\"' -s EXPORT_ES6=1 -s MODULARIZE=1  -s ALLOW_MEMORY_GROWTH=1"
+  LINK_FLAGS "--bind -s ASSERTIONS=1 -s 'EXPORT_NAME=\"createLibCellML\"' -s EXPORT_ES6=1 -s ALLOW_MEMORY_GROWTH=1"
```
nickerso, line 54: is this really part of the parsing of 1.x models?

src/parser.cpp
```diff
@@ -57,23 +59,34 @@ class Parser::ParserImpl: public Logger::LoggerImpl
      * the @c std::string @p input. Any entities or attributes in @p model with names
      * matching those in @p input will be overwritten.
      *
+     * If @p parseVersion1XModels is @c true then if the @p input is a CellML 1.0 or 1.1
+     * model parse it into CellML 2.0 data structures. If @p parseVersion1XModels is @c false
+     * strictly only parse CellML 2.0 models.
+     *
```
nickerso: need to remove

src/parser.cpp
```diff
@@ -84,6 +97,7 @@ class Parser::ParserImpl: public Logger::LoggerImpl
      *
      * @param component The @c ComponentPtr to update.
      * @param node The @c XmlNodePtr to parse and update the @p component with.
+     * @param transforming Boolean indicating whether transforming to CellML 2.0 or not.
```
nickerso: remove

src/parser.cpp
```diff
@@ -96,6 +110,7 @@ class Parser::ParserImpl: public Logger::LoggerImpl
      *
      * @param model The @c ModelPtr to update.
      * @param node The @c XmlNodePtr to parse and update the model with.
+     * @param transforming Boolean indicating whether transforming to CellML 2.0 or not.
```
nickerso: remove

nickerso: and all other occurrences...

src/bindings/javascript/CMakeLists.txt
hsorby: No it isn't.

src/parser.cpp
hsorby: [968bc28](https://github.com/cellml/libcellml/pull/982/commits/968bc28c65fab15cc470b785507326c06d1694d1)

src/parser.cpp
hsorby: [c036462](https://github.com/cellml/libcellml/pull/982/commits/c036462b2ad97fefc97091e6305a0ed6ed0c049a)

src/parser.cpp
hsorby: [c036462](https://github.com/cellml/libcellml/pull/982/commits/c036462b2ad97fefc97091e6305a0ed6ed0c049a)

nickerso (dismissed):
I think this is good to go as is. I do wonder if in future we might want to look at adding more specific tests for the various CellML 1.x features that are largely encompassed in the large HH test (units in components, variables with no interfaces, etc.) but I suspect we'll add those tests if future debugging requires us to ;-)

nickerso (dismissed):

src/api/libcellml/parser.h
```diff
@@ -57,12 +57,39 @@ class LIBCELLML_EXPORT Parser: public Logger
...
+    /**
+     * @brief Create and populate a new model from a @c std::string.
+     *
+     * Creates and populates a new model pointer by parsing CellML
+     * entities and attributes from CellML 1.0 and 1.1 format @p input @c std::string.
+     *
+     * Returns a @c nullptr if the @p input is not a @c std::string representation
+     * of a CellML model.
+     *
+     * All existing issues will be removed before the input is parsed.
```
agarny: What is meant by this?

src/api/libcellml/parser.h
```diff
@@ -57,12 +57,39 @@ class LIBCELLML_EXPORT Parser: public Logger
...
+     *
+     * Set the @p renameNonSiUnits to @c true to rename common alternate spellings
+     * of SI units. Renaming non-SI units is @c false by default. The SI units
+     * that will get renamed if this parameter is true are:
+     *  * liter -> litre
+     *  * meter -> metre
```
agarny: Why do we have `renameNonSiUnits`? What would be the use case where I want to convert a CellML 1.x model to CellML 2.0 and yet don't want to convert non-SI unitts?

Also, why is `renameNonSiUnits` set to `false` by defaulut? If I want to parse a CellML 1.x model, I would by default want to have everything done to end up with a CellML 2.0 model, which means converting non-SI units to SI units.

What do we get if we don't rename non-SI units? An invalid CellML 2.0 model? If so, what would be the use of such a model?

src/api/libcellml/parser.h
```diff
@@ -57,12 +57,39 @@ class LIBCELLML_EXPORT Parser: public Logger
...
+     *
+     * @param input The string to parse into a model.
+     * @param renameNonSiUnits Rename SI units with common alternate spellings.
+     *
+     * @return The new @c ModelPtr deserialised from the input string.
+     */
+    ModelPtr parse1XModel(const std::string &input, bool renameNonSiUnits = false);
```
agarny: Why do we need a new API to parse a CellML 1.x model? Can we not simply use `ModelPtr parseModel(const std::string &input);` above? I.e. non `renameNonSiUnits` parameter and have `parseModel()` to detect for me whether I am trying to parse a CellML 1.x or a CellML 2.0. Anything else would mean that the API user (e.g., OpenCOR) would need to do such a detection while libCellML should really do that for me.

src/xmlattribute.cpp
```diff
@@ -114,4 +115,16 @@ void XmlAttribute::removeAttribute()
     xmlRemoveProp(mPimpl->mXmlAttributePtr);
 }
 
+void XmlAttribute::setNamespacePrefix(const std::string &prefix)
+{
+    std::array<xmlChar, 50> buffer = {};
+    xmlNodePtr parent = mPimpl->mXmlAttributePtr->parent;
+
+    xmlChar *fullelemname = xmlBuildQName(mPimpl->mXmlAttributePtr->name, reinterpret_cast<const xmlChar *>(prefix.c_str()), buffer.data(), static_cast<int>(buffer.size()));
```
agarny: `fullelemname` -> `fullElemName`

src/xmlattribute.cpp
```diff
@@ -114,4 +115,16 @@ void XmlAttribute::removeAttribute()
     xmlRemoveProp(mPimpl->mXmlAttributePtr);
 }
 
+void XmlAttribute::setNamespacePrefix(const std::string &prefix)
+{
+    std::array<xmlChar, 50> buffer = {};
```
agarny: Can we really not get away from that `50` constant?

src/xmlattribute.h
```diff
@@ -148,6 +148,18 @@ class XmlAttribute
      */
     void removeAttribute();
 
+    /**
+     * @brief Set the namespace prefix for this XmlAttribute.
+     *
+     * Set the namespace prefix for this XmlAttribute.
+     * The maximum length allowed for the namespace prefix and attribute name combined is 48 characters.
```
agarny: Not nice to have a limit on the length of the namespace. Also, 48 or 50 characters?

src/xmlnode.h
```diff
@@ -140,6 +140,22 @@ class XmlNode
      */
     bool isElement() const;
 
+    /**
+     * @brief Check if this @c XmlNode is an element node in the
+     * CellML 1.0 or 2.0 namespace with the given local name.
```
agarny: "CellML 1.0" -> "CellML 1.X" (same below)

src/xmlnode.h
```diff
@@ -154,7 +170,55 @@ class XmlNode
      * CellML 2.0 namespace with the given local name @p name; and
      * @c false otherwise.
      */
-    bool isCellmlElement(const char *name = nullptr) const;
+    bool isCellml20Element(const char *name = nullptr) const;
```
agarny, line 173: Minor: might be nice to have this method listed in the same order as in the `.cpp` file. I find it easier to follow the same "logic" in the `.h` and `.cpp` files.

src/xmlutils.cpp
```diff
@@ -79,6 +80,52 @@ XmlNamespaceMap traverseTreeForUndefinedNamespaces(const XmlNodePtr &node)
     return undefinedNamespaces;
 }
 
+std::vector<XmlAttributePtr> findAllAttributesWithOldCellmlNamespace(const XmlNodePtr &node)
```
agarny: Might want to rename this method to `findAllAttributesWith1XCellmlNamespace()` or even simply `attributesWith1XCellmlNamespace()`.

src/xmlutils.cpp
```diff
@@ -79,6 +80,52 @@ XmlNamespaceMap traverseTreeForUndefinedNamespaces(const XmlNodePtr &node)
...
+        // Append attributes found on child nodes.
+        attributes.insert(attributes.end(), subAttributes.begin(), subAttributes.end());
+
+        tempNode = tempNode->next();
+    }
+
+    return attributes;
+}
+
+void traverseTreeRemovingOldCellmlNamespaces(const XmlNodePtr &node)
```
agarny: Might want to rename this method to `traverseTreeRemovingCellml1XNamespaces()` or even simply `removeCellml1XNamespaces()`.

src/xmlutils.h
```diff
@@ -50,6 +50,28 @@ XmlNamespaceMap determineMissingNamespaces(const XmlNamespaceMap &namespaceMap1,
  */
 XmlNamespaceMap traverseTreeForUndefinedNamespaces(const XmlNodePtr &node);
 
+/**
+ * @brief Remove all the CellML 1.0 or CellML 1.1 namespaces from the given node and its children.
```
agarny, line 54: "CellML 1.0 or CellML 1.1" -> "CellML 1.X"? (same below)

src/parser.cpp
```diff
@@ -49,6 +49,8 @@ class Parser::ParserImpl: public Logger::LoggerImpl
 {
 public:
     Parser *mParser = nullptr;
+    bool mRenameNonSiUnits = false;
+    bool mTransformFrom1X = false;
```
agarny: "mTransformFrom1X" -> "mParseFromCellml1X"?

src/parser.cpp
```diff
@@ -63,17 +65,24 @@ class Parser::ParserImpl: public Logger::LoggerImpl
...
-     * in the @p model with names matching those in @p input
-     * will be overwritten.
+     * Takes a @c std::string and attempts to parse it into CellML 2.0 data structures.
+     * Returns @c nullptr if the @p input is not a @c std::string representation of
+     * a CellML model.
      *
-     * @param model The @c ModelPtr to update.
-     * @param input The string to parse and update the @p model with.
+     * The @p parseVersion1XModels if @c true will parse CellML 1.0 and CellML 1.1 elements
+     * into CellML 2.0 as best it can.
```
agarny: "...as best as it can."?

src/parser.cpp
```diff
@@ -241,23 +262,119 @@ ParserPtr Parser::create() noexcept
...
     return model;
 }
 
-void Parser::ParserImpl::updateModel(const ModelPtr &model, const std::string &input)
+/**
+ * @brief Test to determine if the attribute is an XML identifier.
+ *
+ * Test to determine if the attribute is an XML identifier. When
+ * transfoming from CellML 1.0 and CellML 1.1 it will also consider
+ * namespaced identifiers often used in CellML 1.X models.
```
agarny: Might want to mention that we are actually talking about the cmeta namespace since that's the only one we consider.

src/parser.cpp
```diff
@@ -241,23 +262,119 @@ ParserPtr Parser::create() noexcept
...
+std::string convertNonSiUnits(const std::string &unitsName)
+{
+    std::string convertedName = unitsName;
+    if (unitsName == "liter") {
+        convertedName = "litre";
+    } else if (unitsName == "meter") {
+        convertedName = "metre";
+    }
+
+    return convertedName;
```
agarny: Why not simply have:
```c++
if (unitsName == "liter") {
    return "litre";
}

if  (unitsName == "meter") {
    return "metre";
}

return unitsName;
```

src/parser.cpp
```diff
@@ -241,23 +262,119 @@ ParserPtr Parser::create() noexcept
...
+}
+
+/**
+ * @brief Test to determine if the @p node describes an encapsulation relationship.
+ *
+ * Test to determine if the @p node describes an encapsulation relationship.
+ *
+ * @param node The XML node to test.
+ *
+ * @return @c true if the @p node has child node with a relationship that is an encapsulation, @c false otherwise
```
agarny: "...has a child node..."
Missing full stop.

src/parser.cpp
```diff
@@ -297,16 +414,41 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
...
         return;
     }
+    if (mTransformFrom1X) {
+        if (!node->isCellml1XElement("model")) {
+            auto issue = Issue::IssueImpl::create();
+            issue->mPimpl->setDescription("Given model is not a CellML 1.0 or CellML 1.1 model, try parsing with parseModel().");
+            issue->mPimpl->mItem->mPimpl->setModel(model);
+            addIssue(issue);
+            return;
+        }
```
agarny: We really shouldn't have to do something like this.

src/parser.cpp
```diff
@@ -297,16 +414,41 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
...
+        auto issue = Issue::IssueImpl::create();
+        std::string version = "1.1";
+        if (node->isCellml10Element()) {
+            version = "1.0";
+        }
+        issue->mPimpl->setDescription("Given model is a CellML " + version + " model, the parser will try to represent this model in CellML 2.0.");
+        issue->mPimpl->setLevel(Issue::Level::MESSAGE);
+        issue->mPimpl->mItem->mPimpl->setModel(model);
+        addIssue(issue);
+    }
```
agarny: I am not sure this is needed. As an API user, I don't care about such a message. However, if the parsing fails because of a CellML 1.X related issue then I will want and need to know that.

src/parser.cpp
```diff
@@ -318,23 +460,26 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
     std::vector<XmlNodePtr> connectionNodes;
     std::vector<XmlNodePtr> encapsulationNodes;
     while (childNode) {
-        if (childNode->isCellmlElement("component")) {
+        if (childNode->isCellml20Element("component") || (mTransformFrom1X && childNode->isCellml1XElement("component"))) {
```
agarny: Why not simply use `isCellmlElement("component")`? It's not like, here, we need to distinguish between CellML 1.X and CellML 2.0 (same below; e.g., for `units`).

src/parser.cpp
```diff
@@ -318,23 +460,26 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
...
             ImportSourcePtr importSource = ImportSource::create();
             loadImport(importSource, model, childNode);
-        } else if (childNode->isCellmlElement("encapsulation")) {
+        } else if (childNode->isCellml20Element("encapsulation")) {
             // An encapsulation should not have attributes other than an 'id' attribute.
             if (childNode->firstAttribute()) {
                 XmlAttributePtr childAttribute = childNode->firstAttribute();
                 while (childAttribute) {
-                    if (childAttribute->isType("id")) {
+                    if (isIdAttribute(childAttribute, mTransformFrom1X)) {
```
agarny: I would use `false` rather than `mTransformFrom1X` since it's always going to be `false`.

src/parser.cpp
```diff
@@ -373,11 +518,22 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
                 issue->mPimpl->setReferenceRule(Issue::ReferenceRule::MODEL_CHILD);
                 addIssue(issue);
             }
+        } else if (mTransformFrom1X && childNode->isCellml1XElement("group")) {
```
agarny: If we are testing `mTransformFrom1X` for `true` whenever we make a call to `isCellml1XElement()` then shouldn't we test `mTransformFrom1X` for `false` whenever we make a call to `isCellml20Element()` (see line 509 for instance)?

src/parser.cpp
```diff
@@ -466,11 +645,31 @@ void Parser::ParserImpl::loadComponent(const ComponentPtr &component, const XmlN
         } else if (childNode->isComment()) {
             // Do nothing.
         } else {
-            auto issue = Issue::IssueImpl::create();
-            issue->mPimpl->setDescription("Component '" + component->name() + "' has an invalid child element '" + childNode->name() + "'.");
-            issue->mPimpl->mItem->mPimpl->setComponent(component);
-            issue->mPimpl->setReferenceRule(Issue::ReferenceRule::COMPONENT_CHILD);
-            addIssue(issue);
+            if (!(mTransformFrom1X && (childNode->name() == "units"))) {
```
agarny: I would rework this test. I would have something like:
```c++
} else if (childNode->isComment()) {
    // Do nothing.
} else if (mTransformFrom1X && (childNode->name() == "units")) {
    // Do nothing.
} else {
    auto issue = Issue::IssueImpl::create();
    if (mTransformFrom1X) {
        issue->mPimpl->setDescription("Component '" + component->name() + "' ignoring child element '" + childNode->name() + "'.");
        issue->mPimpl->setLevel(Issue::Level::MESSAGE);
    } else {
        issue->mPimpl->setDescription("Component '" + component->name() + "' has an invalid child element '" + childNode->name() + "'.");
    }
    issue->mPimpl->mItem->mPimpl->setComponent(component);
    issue->mPimpl->setReferenceRule(Issue::ReferenceRule::COMPONENT_CHILD);
    addIssue(issue);
}
```
It would make what we want to do clearer and, therefore, easier to maintain.

src/parser.cpp
```diff
@@ -1043,11 +1312,13 @@ void Parser::ParserImpl::loadEncapsulation(const ModelPtr &model, const XmlNodeP
         } else if (componentRefNode->isComment()) {
             // Do nothing.
         } else {
-            auto issue = Issue::IssueImpl::create();
-            issue->mPimpl->setDescription("Encapsulation in model '" + model->name() + "' has an invalid child element '" + componentRefNode->name() + "'.");
-            issue->mPimpl->mItem->mPimpl->setEncapsulation(model);
-            issue->mPimpl->setReferenceRule(Issue::ReferenceRule::ENCAPSULATION_CHILD);
-            addIssue(issue);
+            if (!(mTransformFrom1X && componentRefNode->isCellml1XElement("relationship_ref"))) {
```
agarny: See my comment for line 648.

tests/parser/cellml_1_0.cpp
```diff
@@ -0,0 +1,114 @@
...
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model xmlns=\"http://www.cellml.org/cellml/2.0#\" name=\"name\"/>\n";
+
+    libcellml::ParserPtr parser = libcellml::Parser::create();
+    libcellml::ModelPtr model = parser->parse1XModel(e);
+
+    EXPECT_EQ("", model->name());
+    EXPECT_EQ(size_t(1), parser->issueCount());
+    EXPECT_EQ(libcellml::Issue::Level::ERROR, parser->issue(0)->level());
+    EXPECT_EQ("Given model is not a CellML 1.0 or CellML 1.1 model, try parsing with parseModel().", parser->issue(0)->description());
```
agarny: We should never have to tell a library user to do this. The library should do it automatically for us.

tests/parser/cellml_1_0.cpp
```diff
@@ -0,0 +1,114 @@
...
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model xmlns=\"http://www.cellml.org/cellml/1.0#\"/>\n";
+
+    libcellml::ParserPtr parser = libcellml::Parser::create();
+    libcellml::ModelPtr model = parser->parse1XModel(e);
+    EXPECT_EQ(size_t(1), parser->issueCount());
+    EXPECT_EQ(libcellml::Issue::Level::MESSAGE, parser->issue(0)->level());
+    EXPECT_EQ("Given model is a CellML 1.0 model, the parser will try to represent this model in CellML 2.0.",
+              parser->issue(0)->description());
+}
```
agarny: Unneeded test if we are not to generatet that message (see line 33). Same in `tests/parser/cellml_1_1.cpp`.

tests/parser/cellml_1_1.cpp
```diff
@@ -0,0 +1,154 @@
...
+{
+    const std::string n = "name";
+    const std::string e =
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model xmlns=\"http://www.cellml.org/cellml/1.1#\" xmlns:cmeta=\"http://www.cellml.org/metadata/1.0#\" name=\"name\" cmeta:id=\"great_id\"/>\n";
+
+    libcellml::ParserPtr parser = libcellml::Parser::create();
+    libcellml::ModelPtr model = parser->parseModel(e);
+    EXPECT_EQ(n, model->name());
+}
```
agarny: Delete since exactly the same contents as `parseModelWithCMeta10Id`.

agarny (dismissed):
I would use the same API for parsing a CellML 1.X or 2.0 model. Our internal representation of a CellML model probably ought to keep track of the CellML version of a string that was parsed, if any, so that an end user (e.g., OpenCOR) can make use of it, if ever needed (e.g., someone loads a CellML 1.X model in OpenCOR, in which case OpenCOR would warn that the model has been converted to CellML 2.0 and that if it was to be edited then it would be saved back as a CellML 2.0 file).

src/parser.cpp
hsorby: Because I could have a CellML 1.0 component inside a CellML 2.0 model.  Contrived, sure but possible.

agarny: Contrived indeed, but then following the same contrived logic we should then have:
```c++
if ((!mTransformFrom1X && childNode->isCellml20Element("component"))
    || (mTransformFrom1X && childNode->isCellml1XElement("component"))) {
...
}
```
in case a CellML 1.X model has a CellML 2.0 component... :)

src/xmlattribute.h
hsorby: [e9d9938](https://github.com/cellml/libcellml/pull/982/commits/e9d9938a2e2d1404f7bfdffd38b207f225289fb5)

src/xmlattribute.cpp
hsorby: [911000d](https://github.com/cellml/libcellml/pull/982/commits/911000dab15d965d6dc4167870026a12e695d87e)

src/xmlattribute.cpp
hsorby: [b847544](https://github.com/cellml/libcellml/pull/982/commits/b847544e2fd56cf8025ad31c9238b0e9f495dc6f)

src/api/libcellml/parser.h
hsorby: [129d06e](https://github.com/cellml/libcellml/pull/982/commits/129d06eaed49c034195dcac5d388602d51445b10)

src/xmlnode.h
hsorby: [fd11515](https://github.com/cellml/libcellml/pull/982/commits/fd11515cfe3c4d3c7cdd0460a5fd5a9252dcd222)

src/xmlnode.h
hsorby: [afcf6b2](https://github.com/cellml/libcellml/pull/982/commits/afcf6b2a8096f558b2986257dd76853ed72bb830)

src/xmlutils.cpp
hsorby: [97d55b0](https://github.com/cellml/libcellml/pull/982/commits/97d55b0741f4cfb4bdd28ff91553b9a5afe28b15)

src/xmlutils.cpp
hsorby: [379ed90](https://github.com/cellml/libcellml/pull/982/commits/379ed90e82e53186654a5a6dc814eab026415716)

src/parser.cpp
hsorby: [9536438](https://github.com/cellml/libcellml/pull/982/commits/9536438ce48f8039671a255d28192645edaa4291)

src/parser.cpp
hsorby: [4abf42b](https://github.com/cellml/libcellml/pull/982/commits/4abf42b5ebcd82c0ff7726cccc1a5f5366a59bd0)

src/parser.cpp
hsorby: [e276ddf](https://github.com/cellml/libcellml/pull/982/commits/e276ddf7ae4adadfdd88c3bc47fd6b928a664973)

src/parser.cpp
hsorby: [ed7959a](https://github.com/cellml/libcellml/pull/982/commits/ed7959a69bf1c5bcacaeac2c1776932a84d4fb95)

src/parser.cpp
hsorby: [e26ad07](https://github.com/cellml/libcellml/pull/982/commits/e26ad07e10f77a012d1871991903cc08098820bf)
[8651e3a](https://github.com/cellml/libcellml/pull/982/commits/8651e3a0f5237fbe91e4e82786e2676aa4c9628d)

src/parser.cpp
hsorby: [ddb8153](https://github.com/cellml/libcellml/pull/982/commits/ddb81539339a1820ffbd57fa981f01afb72f34fa)

src/parser.cpp
hsorby: [ddb8153](https://github.com/cellml/libcellml/pull/982/commits/ddb81539339a1820ffbd57fa981f01afb72f34fa)

tests/parser/cellml_1_1.cpp
hsorby: [6662868](https://github.com/cellml/libcellml/pull/982/commits/6662868657288563ab1b5c0e8a6287dac0a85673)

src/parser.cpp
hsorby: [b7bcbb5](https://github.com/cellml/libcellml/pull/982/commits/b7bcbb5b85783422c98c9fa976ce93bbf5d7bc73)

src/parser.cpp
hsorby: [b7bcbb5](https://github.com/cellml/libcellml/pull/982/commits/b7bcbb5b85783422c98c9fa976ce93bbf5d7bc73)

src/parser.cpp
hsorby: [dcb9335](https://github.com/cellml/libcellml/pull/982/commits/dcb933538c32ea696fe8f1bf1a002b8c4d556ebc)

src/parser.cpp
hsorby: [dcb9335](https://github.com/cellml/libcellml/pull/982/commits/dcb933538c32ea696fe8f1bf1a002b8c4d556ebc)

src/xmlattribute.cpp
hsorby: [53c80f1](https://github.com/cellml/libcellml/pull/982/commits/53c80f16700e4de02cd90fa45965f5873b5c1957)

src/api/libcellml/parser.h
hsorby: I think we need new API for CellML 1.X models because we want to know that the user is asking for an interpretation of their version 1.X model in version 2.0 terms.
I feel that the detection burden is very minor and not onerous, and I don't agree that libCellML should do that for you.

src/api/libcellml/parser.h
hsorby: I think the `renameNonSiUnits` is a halfway measure to providing a `Helper` class that can perform fixes to CellML 2.0 models.  I would rather the parser just parsers what it has in front of it and let changes be made by an action from the user.

src/api/libcellml/parser.h
hsorby: That any issues currently in the parser will be removed when this method is used.

src/xmlutils.h
hsorby: I prefer it this way, because I don't remove any `CellML 1.X` namespaces.

tests/parser/cellml_1_0.cpp
hsorby: Possibly redundant depending on other decisions.

tests/parser/cellml_1_0.cpp
hsorby: Dependent on other decisions.

src/parser.cpp
hsorby: [ee3fdba](https://github.com/cellml/libcellml/pull/982/commits/ee3fdbacee4db3808d0df8839b23427248eb090a)

nickerso (dismissed):

src/api/libcellml/parser.h
agarny: Then, I am not liking it at all. libCellML handles CellML files while tools that use libCellML don't, and that's why they use libCellML. So, yes, I would expect libCellML to make their life as simple as possible when it comes to CellML, and it doesn't at this stage.

Tools that rely on libCellML should, using the same API, be able to parse any CellML contents, no matter whether it's CellML 1.X or 2.0. If libCellML parsed some CellML contents, then there should be a way to tell whether that contents was CellML 1.X or 2.0.

With the current API, those tools first need to detect whether they are dealing with a CellML 1.X or 2.0 file, and then call the appropriate API. To me, this is nonsense.

We are supposed to make libCellML's users' life easier...

src/parser.cpp
```diff
@@ -397,31 +397,33 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
...
             if (nodeNamespace.empty()) {
                 nodeNamespace = "null";
             }
-            issue->mPimpl->setDescription("Model element is in invalid namespace '" + nodeNamespace + "'. A valid CellML root node should be in namespace '" + CELLML_2_0_NS + "'.");
+            if (mParseFromCellml1X && node->isCellml20Element("model")) {
+                issue->mPimpl->setDescription("Given model is a CellML 2.0 model, use 'parseModel(...)' instead.");
+            } else if (!mParseFromCellml1X && node->isCellml1XElement("model")) {
+                issue->mPimpl->setDescription("Given model is a CellML 1.0 or CellML 1.1 model, use 'parse1XModel(...)' instead.");
+            } else if (mParseFromCellml1X) {
+                issue->mPimpl->setDescription("Given model is not a CellML 1.0 or CellML 1.1 model.");
```
agarny: What is that description supposed to be about?

src/parser.cpp
```diff
@@ -397,31 +397,33 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
...
+    if (!((!mParseFromCellml1X && node->isCellml20Element("model")) || (mParseFromCellml1X && node->isCellml1XElement("model")))) {
         auto issue = Issue::IssueImpl::create();
         if (node->name() == "model") {
             std::string nodeNamespace = node->namespaceUri();
             if (nodeNamespace.empty()) {
                 nodeNamespace = "null";
             }
-            issue->mPimpl->setDescription("Model element is in invalid namespace '" + nodeNamespace + "'. A valid CellML root node should be in namespace '" + CELLML_2_0_NS + "'.");
+            if (mParseFromCellml1X && node->isCellml20Element("model")) {
+                issue->mPimpl->setDescription("Given model is a CellML 2.0 model, use 'parseModel(...)' instead.");
```
agarny: An issue really ought to be about a CellML document, not about someone's use of libCellML (same for line 411).

agarny (commented):
Still not happy with the current API. Maybe something we can discuss in person at our next libCellML meeting.

src/api/libcellml/parser.h
nickerso: I'm starting to wonder if we actually want three different methods in the API depending on what the user wants to do.

Method 1 - parse a CellML 2 model: this is the current state and works well.

Method 2 - parse a CellML 1.x model and convert as much as possible to CellML 2: the user is knowingly asking libCellML to convert a CellML 1.x model into a CellML 2 model.

Method 3 - parse a CellML model: the user doesn't care what version CellML model they have and are prepared to do any follow up querying that might be needed to work out what is going on. This is really then just a convenience method that wraps the current `parseModel` with a fall-back to `parse1XModel`.

luciansmith: I can say that for myself, I would use Method 3 all the time.  I would assume that it would then create reasonable CellML objects, which could be queried and possibly written back out to whatever CellML's current default level/version was.

I wouldn't use method 1 much, because no CellML 2 models exist :-)

I would use method 2 if method 3 didn't exist, and just assume that any CellML model I came across was 1.1, because that's the most likely scenario.

But really, I agree with @agarny that I don't want to worry about CellML levels and versions.  I just want to read in a file that the user told me is CellML, convert it to objects that I can query, translate those objects to Antimony objects, and be done.

agarny: A "third" method might be the way to go. However, I would still have one method, but with a parameter specifying the "type" of inpuut, i.e. something like:
```c++
ModelPtr parseModel(const std::string &input, Type type);
```
where `Type` would be something like:
```c++
enum class Type
{
    AUTO,
    CELLML_1_X,
    CELLML_2_0
};
```
Also, in the case of `AUTO`, I wouldn't have it as a wrapper for the current `parseModel()` with a fall-back to `parse1XModel()`. Instead, I would have it as a proper parsing for CellML 1.X/2.0 models.

src/parser.cpp
hsorby: It is about passing an XML document that is neither CellML 1.0 or CellML 1.1 to the CellML 1X parser.

src/api/libcellml/parser.h
hsorby: You do need to start caring about such things when you are reading in a CellML 1.0 or CellML 1.1 document, you can't simply write out a CellML 2.0 document to the original location.  That would be disastrous.

src/parser.cpp
hsorby: The first part is about the CellML document? The user has attempted to parse a CellML 2.0 model with the CellML 1X parser. The second part is a suggestion on what should be used instead.

src/parser.cpp
agarny: I meant that the sentence doesn't seem to be complete? I would have expected something like `Given..., [use|do|whatever]...` like on line 411?

src/api/libcellml/parser.h
agarny: > You do need to start caring about such things when you are reading in a CellML 1.0 or CellML 1.1 document, you can't simply write out a CellML 2.0 document to the original location. That would be disastrous.

It's about to the libCellMLuser to decide what to do, not to libCellML. When it comes to OpenCOR, we want to read in any CellML file, be it CellML 1.X or CellML 2.0. When it comes to writing out a CellML file, it will depend on what we had to start with:
- New CellML file or CellML 2.0 file: we save the file as a CellML 2.0 file;
- CellML 1.X file: we warn the user that the file will be upgraded to CellML 2.0, and ask for confirmation to overwrite the current file or offer to save the file under a new name. libCellML has nothing to do with that.

src/parser.cpp
agarny: I wouuld expct any libCellML issue description to be directly reusable in a tool like OpenCOR, to help the user with what went wrong. Such a message wouldn't help an OpenCOR user at all.

nickerso: where does that expectation come from? is this something we are aiming to do with libCellML?

agarny: Which expectation?

nickerso: This one...

> I would expect any libCellML issue description to be directly reusable in a tool like OpenCOR

agarny: It does indeed come from how I would expect libCellML to behave when used by OpenCOR. Now, I appreciate that such an issue description may be useful in another use case. So, maybe it could still be generated, but if so then the issue should have a specific type, so that it can be easily discarded by those who are not interested in it.

hsorby:
This work has been revamped, it now uses a strict mode for the parser and importer with the allowance for setting both parser and importer into a permissive mode (e.g. `parser->setStrict(false)`).  In permissive mode the parser and importer will parse CellML 1.0 and 1.1 models.

src/api/libcellml/importer.h
```diff
@@ -48,6 +49,24 @@ class LIBCELLML_EXPORT Importer: public Logger
...
+     *   ImporterPtr importer = libcellml::Importer::create(false);
+     * @endcode
+     *
+     * @overload
+     *
+     * @param strict The boolean value to set.
+     *
+     * @return A smart pointer to a @c Importer object.
+     */
+    static ImporterPtr create(bool strict) noexcept;
```
agarny: Why not simply have `static ImporterPtr create(bool strict = false) noexcept;` rather than two `create()` methods?

src/api/libcellml/types.h
```diff
@@ -58,6 +58,8 @@ class Parser; /**< Forward declaration of Parser class. */
 using ParserPtr = std::shared_ptr<Parser>; /**< Type definition for shared parser pointer. */
 class Printer; /**< Forward declaration of Printer class. */
 using PrinterPtr = std::shared_ptr<Printer>; /**< Type definition for shared printer pointer. */
+class Strict; /**< Forward declaration of Parser class. */
+using StrictPtr = std::shared_ptr<Strict>; /**< Type definition for shared parser pointer. */
```
agarny: This is not needed.

Also, thinking of it, I would expect any class mentioned in `types.h`, and therefore available through the API, to be useable/relevant by/to libCellML users. This is not the case with `Strict`. So, I would scrap the `Strict` class and implement its (very limited) functionality directly in the `Parser` and `Importer` class.

src/bindings/interface/strict.i
```diff
@@ -0,0 +1,27 @@
...
+%import "types.i"
+
+%feature("docstring") libcellml::Strict
+"Abstract base class for entities that can be imported.";
+
+%feature("docstring") libcellml::Strict::isStrict
+"Returns an ImportSource if this entity is imported, else `None`.";
+
+%feature("docstring") libcellml::Strict::setStrict
+"Set the ImportSource for this entity (use `None` to unset).";
```
agarny: "Bad" copying/pasting job, besides the fact that I wouldn't have the `Strict` class.

src/api/libcellml/parser.h
```diff
@@ -49,6 +49,24 @@ class LIBCELLML_EXPORT Parser: public Logger
...
+     *   ParserPtr parser = libcellml::Parser::create(false);
+     * @endcode
+     *
+     * @overload
+     *
+     * @param strict The boolean value to set.
+     *
+     * @return A smart pointer to a @c Parser object.
+     */
+    static ParserPtr create(bool strict) noexcept;
```
agarny: Same as for the `Importer` class, I would have only one `create()` method.

src/api/libcellml/strict.h
```diff
@@ -0,0 +1,90 @@
+/*
```
agarny, line 1: I would really scrap that class. Overcomplicating things for no real benefit.

src/commonutils.cpp
```diff
@@ -27,7 +27,7 @@ libcellml::ModelPtr owningModel(const libcellml::ParentedEntityConstPtr &entity)
 {
     auto model = std::dynamic_pointer_cast<libcellml::Model>(entity->parent());
     auto component = owningComponent(entity);
-    while ((model == nullptr) && (component != nullptr) && component->parent()) {
+    while ((model == nullptr) && (component != nullptr)) {
```
agarny, line 30: What does it have to do with this PR? Was it breaking something all of a sudden?

src/parser.cpp
```diff
@@ -49,6 +48,8 @@ class Parser::ParserImpl: public Logger::LoggerImpl
 {
 public:
     Parser *mParser = nullptr;
+    bool mParsingOldVersion = false;
+    bool mParsing20Version = true;
```
agarny: Might want, for consistency, to rename `mParsingOldVersion` to `mParsing1XVersion` or, better, might want to make it more "future proof" by having something like:
```c++
enum class ParsingVersion
{
    CELLML_1_X,
    CELLML_2_0.
    UNKNOWN
};

ParsingVersion mParsingVersion = ParsingVersion.UNKNOWN;
```

src/parser.cpp
```diff
@@ -318,23 +460,26 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
...
     std::vector<XmlNodePtr> encapsulationNodes;
     while (childNode) {
-        if (childNode->isCellmlElement("component")) {
+        if (parseNode(childNode, "component")) {
             auto component = Component::create();
             loadComponent(component, childNode);
             model->addComponent(component);
-        } else if (childNode->isCellmlElement("units")) {
+            if (mParsingOldVersion && areUnitsDefinedInComponent(childNode)) {
+                loadUnitsFromComponent(model, childNode);
```
agarny: Would put the contents of `areUnitsDefinedInComponent()` in `loadUnitsFromComponent()`. It's not like `areUnitsDefinedInComponent()` is used more than once.

src/parser.cpp
```diff
@@ -216,6 +229,8 @@ class Parser::ParserImpl: public Logger::LoggerImpl
      * @param component The @c ComponentPtr the reset belongs to.
      */
     void checkResetChildMultiplicity(size_t count, const std::string &childType, const ResetPtr &reset, const ComponentPtr &component);
+
+    bool parseNode(const XmlNodePtr &node, const char *name);
```
agarny, line 245: Need documentation.

src/parser.cpp
```diff
@@ -239,25 +255,122 @@ ParserPtr Parser::create() noexcept
...
+ * XML identifiers in the CMETA 1.0 namespace.
+ *
+ * @param attribute The attribute to test.
+ * @param transforming @true if transfomring from CellML 1.0 or CellML 1.1, @c false otherwise.
+ *
+ * @return @c true if the given attribute is an XML identifier attribute, @c false otherwise.
+ */
+bool isIdAttribute(const XmlAttributePtr &attribute, bool transforming)
+{
+    return attribute->isType("id") || (attribute->isType("id", CMETA_1_0_NS) && transforming);
```
agarny: Slight optimisation:
```c++
return attribute->isType("id") || (transforming && attribute->isType("id", CMETA_1_0_NS));
```

src/parser.cpp
```diff
@@ -239,25 +255,122 @@ ParserPtr Parser::create() noexcept
...
+ * Test to determine if the attribute is an XML identifier. When
+ * transfoming from CellML 1.0 and CellML 1.1 it will also test for
+ * XML identifiers in the CMETA 1.0 namespace.
+ *
+ * @param attribute The attribute to test.
+ * @param transforming @true if transfomring from CellML 1.0 or CellML 1.1, @c false otherwise.
+ *
+ * @return @c true if the given attribute is an XML identifier attribute, @c false otherwise.
+ */
+bool isIdAttribute(const XmlAttributePtr &attribute, bool transforming)
```
agarny, line 307: Rename `transforming` to `cellml1X`? Or, better, make this method part of `ParserImpl` and use `mParsingVersion`.

src/parser.cpp
```diff
@@ -239,25 +255,122 @@ ParserPtr Parser::create() noexcept
...
+    return unitsName;
+}
+
+/**
+ * @brief Test to determine if a units element is present.
+ *
+ * Test to determine if a units element is a child of the given @p node.
+ *
+ * @param node The XML node to test the children of.
+ * @return @c true if units is a child of the node, @c false otherwise.
```
agarny: "... the node..." -> "... [the?] @p node..."

src/parser.cpp
```diff
@@ -239,25 +255,122 @@ ParserPtr Parser::create() noexcept
...
+ */
+bool areUnitsDefinedInComponent(const XmlNodePtr &node)
+{
+    XmlNodePtr childNode = node->firstChild();
+    while (childNode) {
```
agarny: `childNode != nullptr`

src/parser.cpp
```diff
@@ -239,25 +255,122 @@ ParserPtr Parser::create() noexcept
...
+ *
+ * @param node The XML node to test.
+ *
+ * @return @c true if the @p node has a child node with a relationship that is an encapsulation, @c false otherwise.
+ */
+bool isEncapsulationRelationship(const XmlNodePtr &node)
 {
-    loadModel(model, input);
+    XmlNodePtr childNode = node->firstChild();
+    while (childNode) {
```
agarny: `childNode != nullptr`

src/parser.cpp
```diff
@@ -239,25 +255,122 @@ ParserPtr Parser::create() noexcept
...
+        if (childNode->isCellml1XElement("relationship_ref")) {
+            XmlAttributePtr attribute = childNode->firstAttribute();
+            while (attribute) {
```
agarny: `attribute != nullptr`

src/parser.cpp
```diff
@@ -239,25 +255,122 @@ ParserPtr Parser::create() noexcept
...
+                if (attribute->isType("relationship") && attribute->value() == "encapsulation") {
```
agarny: `if (attribute->isType("relationship") && (attribute->value() == "encapsulation")) {`
i.e. use parentheses.

src/parser.cpp
```diff
@@ -281,34 +394,63 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
         addIssue(issue);
         return;
     }
-    if (!node->isCellmlElement("model")) {
+
+    mParsing20Version = node->isCellml20Element("model");
+    if ((mParser->isStrict() && !mParsing20Version) || (!node->isCellmlElement("model"))) {
```
agarny: No need for parentheses around `!node->isCellmlElement("model")`.

src/parser.cpp
```diff
@@ -239,25 +255,122 @@ ParserPtr Parser::create() noexcept
...
+}
+
+ModelPtr Parser::ParserImpl::parseModel(const std::string &input)
+{
+    removeAllIssues();
     ModelPtr model = nullptr;
     if (input.empty()) {
         auto issue = Issue::IssueImpl::create();
         issue->mPimpl->setDescription("Model string is empty.");
         issue->mPimpl->setReferenceRule(Issue::ReferenceRule::XML);
```
agarny: The reference rule should only be set if we are in strict parsing (to be consistent with the rest of the code).

src/parser.cpp
```diff
@@ -281,34 +394,63 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
...
         if (attribute->isType("name")) {
             model->setName(attribute->value());
-        } else if (attribute->isType("id")) {
+        } else if (isIdAttribute(attribute, mParsingOldVersion)) {
             model->setId(attribute->value());
         } else {
             auto issue = Issue::IssueImpl::create();
-            issue->mPimpl->setDescription("Model '" + node->attribute("name") + "' has an invalid attribute '" + attribute->name() + "'.");
+            if (mParsingOldVersion) {
+                issue->mPimpl->setDescription("Model '" + node->attribute("name") + "' ignoring attribute '" + attribute->name() + "'.");
```
agarny, line 454: "... ignored..." (or "... is ignoring..."?)

Same in other places.

src/parser.cpp
```diff
@@ -351,7 +496,7 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
             if (componentRefNode) {
```
agarny: `if (componentRefNode != nullptr) {`

src/parser.cpp
```diff
@@ -465,37 +644,62 @@ void Parser::ParserImpl::loadComponent(const ComponentPtr &component, const XmlN
...
             addIssue(issue);
         }
         childNode = childNode->next();
     }
 }
 
+void Parser::ParserImpl::loadUnitsFromComponent(const ModelPtr &model, const XmlNodePtr &node)
+{
+    XmlNodePtr childNode = node->firstChild();
+    while (childNode) {
```
agarny: `while (childNode != nullptr) {`

src/parser.cpp
```diff
@@ -688,23 +922,50 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr
     std::string component2Name;
     std::string mappingId;
     std::string connectionId;
-    XmlAttributePtr attribute = node->firstAttribute();
+
+    XmlNodePtr componentNode = nullptr;
+    if (mParsingOldVersion) {
+        XmlNodePtr childNode = node->firstChild();
+        while (childNode && componentNode == nullptr) {
```
agarny: `while ((childNode != nullptr) && (componentNode == nullptr)) {`

src/parser.cpp
```diff
@@ -688,23 +922,50 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr
...
+            issue->mPimpl->mItem->mPimpl->setModel(model);
+            addIssue(issue);
+            return;
+        }
+    } else {
+        componentNode = node;
+    }
+
+    XmlAttributePtr attribute = componentNode->firstAttribute();
     while (attribute) {
```
agarny: `while (attribute != nullptr) {`

src/parser.cpp
```diff
@@ -826,11 +1087,18 @@ void Parser::ParserImpl::loadConnection(const ModelPtr &model, const XmlNodePtr
         } else if (childNode->isComment()) {
             // Do nothing.
         } else {
-            auto issue = Issue::IssueImpl::create();
-            issue->mPimpl->setDescription("Connection in model '" + model->name() + "' has an invalid child element '" + childNode->name() + "'.");
-            issue->mPimpl->mItem->mPimpl->setModel(model);
-            issue->mPimpl->setReferenceRule(Issue::ReferenceRule::CONNECTION_CHILD);
-            addIssue(issue);
+            if (!(mParsingOldVersion && childNode->name() == "map_components")) {
```
agarny: `if (!(mParsingOldVersion && (childNode->name() == "map_components"))) {`
or better for me:
`if (!mParsingOldVersion || (childNode->name() != "map_components"))) {`

Also move that `if` statement in the `else` clause in line 1089.

tests/coverage/coverage.cpp
```diff
@@ -105,3 +105,57 @@ TEST(Coverage, unitsItem)
...
     EXPECT_FALSE(unitsItem->isValid());
 }
+
+TEST(Coverage, parserBranchesCellml10RelationshipRef)
+{
+    libcellml::ParserPtr parser = libcellml::Parser::create(false);
+    auto model = parser->parseModel(fileContents("coverage/cellml1x_relationshipref.xml"));
+
+    EXPECT_EQ(size_t(0), model->unitsCount());
+    EXPECT_EQ(size_t(1), model->componentCount());
```
agarny, line 115: I can see 4 components in that model... or am I missing something?

tests/coverage/coverage.cpp
```diff
@@ -105,3 +105,57 @@ TEST(Coverage, unitsItem)
...
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model name=\"bob\" xmlns=\"http://www.cellml.org/cellml/2.0#\"/>\n";
+
+    auto parser = libcellml::Parser::create();
+    auto model = parser->parseModel(in);
+
+    EXPECT_EQ(size_t(0), parser->issueCount());
+
+    EXPECT_EQ(nullptr, parser->error(3));
+    EXPECT_EQ(nullptr, parser->warning(4));
```
agarny, line 160: A bit of a stretch, but I would check the exact errors and warnings, in case our codebase changes and generate 3 errors and 4 warnings that are not necessarily all the same.

tests/importer/file_parser.cpp
```diff
@@ -61,11 +61,12 @@ TEST(Importer, resolveComplexImportsModelFromFile)
 
 TEST(Importer, resolveUnitsImportFromFile)
 {
-    libcellml::ParserPtr p = libcellml::Parser::create();
+    libcellml::ParserPtr p = libcellml::Parser::create(false);
     libcellml::ModelPtr model = p->parseModel(fileContents("import_units_model.cellml"));
     auto importer = libcellml::Importer::create();
 
     EXPECT_EQ(size_t(0), p->issueCount());
+    printIssues(p);
```
agarny: Remove.

tests/parser/parser.cpp
```diff
@@ -1507,6 +1535,11 @@ TEST(Parser, invalidModelWithTextInAllElements)
         "Unit referencing 'ball' in units 'robot' has an invalid non-whitespace child text element '\n      rolls\n    '.",
         "Component 'ship' has an invalid non-whitespace child text element '\n    falcon\n    '.",
         "Variable 'jedi' has an invalid non-whitespace child text element '\n      rey\n    '.",
+        "Reset has an invalid non-whitespace child text element '\n      boba fet\n      '. Either a test_value block or a reset_value block is expected.",
```
agarny, line 1538: Note to ourselves: we definitely to revisit our issue descriptions. We are anything but consistent. Here, for instance, we don't specify which reset we are talking about while we, for instance, do for the model element at line 1532. Conversely, we here specify what we would expect while we don't at line 1532.

tests/parser/parser.cpp
```diff
@@ -1880,6 +1913,112 @@ TEST(Parser, parseResetsWithIssues)
...
+    EXPECT_EQ("variable1", v1->name());
+
+    libcellml::VariablePtr v2 = r->testVariable();
+    EXPECT_EQ("variable2", v2->name());
+
+    std::string testValueString = r->testValue();
+    std::string t =
+        "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
+        "          some condition in mathml\n"
+        "        </math>\n";
```
agarny: Should be made `const` and should be moved at the top of the test. Same for `rt` at line 1967.

tests/parser/parser.cpp
```diff
@@ -1880,6 +1913,112 @@ TEST(Parser, parseResetsWithIssues)
...
+        "      <reset_value>\n"
+        "        <math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
+        "          some value in mathml\n"
+        "        </math>\n"
+        "      </reset_value>\n"
+        "      <initial_value>\n"
+        "        <math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
+        "          some value in mathml\n"
+        "        </math>\n"
+        "      </initial_value>\n"
```
agarny, line 1940: The test is called `parseResetIllegalChild` and I understand that `initial_value` is an illegal child of the `reset` element, yet several things are tested in this test except that illegal child...?

tests/parser/parser.cpp
```diff
@@ -1880,6 +1913,112 @@ TEST(Parser, parseResetsWithIssues)
...
+
+    std::string resetValueString = r->resetValue();
+    std::string rt =
+        "<math xmlns=\"http://www.w3.org/1998/Math/MathML\">\n"
+        "          some value in mathml\n"
+        "        </math>\n";
+    EXPECT_EQ(rt, resetValueString);
+}
+
+TEST(Parser, parseResetsWithOrderIssues)
```
agarny, line 1978: Not testing the orders as such?

tests/parser/parser.cpp
```diff
@@ -2279,3 +2418,30 @@ TEST(Parser, importComponentMadeConcrete)
     EXPECT_FALSE(model->component("childComponent")->isImport());
     EXPECT_FALSE(model->component("parentComponent")->isImport());
 }
+
+TEST(Parser, parseMixedCellMl1XAndCellMl20)
+{
+    auto parser = libcellml::Parser::create(false);
+    auto model = parser->parseModel(fileContents("mixed/sin_approximations_import_1X_and_20.xml"));
+
+    EXPECT_EQ(size_t(12), parser->issueCount());
```
agarny, line 2432: Shouldn't we explicitly test the issue descriptions to ensure that we always get what we would expect, not just a number of issues which might still be the same after working on our codebase, but not the issue descriptions. (Same remark for the two following tests.)

tests/parser/cellml_1_1.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+    const std::string e =
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model xmlns=\"http://www.cellml.org/cellml/1.1#\"/>\n";
+
+    libcellml::ParserPtr parser = libcellml::Parser::create(false);
+    libcellml::ModelPtr model = parser->parseModel(e);
+    EXPECT_EQ(size_t(1), parser->issueCount());
+    EXPECT_EQ(libcellml::Issue::Level::MESSAGE, parser->issue(0)->level());
+    EXPECT_EQ("Given model is a CellML 1.1 model, the parser will try to represent this model in CellML 2.0.", parser->issue(0)->description());
```
agarny: Should make the expected issue description a `const std::string`.

src/parser.cpp
```diff
@@ -281,34 +394,63 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
...
+    if ((mParser->isStrict() && !mParsing20Version) || (!node->isCellmlElement("model"))) {
         auto issue = Issue::IssueImpl::create();
         if (node->name() == "model") {
             std::string nodeNamespace = node->namespaceUri();
             if (nodeNamespace.empty()) {
                 nodeNamespace = "null";
             }
-            issue->mPimpl->setDescription("Model element is in invalid namespace '" + nodeNamespace + "'. A valid CellML root node should be in namespace '" + CELLML_2_0_NS + "'.");
+            if (mParser->isStrict() && node->isCellml1XElement("model")) {
+                issue->mPimpl->setDescription("Given model is a CellML 1.0 or CellML 1.1 model but strict parsing mode is on.");
```
agarny: "Given the model..."
Also, we could specify the exact version of the model (i.e. 1.0 or 1.1), as we do elsewhere in another issue description.

src/parser.cpp
```diff
@@ -281,34 +394,63 @@ void Parser::ParserImpl::loadModel(const ModelPtr &model, const std::string &inp
...
         return;
     }
+    mParsingOldVersion = node->isCellml1XElement("model");
+    if (mParsingOldVersion) {
+        auto issue = Issue::IssueImpl::create();
+        std::string version = "1.1";
+        if (node->isCellml10Element()) {
+            version = "1.0";
+        }
+        issue->mPimpl->setDescription("Given model is a CellML " + version + " model, the parser will try to represent this model in CellML 2.0.");
```
agarny: "Given the model..."

tests/importer/cellml_1_1.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include <libcellml>
+
+TEST(ImporterStrict, sineCellml11Import)
+{
+    libcellml::ParserPtr parser = libcellml::Parser::create(false);
+    libcellml::ImporterPtr importer = libcellml::Importer::create(false);
+    auto model = parser->parseModel(fileContents("cellml1X/sin_approximations_import.xml"));
+
+    EXPECT_EQ(size_t(3), parser->issueCount());
```
agarny, line 46: In this test, and the following one, we might want to test the issue descriptions from both the parser and the importer, especially from the importer since I would expect the issue descriptions to be different for the two tests, although they both have the same number of importer-related issues.

tests/parser/cellml_1_0.cpp
```diff
@@ -0,0 +1,145 @@
...
+*/
+#include "test_utils.h"
+
+#include "gtest/gtest.h"
+
+#include <libcellml>
+
+TEST(ParserTransform, emptyCellml10)
+{
+    const std::string e =
```
agarny: Very minor, but shouldn't in be `in` rather than `e` that we tend to use for an expected output? (Same in the other tests.)

tests/parser/cellml_1_0.cpp
```diff
@@ -0,0 +1,145 @@
...
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model xmlns=\"http://www.cellml.org/cellml/1.3#\" name=\"name\"/>\n";
+
+    libcellml::ParserPtr parser = libcellml::Parser::create();
+    libcellml::ModelPtr model = parser->parseModel(e);
+
+    EXPECT_EQ("", model->name());
+    EXPECT_EQ(size_t(1), parser->issueCount());
+    EXPECT_EQ(libcellml::Issue::Level::ERROR, parser->issue(0)->level());
+    EXPECT_EQ("Model element is in an invalid namespace 'http://www.cellml.org/cellml/1.3#'. A valid CellML root node should be in the namespace 'http://www.cellml.org/cellml/2.0#'.", parser->issue(0)->description());
```
agarny: Replace `A valid CellML root node...` with `A model element...`, to be consistent with the previous sentence (which might start with `A model element...`?).

tests/parser/cellml_1_0.cpp
```diff
@@ -0,0 +1,145 @@
...
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<model xmlns=\"http://www.cellml.org/cellml/1.1#\" name=\"name\"/>\n";
+
+    libcellml::ParserPtr parser = libcellml::Parser::create();
+    libcellml::ModelPtr model = parser->parseModel(e);
+
+    EXPECT_EQ("", model->name());
+    EXPECT_EQ(size_t(1), parser->issueCount());
+    EXPECT_EQ(libcellml::Issue::Level::ERROR, parser->issue(0)->level());
+    EXPECT_EQ("Given model is a CellML 1.0 or CellML 1.1 model but strict parsing mode is on.", parser->issue(0)->description());
```
agarny: As mentioned elsewhere, I would make the issue description clearer by, here, having `... a CellML 1.1 model...`

tests/parser/cellml_1_0.cpp
```diff
@@ -0,0 +1,145 @@
...
+    EXPECT_EQ(size_t(5), model->componentCount());
+    EXPECT_EQ(size_t(4), parser->issueCount());
+
+    auto validator = libcellml::Validator::create();
+    validator->validateModel(model);
+
+    EXPECT_EQ(size_t(0), validator->issueCount());
+}
+
+TEST(ParserTransform, annotatedModel)
```
agarny: `annotatedCellml1XModel` rather than `annotateModel` which is too generic?

tests/parser/cellml_1_0.cpp
```diff
@@ -0,0 +1,145 @@
...
+    auto model = parser->parseModel(fileContents("cellml1X/annotated_model.cellml"));
+
+    EXPECT_EQ(size_t(5), model->unitsCount());
+    EXPECT_EQ(size_t(2), model->componentCount());
+    EXPECT_EQ(size_t(77), parser->issueCount());
+
+    auto validator = libcellml::Validator::create();
+    validator->validateModel(model);
+
+    EXPECT_EQ(size_t(4), validator->issueCount());
```
agarny, line 123: Not clear what we are trying to test in this test...?

tests/parser/cellml_1_0.cpp
```diff
@@ -0,0 +1,145 @@
...
+    auto model = parser->parseModel(fileContents("cellml1X/non_si_units.cellml"));
+
+    EXPECT_EQ(size_t(11), model->unitsCount());
+    EXPECT_EQ(size_t(2), model->componentCount());
+
+    auto unitsMetre = model->units(0);
+    auto unitsLitre = model->units(1);
+
+    EXPECT_EQ("metre", unitsMetre->unitAttributeReference(0));
+    EXPECT_EQ("litre", unitsLitre->unitAttributeReference(0));
```
agarny, line 138: Make `"metre"` and `"litre"` constants (that would then get reused below)?

agarny (changes requested):
Happy with the direction this is going. Just a few suggestions.

src/api/libcellml/types.h
hsorby: I am not going to duplicate code that should be in a class, no matter how limited the functionality.

agarny: I can't argue with the fact that duplicated is and should always be a no-go.

However, as it stands, the `Strict` class is part of the API while it has nothing to do there. So, rather than have the `Parser` and `Importer` classes inherit from `Strict`, I would rather have `ParserImpl` and `ImporterImpl` inherit from it, and thus not have the `Strict` class part of the API (and mentioned in `src/api/libcellml/types.h`), which is just a nonsense to me.

src/commonutils.cpp
hsorby: It was breaking the Buildbot test for branch coverage.  Branch coverage has been set in addition to line coverage and it is set to be very strict.

src/parser.cpp
hsorby: Since we are following a TDD approach, future proofing is unnecessary.

tests/coverage/coverage.cpp
hsorby: Yes, component count is 1. Others are in the component hierarchy.

tests/coverage/coverage.cpp
hsorby: This is a coverage test, there aren't three errors or four warnings.  This is testing limit checking in issue retrieval.

tests/parser/parser.cpp
hsorby: It is testing the the issues with orders and the test looks at the expected behaviour for resets with order issues.

tests/parser/parser.cpp
hsorby: I don't think so, this test is looking at what happens when you mix CellML 1.0 and CellML 2.0 in the same document.  Making sure that we get errors and not a successfully parsed document.  This is an extremely unlikely scenario, but one we wan't to make sure we can deal with.

agarny: Ok

src/api/libcellml/importer.h
```diff
@@ -40,13 +41,23 @@ class LIBCELLML_EXPORT Importer: public Logger
      * @brief Create an @c Importer object.
      *
      * Factory method to create an @c Importer.  Create a
-     * blank importer with::
+     * blank importer with:
```
agarny: Replace "blank" with "strict"?

src/parser.cpp
agarny: Fair enough, but would still rename `mParsingOldVersion` to `mParsing1XVersion` then.

tests/coverage/coverage.cpp
agarny: Oops, my bad!

tests/parser/parser.cpp
```diff
@@ -1880,6 +1913,117 @@ TEST(Parser, parseResetsWithIssues)
     EXPECT_EQ(resetExpected, p->issue(0)->item()->reset());
 }
 
+TEST(Parser, parseResetIllegalChild)
+{
+    const std::string e = "Reset in component 'componentA' has an invalid child 'initial_value'.";
+
```
agarny: May want to remove that empty line.

agarny (dismissed):
A few minor comments, but overall happy with this PR now.

nickerso (dismissed):

src/api/libcellml/importer.h
hsorby: [15daf47](https://github.com/cellml/libcellml/pull/982/commits/15daf476fc1fe5f24f788570548baacde48565bf)

tests/parser/parser.cpp
hsorby: b8921b5

agarny (approved):
All good for me!

agarny:
Hmm, the GCov test failed. Is that a disk problem? I wanted to re-run that particular test, but it looks like I can't. @hsorby

hsorby:
I think so but it is happening quite regularly now.

nickerso (approved):

PR #989, opened 2022-06-07
Add JS bindings Generator test

### LINKED ISSUE #911: Generator JS bindings tests
Implement tests for Generator JS bindings class.

hsorby:
Closes #911.

### DIFF
--- a/tests/bindings/javascript/generator.test.js
+++ b/tests/bindings/javascript/generator.test.js
@@ -13,3 +13,78 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+
+const loadLibCellML = require('libcellml.js/libcellml.common')
+let libcellml = null
+
+const modelString = "<?xml version=\"1.0\" encoding=\"UTF-8\"?> \
+<model name=\"model\" xmlns=\"http://www.cellml.org/cellml/2.0#\" xmlns:cellml=\"http://www.cellml.org/cellml/2.0#\"> \
+    <units name=\"per_second\"> \
+        <unit exponent=\"-1\" units=\"second\"/> \
+    </units> \
+    <component name=\"component\"> \
+        <variable name=\"time\" units=\"second\"/> \
+        <variable name=\"x\" units=\"dimensionless\" initial_value=\"0\"/> \
+        <math xmlns=\"http://www.w3.org/1998/Math/MathML\"> \
+            <apply> \
+                <eq/> \
+                <apply> \
+                    <diff/> \
+                    <bvar> \
+                        <ci>time</ci> \
+                    </bvar> \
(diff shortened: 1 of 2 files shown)


nickerso (dismissed):

tests/bindings/javascript/generator.test.js
```diff
@@ -13,3 +13,78 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+        m = p.parseModel(modelString)
+        a = new libcellml.Analyser()
+
+        a.analyseModel(m)
+
+        g.setModel(a.model())
+
+        expect(g.interfaceCode().length).toBe(996)
+        expect(g.implementationCode().length).toBe(1058)
```
agarny, line 88: Shouldn't we just check that we don't have an empty string? I mean, to change the code generator even a tiny bit might break those tests for no good reason.

agarny (changes requested):

tests/bindings/javascript/generator.test.js
nickerso: isn't that the whole point of having tests?

agarny: For me, the JavaScript tests are here to confirm that what we can do from C++ we can also do from JavaScript, which here means to be able to retrieve some (non-empty) interface/implementation code. The "real" tests are to be done using the C++ implementation.

hsorby: 792040d

agarny (dismissed):

tests/bindings/javascript/generator.test.js
nickerso: surely that the generated code is accurately being put into JavaScript strings is important to test here? We're already less stringent than the C++ (and Python?) testing by not comparing the string directly.

Is the issue that "standard" developers may not have the build environment to test the JS bindings before making a PR and getting buildbot to run the tests to find if the string has changed length?

agarny: > Is the issue that "standard" developers may not have the build environment to test the JS bindings before making a PR and getting buildbot to run the tests to find if the string has changed length?

Indeed, libCellML being a C++ library, I would expect "standard" developers to be all set up for C++ development, but not necessarily for Python and/or JavaScript bindings. Now, if we *really* want the test to be "completel" then I would test against the actual expected code, not its length.

hsorby: Where are we with this?  Do we want the exact line count or not?  I take it from the comments that Andre is keen on the exact count and Alan isn't does that make me the deciding vote?

agarny: The number of lines doesn't mean anything to me, but I can't be bothered to argue it further, so whatever.

agarny (approved):

nickerso (approved):

PR #1008, opened 2022-07-18
Fixes to get the CD working again

### LINKED ISSUE #1007: Continuous deployment needs fixing
The current continuous deployment is not working and needs updating to get it working again.

hsorby:
Closes #1007.


README.rst
```diff
@@ -1,3 +1,7 @@
+.. image:: https://github.com/hsorby/l1bcellml/actions/workflows/deploy-on-release.yml/badge.svg?branch=release
+  :width: 200
+  :target: https://github.com/hsorby/l1bcellml/actions/workflows/deploy-on-release.yml
```
agarny: `cellml` rather than `hsorby`?

.github/workflows/deploy-on-release.yml
```diff
@@ -182,14 +179,14 @@ jobs:
             build_type: "Release", cc: "gcc", cxx: "g++", python_version: 3.8
           }
         - {
-            name: "macOS 10.15 Clang Py38", deploy_libs: false, deploy_python: true,
-            os: macos-10.15,
+            name: "macOS 12.4 Clang Py38", deploy_libs: false, deploy_python: true,
```
agarny: Replace `12.4` with `12` or `12.x` (same at line 188) in case there is a newer 12.x version?

.github/workflows/deploy-on-release.yml
```diff
@@ -164,16 +161,16 @@ jobs:
         - {
             name: "Windows Server 2019 MSVC Py38", deploy_libs: false, deploy_python: true,
             os: windows-2019,
-            libxml2_dir: '-D "LibXml2_DIR=../usr/local/lib/cmake/libxml2-2.9.10/"',
-            zlib_dir: '-D "ZLIB_DIR=../usr/local/lib/cmake/ZLIB/"',
+            libxml2_dir: '-D "LibXml2_DIR=../usr/local/libxml2-2.9.10/CMake/"',
+            zlib_dir: '-D "ZLIB_DIR=../usr/local/ZLIB-1.2.3/CMake/"',
```
agarny: Replace `..` with `$GITHUB_WORKSPACE` (same at lines 172-173) to be consistent (and bullet-proof) with what has been done at line 526 for instance?

agarny (commented):
This seems fine. Just a few minor comments.

Otherwise, regarding the generation of wheels, why not use `cibuildwheel`?

.github/workflows/deploy-on-release.yml
hsorby: I use 12.4 here because it is the version of the virtual machine from GitHub actions.

.github/workflows/deploy-on-release.yml
hsorby: I can't use environment variables here for the matrix strategy we are stuck with this, unless we want to get super fancy and create a dynamic matrix.

README.rst
agarny: I understand that this will eventually get changed, so resolving.

hsorby: When I look at the changes it shows the correct GitHub project location.

agarny: Oops, I hadn't bothered looking thinking it was one of those that might need to be done after merging in or something. Anyway, I can now see that it's ok. :)

cmake/environmentchecks.cmake
```diff
@@ -190,6 +198,12 @@ if(SWIG_EXECUTABLE)
   set(BINDINGS_AVAILABLE TRUE CACHE INTERNAL "Executable required to generate bindings is available.")
 endif()
 
+if(BINDINGS_AVAILABLE AND Python_Development.Module_FOUND)
+  set(PYTHON_BINDINGS_AVAILABLE TRUE CACHE INTERNAL "Requirements for creating Python bindings are available.")
+elseif(BINDINGS_AVAILABLE AND Python_Development_FOUND)
+  set(PYTHON_BINDINGS_AVAILABLE TRUE CACHE INTERNAL "Requirements for creating Python bindings are available.")
+endif()
```
agarny: Woudl rework this to have only one conditional statement..

src/bindings/python/setup.py
```diff
@@ -0,0 +1,29 @@
...
+setup(
+    name="libcellml",
+    version=tag[1:],
+    description="libCellML is a library for manipulating CellML models.",
+    author="libCellML contributors",
+    url="https://libcellml.org",
+    license="Apache 2.0",
+    packages=["libcellml"],
+    include_package_data=True,
+    #package_data={'libcellml': ['libcellml.0.2.0-dev.24.dylib', '_analyser.so', '_analyserequation.so', '_analyserequationast.so', '_analyserexternalvariable.so', '_analysermodel.so', '_analyservariable.so', '_annotator.so', '_component.so', '_componententity.so', '_entity.so', '_enums.so', '_generator.so', '_generatorprofile.so', '_importer.so', '_importsource.so', '_importedentity.so', '_issue.so', '_logger.so', '_model.so', '_namedentity.so', '_parentedentity.so', '_parser.so', '_printer.so', '_reset.so', '_types.so', '_units.so', '_validator.so', '_variable.so', '_version.so']},
```
agarny: Remove?

src/bindings/python/setup.py
```diff
@@ -0,0 +1,29 @@
...
+    cmake_source_dir="../../../",
+    cmake_install_target="install-wheel",
+    cmake_args=cmake_args,
+    #cmake_install_dir="libcellmlXXXXXX",
```
agarny: Remove?

agarny (changes requested):
Looks ok, just some minor suggestions.

agarny (dismissed):

nickerso (approved):
Seem to be a few random whitespace changes in the tests and a new test added for coverage purposes that are likely nothing to do with getting the CD working, but not worth pulling them into a separate PR at this point.

tests/model/model.cpp
```diff
@@ -465,7 +465,7 @@ TEST(Model, constructors)
 
     EXPECT_EQ(e, a);
 
-    //Testing copy constructor
+    // Testing copy constructor
```
agarny, line 468: I guess it's not critical for our tests, but in our codebase we would have a full stop at the end of a comment.

agarny (approved):
Just one comment that is not directly linked to this PR and which doesn't affect this PR, so approving as-is. (Might still want to address the comment at some point though.)

PR #1026, opened 2022-08-24
Fix segfault when printing complete MathML document

### LINKED ISSUE #1024: How to set MathML string to libcellml component?
Following on the discussion in issue #1017 I tried creating math equations using ```libsbml.parseL3Formula```. However, my program crashes on the last line rather than set the equation to libcellml component.

Snippet:
```comp_parameters = libcellml.Component()
    comp_parameters.setName("parameters")
    model.addComponent(comp_parameters)
    math_ast1: libsbml.ASTNode = libsbml.parseL3Formula( "eff / t_ave")
    var_math_ast1 = libsbml.writeMathMLToString(math_ast1)
    print(var_math_ast1)
    comp_parameters.setMath(var_math_ast1)
```
Output:
```<?xml version="1.0" encoding="UTF-8"?>
  <apply>
    <divide/>
    <ci> eff </ci>
    <ci> t_ave </ci>
  </apply>
</math>
```

hsorby:
Fixes #1024.


hsorby:
It appears some OSes give you a segfault and others will return an empty string, both outcomes are undesirable.

agarny (dismissed):

nickerso:
So this PR solves this issue by picking up on any errors when using libXml2 to parse the generated model to "pretty print" the serialised model - which is a good addition. But since we already have an internal `printMath` function in the printer to process the math strings I wonder if we should be using that to remove XML declarations? Seems like it might be a fairly common use-case to use an external library like libSBML to generate MathML to put into components and those fragments may be complete self-contained XML documents.

But perhaps this is really a future addition to that "model fixer" tool that we have previously discussed...

hsorby:
I have re-jigged the logging of errors to include the location (reset, or component) where the MathML error was detected.  I think this makes more sense when the MathML is subsequently not printed but the rest of the model is.

tests/math/math.cpp
```diff
@@ -20,6 +20,20 @@ limitations under the License.
 
 #include <libcellml>
 
+// Depending on the libXML2 version the error messages differ.
+const std::vector<std::string> expectedIssues_2_9_4 = {
```
agarny, line 25: Is 2.9.4 (and 2.9.10 below) related to the version of libxml2 used? If so, it might be worth making it clear or we might, in the future, wonder what it is about?

agarny (dismissed):

tests/math/math.cpp
agarny: I had completely overlooked the original message that mentioned `libxml2` (and not `libXML2` or even `LibXml2`). Still, I was replacing `expectedIssues_2_9_4` with something like `expectedIssues_libxml2_2_9_4`. Anyway, I guess the comment is sufficient. So, resolving this issue.

agarny (dismissed):

src/printer.cpp
```diff
@@ -127,12 +135,32 @@ std::string printConnections(const ComponentMap &componentMap, const VariableMap
...
 }
 
-std::string printMath(const std::string &math)
+std::string Printer::PrinterImpl::printMath(const std::string &math)
 {
     static const std::regex before(">[\\s\n\t]*");
     static const std::regex after("[\\s\n\t]*<");
-    auto temp = std::regex_replace(math, before, ">");
-    return std::regex_replace(temp, after, "<");
+    static const std::string xmlDeclaration = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>";
```
nickerso: are we restricted to XML 1.0 and UTF-8 encoding? I don't think we are...and if we are going to do a string match replacement then probably better to do a regex replace to handle XML 1.1 and other encodings (or no encoding) and the possible `standalone` attribute.

agarny: `xmlDeclaration` is used after we have done a pretty print, which seems to be using UTF-8 (see `XmlDoc::prettyPrint()`), so it looks like it might be fine?

hsorby: This is the relevant code for the XML declaration:
```
	/*
	 * Save the XML declaration
	 */
	if ((ctxt->options & XML_SAVE_NO_DECL) == 0) {
	    xmlOutputBufferWrite(buf, 14, "<?xml version=");
	    if (cur->version != NULL) 
		xmlBufferWriteQuotedString(buf->buffer, cur->version);
	    else
		xmlOutputBufferWrite(buf, 5, "\"1.0\"");
	    if (encoding != NULL) {
		xmlOutputBufferWrite(buf, 10, " encoding=");
		xmlBufferWriteQuotedString(buf->buffer, (xmlChar *) encoding);
	    }
	    switch (cur->standalone) {
		case 0:
		    xmlOutputBufferWrite(buf, 16, " standalone=\"no\"");
		    break;
		case 1:
		    xmlOutputBufferWrite(buf, 17, " standalone=\"yes\"");
		    break;
	    }
	    xmlOutputBufferWrite(buf, 3, "?>\n");
	}
```
This has to be the first thing you see in the text so we could add a regex around this.

nickerso: wait?! the `prettyPrint` adds the XML declaration in? so can't we just expose the `XML_SAVE_NO_DECL` option and not manipulate the string afterwards?

hsorby: I can't seem to find a way to set the `XML_SAVE_NO_DECL` option, if you can tell me how I will do it.

nickerso: can't see an obvious way using the `xmlDocDumpFormatMemoryEnc` approach....would need to switch to the `xmlsave` approach. regex it is :)

nickerso (approved):

agarny (approved):

PR #1028, opened 2022-09-10
Fixing the manylinux tag name

### LINKED ISSUE #1027: cellml pip does not install on linux any more
Hi all,
unfortunately you seem to have broken the pypi package for linux.

I.e.
```
pip install libcellml
```
only installs a very old version.
```
Successfully installed libcellml-0.2.0.dev21
```
but not the latest version on python 3.9 on linux (ubuntu), i.e., https://pypi.org/project/libcellml/

Installation on python 3.10 does not work at all in linux (ubuntu).

Please fix this issues, because the library is not usable at the moment on linux without the pypi package.

Best Matthias

hsorby:
Fixing the manylinux tag name so the Linux wheels are accepted by pypi.

Fixes #1027.


hsorby:
Closing this PR because of the mixing of the mathml-segfault branch.

PR #1040, opened 2022-09-15
Shift to use dynamic lookup for undefined symbols in shared libraries

### LINKED ISSUE #1039: Fix Linux, macOS wheels
The Linux and macOS wheels that we are currently creating are not correct. We need to use `-undefined dynamic_lookup` to link the Python library. This allows undefined symbols to be resolved at run time and not link time, this is crucial for distributing C extension binaries.  With this in place, the undefined symbols will be found in the environment that the process/application is running in.  Well that's the theory, now for the practical part.

hsorby:
When linking to Python we will lookup symbols at runtime and not link time. 
All other library symbols are resolved at link time.

Fixes #1039.


nickerso (approved):

agarny (approved):
Man, those Python bindings with shared libraries (rather than one "big" static library) really seem to be causing all sorts of problems. I really don't understand why we do it that way, but so be it. (That's exactly the kind of thing I wanted to avoid with libOpenCOR's Python bindings.)

PR #1057, opened 2022-11-25
Modify code generation to start Python enums from zero

### LINKED ISSUE #991: Python code generation: Initialise enums from zero
For code generation of the Python profile enumerations of the VariableType are initialised to start at `1`.  However, the C++ equivalent is initialised at `0`.  I think this should be made consistent across the code generation profiles.

This means initialising our Python generated enumeration code to start from `0` and not `1`.

hsorby:
Fixes #991.


hsorby:
Whilst this is slightly un-pythonic it is very useful for checking the tutorial code we provide via the website because it makes the output of both the C++ code generation and Python code generation exactly the same.

nickerso (approved):
Its not clear from #991 that this is something that is needed, but assuming that it is then this change looks ok.

src/generatorprofile.cpp
```diff
@@ -853,7 +853,7 @@ void GeneratorProfile::GeneratorProfileImpl::loadProfile(GeneratorProfile::Profi
                                       "\n";
 
         mInterfaceVersionString = "";
-        mImplementationVersionString = "__version__ = \"0.3.0\"\n";
+        mImplementationVersionString = "__version__ = \"0.3.1\"\n";
```
agarny, line 856: Do we really not want to keep the version numbers to be the same for both C and Python?! I would expect that version to be the version of the generator, not the C or Python generator *per se*.

agarny (commented):
My problem with this PR and the issue it fixes is that it is about making our life, that of developers, easier while we should aim at making our users' life easier.

So, when it comes to generated Python code, I would expect that code to look as native as possible to a Python developer, which means starting enums from one, not zero. After all, this is exactly what `auto()` does in Python (see https://github.com/cellml/libcellml/issues/991#issuecomment-1159865164). So, if anything, I would use `auto()` rather than "hard-coded" values.

src/generatorprofile.cpp
hsorby: I wouldn't agree, they have different version numbers because they can be changed independently of one another.  It is quite likely that they will be synchronised of course but for this sort of change the version number accurately describes the progression of the generated code.

hsorby:
The whole point to the Python enums is to not use a value but the attribute. In 99% of cases you are not concerned with the actual value, simply the attribute.  Our users are not going to notice the value of our enumerations in the code generation.

src/generatorprofile.cpp
agarny: Ok, can't be bothered to argue.

agarny (approved):
I don't like it at all, but I am getting bored.

PR #1058, opened 2022-11-28
Debug utilities library installation

### LINKED ISSUE #980: Windows debug installation misses cellml_debug_utilities.lib
If you build a debug version of libCellML on Windows and install it then `cellml_debug_utilities.lib` doesn't get installed. Not sure whether it should be installed, but if I try to link libOpenCOR against an installed (static) copy of the debug version of libCellML then it complains about `cellml_debug_utilities.lib` not being found. If `cellml_debug_utilities.lib` is needed for the debug version of libCellML then I wouldn't in fact have it installed. I would expect it to be statically linked to `libcellmld.lib`. Whatever the case, I cannot currently use an installed debug version of libCellML while I can use an installed release version without any problems.

hsorby:
Fixes #980.

I have also added a test to try and detect if the debug utilities is being used by the libCellML library. It of course shouldn't be, it is only to be used for debugging.


tests/CMakeLists.txt
```diff
@@ -110,6 +110,9 @@ foreach(TEST ${LIBCELLML_TESTS})
   unset(CURRENT_CATEGORY)
 endforeach()
 
+set(ALL_LIBRARY_SOURCE_FILES ${SOURCE_FILES} ${API_HEADER_FILES} ${MODULE_HEADER_FILES} ${GENERATED_HEADER_FILES} ${HEADER_FILES})
+add_test(NAME are_debug_utilities_not_being_used COMMAND ${CMAKE_COMMAND} -P ${CMAKE_SOURCE_DIR}/cmake/checkfordebugheaderusage.cmake ${ALL_LIBRARY_SOURCE_FILES})
```
agarny: Might just want to rename that test to `are_debug_utilities_used`. It's not like that we are testing the return value of something.

agarny (changes requested):
Just a minor change otherwise all good.

agarny (approved):

nickerso (approved):

PR #1063, opened 2022-11-29
Keep Python generated wrappers at SWIG version 4.0.1

### LINKED ISSUE #1062: SWIG 4.1.0 has changed the way static methods are wrapped
The gihub actions containers we depend on have updated to SWIG 4.1.0. Ubuntu, one of our primary target platforms does not yet have a standard package for SWIG 4.1.0. 

In SWIG 4.0.2 static methods of a class had an extra function created as an alternative?? for the class static method.  SWIG 4.1.0 has removed these extra functions.  This is a problem for us as we are requiring that we have 100% coverage for the Python API.  Now that the Python API is different depending on which version of SWIG has wrapped the library, we will also need to have an adaptive test based on the version of SWIG the library was wrapped with.

The alternative is to require SWIG 4.1.0 as the minimum version of SWIG.

hsorby:
Fixes #1062.


agarny:
I haven't checked in details, but it looks like this PR is aiming at supporting old versions of SWIG. I really don't see the point. Let's be honest, we are likely going to be the only ones ever to build libCellML's Python bindings. So, let's try to make our life as easy as possible rather than unnecessarily complicated. IOW, let's just make the minimum version of SWIG be 4.1.0.

hsorby:
I agree, but the sticking point here is Ubuntu. Ubuntu doesn't have SWIG 4.1.0 available through the package manager and I prefer not to make people install packages that aren't available over adjusting for small changes like this one in the codebase.

agarny:
People? Which people? Me and you, most likely? As for Ubuntu not having that package, we can simply build the version of SWIG we want and use that. After all, this is what we do with libxml2 on Windows, no? So, why would we make our life unnecessarily difficult with SWIG, knowing that people interested in the Python bindings will just wait for us to push things on PyPI rather than going through the trouble of building things themselves.

agarny (approved):
Nice, thanks!

nickerso (approved):

PR #1069, opened 2022-12-04
Tests: added support for libXml2 2.9.11+

### LINKED ISSUE #1068: Tests: add support for libXml2 2.9.11+
Our tests currently work with libXml2 2.9.10 and below, but... with libXml2 2.9.11, [error messages have once again changed](https://gitlab.gnome.org/GNOME/libxml2/-/blob/e1bcffea180d6cc0651757bb64284a763e0e2239/parser.c#L9940-L9962), which means that some of our tests fail if we are using a version of libXml2 which is greater than 2.9.10 (the macOS 13.0 SDK uses libXml2 2.9.13). The current version of libXml2 is version 2.10.3, but thank god their `xmlParseContent()` method is still the same. So, it looks like we might, for now at least, be good with just supporting 2.9.11.

agarny:
Fixes #1068.

### DIFF
--- a/tests/math/math.cpp
+++ b/tests/math/math.cpp
@@ -37,6 +37,14 @@ const std::vector<std::string> expectedIssues_2_9_10 = {
     "LibXml2 error: EndTag: '</' not found.",
 };
 
+// Version 2.9.11 of LibXml2 reports the following errors,
+// used on macOS 13 machines.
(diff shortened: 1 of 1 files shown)


hsorby (approved):
May have to look into libXml2 version aware answers if the output keeps changing.

nickerso (approved):

Issue #1074, opened 2022-12-12
Analyser: rework the analysis of a model with external variables


agarny:
Right now, when we analyse a model, we first determine the type of each variable and then check whether there are variables that should be marked as external which. This works fine even though a `VARIABLE_BASED_CONSTANT` equation may have to be requalified as an `ALGEBRAIC` equation in case it (in)directly depends on an external variable.

Our current approach works fine, but it might become tedious once we have variables that need to be computed using a non-linear algebraic solver (e.g., KINSOL; see issue #882).

So, rather than handling external variables *after* determining the type of each variable, we probably ought to take them into account when actually determing the type of each variable. To do so will prevent us from having to requalify some equations, something that requires us to think about all possible cases where an equation might need to be requalified.

nickerso:
Would it work if users were allowed to specify the type of the external variable? i.e., its different when its a constant computed outside the model versus turning a parameter into a state variable...

agarny:
> Would it work if users were allowed to specify the type of the external variable? i.e., its different when its a constant computed outside the model versus turning a parameter into a state variable...

Right now, external variables are just that: external. As far as the analyser (and generator) is concerned, it just knows that a variable is computed "outside", but has no idea whether it's actually a constant, an algebraic, a state, etc.

What you are suggesting goes quite a bit beyond what I originally had in mind when it comes to external variables. I can see cases where it could be useful to know the exact type of an external variable. For instance, if we knew that it was a constant then code generation could be optimised. However, if the user was to specify that an external variable is actually a state variable, there is no guarantee that it would rely on the same variable of integration. Also, what if our model is a pure algebraic model and an external variable is a state? What about the case where an external variable is "computed" using a lookup table containing some experimental data (e.g., to simulate an action potential clamp protocol)? What would be the type of that external variable?

Yeah, the more I think about typing an external variable, the more questions I have. Still, happy for us to discuss it at one of our next libCellML meetings.

Issue #882, opened 2021-06-29
Analyser/Generator: support systems of non-linear algebraic equations


agarny:
We want to support systems of non-linear algebraic equations. This includes models where we have an equation with only one unknown, with the unknown not being on its own on either side of the equation. For example:
```
a + b = c + d
```
where `b`, `c`, and `d` are known and `a` is unknown. Such an equation is equivalent to:
```
a = c + d - b
```
but we don't want (for now, at least) to rearrange such an equation since it can quickly become very tricky (e.g., we could have `a^2 + b = c + d`, which would result in two solutions). So, instead, such an equation should be solved using a non-linear algebraic solver like [KINSOL](https://computing.llnl.gov/projects/sundials/kinsol).

Other systems of non-linear algebraic equations include the classical systems of `n+` equations with `n` unknowns with `n > 1`. For example:
```
2x + y - 2z = -1
3x - 3y - z = 5
x - 2y + 3z = 6
```

Things to do:
 - [ ] Have the analyser support systems of non-linear algebraic equations;
 - [ ] Have the generator support systems of non-linear algebraic equations; and
 - [ ] Make it possible to mark an algebraic variable (computed through a system of NLA equations) as an external variable.

PR #1077, opened 2022-12-21
Analyser: reworked the analysis of a model with external variables

agarny:
Fixes issue #1074.


agarny:
I will likely have another quick look at it tomorrow morning, but I believe this is ready to be reviewed.

This PR:
1. improves the handling of external variables in the analyser, as well as
2. moves that handling as early as possible in the analysis of a model.

The second point is really the reason behind this PR since right now an equation of type `VARIABLE_BASED_CONSTANT ` has to be requalified as an `ALGEBRAIC ` equation if it relies on external variables (see [here](https://github.com/cellml/libcellml/blob/c8e694f39e07cff554ecd9d58f997ca0cda94aab/src/analyser.cpp#L2590-L2609)).

With this PR, by handling external variables as early as possible, we do NOT need to do any requalification anymore (see [here](https://github.com/agarny/libcellml/blob/issue1074/src/analyser.cpp#L2583-L2584)). This is going to be particularly useful when addressing issue #882 since external variables could otherwise make things rather tricky.

agarny:
Ok, just a minor cleaning up of the code and various dummy commits that eventually resulted in the CI passing as its should have in the first place (what a waste of time!). So, now, all good to review. @hsorby and @nickerso.

luciansmith:
When nothing actually needs to be changed to get the CI to pass, you should be able to just click the 're-run failed tests' button; you don't need a whole dummy commit.  Just FYI; I went through that phase myself before figuring it out.

agarny:
> When nothing actually needs to be changed to get the CI to pass, you should be able to just click the 're-run failed tests' button; you don't need a whole dummy commit. Just FYI; I went through that phase myself before figuring it out.

Agreed when it comes to GitHub Actions. I have done it on occasions, but we are not using GH Actions for our CI. We use Buildbot and, yesterday, I tried to rerun one of the tests and it passed, but its updated status wasn't reflected in GitHub. So, I had to create a dummy commit and hope for the best. Anyway, will see what @hsorby has to say about it. All I know is that it's very frustrating and a waste of people's time.

agarny:
@hsorby and @nickerso: would be nice to have this one reviewed (and then PR #1080).

nickerso (approved):
Assuming CI gets back to showing this as passing, then it looks good to me. I did wonder if there needed to be an update to the docs to go along with this change, but then I found: https://libcellml.org/documentation/v0.4.0/user/asides/analyser#understanding-analyser, so that will be a future issue ;-)

agarny:
> I did wonder if there needed to be an update to the docs to go along with this change, but then I found: https://libcellml.org/documentation/v0.4.0/user/asides/analyser#understanding-analyser, so that will be a future issue ;-)

🤪

hsorby (approved):

PR #1083, opened 2023-01-17
Improve MathML validation

### LINKED ISSUE #1082: Improve MathML validation
We use the MathML DTD to validate a MathML string (see the call to `XmlDoc::parseMathML()` in `Validator::ValidatorImpl::validateMath()`). However, this is not sufficient to ensure that this is valid MathML. Consider the following MathML string:
```xml
    <apply>
        <eq/>
        <ci>y</ci>
        <apply>
            <root/>
            <ci>a</ci>
            <ci>b</ci>
            <ci>c</ci>
        </apply>
    </apply>
</math>
```
According to the MahML DTD (and https://www.mathmlcentral.com/Tools/ValidateMathML.jsp), it's valid, but... a `root` element must have either one or two siblings. If there are two siblings then the first sibling must be a `degree` element with one child corresponding to the degree of the root. The second sibling is what needs to be rooted. If there is only one sibling then it means that we have a square root. In other words:
```xml
    <apply>
        <eq/>
        <ci>y</ci>
        <apply>
            <root/>
            <ci>a</ci>
        </apply>
    </apply>
</math>
```
and:
```xml
    <apply>
        <eq/>
        <ci>y</ci>
        <apply>
            <root/>
            <degree>
                <ci>b</ci>
            </degree>
            <ci>a</ci>
        </apply>
    </apply>
</math>
```
are *really* valid.

So, upon successful DTD "validation", we should confirm that it is indeed *really* valiid by doing some extra checks. When it comes to the `root` element, we should check that it has one or two siblings, etc.

This means that once a MathML string has been *really* validated, something like the analyser can do whatever it wants with it wthout having to confirm whether a `root` element is properly used, etc.

agarny:
Fixes #1082.


src/issue.cpp
```diff
@@ -144,6 +144,7 @@ static const std::map<Issue::ReferenceRule, std::vector<std::string>> ruleToInfo
     {Issue::ReferenceRule::RESET_RESET_VALUE, {"RESET_RESET_VALUE", "2.11", baseSpecificationUrl, "specB11"}},
     {Issue::ReferenceRule::MATH_MATHML, {"MATH_MATHML", "2.12.1", baseSpecificationUrl, "specB12"}},
     {Issue::ReferenceRule::MATH_CHILD, {"MATH_CHILD", "2.12.2", baseSpecificationUrl, "specB12"}},
+    {Issue::ReferenceRule::MATH_CI, {"MATH_CI", "2.12.3", baseSpecificationUrl, "specB12"}},
     {Issue::ReferenceRule::MATH_CI_VARIABLE_REF, {"MATH_CI_VARIABLE_REF", "2.12.3", baseSpecificationUrl, "specB12"}},
```
hsorby: We now have two references to the same rule?

src/utilities.h
```diff
@@ -800,4 +801,50 @@ std::vector<UnitsPtr> unitsUsed(const ModelPtr &model, const ComponentPtr &compo
...
+/**
+ * @brief Return the number of non-comment children.
+ *
+ * Return the number of non-comment children for the given node.
+ *
+ * @param node The node for which we want the number of non-comment children.
+ *
+ * @return The number of non-comment children.
+ */
+size_t nonCommentChildCount(const XmlNodePtr &node);
```
hsorby, line 847: Does this count or skip whitespace nodes?

src/xmlnode.cpp
```diff
@@ -256,6 +266,11 @@ XmlAttributePtr XmlNode::firstAttribute() const
     return attributeHandle;
 }
 
+bool XmlNode::is(const XmlNodePtr &node) const
```
hsorby: This looks like you are creating an equality operator. I am also not keen on the name `is`.  I would prefer if the operator is not to be overridden for the name to be `equals`.

tests/test_utils.h
```diff
@@ -36,7 +36,7 @@ const std::string NON_EMPTY_MATH =
     "    <ci>var</ci>\n"
     "    <apply>\n"
     "      <plus/>\n"
-    "      <cn cellml:units=\"dimensionless\">3.44<sep/>2</cn>\n"
+    "      <cn cellml:units=\"dimensionless\" type=\"e-notation\">3.44<sep/>2</cn>\n"
```
hsorby, line 39: Does the MathML standard require the type to be set for e-notation numbers?

src/issue.cpp
agarny: As you can see, we already have `MATH_CI` which is used in case a `ci` element doesn't have an identifier as a child (see [here](https://github.com/agarny/libcellml/blob/issue1082/src/validator.cpp#L2109-L2111)). However, here, I would like to distinguish that `MATH_CI` case with the case where an identifier is provided, BUT it's not a variable that exists in the component. The MathML rule addresses the first case, but not really the latter one. So, not sure what to do here.

src/utilities.h
agarny: What do you mean by whitespace node exactly? Say that you have:
```xml
<apply>
    <!-- 'eq' element not the first MathML sibling. -->
    <ci>a</ci>
    <eq/>
    <ci>b</ci>
</apply>
```
how would you have an whitespace node within the `apply` element?

src/xmlnode.cpp
agarny: We do indeed use `equals()` in other places, so ok, done in 17155ef4.

tests/test_utils.h
agarny: Yes, see https://www.w3.org/TR/MathML2/chapter4.html#contm.typeattrib.

src/utilities.h
nickerso: each of the newlines and indents would be whitespace nodes in the raw XML DOM, but I have a feeling that maybe we ignore whitespace nodes when parsing math?

src/issue.cpp
nickerso: the CellML spec doesn't make that distinction. The spec just says (in rule 2.12.3) that `The contents of a MathML ci element MUST be a valid variable reference`. Adding a new check on `ci` elements having exactly one non-comment child node is not checking that rule - so I would be in favour of not adding this check...or at least not linking it to rule 2.12.3.

tests/test_utils.h
hsorby: The MathML DTD doesn't seem to care.

src/utilities.h
agarny: Ok, I thought it might be that since indeed libxml2 ignores them entirely.

src/issue.cpp
agarny: We definitely need that check, so that the analyser can safely make use of that non-comment child node (should a model be considered valid).

Otherwise, I have just realised that `MATH_CI_VARIABLE_REF` was already present and that I actually added `MATH_CI` (still for the reason given [above](https://github.com/cellml/libcellml/pull/1083#discussion_r1098154787)). So, I think it should be fine to use `MATH_CI_VARIABLE_REF` where I used `MATH_CI` (see [here](https://github.com/agarny/libcellml/blob/issue1082/src/validator.cpp#L2109-L2111)), and remove `MATH_CI`.

This is what I have done in commit b93075a5.

tests/test_utils.h
agarny: Yeah, it doesn't care about semantics, just syntax, hence this PR indeed. :)

src/issue.cpp
nickerso: Seems reasonable to me.

tests/test_utils.h
agarny: Right now, we have something like:
```xml
<cn cellml:units="dimensionless" type="e-notation">3.44<sep/>2</cn>
```

and the question is indeed to know whether our current validator would accept:
```xml
<cn cellml:units="dimensionless">3.44e2</cn>
```

and it shouln't.

hsorby:
Does a `cn` element that has a value in the `e-notation` format fail if the `type` attribute isn't set?

tests/test_utils.h
hsorby: Does this cover: `<cn cellml:units="dimensionless">3.44<sep/>2</cn>` which should also be unacceptable?

agarny: It does and there is already a test for it.

agarny: Done in c6af868d.

src/utilities.cpp
```diff
@@ -39,6 +39,21 @@ limitations under the License.
 
 namespace libcellml {
 
+bool canConvertToBasicDouble(const std::string &in)
```
nickerso: Do we really need this separate to the below `convertToDouble` method? Which seems to do exactly the same things but also returns that value...

agarny: No, it's not the same as `convertToDouble()` since it allows the exponent part (i.e. `123.45` and `1.23e45` are allowed) while `canConvertToBasicDouble()` doesn't (i.e. `123.45` is allowed, but `1.23e45` is not). Notice that the former relies on `isCellMLReal()` while the latter on `isCellMLBasicReal()`.

(When it comes to returning a `bool` of a `double`, I went for `bool` because our current use of `canConvertToBasicDouble()` doesn't care about the value, just whether it can be converted to a basic double, not to mention that it makes branch coverage much simpler.)

src/xmlnode.h
```diff
@@ -246,6 +246,28 @@ class XmlNode
...
+    /**
+     * @brief Check if this @c XmlNode is a text node representing a basic number.
+     *
+     * Checks whether this @c XmlNode is a text node representing a basic number.
+     * Returns @ true if so, and @c false otherwise.
+     *
+     * @return @c true if this @c XmlNode is a text node representing a basic number
+     * and @c false otherwise.
+     */
+    bool isBasicNumber() const;
```
hsorby: This function has to be called `isBasicReal`, it is not testing if the node is a number.

src/xmlnode.h
```diff
@@ -246,6 +246,28 @@ class XmlNode
...
+
+    /**
+     * @brief Check if this @c XmlNode is a text node representing an integer.
+     *
+     * Checks whether this @c XmlNode is a text node representing an integer.
+     * Returns @ true if so, and @c false otherwise.
```
hsorby: Missing type for `@`

src/utilities.cpp
hsorby: I would like to see this refactored to reuse the code, this function and `convertToDouble` have similarities that can be combined to reduce repetition.

hsorby (changes requested):

src/xmlnode.h
```diff
@@ -246,6 +246,28 @@ class XmlNode
      */
     bool isText() const;
 
+    /**
+     * @brief Check if this @c XmlNode is a text node representing a basic number.
+     *
+     * Checks whether this @c XmlNode is a text node representing a basic number.
+     * Returns @ true if so, and @c false otherwise.
```
hsorby: Missing type for `@`.

src/validator.cpp
```diff
@@ -1635,6 +1708,531 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+                                                             const XmlNodePtr &node,
+                                                             const std::string &name,
+                                                             const ComponentPtr &component)
+{
+    auto index = 0;
+    auto childNode = mathmlChildNode(parentNode, index);
+
+    while (childNode->equals(node)) {
+        childNode = mathmlChildNode(parentNode, ++index);
+    }
```
hsorby: For how long can a child node equal a node. I would have thought that this could only be one?
Doesn't the method name just dictate calling `childNode = mathmlChildNode(parentNode, 1)`?

src/generator.cpp
```diff
@@ -32,6 +32,14 @@ limitations under the License.
...
+#    undef TRUE
+#endif
+
+#ifdef FALSE
+#    undef FALSE
+#endif
+
 #ifdef NAN
 #    undef NAN
 #endif
```
hsorby, line 45: Does anyone else think these need an explanation?

src/utilities.cpp
agarny: What about my comment just above?

src/xmlnode.h
agarny: 3170dc65

src/xmlnode.h
agarny: 830ab595

src/xmlnode.h
agarny: 830ab595

src/generator.cpp
agarny: It has to do with building on Linux. I could confirm (same with `NAN`) and make a comment about it, if really needed.

src/validator.cpp
agarny: Either `childNode = mathmlChildNode(parentNode, 0)` or `childNode = mathmlChildNode(parentNode, 1)`. Improved in 830ab595.

src/utilities.cpp
```diff
@@ -96,29 +101,18 @@ std::string convertToString(double value, bool fullPrecision)
     return strs.str();
 }
 
-int convertToInt(const std::string &in, bool *ok)
+bool convertToInt(const std::string &in, int *out)
```
hsorby: We should pass this int by reference as we have done for the `convertToDouble` function.

src/utilities.h
```diff
@@ -170,41 +171,62 @@ static const std::map<Variable::InterfaceType, std::string> interfaceTypeToStrin
...
  *
  * @sa isCellMLReal
  *
  * @param in The @c std::string value to convert to a @c double.
- * @param ok Optional parameter returns @c true if the conversion was successful and @c false if it wasn't.
+ * @param out The value of the @p in as a doulbe if the conversion was successful,
+ * left unchanged if the conversion fails.
  *
- * @return The double value of @p in.
+ * @return The @c true if the conversion succeeded, @c false otherwise.
```
hsorby: Remove `The`

hsorby (changes requested):

src/utilities.cpp
agarny: 301fb571

src/utilities.h
agarny: 95dbdcab

hsorby (dismissed):
I think this one is good to go!

src/utilities.cpp
```diff
@@ -39,29 +39,34 @@ limitations under the License.
 
 namespace libcellml {
 
-double convertToDouble(const std::string &in, bool *ok)
+bool stringToDoube(const std::string &in, double &out)
```
nickerso: `stringToDoube` ?! really? how'd that get past @hsorby. Guess people rely on autocomplete too much for the coding ;-)

src/utilities.h
```diff
@@ -800,4 +835,50 @@ std::vector<UnitsPtr> unitsUsed(const ModelPtr &model, const ComponentPtr &compo
...
+ * @brief Return the MathML child at a given index.
+ *
+ * Return the MathML child, at @p index, of the given node.
+ *
+ * @param node The node from which we want the MathML child at @p index.
+ * @param index The index of the MathML child.
+ *
+ * @return The MathML child at @p index.
+ */
+XmlNodePtr mathmlChildNode(const XmlNodePtr &node, size_t index);
```
nickerso, line 882: I'm finding the description of this one a bit confusing...from the description I'm expecting to get the index'th child of the node as a MathML node. But the implementation will only index MathML children.

I'd find `Return the index'th MathML child for the given node` more consistent with the implementation. Is it just me?

nickerso: (something to think about moving forward, not needing to be changed in this PR)

src/validator.cpp
```diff
@@ -492,6 +492,69 @@ class Validator::ValidatorImpl: public LoggerImpl
...
+    /**
+     * @brief Add a MathML-related issue.
+     *
+     * Add a MathML-related issue.
+     *
+     * @param description The description for the MathML-related issue.
+     * @param referenceRule The reference rule for the MathML-related issue.
+     * @param component The component where the MathML-related issue occurred.
+     */
+    void addMathMLIssue(const std::string &description,
```
nickerso: Random observation - not consistent in `MathML` vs `Mathml` :)

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+
+    addIssue(issue);
+}
+
+bool Validator::ValidatorImpl::hasOneMathmlSibling(const XmlNodePtr &parentNode,
+                                                   const XmlNodePtr &node,
+                                                   const ComponentPtr &component)
+{
+    if (mathmlChildCount(parentNode) != 2) {
+        addMathMLIssue("Math has a '" + node->name() + "' element without a MathML sibling.",
```
nickerso: Should be "Math has a xxx element without exactly one MathML sibling."

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+
+    return true;
+}
+
+bool Validator::ValidatorImpl::hasTwoMathmlSiblings(const XmlNodePtr &parentNode,
+                                                    const XmlNodePtr &node,
+                                                    const ComponentPtr &component)
+{
+    if (mathmlChildCount(parentNode) != 3) {
+        addMathMLIssue("Math has a '" + node->name() + "' element without two MathML siblings.",
```
nickerso: Should be "Math has a xxx element without exactly two MathML siblings."

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+}
+
+size_t Validator::ValidatorImpl::hasOneOrTwoMathmlSiblings(const XmlNodePtr &parentNode,
+                                                           const XmlNodePtr &node,
+                                                           const ComponentPtr &component)
+{
+    auto childCount = mathmlChildCount(parentNode);
+
+    if ((childCount != 2) && (childCount != 3)) {
+        addMathMLIssue("Math has a '" + node->name() + "' element without one or two MathML siblings.",
```
nickerso: Should be "Math has a xxx element without either exactly one or exactly two MathML siblings." -  I think. Not so certain on the wording for this one.

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+
+    return childCount - 1;
+}
+
+bool Validator::ValidatorImpl::isFirstMathmlChild(const XmlNodePtr &parentNode,
+                                                  const XmlNodePtr &node,
+                                                  const ComponentPtr &component)
+{
+    if (!mathmlChildNode(parentNode, 0)->equals(node)) {
+        addMathMLIssue("Math has a '" + node->name() + "' element which is not the first MathML sibling.",
```
nickerso: Should be "..is not the first MathML child" right? not sibling...

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+
+    return true;
+}
+
+bool Validator::ValidatorImpl::isSecondMathmlChild(const XmlNodePtr &parentNode,
+                                                   const XmlNodePtr &node,
+                                                   const ComponentPtr &component)
+{
+    if (!mathmlChildNode(parentNode, 1)->equals(node)) {
+        addMathMLIssue("Math has a '" + node->name() + "' element which is not the second MathML sibling.",
```
nickerso: child not sibling

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+                       component);
+
+        return false;
+    }
+
+    return true;
+}
+
+bool Validator::ValidatorImpl::hasFirstMathmlSiblingWithName(const XmlNodePtr &parentNode,
+                                                             const XmlNodePtr &node,
```
nickerso, line 1832: This isn't actually checking the "first sibling" its checking the the first or second child of the provided parent node. I suspect this works due to the patterns being checked, but the method name implies something much more general. Something to worry about in future perhaps...

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+    }
+
+    return true;
+}
+
+bool Validator::ValidatorImpl::hasOneMathmlChild(const XmlNodePtr &node,
+                                                 const ComponentPtr &component)
+{
+    if (mathmlChildCount(node) != 1) {
+        addMathMLIssue("Math has a '" + node->name() + "' element without a MathML child.",
```
nickerso: without exactly one MathML child

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+    }
+
+    return true;
+}
+
+bool Validator::ValidatorImpl::hasTwoMathmlChildren(const XmlNodePtr &node,
+                                                    const ComponentPtr &component)
+{
+    if (mathmlChildCount(node) != 2) {
+        addMathMLIssue("Math has a '" + node->name() + "' element without two MathML children.",
```
nickerso: without exactly two MathML children

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+    return true;
+}
+
+bool Validator::ValidatorImpl::hasOneOrTwoMathmlChildren(const XmlNodePtr &node,
+                                                         const ComponentPtr &component)
+{
+    auto childCount = mathmlChildCount(node);
+
+    if ((childCount != 1) && (childCount != 2)) {
+        addMathMLIssue("Math has a '" + node->name() + "' element without one or two MathML children.",
```
nickerso: without exactly one or two children.

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+
+        // Assignment, and relational and logical operators.
+
+    } else if (node->isMathmlElement("eq")
+               || node->isMathmlElement("neq")
+               || node->isMathmlElement("lt")
+               || node->isMathmlElement("leq")
+               || node->isMathmlElement("gt")
+               || node->isMathmlElement("geq")) {
+        auto parentNode = node->parent();
```
nickerso, line 1937: eq, leq, lt, geq, gt are n-ary operators (followed by zero or more child elements) and neq is a binary operator (followed by exactly two child elements). I don't recall the CellML spec putting additional restrictions on the n-ary operators...but I could easily be wrong.

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+
+    } else if (node->isMathmlElement("plus")
+               || node->isMathmlElement("minus")) {
+        auto parentNode = node->parent();
+
+        hasAtLeastOneMathmlSibling(parentNode, node, component)
+            && isFirstMathmlChild(parentNode, node, component);
+    } else if (node->isMathmlElement("times")
+               || node->isMathmlElement("divide")) {
+        auto parentNode = node->parent();
```
nickerso, line 1967: divide is a binary operator (times is n-ary)

src/validator.cpp
nickerso: I'm also starting to think that maybe this validation is designed to validate against what the code generation can work with, rather than strictly validating against the MathML spec...

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+
+        // Piecewise statement.
+
+    } else if (node->isMathmlElement("piecewise")) {
+        for (size_t i = 0, iMax = mathmlChildCount(node); i < iMax; ++i) {
+            validateMathMLElementsChildrenAndSiblings(mathmlChildNode(node, i), component);
+        }
+    } else if (node->isMathmlElement("piece")) {
+        if (hasTwoMathmlChildren(node, component)) {
+            for (size_t i = 0, iMax = mathmlChildCount(node); i < iMax; ++i) {
```
nickerso: is a loop necessary given we know there are exactly two MathML children to validate?

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+                validateMathMLElementsChildrenAndSiblings(mathmlChildNode(node, i), component);
+            }
+        }
+    } else if (node->isMathmlElement("otherwise")) {
+        if (hasOneMathmlChild(node, component)) {
+            for (size_t i = 0, iMax = mathmlChildCount(node); i < iMax; ++i) {
```
nickerso: as above, we know there is exactly one MathML child to validate, so not sure a loop is needed...

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+        //     <ci>a</ci>
+        //   </apply>
+        //
+        // and
+        //
+        //   <apply>
+        //     <log/>
+        //     <logbase>
+        //       <cn cellml:units="dimensionless">3</cn>
+        //     </degree>
```
nickerso: `</logbase>` not degree.

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+            }
+        } else {
+            addMathMLIssue("Math has a 'cn' element which is not of 'real' or 'e-notation' type.",
+                           Issue::ReferenceRule::MATH_CN_FORMAT,
+                           component);
+        }
+
+        // Qualifier elements.
+
+    } else if (node->isMathmlElement("degree")) {
```
nickerso, line 2160: this check is duplicating the checks being done for `root` and `bvar`. Perhaps we can skip these checks?

src/validator.cpp
```diff
@@ -1635,6 +1708,530 @@ void Validator::ValidatorImpl::validateMathMLElements(const XmlNodePtr &node, co
...
+            isSecondMathmlChild(parentNode, node, component)
+                && hasOneMathmlChild(node, component);
+        } else if (siblingCount == 2) {
+            // Used with a 'root' element.
+
+            hasFirstMathmlSiblingWithName(parentNode, node, "root", component)
+                && isSecondMathmlChild(parentNode, node, component)
+                && hasOneMathmlChild(node, component);
+        }
+    } else if (node->isMathmlElement("logbase")) {
```
nickerso, line 2200: as for degree, this duplicates the checks done for `log`...

nickerso (changes requested):
One definite change that is required and a bunch more comments that I think need to be addressed but open to be convinced otherwise.

Also not so keen on the inconsistent `Mathml` vs `MathML` in method names (and the corresponding `mathml`) - but happy to defer that to a future tidy up if needed.

(I haven't checked the tests yet as my comments will likely require a lot of changes there if they are all addressed...)

src/utilities.cpp
agarny: Oopsie! :) That's why we want two people to review everything. :)

c079531a

src/validator.cpp
agarny: Agreed! Changed to `addMathmlIssue`.

d70e227b

src/validator.cpp
agarny: 234b379d

src/validator.cpp
agarny: 472fdbdd

src/validator.cpp
agarny: ef4c8926

src/utilities.h
agarny: Yes, my English wasn't great here.

8f9ab921

src/validator.cpp
agarny: Overuse of copying/pasting. My bad!

1bf3812f

src/validator.cpp
agarny: Overuse of copying/pasting. My bad!

1bf3812f

src/validator.cpp
agarny: Am happy with your wording. :) Also somewhere else as can be seen in the fix.

f85deea6

src/validator.cpp
agarny: Actually, no. It's the name of the method which is wrong.

46c115db

src/validator.cpp
agarny: Nope, wrong name of method in fact.

46c115db

src/validator.cpp
agarny: I called it that way because it's checking for the first MathML sibling of `node`. To do that check, we need to use `node`'s parent and go through its children and skip `node` itself.

src/validator.cpp
agarny: 5a4c96d2

src/validator.cpp
agarny: b4cf1e05

src/validator.cpp
agarny: fc447618

src/validator.cpp
agarny: Actually overriden.

fc447618

src/validator.cpp
agarny: No, it doesn't. Someone could perfectly use `logbase` without using `log` and this is what those checks are about. See [this CellML file](https://github.com/agarny/libcellml/blob/issue1082/tests/resources/invalidmathmlelementschildrenorsiblings.cellml) and search for `logbase` to see what I mean.

src/validator.cpp
agarny: No, it doesn't. Someone could perfectly use `degree` without using `root` or `bvar`, and this is what those checks are about. See [this CellML file](https://github.com/agarny/libcellml/blob/issue1082/tests/resources/invalidmathmlelementschildrenorsiblings.cellml) and search for `degree` to see what I mean.

src/validator.cpp
agarny: Oops!

3786b241

src/validator.cpp
agarny: Oops!

3786b241

src/validator.cpp
agarny: Damn it! I made the same mistake for the `minus` operator which is either a unary or a binary operator (see 5a93957f).

ac71b283

src/validator.cpp
agarny: > eq, leq, lt, geq, gt are n-ary operators (followed by zero or more child elements) and neq is a binary operator (followed by exactly two child elements). I don't recall the CellML spec putting additional restrictions on the n-ary operators...but I could easily be wrong.

Argh, it is clearly a leftover from my COR code! I have just double checked the CellML specs and, as I expected, we don't have any restrictions on those operators.

> I'm also starting to think that maybe this validation is designed to validate against what the code generation can work with, rather than strictly validating against the MathML spec...

Apart from my oversight regarding the `minus` and `divide` operators, I used the MathML specifications to validate things.

Now, I clearly overlooked the type of operators `eq`, `neq`, `lt`, `leq`, `gt`, and `geq` are. `neq` is indeed a binary operator while the others are n-ary operators as confirmed [here](https://www.w3.org/TR/MathML2/chapter4.html#contm.relation).

I am, however, confused as how we can interprete an n-ary operator. If we consider `eq`, it means that we could have:
- No operand, i.e.
```xml
    <apply>
        <eq/>
    </apply>
</math>
```
What do/can we do with that?
- One operand, e.g.
```xml
    <apply>
        <eq/>
        <ci>a</ci>
    </apply>
</math>
```
What do/can we do with that?
- Two operands, e.g.
```xml
    <apply>
        <eq/>
        <ci>a</ci>
        <ci>b</ci>
    </apply>
</math>
```
That's what we support and we interpret it as meaning `a = b` if `eq` is the root element of a MathML equation or `a == b` otherwise.
- Three or more operands, e.g.
```xml
    <apply>
        <eq/>
        <ci>a</ci>
        <ci>b</ci>
        <ci>c</ci>
        <ci>d</ci>
        <ci>e</ci>
    </apply>
</math>
```
I understand this as meaning `a = b = c = d = e` or, once code has been generated for it, as `a = b`, `b = c`, `c = d`, and `d = e`. That is, if `eq` is the root element of a MathML equation. If iis not then I understand this as meaning `a == b == c == d == e` or `(a == b) && (b == c) && (c == d) && (d == e)`.

I guess that in COR, I went for the easiest solution and considered those operators as binary operators, and did the same here.

@nickerso and @hsorby... views?

nickerso: yeah, this was why I was wondering if we really should be validating against what libCellML can handle rather than the MathML spec directly. I'm happy to leave this "as is" and punt it to a future issue to handle the validation and code generation for these cases when a need is demonstrated. We can be fairly confident in knowing that these are currently only used in the binary manner for now....I think :)

agarny: Yeah, I think that after more than 20 years, it's pretty safe to say that people have been using those operators in a binary manner. :)

src/validator.cpp
nickerso: doh! you're right :)

src/validator.cpp
nickerso: ok, I think I'm getting this one. The method is simply checking the first MathML child in practice, but the way this method is used its actually checking that the given node is the first sibling...

src/validator.cpp
nickerso: yeah, I'm understanding this collection of methods now...its not so much about how the checks are being performed but how they are used.

agarny:
Ok, I believe @nickerso's points have been addressed, so time for another round of reviewing. Ping @hsorby.

src/validator.cpp
nickerso: I'll resolve this thread for now and have added #1139 so we might remember the issue in future. @hsorby - feel free to reopen if you think we need to address this in this PR.

nickerso (approved):
nice!

hsorby (approved):

PR #1088, opened 2023-01-26
Fix for CI issues

### LINKED ISSUE #1087: Problems with CI sometimes failing
The CI sometimes fails, this should be fixed as it is causing development issues.

hsorby:
I think our CI issues that we have been experiencing have been due to a race condition in the parallel builds of the coverage testing.  I believe this change will fix the race condition.

Fixes #1087.


agarny:
So, it's not an IT-related issue then? If so then well done on tracking it down!

hsorby:
Unfortunately it seems more of a Hugh issue.  Time will tell if I have indeed fixed it, but so far so good.

agarny:
> Unfortunately it seems more of a Hugh issue. Time will tell if I have indeed fixed it, but so far so good.

Oh well, we all make mistakes, even though it has really driven me mad lately!

agarny (approved):

nickerso (approved):

PR #1112, opened 2023-01-28
Resolve branch coverage in xmldoc.cpp.

### LINKED ISSUE #1106: Clear missed branches from xmldoc.cpp

hsorby:
Fixes #1106.


src/xmldoc.cpp
```diff
@@ -43,11 +44,8 @@ namespace libcellml {
  */
 void structuredErrorCallback(void *userData, xmlErrorPtr error)
 {
-    std::string errorString = std::string(error->message);
     // Swap libxml2 carriage return for a period.
-    if (errorString.substr(errorString.length() - 1) == "\n") {
-        errorString.replace(errorString.end() - 1, errorString.end(), ".");
-    }
+    std::string errorString = std::regex_replace(error->message, std::regex("\\n"), ".");
```
agarny: If I recall correctly, to created a regular expression is somewhat time consuming, so might want to make `[std](std::regex("\\n"))` static.

agarny (changes requested):

src/xmldoc.cpp
hsorby: [bd8b62e](https://github.com/cellml/libcellml/pull/1112/commits/bd8b62e380653253b380c4e941d7cbbc597a7ac3)

src/xmldoc.cpp
```diff
@@ -32,6 +32,8 @@ limitations under the License.
 
 namespace libcellml {
 
+static std::regex newLineRegex("\\n");
```
agarny: Why not simply have it in `structuredErrorCallback()`, at least until we need it elsewhere too? Also, might want to make it `static const` rather than just `static`.

hsorby: [172e825](https://github.com/cellml/libcellml/pull/1112/commits/172e82586d21b14c53d16548d3b684e28336114d)

agarny (approved):

nickerso (approved):

PR #1123, opened 2023-02-05
Resolve branch coverage in printer.cpp

### LINKED ISSUE #1098: Clear missed branches from printer.cpp

hsorby:
Fixes #1098.

### DIFF
--- a/src/debug.cpp
+++ b/src/debug.cpp
@@ -118,6 +118,21 @@ void printConnectionMap(const ConnectionMap &map)
     }
 }
 
+void printComponentMap(const ComponentMap &map)
+{
+    Debug() << "Print out of component map";
+    for (const auto &iter : map) {
+        auto key = iter.first;
+        auto value = iter.second;
+        Debug() << "first: " << key->name();
+        if (value) {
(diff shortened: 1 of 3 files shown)


agarny (approved):

src/debug.cpp
```diff
@@ -118,6 +118,21 @@ void printConnectionMap(const ConnectionMap &map)
     }
 }
 
+void printComponentMap(const ComponentMap &map)
```
nickerso, line 121: what am I missing? Isn't this adding a method that is not used anywhere?

hsorby: You are probably missing that it is a debugging function for debugging only.  We keep some debugging code around in case it is useful, but we don't allow it to be included. We have a test that determines if the debug header file is included to guard against accidental inclusion.

nickerso: (just weird to see a new unused method added in a PR that addresses coverage...)

nickerso (approved):

src/debug.cpp
hsorby: I had to use it to help me figure out if I could trigger the case I was after.

PR #1124, opened 2023-02-08
Resolve branch coverage in importer.cpp

### LINKED ISSUE #1094: Clear missed branches from importer.cpp

hsorby:
Fixes #1094.


src/importer.cpp
```diff
@@ -136,17 +136,11 @@ bool Importer::ImporterImpl::checkUnitsForCycles(const UnitsPtr &units, History
...
-        for (size_t u = 0; u < units->unitCount(); ++u) {
-            std::string ref;
-            std::string prefix;
-            std::string id;
-            double multiplier;
-            double exponent;
-
-            units->unitAttributes(u, ref, prefix, exponent, multiplier, id);
+        for (size_t index = 0; index < units->unitCount(); ++index) {
+            std::string ref = units->unitAttributeReference(index);
```
agarny, line 140: Might want to use `auto ref = ...` (same in other places).

src/model.cpp
```diff
@@ -298,14 +298,27 @@ bool hasComponentImports(const ComponentEntityConstPtr &componentEntity)
     return importsPresent;
 }
 
+bool hasUnitsImports(const UnitsPtr &units)
+{
+    bool importPresent = units->isImport();
+    auto model = owningModel(units);
+    for (size_t index = 0; !importPresent && (index < units->unitCount()); ++index) {
```
agarny: `unitsCount()` does quite a bit of castings. Might want to replace with:
```c++
auto unitsCount = units->unitCount();
for (size_t index = 0; !importPresent && (index < unitsCount); ++index) {
```

tests/importer/importer.cpp
```diff
@@ -1183,3 +1198,193 @@ TEST(Importer, importInvalidComponentFromCellmlModelResetError)
...
+    importModel2->addUnits(thirdUnits);
+    importModel2->addComponent(component3);
+
+    auto importedComponent1 = libcellml::Component::create("component1_import");
+    auto importedComponent2 = libcellml::Component::create("component2_import");
+
+    model->addComponent(importedComponent1);
+    importModel1->addComponent(importedComponent2);
+
+    auto printer = libcellml::Printer::create();
```
agarny: Not used.

tests/importer/importer.cpp
```diff
@@ -1183,3 +1198,193 @@ TEST(Importer, importInvalidComponentFromCellmlModelResetError)
     EXPECT_EQ(size_t(1), importer->errorCount());
     EXPECT_EQ(e, importer->error(0)->description());
 }
+
+TEST(Importer, isResolvedUnitsChildUnresolvedII)
```
agarny: I know that those are tests, but the naming of objects is a bit all over the place. For instance, `component1`, but `variable` rather than `variable1` as is the case with `component2` and `variable2`. Also unnecessarily difficult to read when `importModel1` has `component2` while it might have been easier to read if it had been `component1`. `frstUnits`, etc. might be renamed to `units1`, etc. Could also declare variables just before making use of them (e.g., `importer` is declared at the very beginning and used at the very end).

Otherwise, I find this kind of tests difficult to "visualise". I would much prefer to see some CellML code.

agarny (commented):
Just putting out there some comments I wrote when going through this last week. I am not done though. It's just that I don't want to inadvertently close this tab and lose what I had started doing. Will get back to it once I am done with my current libCellML work...

src/importer.cpp
hsorby: I prefer the `std::string` as it is clearer.

agarny: We are not consistent, but so be it.

src/model.cpp
hsorby: [7916d86](https://github.com/cellml/libcellml/pull/1124/commits/7916d86243527f90f4b7a75a809bdcc31f475a16)

tests/importer/importer.cpp
hsorby: [8e5de50](https://github.com/cellml/libcellml/pull/1124/commits/8e5de508a120dac0c133a2eee14e9c2b07584455)

tests/importer/importer.cpp
hsorby: [dcedc76](https://github.com/cellml/libcellml/pull/1124/commits/dcedc76e66c3d460f59c314e6b0ca5eb021c335e)

This test was only used to explore possibilities and should have been removed. Gone now.

nickerso (approved):

tests/importer/importer.cpp
```diff
@@ -1183,3 +1198,126 @@ TEST(Importer, importInvalidComponentFromCellmlModelResetError)
...
+        "  </component>\n"
+        "</model>\n";
+
+    auto importer = libcellml::Importer::create();
+    auto parser = libcellml::Parser::create();
+
+    auto model = parser->parseModel(fileContents("importer/model_cascaded_units.cellml"));
+
+    importer->resolveImports(model, resourcePath("importer"));
+    EXPECT_EQ(size_t(0), importer->errorCount());
```
agarny, line 1315: Why use `errorCount()` in some places and `issueCount()` in some others? At the end of the day, this file passes all the tests with `issueCount()` being used everywhere.

agarny (approved):

tests/importer/importer.cpp
hsorby: I can't remember exactly why I did this but I think it was for two main reasons:
1. Utilise the `errorCount` API more for coverage
2. Show that no `issues` were created be they `warnings`, or `infos`, show a focus on errors only.

PR #1128, opened 2023-02-14
Resolve branch coverage in validator.cpp

### LINKED ISSUE #1103: Clear missed branches from validator.cpp

hsorby:
Fixes #1103.


nickerso (dismissed):

hsorby:
Excessive number of commits now. Have we lost track of a merging point?

nickerso (dismissed):

nickerso (dismissed):

src/validator.cpp
```diff
@@ -180,7 +205,7 @@ uint32_t convertTextToUint32(const std::string &text, size_t initialValue = 0)
     uint32_t value = 0;
     size_t index = 0;
     for (size_t j = initialValue; j < 4; ++j) {
-        auto tempValue = static_cast<uint32_t>(text[index++]) << bitShifts[j];
+        auto tempValue = static_cast<uint32_t>((unsigned char)text[index++]) << bitShifts[j];
```
agarny: Two castings, including a C-type casting? Why? I am actually surprised that Clang-Tidy allows C-type casting...? (Same below.)

tests/coverage/coverage.cpp
```diff
@@ -756,6 +756,88 @@ TEST(Importer, notMathMLMathNodesInComponentMath)
...
+    component->addVariable(variable);
+    component->appendMath(math);
+
+    model->addComponent(component);
+
+    auto validator = libcellml::Validator::create();
+
+    validator->validateModel(model);
+    EXPECT_EQ(size_t(0), validator->errorCount());
+}
```
agarny, line 784: I would think that we already have that kind of test in https://github.com/cellml/libcellml/issues/1082?

tests/coverage/coverage.cpp
```diff
@@ -756,6 +756,88 @@ TEST(Importer, notMathMLMathNodesInComponentMath)
...
+
+TEST(CoverageValidator, invalidXmlIds)
+{
+    auto parser = libcellml::Parser::create();
+    auto model = parser->parseModel(fileContents("coverage/invalidxmlids.cellml"));
+
+    auto validator = libcellml::Validator::create();
+
+    validator->validateModel(model);
+    EXPECT_EQ(size_t(1), validator->errorCount());
```
agarny, line 794: Which ids in `invalidxmlids.cellml` are invalid? All of them? Only one of them? If so, which one?

tests/coverage/coverage.cpp
```diff
@@ -756,6 +756,88 @@ TEST(Importer, notMathMLMathNodesInComponentMath)
...
+    v3->setUnits("dimensionless");
+
+    c->addVariable(v1);
+    c->addVariable(v2);
+    c->addVariable(v3);
+    c->setMath(math);
+    m->addComponent(c);
+
+    v->validateModel(m);
+    EXPECT_EQ(size_t(3), v->issueCount());
```
agarny, line 845: It would be nice to have the exact issue descriptions rather than just test for the number of issues.

tests/coverage/coverage.cpp
hsorby: No, that is where the hole this test is plugging came from.

src/validator.cpp
hsorby: I did have a reason for the double casting, wish you had asked me this when I made the change.

agarny: Apparently, better late than never, so here we are. :)

tests/coverage/coverage.cpp
hsorby: [528a80f](https://github.com/cellml/libcellml/pull/1128/commits/528a80f1273ba2e874ef174e26b25907d50a9688)

tests/coverage/coverage.cpp
hsorby: [528a80f](https://github.com/cellml/libcellml/pull/1128/commits/528a80f1273ba2e874ef174e26b25907d50a9688)

src/validator.cpp
hsorby: [1291e1f](https://github.com/cellml/libcellml/pull/1128/commits/1291e1f19e6b7020f2b7aacc026247d6f16c0ac0)

hsorby: It is because the text could theoritically be UTF-16 or UTF-32 and you can't cast those to `uint32_t` without warnings, but we don't allow those encodings so we are able to safely make the cast from UTF-8 text.

agarny: Fair enough. Would it be worth make a note of it in our code?

nickerso (dismissed):

src/validator.cpp
```diff
@@ -2612,7 +2608,7 @@ IdMap Validator::ValidatorImpl::buildModelIdMap(const ModelPtr &model)
                 addIdMapItem(id, info, idMap);
             }
         }
-        if (units->isImport() && units->importSource() != nullptr && !units->importSource()->id().empty()) {
+        if (units->importSource() != nullptr && !units->importSource()->id().empty()) {
```
agarny: Forgot that one:
```c++
if ((units->importSource() != nullptr) && !units->importSource()->id().empty()) {
```

agarny (changes requested):

src/validator.cpp
hsorby: [9a5d2f1](https://github.com/cellml/libcellml/pull/1128/commits/9a5d2f1c22b12b128dc26d904a200122b1d3a678)

nickerso (approved):

agarny (approved):

PR #1131, opened 2023-02-20
Analyser/Generator: added some `xxxAsString()` methods

### LINKED ISSUE #446: Name or map function for generator modelType enums?
I'll start by saying I loathe `enum`s ... here's one reason why ... 

I'm writing a basic tutorial on use of the code generation, and I want to show how to return the `modelType()` (this being an `enum` of UNKNOWN, ODE, ALGEBRAIC, etc etc).  I think it's really awkward for the user to have to retrieve the model type, then write a big switch statement to find out about their code.  There have been all kinds of reflection-ish macros defined around the place, but none seems to be very clean either.

Can I propose either:

- return a string from the public-facing `modelType()` rather than the enum.  We can keep the current modelType functionality as `modelTypeEnum()` or something ... or create a new function called `modelTypeAsString()` or whatever
- provide a map from the enum to something more useful to the user, in the same way as we've done for the `Prefix` enums.
- overload the utility `convertToString()` to take `ModelType` enums as argument (not really a fan of this)

Thoughts?  @agarny especially?

agarny:
Fixes #446.


src/api/libcellml/generatorprofile.h
```diff
@@ -72,6 +72,15 @@ class LIBCELLML_EXPORT GeneratorProfile
...
     Profile profile() const;
 
+    /**
+     * @brief Get the @c Profile as a string for this @c GeneratorProfile.
+     *
+     * Return the @c Profile as a string for this @c GeneratorProfile.
+     *
+     * @return The @c Profile as a string for this @c GeneratorProfile.
+     */
+    std::string profileAsString() const;
```
nickerso: Is this more getting the profile name as a string? I'd half expect that getting the profile as a string would return the profile definition...

agarny: What do you mean by the profile definition? A profile contains a lot, I mean a lot, of information...

nickerso: exactly, so if asking for a profile as a string I'd expect to get a string version of all that information. Whereas the `C` or `Python` is just the name of the profile..

src/analyserequation.cpp
```diff
@@ -68,6 +68,29 @@ AnalyserEquation::Type AnalyserEquation::type() const
     return mPimpl->mType;
 }
 
+std::string AnalyserEquation::typeAsString() const
+{
+    if (mPimpl->mType == Type::TRUE_CONSTANT) {
+        return "TRUE_CONSTANT";
```
hsorby: Are we looking for a direct string form here or a more readable form "True constant"?

src/api/libcellml/generatorprofile.h
agarny: The idea, at this stage, was merely to convert an `enum class` value to a string. Nothing more nothing less.

`GeneratorProfile` has such an `enum class` which is used by `GeneratorProfile::profile()` to return the (original) profile used by an instance of `GeneratorProfile`. So, for that class, I merely added `profileAsString()`.

If I understood you right, what you are after is quite different...?

src/analyserequation.cpp
agarny: This is open for debate. Truth be told, when I first started working on this PR, I had a "proper" string for each `enum class` then when I got to `AnalyserEquationAst`'s `enum class`, I decided to simply have a direct string form. I mean, for `AnalyserEquationAst::Type::ACSCH`, do we really want something like `The arc hyperbolic cosecant function.`? Also, should we start with a capital letter? Should have the `the` article? Should we end up with a full stop? If anything, I might want `arc hyperbolic cosecant function`, but that's rather long.

Otherwise, I have just noticed that for `version()`, we have `versionString()` as opposed to `versionAsString()`. So, if anything, we might to rename the current `xxxAsString()` to `xxxString()`?

Also, looking at https://github.com/cellml/libcellml/pull/1132, I can't say that I am keen on `enumerationAsString()`. It's not part of a class itself, which means that an IDE won't suggest when we type something like `myAnalyserModel->type...` and, instead, the user needs to know about it (who reads documentation...). (Also, had I know that such a PR was in the work (no associated issue?), I wouldn't have wasted my time on this PR...)

hsorby: It wasn't in the works, I just wanted to try out overloading on enumerations. with `enumerationAsString` once I know it is there I can use it for any and all enumerations.

I don't think it should be part of the class, it doesn't sit right for me.

I would probably argue that we should change to `xxxAsString()` rather than the alternative.

hsorby: I wouldn't go for super long names myself the shorter forms are acceptably common, perhaps more so than the formal term.

hsorby: There doesn't appear to be an associated issue with this PR?

src/api/libcellml/generatorprofile.h
hsorby: I agree that the resolution of this issue is about converting enums to strings, not about poorly named enums.  That is a different can of worms.

src/analyserequation.cpp
agarny: > I don't think it should be part of the class, it doesn't sit right for me.

I don't see anything wrong with it myself, but more important I am not keen on "isolated" methods like `enumerationAsString()`.

> I would probably argue that we should change to xxxAsString() rather than the alternative.

I actually agree.

> I wouldn't go for super long names myself the shorter forms are acceptably common, perhaps more so than the formal term.

Ok, we can discus the details tomorrow. The most important, I suppose, is that all the tests (incl. coverage) are all there.

> There doesn't appear to be an associated issue with this PR?

My bad, it was discussed in Slack and forgot to mention it in the PR (I have edited my "first" comment).

src/analyser.cpp
```diff
@@ -34,6 +34,7 @@ limitations under the License.
 #include "analysermodel_p.h"
 #include "analyservariable_p.h"
 #include "anycellmlelement_p.h"
+#include "commonutils.h"
```
hsorby, line 37: Why has this header been added?

src/analyserequationast.cpp
```diff
@@ -18,6 +18,10 @@ limitations under the License.
 
 #include "analyserequationast_p.h"
 
+#ifdef NAN
+#    undef NAN
+#endif
+
```
hsorby, line 24: Hasn't this already been done elsewhere? Why the repatiion

src/analyserexternalvariable.cpp
```diff
@@ -16,9 +16,8 @@ limitations under the License.
 
 #include "libcellml/analyserexternalvariable.h"
 
-#include "libcellml/component.h"
-
 #include "analyserexternalvariable_p.h"
+#include "commonutils.h"
```
hsorby, line 20: Added for what reason?

src/api/libcellml/analyserequation.h
```diff
@@ -61,6 +61,17 @@ class LIBCELLML_EXPORT AnalyserEquation
      */
     Type type() const;
 
+    /**
+     * @brief Get the string version of a @c Type.
```
hsorby: The `@c` is a code construct do you mean `@ref`?

src/enums.cpp
```diff
@@ -20,25 +20,25 @@ limitations under the License.
...
+        {CellmlElementType::MATH, "math"},
+        {CellmlElementType::MAP_VARIABLES, "map_variables"},
+        {CellmlElementType::MODEL, "model"},
+        {CellmlElementType::RESET, "reset"},
+        {CellmlElementType::RESET_VALUE, "reset_value"},
+        {CellmlElementType::TEST_VALUE, "test_value"},
+        {CellmlElementType::UNDEFINED, "undefined"},
+        {CellmlElementType::UNIT, "unit"},
+        {CellmlElementType::UNITS, "units"},
+        {CellmlElementType::VARIABLE, "variable"}};
```
hsorby: These strings are all lowercase different from every other mapping.

hsorby (commented):
The enumerations could probably all have there doxygen comments updated to the correct line reference.

src/analyser.cpp
agarny: As a result of some cleaning up (see 57289480).

src/analyserequationast.cpp
agarny: As a result of some cleaning up (see 57289480).

src/analyserexternalvariable.cpp
agarny: As a result of some cleaning up (see 57289480).

src/enums.cpp
agarny: Fair point. Consider it fixed in a few minutes.

agarny: 15ebc22e

nickerso (dismissed):
I think this is looking ok. The one concern I have is that the use of all the `typeToString.at()` will throw an exception if called with a non-existent type. I think we have previously agreed to avoid throwing exceptions - maybe? Should we either be protecting from throwing exceptions or documenting that they could happen?

Also fine with this going in as is and addressing this as part of that plan to test the API with null values...

agarny:
> I think this is looking ok. The one concern I have is that the use of all the `typeToString.at()` will throw an exception if called with a non-existent type. I think we have previously agreed to avoid throwing exceptions - maybe? Should we either be protecting from throwing exceptions or documenting that they could happen?

Those `typeToString.at()` calls are made within methods that take an `enum class` value as a parameter which is then used as a parameter to the `at()` method. So, we will always be fine.

nickerso:
> > I think this is looking ok. The one concern I have is that the use of all the `typeToString.at()` will throw an exception if called with a non-existent type. I think we have previously agreed to avoid throwing exceptions - maybe? Should we either be protecting from throwing exceptions or documenting that they could happen?
> 
> Those `typeToString.at()` calls are made within methods that take an `enum class` value as a parameter which is then used as a parameter to the `at()` method. So, we will always be fine.

doh! thanks. Is that enforced across to the Python bindings too?

agarny:
> > > I think this is looking ok. The one concern I have is that the use of all the `typeToString.at()` will throw an exception if called with a non-existent type. I think we have previously agreed to avoid throwing exceptions - maybe? Should we either be protecting from throwing exceptions or documenting that they could happen?
> > Those `typeToString.at()` calls are made within methods that take an `enum class` value as a parameter which is then used as a parameter to the `at()` method. So, we will always be fine.
> doh! thanks. Is that enforced across to the Python bindings too?

Hmm, good question! I would guess that it isn't. I mean, `enum class` values are converted to integers in Python, right? So... anything could go there. @hsorby?

hsorby:
Yes, I think we have protections on some of the API for the bindings covering enumerations: see here https://github.com/cellml/libcellml/blob/f34f4d9854733f119e2d4bc8bde6933809d6b498/src/bindings/interface/types.i#L119 I think this should be added for these methods as well.

agarny:
> Yes, I think we have protections on some of the API for the bindings covering enumerations: see here
> 
> https://github.com/cellml/libcellml/blob/f34f4d9854733f119e2d4bc8bde6933809d6b498/src/bindings/interface/types.i#L119
> 
> I think this should be added for these methods as well.

Ok, I believe I have added everything that is needed for our Python bindings. I have also checked for our JavaScript bindings and it looks like we are already all good for that one.

nickerso (dismissed):

hsorby (changes requested):
I know we don't test test our Python bindings but in the case of the `typeAsString` methods and their ilk I think we should test the extents in the Python bindings as this is something that we have added to the interface files to (hopefully) avoid segfaults.

agarny:
> I know we don't test test our Python bindings but in the case of the `typeAsString` methods and their ilk I think we should test the extents in the Python bindings as this is something that we have added to the interface files to (hopefully) avoid segfaults.

I believe it has already been done, not only for Python (e.g., https://github.com/cellml/libcellml/pull/1131/files#diff-6217397c5913cb67ab08329d9d5b36e17a6a2367ac49c585ae580aa1fd11c328), but also for JavaScript (e.g., https://github.com/cellml/libcellml/pull/1131/files#diff-b49b859ee2ff441946db1f6ea8d135398596fccd8b5b023b6811c7e37b253535)... or am I missing something?

src/units.cpp
```diff
@@ -346,51 +399,41 @@ void Units::addUnit(const std::string &reference)
...
+    addUnit(standardUnitToString.at(standardUnit), "0", 1.0, 1.0, "");
 }
 
 void Units::unitAttributes(StandardUnit standardUnit, std::string &prefix, double &exponent, double &multiplier, std::string &id) const
 {
     std::string dummyReference;
-    const std::string reference = standardUnitToString.find(standardUnit)->second;
-    auto result = pFunc()->findUnit(reference);
-    unitAttributes(size_t(result - pFunc()->mUnitDefinitions.begin()), dummyReference, prefix, exponent, multiplier, id);
+    unitAttributes(size_t(pFunc()->findUnit(standardUnitToString.at(standardUnit)) - pFunc()->mUnitDefinitions.begin()), dummyReference, prefix, exponent, multiplier, id);
```
hsorby: Should probably be written as an explicit cast such as: `static_cast<sizet>()`.

src/units.cpp
```diff
@@ -346,51 +399,41 @@ void Units::addUnit(const std::string &reference)
...
 }
 
 void Units::unitAttributes(const std::string &reference, std::string &prefix, double &exponent, double &multiplier, std::string &id) const
 {
     std::string dummyReference;
-    auto result = pFunc()->findUnit(reference);
-    unitAttributes(size_t(result - pFunc()->mUnitDefinitions.begin()), dummyReference, prefix, exponent, multiplier, id);
+    unitAttributes(size_t(pFunc()->findUnit(reference) - pFunc()->mUnitDefinitions.begin()), dummyReference, prefix, exponent, multiplier, id);
```
hsorby: Should probably be written as an explicit cast such as: `static_cast<sizet>()`.

hsorby (changes requested):

src/units.cpp
agarny: 80e32368

src/units.cpp
agarny: 80e32368

nickerso (dismissed):

nickerso (approved):

hsorby (approved):

nickerso (approved):

PR #1133, opened 2023-02-21
Analyser/Generator: don't use weak pointers

### LINKED ISSUE #1130: AnalyserModel returned from Analyser gets broken when the Analyser goes out of scope
After some crafty debugging from @hsorby, he found that the underlying cause of an issue I have been seeing is that the analyser model returned from an analyser we lose some critical information, needed by the generator, once the originating analyser goes out of scope (see test below). This is not typically an issue in our current testing as analysing and code generation tends to happen in a single chunk of code. The following test will fail when retrieving the `generator->implementationCode()` string when the `GeneratorImpl::generateCode()` method attempts to navigate the AST.

```C++
TEST(Generator, analyserModelScopeTest)
{
    libcellml::ParserPtr parser = libcellml::Parser::create(false);

    auto model = parser->parseModel(fileContents("model.cellml"));

    libcellml::AnalyserModelPtr analyserModel = nullptr;
    {
        libcellml::AnalyserPtr analyser = libcellml::Analyser::create();
        analyser->analyseModel(model);
        EXPECT_EQ(size_t(0), analyser->errorCount());
        analyserModel = analyser->model();
    }

    auto generator = libcellml::Generator::create();

    generator->setModel(analyserModel);

    EXPECT_EQ(EMPTY_STRING, generator->interfaceCode());
    EXPECT_EQ(EMPTY_STRING, generator->implementationCode());
}
```

agarny:
Fixes #1130.


src/analyserexternalvariable.cpp
```diff
@@ -28,23 +28,23 @@ AnalyserExternalVariable::AnalyserExternalVariableImpl::AnalyserExternalVariable
...
-                                                                                                              const std::string &componentName,
-                                                                                                              const std::string &variableName)
+std::vector<VariablePtr>::iterator AnalyserExternalVariable::AnalyserExternalVariableImpl::findDependency(const ModelPtr &model,
+                                                                                                          const std::string &componentName,
+                                                                                                          const std::string &variableName)
 {
-    return std::find_if(mDependencies.begin(), mDependencies.end(), [=](const VariableWeakPtr &v) {
-        auto variable = v.lock();
+    return std::find_if(mDependencies.begin(), mDependencies.end(), [=](const auto &v) {
+        auto variable = v;
```
hsorby: Not quite sure what this statement is doing now?

agarny: `variable = v`? Clearly nothing. Will be removed. :)

agarny: 087f9c83

nickerso (dismissed):
Not sure I'm completely getting the subtleties of the pointers here, but I like the updated tests :) once the CI is passing, this looks good to go.

agarny:
> Not sure I'm completely getting the subtleties of the pointers here, but I like the updated tests :) once the CI is passing, this looks good to go.

To fix your original issue was actually very quick. Then, I thought I would get rid of all the weak pointers (in the analyser/generator) since the use of local scopes might cause us a similar problem.

But, I should have clearly been more patient and done it carefully since I now have memory leaks. So, I have now reverted things locally and going through things slowly, but surely.

This PR should therefore be ready for review later today. 🤞 In the meantime, https://github.com/cellml/libcellml/pull/1131 is ready. 🙂

nickerso (dismissed):

src/api/libcellml/generator.h
```diff
@@ -95,7 +95,7 @@ class LIBCELLML_EXPORT Generator
      *
      * @return The interface code as a @c std::string.
      */
-    std::string interfaceCode() const;
+    std::string interfaceCode();
```
hsorby: I think this method and `implementationCode` should retain their constness.

agarny: I completely agree! Done in 81bc8f5d.

nickerso (dismissed):

src/analyserexternalvariable.cpp
```diff
@@ -156,21 +152,15 @@ VariablePtr AnalyserExternalVariable::dependency(const ModelPtr &model,
     auto result = mPimpl->findDependency(model, componentName, variableName);
 
     if (result != mPimpl->mDependencies.end()) {
-        return result->lock();
+        return *result;
     }
 
     return nullptr;
```
hsorby: Is `return (result != mPimpl->mDependencies.end()) ? *result : nullptr;` cleaner?

agarny: 1def632e

hsorby (dismissed):

nickerso (dismissed):

hsorby (approved):

nickerso (approved):

PR #1147, opened 2023-04-05
Fix an issue when initialising a state variable using a variable instead of a constant

### LINKED ISSUE #1144: An issue when initialising a state variable using a variable instead of a constant
`libcellml 0.4.0`

When I tried to write python code from the attached CellML model, I got the issues:
The method "analyse_model" found 2 issues:
    - Variable 'q_Ai' in component 'BG_initTest' is initialised using variable 'q_Ai_init', which is not a constant.
    - Variable 'q_Ao' in component 'BG_initTest' is initialised using variable 'q_Ao_init', which is not a constant.
 
[cellml.zip](https://github.com/cellml/libcellml/files/11143983/cellml.zip)

agarny:
Fixes #1144.


hsorby (commented):
My first comment would be is this a minimal example that illustrates the issue?  It seems to me to be a model that highlights the bug in question.

agarny:
> My first comment would be is this a minimal example that illustrates the issue? It seems to me to be a model that highlights the bug in question.

It is indeed. I just went for the model given to us by @WeiweiAi in issue #1144. I am happy to simplify it, if really needed.

nickerso (dismissed):
I agree that the model to illustrate this issue could be much more minimal. But I can also see benefit in using models contributed by users as part of the bug report...

hsorby:
I think this full model should be added to tutorials or as an end-to-end test.  These tests are unit tests, that is, trying to put the minimal groundwork in place to illustrate a particular issue.  I am in favour of reducing the model to the bare minimum required to create the failing situation.

agarny:
Ok, now using a minimal model.

nickerso (approved):

hsorby (approved):

PR #1157, opened 2023-04-28
Fix importing in permissive mode

### LINKED ISSUE #1156: Segfault when importing units from two models
I am getting a segfault when importing units defined in model `U` when importing them from model `A` and model `B` where 
model `A` imports a component from model `B`.

hsorby:
Fixes #1156.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 3.18.0)
 
-set(CMAKE_OSX_DEPLOYMENT_TARGET 10.14 CACHE STRING "Minimum OS X deployment version.")
+set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Minimum OS X deployment version.")
 
 set(PROJECT_NAME libCellML)
 set(PROJECT_URL https://libcellml.org)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 3.18.0)
 
-set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Minimum OS X deployment version.")
+set(CMAKE_OSX_DEPLOYMENT_TARGET 10.14 CACHE STRING "Minimum OS X deployment version.")
 
 set(PROJECT_NAME libCellML)
 set(PROJECT_URL https://libcellml.org)
--- a/tests/importer/importer.cpp
+++ b/tests/importer/importer.cpp
@@ -1467,20 +1467,6 @@ TEST(Importer, importCascadingUnitsImports)
     EXPECT_EQ(e, a);
 }
 
-TEST(Importer, importingCommonUnitsFromTriangleImportStructure)
-{
-    const std::string e = "The attempt to resolve imports with the model at '" + resourcePath("importer/") + "' failed: the file could not be opened.";
-
-    auto importer = libcellml::Importer::create(false);
-    auto parser = libcellml::Parser::create(false);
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 3.18.0)
 
-set(CMAKE_OSX_DEPLOYMENT_TARGET 10.14 CACHE STRING "Minimum OS X deployment version.")
+set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Minimum OS X deployment version.")
```
agarny: Is that still needed? You increased macOS requirements when you used `std::filesystem::is_regular_file()`, but you are not anymore.

tests/importer/importer.cpp
```diff
@@ -1466,3 +1466,27 @@ TEST(Importer, importCascadingUnitsImports)
     const std::string a = printer->printModel(flatModel);
     EXPECT_EQ(e, a);
 }
+
+void testImporterWithInvalidImportedModels(bool strict)
+{
+    const std::string e = "The attempt to import the model at '" + resourcePath("importer/triangle_units_opposite.cellml") + "' failed: the file is not valid XML.";
```
agarny, line 1472: Is there no way to get more information from the parser as why the file is not valid XML? I imagine that this is because `xmlns:xlink="http://www.w3.org/1999/xlink"` is missing from the `import` element? What kind of error are we getting from libxml2, if any?

agarny (changes requested):

CMakeLists.txt
hsorby: Not for this PR anymore, I have to make another PR for the Julia packaging that does require this though.

tests/importer/importer.cpp
hsorby: I don't think we want to start trying to interpret LibXml2 error messages.  We expect a valid XML document, if we don't get one users are going to have to fix that themselves.  I really feel that XML errors are not our domain.

CMakeLists.txt
agarny: Should be done for that other PR then.

hsorby: Yes, it should.

tests/importer/importer.cpp
agarny: Agreed about not wanting to interpret libxml2 error messages, but I would still provide them, maybe as some kind of additional information. As it stands, the current error message is likely to be of no use to someone who is not familiar with XML and we don't expect CellML users to be familar with XML, so...?

hsorby: I think this is a different issue.  We have talked about multi-level error messages previously.  My recollection is that we would consider this post version 1.0.0.

agarny: Agreed that this has nothing to do with this PR _per se_ and if this is something that we are willing to consider post 1.0.0 then I am fine with it.

luciansmith: I will say that libsbml also does indeed report libxml2 error messages to the user.

Also, the most common XML error is the user giving the library a file that is not XML at all, so that might be worth calling out:  either say 'this file isn't in the correct format'  or 'this file is empty'.  (I forget if an empty file qualifies as 'xml' in libxml2's world.)

CMakeLists.txt
hsorby: [8cdb5b8](https://github.com/cellml/libcellml/pull/1157/commits/8cdb5b8d2439138a2b8951ca36e217f2e9287fe9)

agarny (approved):

nickerso (approved):

PR #1179, opened 2023-10-04
Fix flattening segfault

### LINKED ISSUE #1178: Complete HH model using imports is causing a segfault
The complete HH model written using imports causes a segfault.  The issue comes about when trying to replace equivalent units in imported components that themselves have imported component children.

hsorby:
Fixes #1178.


agarny:
@hsorby: I leave it to you to resolve the conflicts.

hsorby:
Fixed the conflict, ready for review @agarny , @nickerso.

tests/resources/importer/HHComplete/PotassiumChannelModel.cellml
```diff
@@ -0,0 +1,186 @@
...
+              </apply>
+              <cn cellml:units="mV">-80</cn>
+            </apply>
+          </apply>
+        </apply>
+      </apply>
+    </math>
+  </component>
+  <component name="nGateParameters">
+    <variable name="n" units="dimensionless" initial_value="0.325" interface="public"/>
```
nickerso, line 133: interesting to see how the analyser would handle this model... As far as I can tell, `n` is connected to `X` in the imported model, which in turn has its own initial value set... (doesn't affect this PR, just a general interest question!)

nickerso (dismissed):

src/importer.cpp
```diff
@@ -714,6 +714,11 @@ void updateComponentsVariablesUnitsNames(const std::string &name, const Componen
 {
     for (size_t variableIndex = 0; variableIndex < component->variableCount(); ++variableIndex) {
         auto variable = component->variable(variableIndex);
+        if (component->isImport()) {
+            auto importModel = component->importSource()->model();
+            auto importComponent = importModel->component(component->importReference());
+            variable = importComponent->variable(variable->name());
+        }
```
agarny, line 721: Might want to replace with something like:
```c++
VariablePtr variable;
if (component->isImport()) {
    auto importModel = component->importSource()->model();
    auto importComponent = importModel->component(component->importReference());
    variable = importComponent->variable(variable->name());
} else {
    variable = component->variable(variableIndex);
}
```

tests/resources/importer/HHComplete/MembraneModelController.cellml
```diff
@@ -0,0 +1,42 @@
...
+            </apply>
+          </piece>
+          <otherwise>
+            <cn cellml:units="microA_per_cm2">0</cn>
+          </otherwise>
+        </piecewise>
+      </apply>
+    </math>
+  </component>
+</model>
```
agarny: Need an empty line at the end of the file.

agarny (changes requested):
Just a couple of minor suggestions.

src/importer.cpp
hsorby: I think I prefer the current implementation.  I'm not sure you can get the correct variable name from that variable anyway.

agarny: To compute `component->variable(variableIndex);` every time is a waste of time when you have a lot of imports, hence my suggestion.

agarny: Oops, I hadn't noticed the call to `variable->name()` in the case of an import. All good then!

agarny (approved):

nickerso (approved):

PR #1181, opened 2023-10-04
Set macOS 10.15 as the minimum deployment target for macOS.

### LINKED ISSUE #1180: Set minimum OSX deployment target to 10.15
Setting the minimum OSX deployment target to 11 would simplify or maintenance slightly.
According to [wikipedia](https://en.wikipedia.org/wiki/MacOS_version_history) macOS 11 is the oldest currently supported software version, it seems reasonable to set this version as the minimum deployment target for libCellML.

hsorby:
Fixes #1180.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 3.18.0)
 
-set(CMAKE_OSX_DEPLOYMENT_TARGET 10.14 CACHE STRING "Minimum OS X deployment version.")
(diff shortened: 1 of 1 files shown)


nickerso (approved):

agarny (commented):
Is that supposed to fix https://github.com/cellml/libcellml/issues/1180? If so, are we not supposed to target macOS 11 and not macOS 10.15?

nickerso:
#1180 just updated to reflect this change.

agarny (approved):

PR #1184, opened 2023-10-05
Analyser: add support for unknown variables that have been marked as external

### LINKED ISSUE #1182: Analyser: add support for unknown variables that have been marked as external
Say that you have a model with an unknown variable (e.g., it misses an initial variable for the model to make sense). To analyse such a model will rightly result in an error being generated.

However, the analyser should be fine with the model if we were to flag such a variable as being external while, at the moment, it will still complain about the variable being unknown.

agarny:
Fixes #1182.

### DIFF
--- a/src/analyser.cpp
+++ b/src/analyser.cpp
@@ -2522,7 +2522,7 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
 
     auto stateIndex = MAX_SIZE_T;
     auto variableIndex = MAX_SIZE_T;
-    auto loopNb = 1;
+    auto loopNumber = 1;
     bool relevantCheck;
     auto checkNlaSystems = false;
 
@@ -2534,12 +2534,12 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
                             || relevantCheck;
         }
 
-        if (((loopNb == 1) || (loopNb == 3)) && !relevantCheck) {
-            ++loopNb;
+        if (((loopNumber == 1) || (loopNumber == 3)) && !relevantCheck) {
+            ++loopNumber;
 
             relevantCheck = true;
             checkNlaSystems = true;
-        } else if ((loopNb == 2) && !relevantCheck) {
+        } else if ((loopNumber == 2) && !relevantCheck) {
             // We have gone through the two loops and we still have some unknown
             // variables, so we consider as initialised those that have been
             // marked as external and we go through the two loops one more time.
@@ -2552,7 +2552,7 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
             }
 
             if (hasExternalVariables) {
-                ++loopNb;
+                ++loopNumber;
 
                 relevantCheck = true;
                 checkNlaSystems = false;
--- a/src/analyser.cpp
+++ b/src/analyser.cpp
@@ -2502,6 +2502,12 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
         }
     }
 
+    // Detmerine whether some variables have been marked as external.
+
+    auto hasExternalVariables = std::any_of(mInternalVariables.begin(), mInternalVariables.end(), [](const auto &iv) {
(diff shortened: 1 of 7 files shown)


agarny:
@hsorby and @nickerso: Weiwei said that she could get our test case to work without this fix by initialising that `P_3` parameter. So, one might think that we could consider all the variables that have been marked as external as being initialised, i.e. something like:

```c++
for (const auto &internalVariable : mInternalVariables) {
    if (internalVariable->mIsExternal) {
        internalVariable->mType = AnalyserInternalVariable::Type::INITIALISED;
    }
}

...

do {
    relevantCheck = false;
    for (const auto &internalEquation : mInternalEquations) {
        relevantCheck = internalEquation->check(mModel, stateIndex, variableIndex, checkNlaSystems)
                        || relevantCheck;
    }

    if (!relevantCheck && !checkNlaSystems) {
        relevantCheck = true;
        checkNlaSystems = true;
    }
} while (relevantCheck);

```
but this will fail for more "complicated" cases (e.g., `TEST(Generator, hodgkinHuxleySquidAxonModel1952WithComputedConstantAsExternalVariable)`). So, the fact that Weiwei's fix works for our test case is just a coincidence and we need the current fix to account for the general case.

Anyway, it's now ready for review.

src/analyser.cpp
```diff
@@ -2524,9 +2528,29 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
...
 
-        if (!relevantCheck && !checkNlaSystems) {
+        if (((loopNb == 1) || (loopNb == 3)) && !relevantCheck) {
+            ++loopNb;
+
             relevantCheck = true;
             checkNlaSystems = true;
+        } else if ((loopNb == 2) && !relevantCheck) {
+            // We have gone through the two loops and we still have some unknown
+            // variables, so we consider them as initialised and we go through
```
nickerso: should this mention external variables? i.e., `so we consider external variables as initialised and we go...`

agarny: No, but I agree that this could be made clearer, hence commit 31f3d228 (which also improves the analyser a bit).

nickerso (dismissed):

src/analyser.cpp
```diff
@@ -2502,17 +2502,27 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
...
+    //       After those two loops, if we still have some unknown variables and
+    //       they have been marked as external, then we consider them as
+    //       initialised and we go through the two loops one more time. This is
+    //       to account for models that have unknown variables (rendering the
+    //       model invalid) that have been marked as external (rendering the
+    //       model valid).
 
     auto stateIndex = MAX_SIZE_T;
     auto variableIndex = MAX_SIZE_T;
+    auto loopNb = 1;
```
hsorby: No super keen on this style of variable naming, but I knew what it was when I read the code so I guess it works.

hsorby (changes requested):
I think the documentation should mention that a variable marked as external is the responsibility of the user. It should also be made clear that an external variable is considered as initialised and again it is the responsibility of the user to make sure relevant conditions for its use are met.

My concern is if we don't do it now we will forget about it and I feel it is rather an important piece of information for users to have.

src/analyser.cpp
agarny: Happy to change it to something else that we are both happy with.

hsorby: `loopNumber`?

agarny: Fine by me: b618810f.

agarny:
> I think the documentation should mention that a variable marked as external is the responsibility of the user. It should also be made clear that an external variable is considered as initialised and again it is the responsibility of the user to make sure relevant conditions for its use are met.
> 
> My concern is if we don't do it now we will forget about it and I feel it is rather an important piece of information for users to have.

Fair point: 4d28f2a0.

hsorby (dismissed):

hsorby (approved):

nickerso (approved):

PR #1185, opened 2023-10-05
AnalyserVariable: hold onto the owning component reference.

### LINKED ISSUE #1183: Analyser: Not holding onto references
When analysing a model the analyser does not hold onto object references that it later needs when generating code. This results in generated code that cannot be simulated.

agarny:
Fixes #1183.

### DIFF
--- a/src/analyservariable.cpp
+++ b/src/analyservariable.cpp
@@ -16,9 +16,12 @@ limitations under the License.
 
 #include "libcellml/analyservariable.h"
 
+#include "libcellml/variable.h"
+
 #include <iterator>
 
 #include "analyservariable_p.h"
+#include "commonutils.h"
 
 namespace libcellml {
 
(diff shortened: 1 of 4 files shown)


src/analyservariable.cpp
```diff
@@ -16,9 +16,12 @@ limitations under the License.
 
 #include "libcellml/analyservariable.h"
 
+#include "libcellml/variable.h"
+
```
hsorby: This include should be below the system include.

hsorby (changes requested):

src/analyservariable.cpp
agarny: f24e786c

nickerso (approved):
looks good to me, assuming CI passes...

agarny:
> looks good to me, assuming CI passes...

Phew, it still does! :)

hsorby (approved):

PR #1212, opened 2024-03-05
Analyser: improve the reporting of unit-related issues with the pow() function

### LINKED ISSUE #1211: Analyser: improve the reporting of unit-related issues with the `pow()` function
For instance, if we have an equation like `y = pow(a, 1.0/b)` with everything dimensionless then we will get a warning that reads `The units in 'y = pow(a, 1.0/b)' in component 'my_component' are not equivalent. 'y' is 'dimensionless' while 'pow(a, 1.0/b)' is 'dimensionless'.`

FWIW, for something like `y = pow(a, 1.0/b)`, we should still generate some kind of a warning since `b` might be a computed variable.

However, for something like `y = pow(a, sin(3.0))` (with `y` and `a` in `second` and `3.0` `dimensionless`), we should generate a warning that reads something like `The units in 'y = pow(a, sin(3.0))' in component 'my_component' are not equivalent. 'y' is 'dimensionless' while 'pow(a, sin(3.0))' is 'second^0.14112'.` while we are currently getting `The units in 'y = pow(a, sin(3.0))' in component 'main' are not equivalent. 'y' is in 'second' while 'pow(a, sin(3.0))' is 'dimensionless'.`

agarny:
Fixes #1211.


hsorby (dismissed):

src/analyser.cpp
```diff
@@ -2235,59 +2257,66 @@ void Analyser::AnalyserImpl::analyseEquationUnits(const AnalyserEquationAstPtr &
...
                     unitsMultipliers = rightUnitsMultipliers;
 
-                    powerRootValue = powerValue(ast->mPimpl->mOwnedLeftChild, noPowerValueOrPowerValueConstant, powerValueAst);
+                    powerData.mExponentValue = powerValue(ast->mPimpl->mOwnedLeftChild, powerData);
                 } else {
                     // No DEGREE element, which means that we are dealing with a
                     // square root.
 
-                    powerRootValue = 2.0;
+                    powerData.mExponentValue = 2.0;
```
hsorby, line 2312: Can someone save my some thinking and tell my why this is a `2.0` and not a `0.5` please.

src/generator_p.h
```diff
@@ -109,7 +109,7 @@ struct Generator::GeneratorImpl
 
     std::string generateMethodBodyCode(const std::string &methodBody) const;
 
-    std::string generateDoubleCode(const std::string &value) const;
+    static std::string generateDoubleCode(const std::string &value);
```
hsorby: As an alternative this method could be made a function and then the `Analyzer` class wouldn't need to be a friend of the `Generator class`.

src/analyser.cpp
agarny: Because we are dealing with a `root` element when we reach that point and if there is no `degree` element then it means that we want the square root of something, hence the "exponent" value being 2.0.

src/generator_p.h
agarny: Agreed. See 35c8b0c7.

agarny:
@hsorby and @nickerso: ready to be re-reviewed. FWIW, yesterday, we talked about adding a reference rule, but I don't think there is a need for that. There is already a special reference rule (`Issue::ReferenceRule::ANALYSER_UNITS`) for all units-related issues found by the analyser.

hsorby (approved):

nickerso (approved):

PR #1220, opened 2024-04-07
Make it possible to interpret a model

### LINKED ISSUE #1218: Make it possible to interpret a model (initially algebraic and ODE models)
Depends on issue #1260.

---

In some cases (e.g., from WebAssembly), to generate some code with the view of compiling it on the fly is next to impossible at this stage (at least, using LLVM+Clang; see https://github.com/opencor/libopencor/issues/223). For this reason, it would be nice to have a way to interpret a model. This could (greatly) take advantage of what has been done for the `Generator`, especially when it comes to which equation should be evaluated and when.

---

Technically speaking, the interpreter should probably not be part of libCellML but rather libOpenCOR. However, to have it in libOpenCOR would require duplicating most of the current `Generator` code, which would be a very bad idea. So, so far, the plan has been on refactoring the `Generator` code so that it can not only generate C and Python code, but also generate the structure needed to interpret a model.

agarny:
Fixes #1218.


agarny:
We don't need an interpreter for libCellML anymore, so closing (see https://github.com/opencor/libopencor/issues/468 for more details).

PR #1234, opened 2024-05-07
Add Codespell GitHub action

### LINKED ISSUE #1233: Add spell checking test
Use GitHub actions to spell check the codebase.

hsorby:
Fixes #1233.


nickerso (dismissed):

.github/workflows/spell-check.yml
```diff
@@ -0,0 +1,21 @@
...
+  contents: read
+
+jobs:
+  codespell:
+    if: github.repository == 'cellml/libcellml'
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: codespell-project/actions-codespell@94259cd8be02ad2903ba34a22d9c13de21a74461 # v2.0
```
agarny: Why use (old) specific commits for those actions? Why not simply use:
```
- uses: actions/checkout@v4.1.5
- uses: codespell-project/actions-codespell@v2.0
```

agarny (commented):
Looks fine if we ignore my comment about the specific commits being used for the two actions.

.github/workflows/spell-check.yml
hsorby: Because `@v4.1.5` is not permanent, whereas the SHA of the commit is. Some people suggest that this is best practice for using actions.  For actions/checkout I usually just use `@v4`

agarny: Actually, yes, I would be using `@v4` rather than a specific version, but definitely not a specific commit as is the case here.

nickerso: the usage instructions (https://github.com/marketplace/actions/codespell-with-annotations) are just using `@v2` rather than the SHA...

agarny (approved):

nickerso (approved):

PR #1252, opened 2024-10-15
API: simplify/improve the API for external variables

### LINKED ISSUE #1251: API: simplify/improve the API for external variables
While working on issue #1244, I can see that some of the API for external variables could be improved:
- `bool addExternalVariable(const AnalyserExternalVariablePtr &externalVariable)`
  An `AnalyserExternalVariablePtr` is currently needed, which is needed when an external variable has some dependencies, but not if it doesn't have any. In the latter case, we should be able to add an external variable using just a `VariablePtr`.
  - [x] Add `bool addExternalVariable(const VariablePtr &variable)`
- `bool removeExternalVariable(const ModelPtr &model, const std::string &componentName, const std::string &variableName)`
  Rather than passing a `ModelPtr`, a component name, and a variable name, we could simply pass a `VariablePtr`.
  - [x] Replace with `bool removeExternalVariable(const VariablePtr &variable)`
- `bool containsExternalVariable(const ModelPtr &model, const std::string &componentName, const std::string &variableName)`
  Same as above.
  - [x] Replace with `bool containsExternalVariable(const VariablePtr &variable)`
- `AnalyserExternalVariablePtr externalVariable(const ModelPtr &model, const std::string &componentName, const std::string &variableName)`
  Similar to above.
  - [x] Replace with `AnalyserExternalVariablePtr externalVariable(const VariablePtr &variable)`

agarny:
Fixes #1251.


nickerso (dismissed):

src/analyser.cpp
```diff
@@ -3405,6 +3397,19 @@ void Analyser::analyseModel(const ModelPtr &model)
     }
 }
 
+bool Analyser::addExternalVariable(const VariablePtr &variable)
+{
+    for (const auto &externalVariable : pFunc()->mExternalVariables) {
+        if (externalVariable->variable() == variable) {
+            return false;
+        }
+    }
```
hsorby: Could this be:

if (containsExternalVariable(variable) {
  return false;
}

?

agarny: Well spotted! Fixed in 590d5e7a9 (as well as for the next `addExternalVariable()` method).

src/analyser.cpp
```diff
@@ -3308,16 +3306,10 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
...
-                                                                                          const std::string &variableName) const
+AnalyserExternalVariablePtrs::const_iterator Analyser::AnalyserImpl::findExternalVariable(const VariablePtr &variable) const
 {
     return std::find_if(mExternalVariables.begin(), mExternalVariables.end(), [=](const auto &ev) {
-        auto variable = ev->variable();
-
-        return (owningModel(variable) == model)
-               && (owningComponent(variable)->name() == componentName)
-               && (variable->name() == variableName);
+        return ev->variable() == variable;
```
hsorby, line 3312: It is now no longer a requirement that an external variable needs to be from the same model?

agarny: Not so much that but rather the fact that before our API required a model, a component name, and a variable name to identify a variable and then check whether it was associated to one of our external variables.

Now, we directly pass a variable, which means that we can now directly check whether that variable is associated with one of our external variables.

hsorby: What happens if I add an external variable from another model? Good things?

agarny: We have a test for that (see [here](https://github.com/agarny/libcellml/blob/issue1251/tests/analyser/analyser.cpp#L439-L472)).

hsorby:
I wouldn’t mind seeing the actual return value of the call to addExternalVariable being checked. I can’t see anywhere where this is being done. I see checks that look at the number of external variables after the fact.

agarny:
> I wouldn’t mind seeing the actual return value of the call to addExternalVariable being checked. I can’t see anywhere where this is being done. I see checks that look at the number of external variables after the fact.

Indeed, we do that for `removeExternalVariable()` and `containsExternalVariable()`. Not sure why I didn't for `addExternalVariable()` but it is now done in 42a0a2edc.

Not sure why, but our codespell test is now failing? See https://github.com/cellml/libcellml/actions/runs/17993176962/job/51187518022: `401 Unauthorized`!?

hsorby:
Probably an expired token. No, not an expired token we don’t currently use one for DockerHub.

hsorby (approved):

tests/coverage/coverage.cpp
```diff
@@ -811,7 +811,7 @@ TEST(Coverage, generator)
 
     // Coverage for various profile settings.
 
-    analyser->addExternalVariable(libcellml::AnalyserExternalVariable::create(model->component("my_component")->variable("eqnEq")));
+    EXPECT_TRUE(analyser->addExternalVariable(libcellml::AnalyserExternalVariable::create(model->component("my_component")->variable("eqnEq"))));
```
nickerso, line 814: given all the tests added to the analyser, is this still needed?

nickerso (commented):
Do we really need to be testing every instance of `addExternalVariable`? Doesn't seem to be what we do in general...

tests/coverage/coverage.cpp
agarny: We currently test every instance of `removeExternalVariable()` and `containsExternalVariable()`. So, was merely being consistent.

nickerso: in case it got missed, the coverage question here is different to the more general comment...

agarny: Ok, if you want, I can remove this particular instance then. Let me know. :)

nickerso: huh? its not what I want - is it needed for coverage? thats the question I'm asking. We only use coverage tests to pick up things that are specifically needed to ensure 100% coverage which can't conveniently be included in the regular tests...at least that is my understanding :)

hsorby: The coverage tests should only be ones that are anomalous, things where we have to cast to `const`, where we are forcing a state simply for coverage’s sake.

agarny: Oops, sorry, I thought you were wondering about the need for `EXPECT_TRUE()` to ensure that `addExternalVariable()` returned true.

All I can say for sure is that the call to `addExternalVariable()` in this coverage test was needed at some point.

Ok, I have just given it a try and, yup, it's still needed. Without it, we are not covering the `false` branch of [this `if` statement](https://github.com/agarny/libcellml/blob/issue1251/src/generator.cpp#L724).

nickerso (approved):

Issue #1056, opened 2022-11-22
Change Anaylser API: model() -> analyserModel()


hsorby:
```
   /**
     * @brief Get the analysed model.
     *
     * Get the analysed model that is the result of the model analysis. This
     * allows to retrieve some information about the type of the model, its
     * variable of integration, states, variables, equations, and whether it
     * needs some specific mathematical functions.
     *
     * @return The analysed model for the @c Model analysed by this @ref Analyser.
     */
    AnalyserModelPtr model() const;
```
The `Analyser` class has the method `model()` which returns an `AnalyserModel`.  I think the method should be called `analyserModel()`.

agarny:
Closing as it has since been done.

PR #1256, opened 2024-10-16
Code generator: allow for variables to be tracked/untracked

### LINKED ISSUE #1244: Code generator: allow for variables to be tracked/untracked
Depends on issue #1243.

---

See issue #1242 for some background discussion. Also, this issue requires issue #1243 to be implemented first.

Note that in the case of a differential model, the VOI and the states are **_always_** to be tracked.

agarny:
Fixes #1244.


src/api/libcellml/analyserequation.h
```diff
@@ -29,23 +29,23 @@ namespace libcellml {
...
     enum class Type
     {
-        TRUE_CONSTANT, /**< An equation that defines a true constant, e.g. x = 3. */
-        VARIABLE_BASED_CONSTANT, /**< An equation that describes a variable-based constant, e.g. x = y+z where y and z are true constants. */
-        ODE, /**< An equation that describes an ordinary differential equation, e.g. d(y)/dt = f(t, x). */
-        NLA, /**< An equation that is used to compute one or several algebraic variables, e.g. x+y = f(z) where x and y are known, but not z. */
-        ALGEBRAIC, /**< An equation that is used to compute an algebraic variable, e.g. y = f(x). */
-        EXTERNAL /**< A placeholder equation to capture the dependency of the model on an external variable. */
+        CONSTANT, /**< An equation that computes a (true) constant (e.g. x = 3). */
+        COMPUTED_CONSTANT, /**< An equation that computes a computed constant (e.g., x = 3+5, x = 3+z, x = y+z where y and z are constants). */
```
hsorby: Are y and z true constants or themselves computed constants?

src/api/libcellml/analyservariable.h
```diff
@@ -120,33 +120,42 @@ class LIBCELLML_EXPORT AnalyserVariable
...
+     * @brief Get the @ref AnalyserModel for this @ref AnalyserVariable.
      *
-     * Return the number of equations used to compute this @ref AnalyserVariable.
+     * Return the @ref AnalyserModel for this @ref AnalyserVariable.
      *
-     * @return The number of equations used to compute this @ref AnalyserVariable.
+     * @return The @ref AnalyserModel.
      */
-    size_t equationCount() const;
+    AnalyserModelPtr model() const;
```
hsorby: For consistency reasons should this not be `analyserModel()`?

src/api/libcellml/generator.h
```diff
@@ -68,49 +65,245 @@ class LIBCELLML_EXPORT Generator
...
+
+    /**
+     * @brief Track all the algebraic variables in the given @p analyserModel.
+     *
+     * Track all the algebraic variables in the given @p analyserModel. This will add all the algebraic variables in the
+     * analyser model to the list of tracked variables.
+     *
+     * @param analyserModel The pointer to the @ref AnalyserModel which all the algebraic variables are to be tracked.
+     */
+    void trackAllAlgebraic(const AnalyserModelPtr &analyserModel);
```
hsorby: Should all of these be like `trackAllAlgebraicVariables()` is that more consistent with the previous PR?

src/analyserequation.cpp
```diff
@@ -29,20 +29,26 @@ AnalyserEquationPtr AnalyserEquation::AnalyserEquationImpl::create()
     return std::shared_ptr<AnalyserEquation> {new AnalyserEquation {}};
 }
 
-bool AnalyserEquation::AnalyserEquationImpl::isEmptyDependency(const AnalyserEquationWeakPtr &dependency)
+bool AnalyserEquation::AnalyserEquationImpl::isDummyDependency(const AnalyserEquationWeakPtr &dependency)
```
hsorby: Are these more accurately called staging dependencies? We are staging a dependency for future analysis?

src/generator.cpp
```diff
@@ -42,62 +44,510 @@ void Generator::GeneratorImpl::reset()
...
+
+        if (analyserVariable == var) {
+            break;
+        }
+    }
+
+    return convertToString(res);
+}
+
+bool Generator::GeneratorImpl::doIsTrackedEquation(const AnalyserEquationPtr &analyserEquation, bool tracked)
```
hsorby: I’m not sure that this function requires the ‘do’ qualifier. It is not virtual and overloaded parameter list will differentiate it.

src/generator.cpp
```diff
@@ -42,62 +44,510 @@ void Generator::GeneratorImpl::reset()
...
+                                ((specialAnalyserVariable->type() == AnalyserVariable::Type::STATE) ?
+                                     "a state variable" :
+                                     "an external variable");
+
+        issue->mPimpl->setDescription("Variable '" + analyserVariable->variable()->name()
+                                      + "' in component '" + owningComponent(analyserVariable->variable())->name()
+                                      + "' is " + variableType + " and "
+                                      + (tracked ?
+                                             "is therefore always tracked." :
+                                             "cannot therefore be untracked."));
```
hsorby: I have seen this bit of code a few times, is it worth a little function to stop the repeats?

src/generator_p.h
```diff
@@ -31,22 +33,77 @@ std::string generateDoubleCode(const std::string &value);
...
+    std::map<AnalyserModelPtr, std::map<AnalyserVariablePtr, bool>> mTrackedVariables;
+
     void reset();
 
-    bool modelHasOdes() const;
-    bool modelHasNlas() const;
+    std::string analyserVariableIndexString(const AnalyserModelPtr &analyserModel,
+                                            const AnalyserVariablePtr &analyserVariable);
+
+    bool doIsTrackedEquation(const AnalyserEquationPtr &analyserEquation, bool tracked);
```
hsorby: Not a virtual function and parameter overloading will differentiate this method.

src/generator_p.h
```diff
@@ -31,22 +33,77 @@ std::string generateDoubleCode(const std::string &value);
...
+
+    bool isTrackedEquation(const AnalyserEquationPtr &analyserEquation);
+    bool isUntrackedEquation(const AnalyserEquationPtr &analyserEquation);
+
+    bool doIsTrackedVariable(const AnalyserModelPtr &analyserModel, const AnalyserVariablePtr &analyserVariable,
+                             bool tracked = true);
+    bool doIsTrackedVariable(const AnalyserVariablePtr &analyserVariable, bool tracked);
```
hsorby: Not a virtual function and parameter overloading will differentiate these methods.

src/generator_p.h
```diff
@@ -31,22 +33,77 @@ std::string generateDoubleCode(const std::string &value);
...
+
+    bool isTrackedVariable(const AnalyserVariablePtr &analyserVariable);
+    bool isUntrackedVariable(const AnalyserVariablePtr &analyserVariable);
```
hsorby: Not sure these are needed if `doIsTrackedVariable(varaible, false)` can be called from the interface class.

src/generator_p.h
```diff
@@ -31,22 +33,77 @@ std::string generateDoubleCode(const std::string &value);
...
+                         Issue::ReferenceRule trackedReferenceRule, Issue::ReferenceRule untrackedReferenceRule);
+
+    void trackVariable(const AnalyserVariablePtr &analyserVariable, bool tracked, bool needRemoveAllIssues = true);
+
+    void trackVariable(const AnalyserVariablePtr &analyserVariable);
+    void untrackVariable(const AnalyserVariablePtr &analyserVariable);
+
+    void trackVariables(const std::vector<AnalyserVariablePtr> &analyserVariables, bool tracked);
+
+    bool validModel(const AnalyserModelPtr &analyserModel);
```
hsorby: `validAnalyserModel()`?

src/generator_p.h
```diff
@@ -31,22 +33,77 @@ std::string generateDoubleCode(const std::string &value);
...
+
+    void trackAllAlgebraic(const AnalyserModelPtr &analyserModel);
+    void untrackAllAlgebraic(const AnalyserModelPtr &analyserModel);
+
+    std::vector<AnalyserVariablePtr> trackableVariables(const AnalyserModelPtr &analyserModel) const;
+
+    void trackAllVariables(const AnalyserModelPtr &analyserModel);
+    void untrackAllVariables(const AnalyserModelPtr &analyserModel);
+
+    size_t doTrackedVariableCount(const AnalyserModelPtr &analyserModel,
```
hsorby: The `do` prefix is unnecessary here?

hsorby (commented):
What is the thinking behind always passing an analyser model to the generator class, as opposed to setting the analyser model for the generator to work on?

src/api/libcellml/analyserequation.h
agarny: They can be either a true or a computed constant. We have tests that for both (see [here](https://github.com/cellml/libcellml/blob/main/tests/resources/generator/algebraic_eqn_computed_var_on_rhs/model.c#L64-L72) for instance).

But, fair enough, I have made it clear that it can be either. Done in 5b6dc32b0.

src/api/libcellml/analyservariable.h
agarny: Agreed! I did spend a fair amount of time on this kind of stuff yesterday, but it could clearly be improved even further. Anyway, done in c7aa88bb1.

src/api/libcellml/generator.h
agarny: Well spotted! Done in ab86f4b5c.

src/analyserequation.cpp
agarny: Sure. Done in e887093c3.

src/generator.cpp
agarny: It doesn't. I had a few `doXxx()` methods. I remember dealing with one of them but then clearly forgot about the remaining ones! Anyway, done in cd2772d18.

src/generator.cpp
agarny: Fair enough. Done in 88967565a.

src/generator_p.h
agarny: Done in cd2772d18.

src/generator_p.h
agarny: Done in cd2772d18.

src/generator_p.h
agarny: Agreed, they (and other similar ones) are not needed _per se_. Done in 5ec5f4b0e.

src/generator_p.h
agarny: Agreed. Done in eb2e2323b.

src/generator_p.h
agarny: Done in cd2772d18.

hsorby:
Other than this question:

What is the thinking behind always passing an analyser model to the generator class, as opposed to setting the analyser model for the generator to work on?


I am happy with this work.

agarny:
> What is the thinking behind always passing an analyser model to the generator class, as opposed to setting the analyser model for the generator to work on?

Oops, I went through the inline comments completely overlooking this one.

Anyway, are you talking about methods like `un/trackAllConstants()`, `un/trackAllComputedConstants()`, etc. that take an analyser model as an input? If so, it's because a generator can be reused with different analyser models. This is why we don't allow the creation of a generator for a given analyser model.

hsorby:
You are asking the Generator to hold a map of analysermodels and their map of tracked and untracked variables. The Generator should really be stateless. I have three problems with this approach:

1. mTrackedVariables will never shrink while the generator object is alive
2. This is not thread safe
3. The genereator is responsible for the lifecycle management of the mTrackedVariables when that is not its concern

For me the tracked varaibles belong with the AnalyserModel itself. Alternatively, you could create a generation context that holds the tracked variable information, and generator profile and you pass the AnalyserModel and GenerationContext when asking the generator to generate. This also means that the Generator doesn’t hold a reference to a generator profile this is passed in when the generation happens. For me this would be a massive archetictural improvement for the library.

agarny:
The un/tracking of variables is only relevant in the context of code generation, which is why `mTrackedVariables` is managed by the generator. To have `mTrackedVariables` managed by the analyser model would prevent the analyser model from being used to generate code with different un/tracked variables, so I don't think the analyser model should manage `mTrackedVariables`.

> You are asking the Generator to hold a map of analysermodels and their map of tracked and untracked variables. The Generator should really be stateless. I have three problems with this approach:
> 
> 1. mTrackedVariables will never shrink while the generator object is alive

Correct, although we could always a method to reset `mTrackedVariables`, if needed.

> 2. This is not thread safe

True, but then again the generator is not thread-safe to start with.

> 3. The genereator is responsible for the lifecycle management of the mTrackedVariables when that is not its concern
> 
> For me the tracked varaibles belong with the AnalyserModel itself. Alternatively, you could create a generation context that holds the tracked variable information, and generator profile and you pass the AnalyserModel and GenerationContext when asking the generator to generate. This also means that the Generator doesn’t hold a reference to a generator profile this is passed in when the generation happens. For me this would be a massive archetictural improvement for the library.

I agree, a better approach would be to have a new class (`GeneratorContext`) that manages un/tracked variables for a given analyser model. Kind of similar to what was done with `GeneratorProfile`, which means that we would then have `Generator::context()` and `Generator::setContext(const GeneratorContextPtr &context)`.

However, I would do that as part of a different GitHub issue. Time is pressing.

hsorby (approved):

hsorby:
This PR also:

* Fixes #1056.

nickerso (approved):

Issue #1253, opened 2024-10-15
CI: add some "basic" GitHub Actions


agarny:
Our Buildbot has been "down" for weeks (months?), making it unnecessarily difficult to test our PRs. Our GitHub Actions should include the following:
- [x] Code formatting (Linux; target `test_clang_format`)
- [x] Documentation (Linux; target `docs`)
- [x] Code coverage (macOS; target `llvm_coverage`)
- [x] Memory leaks (Linux; target `memcheck`)
- C++ (Windows, Linux, and macOS; target `test` --  both static and shared libraries)
  - [x] Windows static library
  - [x] Windows shared library
  - [x] Linux static library
  - [x] Linux shared library
  - [x] macOS static library (Intel)
  - [x] macOS shared library (Intel)
  - [x] macOS static library (ARM)
  - [x] macOS shared library (ARM)
- [x] Python (Windows, Linux, and macOS; target `test`)
- [x] JavaScript (macOS; target `jest_test`)

PR #1258, opened 2024-10-17
LLVM coverage: don't use hints to find llvm-cov and llvm-profdata and let the user know where the report can be found

### LINKED ISSUE #1257: LLVM coverage: don't use hints to find llvm-cov and llvm-profdata and let the user know where the report can be found
Someone might have a version of `llvm-cov` and `llvm-profdata` that they use and which is not located in `/Library/Developer/CommandLineTools/usr/bin`. This means that, right now, they would need to provide `LLVM_BIN_DIR` which is not ideal, not least since there are `PREFERRED_LLVM_COV_NAMES` and `PREFERRED_LLVM_PROFDATA_NAMES` to play with, if needed.

agarny:
Fixes #1257.


agarny:
This is now done (and improved) as part of issue #1253, so closing.

PR #1261, opened 2024-10-21
Reorganise our code in preparation for our work on the interpreter

### LINKED ISSUE #1260: Reorganise our code in preparation for our work on the interpreter
Depends on issue #1244.

---

Work on issue #1218 (see PR #1220) requires quite a bit of reworking of our code. For instance, we want to reuse several parts of the generator so that it can be both used to generate code (as before) and "interpreted" code.

agarny:
Fixes #1260.


agarny:
We don't need an interpreter for libCellML anymore, so closing (see https://github.com/opencor/libopencor/issues/468 for more details).

PR #1305, opened 2024-12-05
Reverted some "cleaning up" of the analyser.

agarny:
Effectively reverting 4357d2b84. Indeed, the code is less efficient: there are more `if` statements and we make many calls (~94% of the time) to AnalyserEquationAst::swapLeftAndRightChildren(). Also, if we were to do that cleaning up then we also need to clean in the case of a piecewise statement.


nickerso (approved):

agarny:
Closing since this is likely never going to be approved, even though that original "cleaning" is pointless, less efficient, and incomplete (since not done for piecewise statements), not to mention that I have other PRs that need to be merged in.

PR #1315, opened 2025-06-24
CI

agarny:


agarny:
Eh, @hsorby, you are hijacking my PR!

Anyway, for some reasons, everything is working fine, except for the Python tests on Windows. Not sure what is going on there. Well, actually, the problem is that Python cannot find the (libCellML?) DLL. For instance:
```
67: ======================================================================
67: ERROR: test_create_destroy (__main__.ParserTestCase.test_create_destroy)
67: ----------------------------------------------------------------------
67: Traceback (most recent call last):
67:   File "C:\Users\Alan\Desktop\libCellML\build\tests\bindings\python\test_parser.py", line 10, in test_create_destroy
67:     from libcellml import Parser
67:   File "C:\Users\Alan\Desktop\libCellML\build\src\bindings\python\libcellml\__init__.py", line 12, in <module>
67:     from libcellml.analyser import Analyser
67:   File "C:\Users\Alan\Desktop\libCellML\build\src\bindings\python\libcellml\analyser.py", line 10, in <module>
67:     from . import _analyser
67: ImportError: DLL load failed while importing _analyser: The specified module could not be found.
```

Otherwise, regarding https://github.com/cellml/libcellml/pull/1315/commits/6e93bbe3f582d7f5b2ade44944a882236592362d, I purposedly didn't do to be version agnostic. If you were to find libxml2 and zlib in config mode then we would need to update the version of libxml2 and zlib both in .github/workflows/ci.yml and in https://github.com/cellml/gha/blob/main/.github/workflows/cd.yml while I would only worry about the version of libxml2 and zlib in the latter.

hsorby:
I have made changes to the libxml2 library to work (maybe) for Windows. It needs updating on the cellml/gha repository. I don't have access to that repo to do anything myself.

The config mode is needed to run the Python tests on Windows, so the path for the tests can be set correctly.
I was first trying to see if this would fix the testing error and then consider if this is a good solution or not.

hsorby (approved):
This PR looks good to me and addresses a long-standing issue with zlib and libxml2 on Windows (by making me fix them so that they actually work as intended).

hsorby:
Re: version numbers in ci.yml.
Finding the zlib and libxml2 libraries in CMake config mode is preferred for running the tests. While specifying the version in the path is not ideal, it is better than finding the libraries in module mode and then manually dealing with the associated DLLs. With this change, we are closer to what we expect a user or developer to set up.

agarny:
> Re: version numbers in ci.yml. Finding the zlib and libxml2 libraries in CMake config mode is preferred for running the tests. While specifying the version in the path is not ideal, it is better than finding the libraries in module mode and then manually dealing with the associated DLLs. With this change, we are closer to what we expect a user or developer to set up.

Agreed.

The only downside is that the paths for `LibXml2_DIR` and `ZLIB_DIR` currently include the version of those libraries, which means that it will break if we ever decide to use a different version of our prebuilt binaries for libxml2 and zlib.

So, any chance we could get rid of the version information in those paths? Not a biggy, just a nice-to-have.

nickerso (approved):

PR #1325, opened 2025-09-16
Allow a constant, a computed constant, or a state variable (but not an algebraic variable) to be used as an initial value

### LINKED ISSUE #1322: Allow a constant, a computed constant, or a state variable (but not an algebraic variable) to be used as an initial value
[And now in the right place!]

In the attached file (rename it .cellml), I attemt to initialise a state variable with a computed constant variable i.e., something along the lines of

    var time: second {pub:in};
    var state: second {pub: in};
    var initial_value: dimensionless;
    initial_value = 2.0{per_second}*state;
    var state_variable: dimensionless {init: initial_value};
    ode(state_variable,time) = 0.005{per_second} * state_variable;

This was previously possible in the old OpenCOR (and, as I understand it, you can also use state variables to initialise other state variables). However, in the new OpenCOR web app (version 0.20250714.0) I get the error

"Variable 'state_variable' in component 'test_component' is initialied using variable 'initial_value' which is not a constant."

It would be useful to allow state variables to by initialised by computed constant (and other state variables) variables as well as constant variables as in the old OpenCOR.

[test_initial.txt](https://github.com/user-attachments/files/21422683/test_initial.txt)

agarny:
Fixes #1322.


src/analyser.cpp
```diff
@@ -2339,8 +2328,13 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
     }
 
     // Recursively analyse the model's components' variables.
-    // Note: we can't do this as part of analyseComponent() since we don't
-    //       necessarily know the state of all the variables.
+    // Note #1: we can't do this as part of analyseComponent() since we don't
+    //          necessarily know the state of all the variables.
+    // Note #2: when it comes to variables initialised using another variable,
```
hsorby: Capitol W for When

src/generator_p.h
```diff
@@ -188,19 +187,34 @@ struct Generator::GeneratorImpl: public Logger::LoggerImpl
...
+                                                             std::vector<AnalyserVariablePtr> &remainingStates,
+                                                             std::vector<AnalyserVariablePtr> &remainingConstants,
+                                                             std::vector<AnalyserVariablePtr> &remainingComputedConstants,
+                                                             std::vector<AnalyserVariablePtr> &remainingAlgebraic);
     void addImplementationComputeRatesMethodCode(const AnalyserModelPtr &analyserModel,
-                                                 std::vector<AnalyserEquationPtr> &remainingAnalyserEquations);
+                                                 std::vector<AnalyserEquationPtr> &remainingEquations);
     void addImplementationComputeVariablesMethodCode(const AnalyserModelPtr &analyserModel,
-                                                     std::vector<AnalyserEquationPtr> &remainingAnalyserEquations);
+                                                     std::vector<AnalyserEquationPtr> &remainingEquations);
```
hsorby: This seems an odd reduction, contrary  to typical parameter naming.

tests/analyser/analyser.cpp
```diff
@@ -255,22 +255,29 @@ TEST(Analyser, nonConstantInitialisingVariable)
...
+
     auto parser = libcellml::Parser::create();
-    auto model = parser->parseModel(fileContents("analyser/non_existing_initialising_variable.cellml"));
+    auto model = parser->parseModel(fileContents("analyser/variable_initialised_using_another_variable.cellml"));
 
     EXPECT_EQ(size_t(0), parser->issueCount());
 
     const std::vector<std::string> expectedIssues = {
-        "Variable 'x' in component 'main' has an invalid initial value 'k'. Initial values must be a real number string or a variable reference.",
+        "Variable 'kStateStateAlgebraic' in component 'main' is initialised using variable 'kStateAlgebraic', which is an algebraic variable. Only a reference to a constant, a computed constant, a state variable, or a non-linear algebraic variable is allowed.",
```
hsorby: Wait so you can initialise with a non-linear algebraic variable but no a linear one?

src/generator.cpp
```diff
@@ -2242,125 +2354,127 @@ void Generator::GeneratorImpl::addInterfaceComputeModelMethodsCode(const Analyse
...
-        // initial value of 3).
+    for (const auto &equation : analyserModel->analyserEquations()) {
+        if (equation->type() == AnalyserEquation::Type::CONSTANT) {
+            methodBody += generateEquationCode(analyserModel, equation, remainingAnalyserEquations, generatedConstantDependencies);
+        }
+    }
 
-        std::vector<AnalyserVariablePtr> generatedConstantDependencies;
+    // Initialise our algebraic variables that have an initial value. Also use an initial guess of zero for algebraic
+    // variables computed using an NLA system.
```
hsorby, line 2406: I thought we weren’t doing this? I thought the initial value was the initial guess?

src/generatorprofile.cpp
```diff
@@ -489,11 +489,17 @@ void GeneratorProfile::GeneratorProfileImpl::loadProfile(GeneratorProfile::Profi
                                                          "[CODE]"
                                                          "}\n";
 
-        mInterfaceComputeComputedConstantsMethodString = "void computeComputedConstants(double *constants, double *computedConstants);\n";
-        mImplementationComputeComputedConstantsMethodString = "void computeComputedConstants(double *constants, double *computedConstants)\n"
-                                                              "{\n"
-                                                              "[CODE]"
-                                                              "}\n";
+        mInterfaceComputeComputedConstantsMethodFamString = "void computeComputedConstants(double *constants, double *computedConstants, double *algebraic);\n";
```
hsorby: Is the Fam in this name For Analyser Model? What about using Diff and NonDiff if looking to reduce length.

src/generator_p.h
agarny: It must have been the indirect result of merging things since even I wouldn't have done such a thing! Anyway, fixed in ec4c74213 (as well as other similar issues).

tests/analyser/analyser.cpp
agarny: The wording of the message is probably not the best, but what I meant to say is that we can't initialise using an algebraic variable while it's ok to initialise using a non-linear algebraic (NLA) variable that is computed using a system of NLA equation(s) (i.e. a system solved using a solver such as KINSOL). Happy to reword the message if someone has a better way to say what I meant.

src/generator.cpp
agarny: That is correct... for "proper" NLA variables that need to be computed using a "proper" NLA system. In such a case, we MUST use the initial value as the initial guess.

Here, I am talking about the case where we have an equation like `x = y + z` with `x` and `y` known and `z` unknown. The analyser doesn't do any rearranging of equations, which means that we (currently) expect the unknown variable to be on its own on either the LHS or RHS of the equation. In the current case, the analyser would have like z to be expressed as `z = x - y` or `x - y = z`. However, this is not the case here, so the analyser decides to solve for `z` using an NLA system with one NLA equation (`x = y + z`) for which the objective function is `f = x - (y + z)` and for which the initial guess of `z` is zero, which (I think!) is fine in THIS particular case.

src/generatorprofile.cpp
agarny: It has been here for a very long time, but still `Fam` stands for "for algebraic models" while `Fdm` stands for "for differential models". However, I like your suggestion of `Diff` (instead of `Fdm`) and `NonDiff` (instead of `Fam`), so done in 2dd9fcc33.

tests/analyser/analyser.cpp
hsorby: I think I would add the ‘computable' word in before the non-linear algebraic variable is allowed.

agarny: Sounds good to me. Done in 12825421e.

hsorby (approved):
Looks good to me.

nickerso (approved):

PR #1327, opened 2025-09-25
Replace DockerHub based codespell with local variant

### LINKED ISSUE #1326: Codespell CI getting rate limited
It has been suggested by AI that the sporadic failure of codespell is a rate limit problem for getting images from dockerhub.
The solution is to add a PAT so the rate limit is personalised and shouldn’t be hit by libCellML typical (or even atypical ) usage.

hsorby:
* Fixes #1326


hsorby:
Instead of using a DockerHub based running of codespell this PR replaces it with a local variant which performs the same job.
It adds a codespell dictionary for any additional words that we might wish to add. It also renames the .codespellexclude to .codespell-exclude for consistency.
Shouldn’t make any difference to the output and should run faster.

hsorby:
Now of course it can no longer pass the ‘Check for spelling errors’ check as it no longer exists and is replaced by ‘Check Spelling / codespell’.

nickerso:
> Now of course it can no longer pass the ‘Check for spelling errors’ check as it no longer exists and is replaced by ‘Check Spelling / codespell’.

so we should be reviewing and merging this one without waiting for that test to pass?

hsorby:
Yes, in essence. I can change the required check but then that would change it for all PRs.  In this case, you need to make sure that the ‘Check Spelling / codespell’ check is passing in lieu of the currently required ‘Check for spelling errors’.

hsorby:
Closing this as the associated issue has been closed.

PR #1334, opened 2025-10-30
Syntax error in generated Python code

### LINKED ISSUE #1333: Syntax error in generated python code
When generating Python code from a CellML model using libCellML, parameter values expressed in scientific notation (e.g. **6E-1**) are incorrectly rendered in the generated code, resulting in a syntax error.

Image

Image

This CellML model can used to reproduce the error: [lotka-volterra.zip](https://github.com/user-attachments/files/23222320/lotka-volterra.zip)

**Note:** This can be temporarily fixed by changing the value to **6.0E-1** in the CellML model.

agarny:
Fixes #1333.

### DIFF
--- a/src/generator.cpp
+++ b/src/generator.cpp
@@ -885,7 +885,7 @@ std::string generateDoubleCode(const std::string &value)
         return value;
     }
 
-    auto ePos = value.find('e');
+    auto ePos = value.find_first_of("eE");
 
     if (ePos == std::string::npos) {
         return value + ".0";
--- a/tests/resources/coverage/generator/model.c
+++ b/tests/resources/coverage/generator/model.c
@@ -381,7 +381,7 @@ void initialiseVariables(double *states, double *rates, double *variables)
     variables[1] = 123.0;
     variables[5] = 1.0e1;
     variables[17] = 1.23e1;
-    variables[178] = 1E1.0;
+    variables[178] = 1.0E1;
     variables[179] = 1.23E1;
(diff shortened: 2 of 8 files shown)


hsorby (approved):
Big change! I approve.

nickerso (approved):
is there no standard method to convert a string to a number?

agarny:
> is there no standard method to convert a string to a number?

Yes, there is. Still, the idea is for the generator to generate "standardised" code. For instance, generate `1.0e-1` no matter whether the model contains `1e-1` or `1.0e-1`. This being said, we don't "standardise" the exponent. So, if the model contains `1E-1` or `1.0E-1`, the generator will (currently) generate `1.0E-1` (and not `1.0e-1`).

agarny:
I suppose that to "standardise" things, we could convert a string to a number and then back to a string.

hsorby:
WHat does `std::stod` do for us?

hsorby:
We have the utility `convertToDouble` that we should? use?

agarny:
Ok, it's slowly coming back to me. I did use `convertToDouble()` (or `std::stod()` directly) and then convert it back to a string (since that's what we need in the end), but that can yield some unexpected results (see the bottom of https://en.cppreference.com/w/cpp/string/basic_string/to_string), hence ended up doing what I am doing.

hsorby:
Hopefully at least one person remembers that we are handling doubles in this way so we can consistently print out the numbers in the code generation.

PR #1335, opened 2025-10-30
Analyser: ensure that a variable computed using an NLA system has an initial guess

### LINKED ISSUE #1332: Analyser: ensure that a variable computed using an NLA system has an initial guess
Right now, if variable `x` is to be computed using an NLA system then it will be initialised using the value of its `initial_value` attribute or `0` if it has no such attribute. However, this default initial guess of `0` may result in sloppy modelling practices. So, we should ensure that `x` has its `initial_value` attribute set if it is to be computed using an NLA system.

agarny:
Fixes #1332.


agarny:
The [original issue](https://github.com/cellml/libcellml/issues/1332) was a red herring, so closing this PR.

Issue #1339, opened 2025-11-20
Add a `GeneratorContext` class to manage un/tracked variables


agarny:
As part of reviewing PR #1256, @hsorby made a valid point about the fact that un/tracked variables shouldn't be managed by the `Generator` (see his original comment [here](https://github.com/cellml/libcellml/pull/1256#issuecomment-3555835780)). He suggested that it should be managed by `AnalyserClass`, but I don't agree with that. However, I do agree with his other option: have a new class that manages the context under which some code should be generated. IOW, we should have a `GeneratorContext` class to manage un/tracked variables in the same way that we have the `GeneratorProfile` class to manage the kind of code that should be generated.

hsorby:
We decided here that having a `GeneratorContext` class simply added an extra layer of requirement for using the `Generator`. And that it would be better for users of the library to be able to pass an `AnalyserModel` directly to the `Generator` as it is now. With the API presented in the PR we have a way to generate code at a variety of complexities with ease of access.

PR #1340, opened 2025-11-20
Add a `GeneratorVariableTracker` class to manage un/tracked variables

agarny:
Fixes issue #1339.


src/api/libcellml/generatorvariabletracker.h
```diff
@@ -0,0 +1,270 @@
...
+ *
+ * The GeneratorVariableTracker class is for representing a CellML GeneratorVariableTracker.
+ */
+class LIBCELLML_EXPORT GeneratorVariableTracker: public Logger
+{
+public:
+    ~GeneratorVariableTracker(); /**< Destructor, @private. */
+    GeneratorVariableTracker(const GeneratorVariableTracker &rhs) = delete; /**< Copy constructor, @private. */
+    GeneratorVariableTracker(GeneratorVariableTracker &&rhs) noexcept = delete; /**< Move constructor, @private. */
+    GeneratorVariableTracker &operator=(GeneratorVariableTracker rhs) = delete; /**< Assignment operator, @private. */
```
hsorby, line 34: These should all be private.

tests/bindings/javascript/generatorvariabletracker.test.js
```diff
@@ -0,0 +1,88 @@
...
+        const p = new libcellml.Parser(true)
+        const m = p.parseModel(hhSquidAxon1952)
+        const v = m.componentByName("membrane", true).variableByName("Cm")
+        const a = new libcellml.Analyser()
+
+        a.analyseModel(m)
+
+        const am = a.analyserModel()
+        const av = am.analyserVariable(v)
+        const gc = new libcellml.GeneratorVariableTracker()
```
hsorby: `gc` is an interesting choice.

agarny: Fixed.

src/api/libcellml/generatorvariabletracker.h
agarny: Why? None of our other classes does that or am I missing something?

hsorby: Now that you have made me think about it some more, we mark it private for the generation of the Doxygen documentation, not as a reminder that it should actually be private.

src/api/libcellml/generator.h
```diff
@@ -65,216 +66,22 @@ class LIBCELLML_EXPORT Generator: public Logger
...
-     * Get the number of tracked computed constants in the given @p analyserModel.
-     *
-     * @param analyserModel The pointer to the @ref AnalyserModel for which to get the number of tracked computed
-     * constants.
-     *
-     * @return The number of tracked computed constants in the analyser model.
+     * @return The @ref GeneratorVariableTracker used.
      */
-    size_t trackedComputedConstantCount(const AnalyserModelPtr &analyserModel);
+    GeneratorVariableTrackerPtr variableTracker();
```
hsorby: We are happy having `variableTracker` and not `generatorVariableTracker`? I see we have `profile` and not `generatorProfile`. Is this the only class that does this, if anybody knows off the top of their head.

agarny: This method is to be removed, so not a concern anymore. (I will finish this PR later tonight. Still need to finish my dinner...)

agarny:
Note that I am not done yet...

nickerso:
Why the change from `GeneratorContext` to `GeneratorVariableTracker`? Some useful commit messages would be great to start seeing. I am not a fan of that change and really struggling to see a need for this at all from the description of #1339...starting to sound more and more like some internal data management that needs no visible presence in the API.

agarny:
> Why the change from `GeneratorContext` to `GeneratorVariableTracker`? Some useful commit messages would be great to start seeing. I am not a fan of that change and really struggling to see a need for this at all from the description of #1339...starting to sound more and more like some internal data management that needs no visible presence in the API.

`GeneratorContext` is a bit too generic while `GeneratorVariableTracker` describes what the class is really about.

In the end, the idea is for the `Generator` class not to hold anything (i.e. a profile and a variable tracker are to be passed to `interfaceCode()` and `implementationCode()` rather than applied to a `Generator` object using `setProfile()` and `setVariableTracker()`).

Now, I agree that this is getting a bit out of hands and this is why I am finishing this cosmetic work (!!) now and this is it.

hsorby (approved):
I like this interface for the generator, I think it looks great.

nickerso (approved):
I don't think this is a good change, but happy to let it through. I assume documentation and tutorial updates are imminent that will help users transition our code to all these new APIs.

PR #1367, opened 2026-04-09
JavaScript: account for the fact that our JavaScript bindings use a 32-bit address space

### LINKED ISSUE #1366: JavaScript: account for the fact that our JavaScript bindings use a 32-bit address space
`AnalyserModel::areEquivalentVariables()` is a cached version of the areEquivalentVariables() utility, but it doesn't account for the fact that our JavaScript bindings use a 32-bit address space. This means that sometimes (especially with "big" CellML files) it causes collisions and incorrect cache hits like the one described at https://github.com/opencor/webapp/issues/374.

agarny:
Fixes #1366.


src/analysermodel_p.h
```diff
@@ -45,6 +46,42 @@ struct AnalyserModel::AnalyserModelImpl
...
     std::vector<AnalyserEquationPtr> mAnalyserEquations;
 
+    struct VariableKeyPair
+    {
+        uintptr_t first;
+        uintptr_t second;
+
+        bool operator==(const VariableKeyPair &other) const
+        {
+#ifdef CODE_COVERAGE_ENABLED
```
nickerso: is this a new thing that we haven't done before? or doing something we've done before in a different way?

nickerso (commented):
I think this looks ok, just trying to understand the introduction and use of the new code coverage define.

src/analysermodel_p.h
agarny: `CODE_COVERAGE_ENABLED` is indeed a new thing. It is needed for our code coverage test (in case the name didn't give it away! :)). We have a line that normally reads:
```c++
return first == other.first && second == other.second;
```
`second == other.second` is some kind of guard, but if I recall correctly code coverage tells us that it is is never `false` and that's because if `first == other.first` is `false` then `second == other.second` doesn't get evaluated. So, I use `CODE_COVERAGE_ENABLED` for the case the code is use during code coverage and, in this case, rather than executing:
```c++
return first == other.first && second == other.second;
```
we execute:
```c++
const auto firstEqual = static_cast<uintptr_t>(first == other.first);
const auto secondEqual = static_cast<uintptr_t>(second == other.second);

return firstEqual * secondEqual != 0;
```
which is the same (albeit less efficient) and has 100% coverage.

nickerso (dismissed):

src/analysermodel_p.h
hsorby: I am super unkeen to see this type of switching happening. We have avoided this as it simply makes a mockery of claiming that we have 100% line coverage. Have you had a look at the generated machine code to actually see if the second is less efficient (has more machine code), quite often the compiler generates the same code when optimisation is turned on.

hsorby: As far as my analysis goes we can replace && with &:

// Bitwise test on booleans.
(first == other.first) & (second == other.second)

This will not create a branch in the debug build.

The optimised builds of either way will produce very similar machine code.

agarny: I have just checked with https://godbolt.org/ and the generated machine code is different for `&` and `&&` in debug mode BUT the same in release mode. So, thanks for that, will implement your suggestion.

hsorby: It may be of interest that with C++20 you can use the default compartor of == and this standard means the compiled code will do the member comparison for you.

nickerso (approved):

src/analysermodel_p.h
agarny: Interesting, not least since I am requesting a switch to C++20 in https://github.com/cellml/libcellml/pull/1372 (see at the very bottom of https://github.com/cellml/libcellml/pull/1372/changes).

agarny: So, should we upgrade to C++20 in this PR?

hsorby: Should we do it as a separate PR so it is explicit and appears in our release notes/changelog?

hsorby: It seems like a significant change that we shouldn’t bury under other changes.

nickerso: for sure! We should have a single clear PR that makes a change like that isolated from any other code changes.

hsorby: And do we want that done before this goes through? So that this PR can take advantage of C++20?

nickerso: I wouldn't think so. You'd want to update all similar comparisons, right? This particular issue has already been resolved, right? so its not blocking this PR being merged.

agarny: Happy to leave this PR as-is and have C++20 added in another PR that will also remove those `operator==()` methods.

hsorby:
For me this looks like a good solution but adding a big file that is not used doesn’t seem to me to be the way to go. 
What I read into this is that we want this test but it takes too long to run in a practical development cycle and it is not needed to cover lines of code that are not already covered. I also see that we are showing a need to have some real world examples or tests that are not run as part of the development cycle but perhaps nightly or weekly on the current main branch.  We could also add the documentation and tutorial tests into this scheme.

So for me this PR needs to get rid of the big file and the unused/commented out tests (from history as well) and just keep the cache changes.

I will add some other issues to capture these test change idead, where people can add their views in a more appropriate  place than this PR comment chain.

hsorby:
Real world example issue is available under here: #1380. Comments for that idea should be sent there.

agarny:
> For me this looks like a good solution but adding a big file that is not used doesn’t seem to me to be the way to go. What I read into this is that we want this test but it takes too long to run in a practical development cycle and it is not needed to cover lines of code that are not already covered. I also see that we are showing a need to have some real world examples or tests that are not run as part of the development cycle but perhaps nightly or weekly on the current main branch. We could also add the documentation and tutorial tests into this scheme.

100% agreed.

> So for me this PR needs to get rid of the big file and the unused/commented out tests (from history as well) and just keep the cache changes.

Agreed, am going to remove them and force push things.

agarny:
Closing this PR in favour of PR #1384. I couldn't "easily" remove the very big CellML file from the Git history. So, faster and easier to create a new PR.

Issue #1370, opened 2026-04-13
Update version identifiers after a release.


dbrnz:
Version `0.6.3` was released on 11 November 2024 and although some significant code changes have been made since, `CMakeLists.txt` (at the HEAD of `main`) still has `_PROJECT_VERSION` set at `0.6.3`, . 

Shouldn't `_PROJECT_VERSION` be updated immediately after a release is made? With `PROJECT_DEVELOPER_VERSION` also being updated? And shouldn't the version now resolve to `0.7.0-rc.1`??

agarny:
That's the thing, no version has been released in a very, very long time. Our current release process relies on the ABI's local infrastructure and... well, it's broken, unavailable, or whatever. So, I am personally waiting on @hsorby doing something about it...

hsorby:
I am working on moving the release process to GitHub Actions, the currently unavailable Buildbot based release process did a lot of work checking and preparing for a release. I’m plugging away at this as a matter of urgency.

hsorby:
As to the update of the version number, this definitely shouldn’t be 0.7.0-rc.1. 
You cannot say that the next release will be a version 0.7.0 release candidate. At best you can say that the current code 0.6.3.post.1. The original idea was to determine the repository status from the last version tag, much like what pyproject.toml does for Python packages using SCM.

dbrnz:
My main point is that it definitely shouldn't have stayed at `0.6.3`. Given though that code from HEAD is being deployed into (semi-)released applications (e.g. simlulationvuer via libopencor), and that HEAD contains some significant changes from the `0.6.x`, shouldn't the current version be at least `0.7.0-alpha.1`??

agarny:
> I am working on moving the release process to GitHub Actions, the currently unavailable Buildbot based release process did a lot of work checking and preparing for a release. I’m plugging away at this as a matter of urgency.

Thanks Hugh!

hsorby:
The release process that has now been implemented adjusts the version number if the release has been successful before merging the release changes back into the main branch.

I think when we make the next official release and confirm that this is indeed happening for all release types, we can close this issue and related PR as complete.

PR #1371, opened 2026-04-14
Update version to `0.7.0-rc.1`

dbrnz:
There still needs to be some process which updates the version identifier immediately after a release is made (see #1370).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,8 +18,8 @@ set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Minimum OS X deployment vers
 
 set(PROJECT_NAME libCellML)
 set(PROJECT_URL https://libcellml.org)
-set(_PROJECT_VERSION 0.6.3)
-set(PROJECT_DEVELOPER_VERSION)
+set(_PROJECT_VERSION 0.7.0)
+set(PROJECT_DEVELOPER_VERSION -rc.1)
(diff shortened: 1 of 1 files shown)


agarny:
No release has been made. If you are talking about the version used in [lib]OpenCOR then think of that version as being a special port of libCellML. I don't like it and would much prefer to use an official version of libCellML, but... I am clearly not in charge of libCellML, so...

dbrnz:
>> ... updates the version identifier immediately after a release is made
>
> No release has been made. 

I meant that it should have been updated immediately after the `0.6.3` release and then each time a new "official" release is made.

> If you are talking about the version used in [lib]OpenCOR then think of that version as being a special port of libCellML. 

Being the HEAD of https://github.com/agarny/libcellml/tree/fixes ??

> I don't like it and would much prefer to use an official version of libCellML

Agree 100%

agarny:
> > > ... updates the version identifier immediately after a release is made
> > No release has been made.
> I meant that it should have been updated immediately after the `0.6.3` release and then each time a new "official" release is made.

Yeah, I actually agree (and this is what I do for [lib]OpenCOR), but I believe versioning is handled differently in libCellML and the version only gets updated when actually creating a version? I might be wrong though! @hsorby to chip in, if needed.

> > If you are talking about the version used in [lib]OpenCOR then think of that version as being a special port of libCellML.
> Being the HEAD of https://github.com/agarny/libcellml/tree/fixes ??

That's correct, this is the version libOpenCOR currently uses. It will soon be using the HEAD of another branch (that contains various optimisations -- I am currently going through a final round of checks).

Issue #1374, opened 2026-04-16
Analyser: cannot detect global NLA systems


dbrnz:
[test_RRC.zip](https://github.com/user-attachments/files/26767236/test_RRC.zip)

Here is a simple test case for issue #1362. The attached contains two models and code to validate them using `libcellml`:
1. `RC.cellml` simulates the charging of a 4F capacitor from a 12V battery via a 8Ω resistor.
2. `RRC.cellml` replaces the 8Ω resistor by two 4Ω resistors in series.

Both models load and run just fine in legacy OpenCOR (simulate for 200 seconds; in the RC model, `u_2` is the voltage across the capacitor; in the RRC model, `u_2` is the voltage at the connection between the two resistors and `u_3` that across the capacitor).

LibCellML though reports issues with the RRC model (`test.cpp` was built and linked against the repo HEAD):

```sh
$ ./test
./RC.cellml

./RRC.cellml
The type of variable 'v_q_2' in component 'main' is unknown.
The type of variable 'u_R_1' in component 'main' is unknown.
The type of variable 'v_R_1' in component 'main' is unknown.
The type of variable 'u_R_3' in component 'main' is unknown.
The type of variable 'v_R_3' in component 'main' is unknown.
The type of variable 'v_0' in component 'main' is unknown.
The type of variable 'v_S_u' in component 'main' is unknown.
The type of variable 'u_2' in component 'main' is unknown.
The type of variable 'v_1' in component 'main' is unknown.
The type of variable 'v_2' in component 'main' is unknown.
The type of variable 'v_3' in component 'main' is unknown.
```

Why??

agarny:
As you found out, it was because the analyser can only indentify NLA systems using an equation's structure, NOT by creating a graph of unresolved equations and variables. Anyway, this has been fixed in PR #1377 (note that it relies on other PRs so it contains way more modifiled files that it actually should!). However, a few things to keep in mind:
- So far, the analyser is currently relying on NLA variables having an `initial_value` (to act as an initial guess) to detect NLA systems.
- The NLA variables in your model don't have an `initial_value`, so it was tripping up the analyser. This being said, if I give your NLA variables an `initial_value`, then libCellML tells me that both `u_3` and `u_R_3` are overconstrained.
[...]

Issue #1050, opened 2022-11-15
Include functionality for rearranging algebraic loops to turn certain DAE's into ODE's


FinbarArgus:
Sometimes a system of equations will have algebraic loops that causes them to be seen as DAE's by the compiler. The process of doing the algebraic manipulation, to turn the system of DAE's into a system of ODE's would be very useful to have as a part of libcellml.

The attached file (change ext to cellml) is seen as a DAE when the line below "DAE Below" is uncommented and an ODE when instead the rearranged equation under "ODE below" is uncommented.

[Checking_V3.txt](https://github.com/cellml/libcellml/files/10008452/Checking_V3.txt)

hsorby:
Thanks for this, we are not yet able to deal with basic algebraic equations so this one is a little down the road.  We will endeavour to address this issue when we get to DAEs proper.

gbernardino:
I  have been struggling with this for quite some time already, and finally  I use a hacky solution. Current cellml does not seem very adapted to  work with algebraic equations, at least opencor (where I write the  cellml) constraints what you put in the LHS of an equation.
I wrote a python script that parses the model to sympy, and then I can use  their symbolic solver. It is not ideal, but works while waiting. After  solving the equation I generate C-code, but it should be relatively easy  to regenerate a cellml file, since sympy allows exortation  to vanilla mathml. Just keep in mind that in the process of solving the  equations, they are modified and might use more variables, so the  module structure can be broken. I tried, and it works in your example. I  can share the code if you want.
Some comments:
1. It is not  guaranteed to find solutions in all cases, but works well in linear  cases which are the ones I use the most (basically bifurcations in 0D  models). Also, keep in mind the sympy solve used to give problems with  piecewise defined functions.
2. It is slow for a high number of variables.
3.  I only tested in a reduced set of cellml (basically the one I need for my models), and I do not deal with imports.

agarny:
libCellML has changed a lot since @FinbarArgus created this issue, but most importantly we had a summer student (@RLee64; see his [technical report](https://github.com/user-attachments/files/26878584/2026.ABI.libCellML.Technical.Report.pdf)) who worked on converting some DAE models into ODE models. Rayen finished in early February and since, libCellML has changed even further. We should therefore merge his work into our [main](https://github.com/cellml/libcellml/tree/main) branch before things diverge too much.

agarny:
[...]

PR #1387, opened 2026-04-24
Analyser: cannot detect global NLA systems

agarny:
Fixes issue #1374.


agarny:
Closing since this has become obsolete as a result of our work on issue #1050. Indeed, global NLA systems can now easily be broken down.

Issue #1388, opened 2026-04-27
Improve caching of equivalent variables for Analyser


hsorby:
Whilst reviewing PR #1366, I played around with alternatives for caching equivalent variables.
I found that I could reduce memory usage and improve speed by changing the data structure for caching equivalent variables.

agarny:
Do you have some metrics to share?

hsorby:
For the big model test the cache drops from ~7.5 million entries to ~5K entries. The time taken to run the test is halved. For smaller tests the difference is negligible.

agarny:
Ok, nice!

agarny:
Closing as a result of PR #1389 having been merged.

Issue #1379, opened 2026-04-22
Update C++ standard to version 20


hsorby:
Currently we are setting the C++ standard to version 17 (is that 2017 without the 20). 
We could update to 20 (2020) to make use of some of the new standard compiler features.

agarny:
FWIW, for me, this means:
- updating our CMake scripts ; and then
- going through our whole codebase to see what can be improved thanks to the "new" C++20 standard.

hsorby:
The work on #1388 could render the original motivation for upgrading to C++ 20 moot.
As the motivation, that of replacing the line `(first == other.first) & (second == other.second)` with a default comparator (which requires C++20) would no longer be needed.

nickerso:
so this just becomes a ticket to update the CMake scripts and make sure tests still pass, leaving specific future optimisations to new tickets?

hsorby:
This ticket was always about just updating the C++ standard from 17 to 20.
That is the requirement meeting criteria for me.

agarny:
> so this just becomes a ticket to update the CMake scripts and make sure tests still pass, leaving specific future optimisations to new tickets?

For me, it is, yes. No new features, no nothing. Just add support for C++20 and take advantage of it wherever we can.

nickerso:
> Just add support for C++20 and take advantage of it wherever we can.

this seems in conflict with @hsorby's comment above.

PR #1389, opened 2026-04-27
Improve caching of equivalent variables in analyser model.

hsorby:
* Addresses #1388.


src/api/libcellml/analysermodel.h
```diff
@@ -600,14 +600,14 @@ class LIBCELLML_EXPORT AnalyserModel
...
-     * @ref Model. However, this might break if a @ref Model is modified after it
-     * has been analysed.
+     * This test is a cached version of the internal areEquivalentVariables() utility.
+     * The cache is built when the model is analysed.
+     * If the model has changed since it was analysed, then the cache might not be valid
+     * and the result of this test may be incorrect.
+     *
+     * This function is intended to be used by the @ref Analyser whilst it is analysing the model.
+     * It is not intended to be used by external code.
+     * Although, we do not restrict the use of this function in case it does have other applications.
```
agarny: People who look at the API will not know (and don't need to know) about our internal `areEquivalentVariables()` utility, so I would remove that mention.

Also, don't insert new lines within a paragraph (i.e. between lines 603 & 604, 604 & 605, 608 & 609, and 609 & 610).

agarny (commented):
Before reviewing further, I am going to wait for PR #1390 (for issue #1379) to be reviewed and merged in since it affects the caching code.

src/analyser.cpp
```diff
@@ -2324,6 +2324,8 @@ void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
 
     mCiCnUnits.clear();
 
+    mAnalyserModel->mPimpl->buildEquivalentVariablesCache();
+
```
agarny: I would move this to `AnalyserModel::AnalyserModelImpl::create()`, i.e.
```c++
AnalyserModelPtr AnalyserModel::AnalyserModelImpl::create(const ModelPtr &model)
{
    auto res = std::shared_ptr<AnalyserModel> {new AnalyserModel(model)};

    res->mPimpl->buildEquivalentVariablesCache();

    return res;
}
```

src/analysermodel.cpp
```diff
@@ -43,6 +43,33 @@ AnalyserModel::~AnalyserModel()
     delete mPimpl;
 }
 
+void AnalyserModel::AnalyserModelImpl::buildEquivalentVariablesCache()
+{
+    mEquivalentVariableCache.clear();
+    for (size_t i = 0; i < mModel->componentCount(); ++i) {
+        buildEquivalentVariablesCache(mModel->component(i));
+    }
+}
```
agarny: I prefer my code not to be compact, i.e.
```c++
    mEquivalentVariableCache.clear();

    for (size_t i = 0; i < mModel->componentCount(); ++i) {
        buildEquivalentVariablesCache(mModel->component(i));
    }
```

src/analysermodel.cpp
```diff
@@ -43,6 +43,33 @@ AnalyserModel::~AnalyserModel()
...
+            auto v2 = reinterpret_cast<uintptr_t>(equivalentVariable.get());
+            if (v2 < v1) {
+                std::swap(v1, v2);
+            }
+            unite(v1, v2);
+        }
+    }
+    for (size_t i = 0; i < component->componentCount(); ++i) {
+        buildEquivalentVariablesCache(component->component(i));
+    }
```
agarny, line 71: I prefer my code not to be compact, i.e.
```c++
    for (size_t i = 0; i < component->variableCount(); ++i) {
        auto variable = component->variable(i);

        for (size_t j = 0; j < variable->equivalentVariableCount(); ++j) {
            auto equivalentVariable = variable->equivalentVariable(j);
            auto v1 = reinterpret_cast<uintptr_t>(variable.get());
            auto v2 = reinterpret_cast<uintptr_t>(equivalentVariable.get());

            if (v2 < v1) {
                std::swap(v1, v2);
            }

            unite(v1, v2);
        }
    }

    for (size_t i = 0; i < component->componentCount(); ++i) {
        buildEquivalentVariablesCache(component->component(i));
    }
```

src/analysermodel.cpp
```diff
@@ -497,26 +524,10 @@ bool AnalyserModel::areEquivalentVariables(const VariablePtr &variable1,
     // means that we can safely cache the result of a call to that utility. In
     // turn, this means that we can speed up any feature (e.g., code generation)
     // that also relies on that utility.
-
```
agarny, line 500: Put that empty line back.

src/analysermodel_p.h
```diff
@@ -43,35 +43,31 @@ struct AnalyserModel::AnalyserModelImpl
     std::vector<AnalyserVariablePtr> mComputedConstants;
     std::vector<AnalyserVariablePtr> mAlgebraicVariables;
     std::vector<AnalyserVariablePtr> mExternalVariables;
-
```
agarny, line 46: Put that empty line back. Different group of variables.

src/analysermodel.cpp
```diff
@@ -497,26 +524,10 @@ bool AnalyserModel::areEquivalentVariables(const VariablePtr &variable1,
...
-    if (it != mPimpl->mCachedEquivalentVariables.end()) {
-        return it->second;
-    }
-
-    auto res = libcellml::areEquivalentVariables(variable1, variable2);
-
-    mPimpl->mCachedEquivalentVariables.emplace(key, res);
-
-    return res;
+    return mPimpl->find(v1) == mPimpl->find(v2);
```
agarny: This is part of the API, so we probably want some safeguards. So, maybe have the following:
```c++
bool AnalyserModel::areEquivalentVariables(const VariablePtr &variable1,
                                           const VariablePtr &variable2)
{
    // This is a cached version of the areEquivalentVariables() utility. Indeed,
    // an AnalyserModel object refers to a static version of a model, which
    // means that we can safely cache the result of a call to that utility. In
    // turn, this means that we can speed up any feature (e.g., code generation)
    // that also relies on that utility.

    if (variable1 == variable2) {
        return true;
    }

    if ((variable1 == nullptr) || (variable2 == nullptr)) {
        return false;
    }

    const auto v1 = reinterpret_cast<uintptr_t>(variable1.get());
    const auto v2 = reinterpret_cast<uintptr_t>(variable2.get());

    return mPimpl->find(v1) == mPimpl->find(v2);
}

```

src/analysermodel_p.h
```diff
@@ -43,35 +43,31 @@ struct AnalyserModel::AnalyserModelImpl
...
-            return hash;
+        const uintptr_t &rootX = find(x);
+        const uintptr_t &rootY = find(y);
+        if (rootX != rootY) {
+            mEquivalentVariableCache[rootY] = rootX;
         }
-    };
-
-    std::unordered_map<VariableKeyPair, bool, VariableKeyPairHash> mCachedEquivalentVariables;
+    }
```
agarny, line 76: I would put some empty lines here and there:
```c++
    std::unordered_map<uintptr_t, uintptr_t> mEquivalentVariableCache;

    uintptr_t find(uintptr_t x)
    {
        auto it = mEquivalentVariableCache.find(x);

        if (it == mEquivalentVariableCache.end()) {
            mEquivalentVariableCache[x] = x;

            return x;
        }

        if (it->second != x) {
            it->second = find(it->second);
        }

        return it->second;
    }

    void unite(uintptr_t x, uintptr_t y)
    {
        const auto rootX = find(x);
        const auto rootY = find(y);

        if (rootX != rootY) {
            mEquivalentVariableCache[rootY] = rootX;
        }
    }
```

src/analysermodel_p.h
```diff
@@ -43,35 +43,31 @@ struct AnalyserModel::AnalyserModelImpl
...
-        {
-            // A simple and portable hash function for a pair of pointers.
-
-            size_t hash = pair.first;
-
-            hash ^= pair.second + 0x9e3779b9 + (hash << 6) + (hash >> 2);
-
-            return hash;
+        const uintptr_t &rootX = find(x);
+        const uintptr_t &rootY = find(y);
```
agarny: Can be replaced with:
```c++
        const auto rootX = find(x);
        const auto rootY = find(y);
```

src/analysermodel_p.h
```diff
@@ -102,6 +98,9 @@ struct AnalyserModel::AnalyserModelImpl
 
     static AnalyserModelPtr create(const ModelPtr &model = nullptr);
 
+    void buildEquivalentVariablesCache();
+    void buildEquivalentVariablesCache(const ComponentPtr &component);
```
agarny: I would swap them, i.e. start with the "specialised" version and finish with the general version (same in the implementation).

src/analysermodel.cpp
```diff
@@ -497,26 +524,10 @@ bool AnalyserModel::areEquivalentVariables(const VariablePtr &variable1,
     // means that we can safely cache the result of a call to that utility. In
     // turn, this means that we can speed up any feature (e.g., code generation)
     // that also relies on that utility.
-
     auto v1 = reinterpret_cast<uintptr_t>(variable1.get());
     auto v2 = reinterpret_cast<uintptr_t>(variable2.get());
```
agarny: Can actually be replaced with:
```c++
    const auto v1 = reinterpret_cast<uintptr_t>(variable1.get());
    const auto v2 = reinterpret_cast<uintptr_t>(variable2.get());
```

src/analysermodel.cpp
agarny: Alternatively, we could have those tests here and then have the "meat" of the method in the private implementation, which could be called directly by the analyser to avoid those safeguards.

agarny (changes requested):
Looks good and definitely faster! Nice!

`AnalyserModel::AnalyserModelImpl::buildEquivalentVariablesCache()` should probably be called in `AnalyserModel::areEquivalentVariables()` (if the cache hasn't already been built), but having said that we know that the analyser is going to call `AnalyserModel::areEquivalentVariables()` many times, so I guess it's fine to call `AnalyserModel::AnalyserModelImpl::buildEquivalentVariablesCache()` upon retrieving the analyser model?

src/analysermodel_p.h
```diff
@@ -43,35 +43,31 @@ struct AnalyserModel::AnalyserModelImpl
...
-    struct VariableKeyPair
-    {
-        uintptr_t first;
-        uintptr_t second;
+    std::unordered_map<uintptr_t, uintptr_t> mEquivalentVariableCache;
 
-        bool operator==(const VariableKeyPair &other) const
-        {
-            return (first == other.first) & (second == other.second);
+    uintptr_t find(uintptr_t x)
```
agarny: Might want to use a different name? Maybe `findVariableAddres()`?

src/analysermodel_p.h
```diff
@@ -43,35 +43,31 @@ struct AnalyserModel::AnalyserModelImpl
...
         }
-    };
+        if (it->second != x) {
+            it->second = find(it->second);
+        }
+        return it->second;
+    }
 
-    struct VariableKeyPairHash
+    void unite(uintptr_t x, uintptr_t y)
```
agarny: Might want to use a different name? Maybe `uniteVariableAddreses()`?

src/analyser.cpp
hsorby: The `buildEquivalentVariablesCache` function expects the model to be a valid model. At this point this is not the case. I think the better place to build the cache is where it currently is and the state of the model is known.

agarny: I don't understand your justification, not least since we have:
```c++
void Analyser::AnalyserImpl::analyseModel(const ModelPtr &model)
{
    // Reset a few things in case this analyser was to be used to analyse more
    // than one model.

    mAnalyserModel = AnalyserModel::AnalyserModelImpl::create(model);

    mInternalVariables.clear();
    mInternalEquations.clear();

    mCiCnUnits.clear();

    mAnalyserModel->mPimpl->buildEquivalentVariablesCache();

    ...
```
So, whether we execute `mAnalyserModel->mPimpl->buildEquivalentVariablesCache();` where it currently is or as part of `AnalyserModel::AnalyserModelImpl::create()` is the same, no?

If `buildEquivalentVariablesCache()` really expects a valid model, then we are not currently checking that `model` (which is passed to `analyseModel()` is indeed valid. Also, by calling `buildEquivalentVariablesCache()` from `analyseModel()` as opposed to from `AnalyserModel::AnalyserModelImpl::create()`, this means that if an API user retrieves an analyser model then it won't benefit from the cache. I don't think we want that.

src/analysermodel_p.h
hsorby: These things are not the same, one is a reference the other is a value.

src/analysermodel_p.h
hsorby: 2f6ac37

src/analysermodel_p.h
hsorby: 2f6ac37

src/analysermodel_p.h
hsorby: 2f6ac37

src/analysermodel_p.h
hsorby: 2f6ac37

src/analysermodel.cpp
hsorby: 2f6ac37

src/analysermodel_p.h
hsorby: 2f6ac37

src/analysermodel.cpp
hsorby: 2f6ac37

src/analysermodel.cpp
hsorby: 2f6ac37

src/analysermodel.cpp
hsorby: 2f6ac37

src/analysermodel_p.h
agarny: Sorry, I was meaning:
```c++
const auto &rootX = find(x);
const auto &rootY = find(y);
```

src/api/libcellml/analysermodel.h
hsorby: 2f6ac37

src/analyser.cpp
hsorby: No, it is not the same.

Yes we are.

You cannot retrieve an analyser model until a model has been analysed.

src/analysermodel.cpp
hsorby: [2f6ac37](https://github.com/cellml/libcellml/pull/1389/commits/2f6ac3791d694fcfb0718ffe068e61ad9650befe)

src/analyser.cpp
agarny: > No, it is not the same.
> 
> Yes we are.

This is not helping me understand what I am missing.

> You cannot retrieve an analyser model until a model has been analysed.

Ok, I have just tried what I originally suggested (albeit I check that `model` does indeed exist; I had forgotten that it is an optional parameter!):
```c++
AnalyserModelPtr AnalyserModel::AnalyserModelImpl::create(const ModelPtr &model)
{
    auto res = std::shared_ptr<AnalyserModel> {new AnalyserModel(model)};

    if (model != nullptr) {
        res->mPimpl->buildEquivalentVariablesCache();
    }

    return res;
}
```
and it's all working fine. So, what am I missing exactly?

src/analysermodel_p.h
hsorby: I prefer not to user `auto` here as this is a simple type, I think the code reads better being clear in this case.

src/analyser.cpp
hsorby: Nothing now as far as I can tell, that you are performing the check for a nullptr.

To me this seems the less obvious place to create the cache, given that the model argument is optional to this create method.

agarny: The reason `model` is optional is so that we can have a default `mAnalyserModel`, making it possible to use the analyser with an invalid model, among other things.

But, ok, I am not going to argue this further.

src/analysermodel_p.h
agarny: The method is now called `findRootAddress()` and even though the name doesn't make sense to me, it kind of hints to the type it returns. But, ok, not going to argue this further.

src/api/libcellml/analysermodel.h
```diff
@@ -600,14 +600,13 @@ class LIBCELLML_EXPORT AnalyserModel
...
      *
-     * To test for equivalence is time consuming, so caching is used to speed
-     * things up. During the analysis of a model, various tests are performed
-     * and their result cached. So, if you test two variables that were tested
-     * during the analysis then the cached result will be returned otherwise the
-     * two variables will be properly tested and their result cached. This works
-     * because an @ref AnalyserModel always refers to a static version of a
-     * @ref Model. However, this might break if a @ref Model is modified after it
-     * has been analysed.
+     * The function utilizes caching which is constructed during the model
```
agarny: "utilises", not "utilizes".

src/api/libcellml/analysermodel.h
```diff
@@ -600,14 +600,13 @@ class LIBCELLML_EXPORT AnalyserModel
...
+     * analysis phase (@ref Analyser::analyseModel). The cache may become
+     * out-of-date if the model is changed after the model has been analysed.
```
agarny: "out of date", not "out-of-date" since it is not followed by a noun.

src/analysermodel_p.h
```diff
@@ -46,32 +46,33 @@ struct AnalyserModel::AnalyserModelImpl
...
-        uintptr_t first;
-        uintptr_t second;
+        auto it = mEquivalentVariableCache.find(x);
 
-        bool operator==(const VariableKeyPair &other) const
-        {
-            return (first == other.first) & (second == other.second);
+        if (it == mEquivalentVariableCache.end()) {
+            mEquivalentVariableCache[x] = x;
+            return x;
```
agarny, line 58: I would put an empty line before the`return` statement (to make things stand out more).

src/analysermodel.cpp
```diff
@@ -43,6 +48,38 @@ AnalyserModel::~AnalyserModel()
...
+void AnalyserModel::AnalyserModelImpl::buildEquivalentVariablesCache()
+{
+    mEquivalentVariableCache.clear();
+
+    for (size_t i = 0; i < mModel->componentCount(); ++i) {
+        buildEquivalentVariablesCache(mModel->component(i));
+    }
+}
+
+void AnalyserModel::AnalyserModelImpl::buildEquivalentVariablesCache(const ComponentPtr &component)
```
agarny, line 51: I would put this method before the previous one (at line 51), to respect the order in which they are declared in `src/analysermodel_p.h`.

src/analysermodel_p.h
agarny, line 1: Do we really want to put some actual code in a header file? I am talking about `findRootAddress()` and `uniteEquivalentAddresses()`.

src/analysermodel_p.h
```diff
@@ -46,32 +46,33 @@ struct AnalyserModel::AnalyserModelImpl
 
     std::vector<AnalyserEquationPtr> mAnalyserEquations;
 
-    struct VariableKeyPair
+    std::unordered_map<uintptr_t, uintptr_t> mEquivalentVariableCache;
+
+    uintptr_t findRootAddress(uintptr_t x)
```
agarny: Would rename this method to `findVariableAddress()`. I am not sure what is meant by "root" here.

src/analysermodel_p.h
```diff
@@ -46,32 +46,33 @@ struct AnalyserModel::AnalyserModelImpl
...
+        if (it->second != x) {
+            it->second = findRootAddress(it->second);
+        }
 
-            size_t hash = pair.first;
+        return it->second;
+    }
 
-            hash ^= pair.second + 0x9e3779b9 + (hash << 6) + (hash >> 2);
+    void uniteEquivalentAddresses(uintptr_t x, uintptr_t y)
```
agarny: I would rename this method to `uniteEquivalentVariableAddresses()` to be consistent with line 51.

.actrc
agarny, line 1: I imagine that this file (as well as `act-end-2-end-input.txt` and `act-payload.json`) were used for profiling? If so, then they should be removed. If not, then they should somehow be modified so that they are neither specific to your environment nor to a specific version of libCellML.

agarny (changes requested):
This looks. "Just" some nitpicking.

src/analysermodel_p.h
hsorby: For a private header file? Yes.
I am going to remove these functions and use a differenct caching technique for further optimastions in another pull request so these won’t last long.

.actrc
hsorby: [1221fcd](https://github.com/cellml/libcellml/pull/1389/commits/1221fcd94891c05c274afd078c41eddf3a07f3a5)

src/analysermodel_p.h
hsorby: [1221fcd](https://github.com/cellml/libcellml/pull/1389/commits/1221fcd94891c05c274afd078c41eddf3a07f3a5)

src/analysermodel_p.h
hsorby: [1221fcd](https://github.com/cellml/libcellml/pull/1389/commits/1221fcd94891c05c274afd078c41eddf3a07f3a5)

src/analysermodel.cpp
hsorby: [1221fcd](https://github.com/cellml/libcellml/pull/1389/commits/1221fcd94891c05c274afd078c41eddf3a07f3a5)

src/analysermodel_p.h
hsorby: [1221fcd](https://github.com/cellml/libcellml/pull/1389/commits/1221fcd94891c05c274afd078c41eddf3a07f3a5)

src/api/libcellml/analysermodel.h
hsorby: [1221fcd](https://github.com/cellml/libcellml/pull/1389/commits/1221fcd94891c05c274afd078c41eddf3a07f3a5)

src/api/libcellml/analysermodel.h
hsorby: [1221fcd](https://github.com/cellml/libcellml/pull/1389/commits/1221fcd94891c05c274afd078c41eddf3a07f3a5)

agarny (approved):
Looks good! Thanks!

src/api/libcellml/analysermodel.h
```diff
@@ -600,14 +600,13 @@ class LIBCELLML_EXPORT AnalyserModel
...
-     * because an @ref AnalyserModel always refers to a static version of a
-     * @ref Model. However, this might break if a @ref Model is modified after it
-     * has been analysed.
+     * The function utilises caching which is constructed during the model
+     * analysis phase (@ref Analyser::analyseModel). The cache may become
+     * out of date if the model is changed after the model has been analysed.
+     *
+     * @note This function is primarily designed for use during model analysis
+     *       by the @ref Analyser. While external usage is not programmatically
+     *       restricted, it is not the primary intended use case.
```
nickerso, line 609: is there a reason to "restrict" this? variable equivalence is a common aspect of CellML and many uses outside of libCellML could take advantage of a convenient and fast equivalence cache if it was pulled out of the analyser and made available as a first class part of the libCellML API. Similar to the CellML API's `CeVAS` service to manage connected variables...

hsorby: It isn’t “restricted”, the note is trying to explain that the function may find uses by the user that the developers have not considered, and that its primary function is to facilitate fast equivalent variable lookups for the analyser. The user needs to understand the limitations of this function if they do use if for their own purposes.

If this note doesn’t convey that message in a shorter form then maybe it needs re-writing.

nickerso: restricted in that its part of the analyser. I'm wondering how hard it would be to pull the equivalence cache out into its own class or maybe part of a model or something... Analysing models can only be done on complete models, whereas variable equivalence can be done any time.

hsorby: I don’t think that would be hard at all. Either option model based cache or first class object `EquivalentVariableCache` would work. 

My initial thought would be to have it as a first class object to keep the model as a plain model (arguments here could be made for some of the action API that currently exist on the model). I think I would prefer to see the model become plainer over time and have actioners do work on the model. The equivalent variable cache could be one of those. I expect this would become clearer once some exploratory tests were written.

I do see this query as being a separate issue to what this pull request is addressing.

nickerso: see #1435

jmdowrick (approved):
Looks great!

Issue #1396, opened 2026-05-14
Exponential scaling of Analyser() computation time for large-scale vascular network models


dsas627:
# Exponential scaling of `Analyser()` computation time for large-scale vascular network models

### Background / Context
I am using Circulatory Autogen to model the time-dependent haemodynamics of complex vascular networks. In this approach, each vessel segment within the vascular network is represented as an individual component within OpenCOR, where each vessel component is mapped to a separate module/model sourced from an external CellML modules file.

To sufficiently capture the haemodynamics of dense vascular anatomies, the resulting CellML models require a significant number of these interconnected vessel segments (scaling into the hundreds or more).

### The Issue
Currently, generating these CellML models places a significant computational load on Circulatory Autogen. As the size of the vascular network (and consequently, the number of vessel segments) increases, the computation time required to fully construct the CellML models scales exponentially.

While I can feasibly generate CellML models for smaller sub-volumes of the network, which is acceptable for early-stage prototyping, scaling this up to represent the entire anatomical volume (which is the ultimate goal of the project) is currently computationally infeasible.

### Investigation & Root Cause
After profiling the performance within Circulatory Autogen, I identified that the primary bottleneck contributing to the vast majority of this computation time is the `Analyser()` function from `libcellml`. As shown in the benchmarking figure below, the execution time for `Analyser()` scales exponentially relative to the number of generated vessel segments:

<img width="3552" height="2064" alt="Image" src="https://github.com/user-attachments/assets/64a53ae1-f38d-4805-9078-4fe9fbe594f9" />

**Figure 1:** A plot showcasing computation time as a function of the number of vessel segments for various helper functions within the Circulatory Autogen pipeline, demonstrating the significantly larger exponential scaling of computation time associated with the `Analyser()` function as the number of vessel segments increases. Sub-volumes between 5% to 20% were benchmarked.

### Impact & Request
Because the `Analyser()` function currently bottlenecks the model generation pipeline, it strictly limits the complexity and scale of the models we can build.

To better facilitate the modeling of larger, denser, and more anatomically complete vascular networks, there is a critical need to improve the computational efficiency and scaling behavior of the `Analyser()` function. Any insights, optimizations, or workarounds to help mitigate this exponential scaling for highly modular/component-heavy models would be greatly appreciated.

If you have any questions please let myself or @FinbarArgus know - thanks ((:

FinbarArgus:
[...]

PR #1397, opened 2026-05-19
Improve the time taken to validate and analyse large models

hsorby:
- Addresses #1396.

To run the investigation tests set an environment variable `BENCHMARKING_MODEL_ROOT` to the root directory of the unzipped contents of the benchmarking files zip in the associated issue.

The longer tests are disabled with the DISABLED_ prefix on the test name, remove this prefix to run the test.


agarny (dismissed):
Looks good although I have sped up a few things here and there (see commit messages).

agarny:
> To run the investigation tests set an environment variable `BENCHMARKING_MODEL_ROOT` to the root directory of the unzipped contents of the benchmarking files zip in the associated issue.
> 
> The longer tests are disabled with the DISABLED_ prefix on the test name, remove this prefix to run the test.

I can't see where `BENCHMARKING_MODEL_ROOT` and the `DISABLED_` prefix are being used in this PR...?

hsorby:
> > To run the investigation tests set an environment variable `BENCHMARKING_MODEL_ROOT` to the root directory of the unzipped contents of the benchmarking files zip in the associated issue.
> > The longer tests are disabled with the DISABLED_ prefix on the test name, remove this prefix to run the test.
> 
> I can't see where `BENCHMARKING_MODEL_ROOT` and the `DISABLED_` prefix are being used in this PR...?

I have taken that bit out now, as this was only part of the investigation phase of this work.
It was there for people to be able to run the models given in the associated issue easily.

agarny (approved):
Looks good to me.

agarny (dismissed):

src/analysermodel.cpp
```diff
@@ -98,20 +113,21 @@ AnalyserModel::Type AnalyserModel::type() const
...
-    {AnalyserModel::Type::ALGEBRAIC, "algebraic"},
-    {AnalyserModel::Type::INVALID, "invalid"},
-    {AnalyserModel::Type::UNDERCONSTRAINED, "underconstrained"},
-    {AnalyserModel::Type::OVERCONSTRAINED, "overconstrained"},
-    {AnalyserModel::Type::UNSUITABLY_CONSTRAINED, "unsuitably_constrained"}};
-
 std::string AnalyserModel::typeAsString(Type type)
 {
-    return typeToString.at(type);
+    static constexpr const char *names[] = {
```
jmdowrick, line 118: Have to be careful with enum ordering now - could be worth documenting this assumption for future Hugh?

src/validator.cpp
```diff
@@ -639,8 +639,9 @@ class Validator::ValidatorImpl: public LoggerImpl
      * @param component The component to check.
      * @param idMap The IdMap object to construct.
      * @param reportedConnections A set of connection identifiers to prevent duplicate reporting.
+     * @param connectionIds A map of connection identifiers to prevent duplicate reporting of connections.
```
jmdowrick: Could the description of the map be clearer? It sounds like a duplication of the description above.

jmdowrick (dismissed):
Looks OK to me. Some nice changes.

src/validator.cpp
hsorby: Fixed in [a0a20e2](https://github.com/cellml/libcellml/pull/1397/commits/a0a20e2c4d866e3cb11879fb63a76b1fa3b2e6b8)

jmdowrick (approved):

src/analysermodel.cpp
hsorby: I think this one is okay, we want to test the typeToString methods specifically to test for mismatches. A future issue perhaps.

agarny (approved):

src/analysermodel.cpp
hsorby: https://github.com/cellml/libcellml/issues/1449

PR #1461, opened 2026-09-18
Fix the Windows Python wheel generation

### LINKED ISSUE #1460: Windows wheels built by cibuildwheel are clashing
The wheel that cibuildwheel is creating for Windows is not behaving as expected.
We are currently asking it to build wheels for both the ARM64 and AMD64 archetictures.
This appears to be happening but the final wheel created has the same name for both architectures.
So we are getting a wheel named libcellml-0.7.1-cp310-cp310-win_amd64.whl for both.
PyPI is rejecting a wheel with this name if it is actually an ARM64 build.

This issue is to get the GitHub publish action to create a correctly named Windows built wheel file and have it successfully upload to PyPI from a GitHub action run.

hsorby:
The Windows wheels are currently always building the AMD64 architecture even when it is meant to be ARM64. This PR fixes that.

* Fixes #1460 

In the fixing of this problem I have upgraded to full pyproject.toml implementation using the latest build backend.


hsorby:
@agarny and @nickerso this PR will fix up the 0.7 series releases. Once it is merged we should be able to release 0.7.2 and then announce it because we won’t knowingly have any problems with it.

agarny (approved):

